Issue #35, opened 2014-08-20
Middleware Extensibility: Support "Multi-tenancy" In AuthenticationOptions (and more)


kingdango:
Copied from https://katanaproject.codeplex.com/discussions/561673 as I was advised this project is moving from CodePlex to GitHub.

Hello Katana team,
# Problem

I've been unable to cleanly extend the "built-in" authentication middleware because of its dependency on a concrete instance of AuthenticationOptions (and subclasses per implementation).

Presently the auth middleware is initialized at start-up and any configuration (API KEY, API SECRET, etc) cannot be changed on a per-request or per-tenant basis.  I've seen some suggestion of multiple middleware instances per tenant as a solution but this is not very efficient or (feasible) when you have N tenants.

In my case each of my customers has their own OAUTH api credentials for each service and I need to be able to set them per-request based on the tenant scope.  My problem, although seemingly common, is not the only reason to consider my proposed change.  

**The current dependency on a concrete implementation of [Middleware]AuthenticationOptions and AuthenticationOptions limits extensibility to the point that many are forced to "roll their own" copies of existing middleware with minor changes for their individual needs.** 
# Proposed Design Change

I propose a design change, of which I'd like to contribute code, to address this limitation and allow for an optional alternative way to load middleware AuthenticationOptions.  To achieve this we would depend on an interface **IAuthenticationOptions** and **I[MiddlewareName]AuthenticationOptions** for each specific middleware implementation.

These interfaces would have default implementations for each middleware.  For example, for Facebook there would be a DefaultFacebookAuthenticationOptions which would implement the interface and use the current (static) approach to loading AppId and AppSecret.  

A second scenario-specific implementation (which would be left to someone else to implement) could be MultitenantFacebookAuthenticationOptions which also implements the interface and identifies the tenant based on the request and then uses a datastore to load the AppId and AppSecret.  

It would probably make most sense to have a FacebookAuthenticationOptionsBase abstract class which contains most of the needed functionality.  We would also need to modify the existing FacebookAuthenticationOptions class to support the new interface but without breaking backward compatibility.

I have already proven this concept in my own project.  In order to implement this we would need to change several authentication middleware classes which have dependencies on AuthenticationOptions as well as existing middleware which has dependencies on concretes of their respective implementations (e.g. FacebookAuthenticationOptions).  Likewise, Extension Methods related to adding middleware to the pipeline would be modified to depend on the aforementioned interfaces instead of concretes. 
# Backwards Compatibility

[...]

PR #44, opened 2014-08-21
Improve extensibility of middleware by introducing interfaces for AuthenticationOptions

kingdango:
Resolve https://github.com/aspnet/Security/issues/35


brentschmaltz:
A solution may lie in providing delegates for obtaining the 'appId' etc. or the entire Options for that matter, the default would return the Options set on the Middleware. I think we can provide a solution without another interface.

Tratcher's observation that the majority of users are developing apps that function as a single 'client' seems correct.
I think the scenario that this discussion brings up, is one that will help some users.

Eilon:
@kingdango we're closing out this PR for now because we would like to take a different approach to multi-tenancy for the auth middleware. As we make progress on any future design ideas, we'll keep https://github.com/aspnet/Security/issues/35 updated with notes.

Issue #82, opened 2014-10-25
AuthenticationHandler.TeardownCoreAsync - Not getting called on exception


BadriL:
In my derived authentication handler, I'm taking some action during request processing and it requires a compensating action during response processing. I would like to do this compensating action in `TeardownCoreAsync`. But then the problem is `TeardownCoreAsync` is not guaranteed to be called. For example, if a downstream middleware throws an exception, `TeardownCoreAsync` does not get called. Is this by design? In authentication middleware, you could have a try-finally and call  `TeardownAsync` in the `finally` block. If `TeardownCoreAsync` is not guaranteed to be called by design, please let me know the recommended way of ensuring a block code to be called during response processing from my derived authentication handler.

PR #89, opened 2014-10-31
Make AuthHandler call Teardown on unhandled exceptions.

Tratcher:
#82 - AuthHandler needs to clean up after itself even for unhandled exceptions. This is especially true if the ErrorHandler starts re-executing requests.
@badriL


Tratcher:
@Praburaj @muratg

Praburaj:
Looks good to me :shipit:

PR #103, opened 2014-12-03
Initial set of Authorization changes

HaoK:
First pass at implementing proposed Authorization changes with some tweaks:
- Renamed AuthorizationPolicyOptions -> AuthorizationOptions with Policies property
- Deleted old implementation, kept the basic tests, removed the old Policy/retry stuff

<b>Issues:</b>
- Without the Before/After hooks, do we still need to separate IAuthorizationService and IAuthorizationPolicyHandler, collapse for now?

- Nothing is consuming the policy options right now, do we need an AuthorizationManager that Authorize attribute would use to do the more friendly Authorize via policyName (instead of via policy instance)?

- AuthorizationContext.Authorized isn't really useful if we immediately stop the logic if any handler fails to authorize.  Should we invoke all handlers letting them change the Authorized state even if previous handlers failed to authorize? (this is would give the handler interface a reason to exist)

cc @loudej @yishaigalatzer @rynowak @blowdart @divega


HaoK:
Replaced with https://github.com/aspnet/Security/pull/110

PR #110, opened 2014-12-30
Current iteration of AuthZ changes

HaoK:
IAuthorizationPolicy has IAuthorizationRequirements
AuthorizationPolicy has RequireClaims/RequireRoles/Require(IAuthorizationRequirements)
IAuthorizationPolicyHandler are the global policy handlers

cc @lodejard @blowdart @divega


HaoK:
Updated with latest conversation with @loudej @lodejard 

cc @yishaigalatzer

samples/SocialSample/SocialSample.kproj
```diff
@@ -12,6 +12,7 @@
   </PropertyGroup>
   <PropertyGroup>
     <SchemaVersion>2.0</SchemaVersion>
+    <DevelopmentServerPort>13232</DevelopmentServerPort>
```
davidfowl: zomg

src/Microsoft.AspNet.Security/AuthorizationOptions.cs
```diff
@@ -0,0 +1,22 @@
...
+        private IDictionary<string, AuthorizationPolicy> PolicyMap { get; } = new Dictionary<string, AuthorizationPolicy>();
+
+        public void AddPolicy([NotNull] string name, [NotNull] AuthorizationPolicy policy)
+        {
+            PolicyMap[name] = policy;
+        }
+
+        public AuthorizationPolicy GetPolicy([NotNull] string name)
+        {
+            return PolicyMap.ContainsKey(name) ? PolicyMap[name] : null;
```
davidfowl: TryGetValue

src/Microsoft.AspNet.Security/Properties/debugSettings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "Profiles": []
```
davidfowl: ignore

src/Microsoft.AspNet.Security/AuthorizationOptions.cs
HaoK: TryGetValue is just more characters no?

```
            AuthorizationPolicy result = null;
            return PolicyMap.TryGetValue(name, out result) ? result : null;

lodejard:
:shipit:

HaoK:
Note, one change Yishai suggested which I'm going to put in before I push, is that we move the DenyAnonymousRequirement and handler to Security since its not specific to MVC and has no depedencies.  Seemed ok so I'm going to move that code over here

HaoK:
Merged 4a635835af1049679103fdca97716e814408addc

PR #111, opened 2015-01-06
OpenIdConnect modifications for K.

brentschmaltz:


src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationHandler.cs
```diff
@@ -1,17 +1,28 @@
...
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
-using System.Threading.Tasks;
+using Microsoft.AspNet.Http.Security;
 using Microsoft.AspNet.Security.Infrastructure;
+using Microsoft.AspNet.Security.Notifications;
 using Microsoft.Framework.Logging;
+using Microsoft.IdentityModel.Protocols;
+using System;
```
davidfowl: Usings sorted incorrectly. System* is first

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationHandler.cs
```diff
@@ -1,17 +1,28 @@
...
+using System.Runtime.ExceptionServices;
+using System.Security.Claims;
+using System.Threading.Tasks;
 
 namespace Microsoft.AspNet.Security.OAuth
 {
     internal class OAuthBearerAuthenticationHandler : AuthenticationHandler<OAuthBearerAuthenticationOptions>
     {
-        private readonly ILogger _logger;
+		private const string HandledResponse = "HandledResponse";
```
davidfowl: formatting is all messed up.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationMiddleware.cs
```diff
@@ -0,0 +1,146 @@
...
+using Microsoft.AspNet.Security.DataProtection;
+using Microsoft.AspNet.Security.Infrastructure;
+using Microsoft.Framework.Logging;
+using Microsoft.Framework.OptionsModel;
+using Microsoft.IdentityModel.Protocols;
+
+namespace Microsoft.AspNet.Security.OpenIdConnect
+{
+    /// <summary>
+    /// OWIN middleware for obtaining identities using OpenIdConnect protocol.
```
Praburaj: ~~OWIN~~

src/Microsoft.AspNet.Security.OAuth/Notifications/OAuthBearerAuthenticationNotifications.cs
```diff
@@ -3,64 +3,61 @@
...
-        /// <param name="context"></param>
-        /// <returns></returns>
-        public virtual Task RequestToken(OAuthRequestTokenContext context)
-        {
-            return OnRequestToken(context);
-        }
+		/// <summary>
+		/// Invoked if exceptions are thrown during request processing. The exceptions will be re-thrown after this event unless suppressed.
+		/// </summary>
+		public Func<AuthenticationFailedNotification<OAuthRequestTokenContext, OAuthBearerAuthenticationOptions>, Task> AuthenticationFailed { get; set; }
```
Tratcher: Formatting issues.

src/Microsoft.AspNet.Security.OAuth/Notifications/OAuthBearerAuthenticationNotifications.cs
```diff
@@ -69,5 +66,5 @@ public Task ApplyChallenge(OAuthChallengeContext context)
         {
```
Tratcher: You don't need this method anymore.

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationHandler.cs
```diff
@@ -24,87 +35,167 @@ protected override AuthenticationTicket AuthenticateCore()
...
-                {
-                    _logger.WriteWarning("invalid bearer token received");
-                    return null;
-                }
+				var validationParameters = Options.TokenValidationParameters.Clone();
+				if (validationParameters.ValidIssuer == null && !string.IsNullOrWhiteSpace(_configuration.Issuer))
+				{
+					validationParameters.ValidIssuer = _configuration.Issuer;
+				}
+				else
```
Tratcher: `else if (!string.IsNullOrWhiteSpace(_configuration.Issuer))`

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationMiddleware.cs
```diff
@@ -2,21 +2,25 @@
...
-    /// created by application code directly, instead it is added by calling the the IAppBuilder UseOAuthBearerAuthentication
-    /// extension method.
-    /// </summary>
-    public class OAuthBearerAuthenticationMiddleware : AuthenticationMiddleware<OAuthBearerAuthenticationOptions>
+	/// <summary>
+	/// Bearer authentication middleware component which is added to an HTTP pipeline. This class is not
+	/// created by application code directly, instead it is added by calling the the IAppBuilder UseOAuthBearerAuthentication
+	/// extension method.
+	/// </summary>
+	public class OAuthBearerAuthenticationMiddleware : AuthenticationMiddleware<OAuthBearerAuthenticationOptions>
```
Tratcher: It looks like you've merged a lot of the old AAD and ADFS scenarios down into this component. Can is still be used as a simple validator without any of the metdata stuff? E.g. can I still just provide my own token validator without setting any of the metadata, configuration, or backchannel properties? It looks like it will always try to set up the configuration manager.

src/Microsoft.AspNet.Security.OAuth/project.json
```diff
@@ -2,19 +2,47 @@
...
     "description": "ASP.NET 5 middleware that enables an application to support any standard OAuth 2.0 authentication workflow.",
     "dependencies": {
+        "Microsoft.AspNet.Http.Extensions": "1.0.0-*",
         "Microsoft.AspNet.Security": "1.0.0-*",
-        "Microsoft.AspNet.Security.DataProtection": "1.0.0-*"
+        "Microsoft.AspNet.Security.DataProtection": "1.0.0-*",
+        "Microsoft.AspNet.WebUtilities": "1.0.0-*",
+        "Microsoft.Framework.Logging": "1.0.0-*",
+        "Newtonsoft.Json": "6.0.6",
+        "System.IdentityModel.Tokens": "5.0.2-*",
```
Tratcher: Do you have your packages pushed to myget yet? On what feed?

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationMiddleware.cs
Tratcher: Everywhere.

PinpointTownes:
Formatting issues put aside (please kill those tabs with fire! :fire:), this PR introduces questionable changes: what's the point of deeply changing the OAuth2 client middleware to use OIDC-specific concepts if you also provide an OpenID Connect client middleware, that doesn't even inherit from the OAuth2 one?

src/Microsoft.AspNet.Security.OAuth/Notifications/OAuthBearerAuthenticationNotifications.cs
```diff
@@ -3,64 +3,61 @@
...
-                context.HttpContext.Response.Headers.AppendValues("WWW-Authenticate", context.Challenge);
-                return Task.FromResult(0);
-            };
-        }
-
-        /// <summary>
-        /// Handles processing OAuth bearer token.
-        /// </summary>
-        public Func<OAuthRequestTokenContext, Task> OnRequestToken { get; set; }
+			AuthenticationFailed = notification => Task.FromResult(0);
```
PinpointTownes: You can easily replace with constructor with the new inline properties setting stuff offered by Roslyn.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -0,0 +1,621 @@
...
+                return null;
+            }
+
+            OpenIdConnectMessage openIdConnectMessage = null;
+
+            // assumption: if the ContentType is "application/x-www-form-urlencoded" it should be safe to read as it is small.
+            if (string.Equals(Request.Method, "POST", StringComparison.OrdinalIgnoreCase)
+              && !string.IsNullOrWhiteSpace(Request.ContentType)
+                // May have media/type; charset=utf-8, allow partial match.
+              && Request.ContentType.StartsWith("application/x-www-form-urlencoded", StringComparison.OrdinalIgnoreCase)
```
Tratcher: FYI: I'm redoing the GetFormAsync right now so we'll be able to remove much of this code.

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationHandler.cs
```diff
@@ -24,87 +35,167 @@ protected override AuthenticationTicket AuthenticateCore()
...
+						return ticket;
+					}
+				}
+
+				throw new InvalidOperationException("No SecurityTokenValidator available for token: " + requestTokenContext.Token);
+            }
+            catch (Exception ex)
+            {
+				// We can't await inside a catch block, capture and handle outside.
+				authFailedEx = ExceptionDispatchInfo.Capture(ex);
```
PinpointTownes: With Roslyn, you can now `await` stuff in a `catch` block.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -0,0 +1,621 @@
...
+
+            return null;
+        }
+
+        /// <summary>
+        /// 'Remembers' the nonce associated with this message. By default the nonce added as a secure cookie.
+        /// </summary>
+        /// <param name="message"><see cref="OpenIdConnectMessage"/> associatied with the nonce.</param>
+        /// <remarks>A cookie is added with the name obtained from  <see cref="GetNonceKey"/>.</remarks>
+        public virtual void AddNonce(string nonce)
```
Tratcher: We discussed removing all these virtual methods and moving all the extensibility to Notifications, no?

src/Microsoft.AspNet.Security/AuthenticationTicket.cs
```diff
@@ -29,11 +29,33 @@ public AuthenticationTicket(ClaimsIdentity identity, AuthenticationProperties pr
...
+
+        /// <summary>
+        /// Gets the authenticated user identity.
+        /// </summary>
+        public string AuthenticationType { get; private set; }
+
+        /// <summary>
         /// Gets the authenticated user identity.
         /// </summary>
         public ClaimsIdentity Identity { get; private set; }
```
Tratcher: We shouldn't have both Identity and Principal.

src/Microsoft.AspNet.Security/AuthenticationTicket.cs
```diff
@@ -29,11 +29,33 @@ public AuthenticationTicket(ClaimsIdentity identity, AuthenticationProperties pr
...
+        /// <param name="properties"></param>
+        public AuthenticationTicket(ClaimsPrincipal principal, AuthenticationProperties properties, string authenticationType)
+        {
+            AuthenticationType = authenticationType;
+            Principal = principal;
+            Properties = properties ?? new AuthenticationProperties();
+        }
+
+        /// <summary>
+        /// Gets the authenticated user identity.
```
Tratcher: Wrong comment. Who is this auth type intended for? What value would it contain?

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
```diff
@@ -22,20 +22,38 @@ public static class SecurityHelper
...
             {
-                foreach (var existingClaimsIdentity in existingPrincipal.Identities)
+                ClaimsIdentity claimsIdentity = identity as ClaimsIdentity;
+                if (claimsIdentity == null)
+                {
+                    context.User.AddIdentity(new ClaimsIdentity(identity));
+                }
+                else
                 {
-                    if (existingClaimsIdentity.IsAuthenticated)
```
Tratcher: The IsAuthenticated check here was being used to filter out the default anonymous identity.

src/Microsoft.AspNet.Security/Notifications/BaseNotification.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// </summary>
+        public void SkipToNextMiddleware()
+        {
+            State = NotificationResultState.Skipped;
+        }
+
+		/// <summary>
+		/// Gets or set the <see cref="AuthenticationTicket"/> to return if this notification signals it handled the notification.
+		/// </summary>
+        public AuthenticationTicket AuthenticationTicket { get; set; }
```
Tratcher: Is an auth ticket appropriate data for every notification? I don't think it applies to OnApplyRedirect for example.

src/Microsoft.AspNet.Security/Notifications/RedirectFromIdentityProviderNotification.cs
```diff
@@ -1,17 +1,21 @@
...
+        public RedirectFromIdentityProviderNotification(HttpContext context, TOptions options)
+            : base(context, options)
+        {
+        }
 
         public string SignInAsAuthenticationType { get; set; }
 
-        public bool Cancel { get; set; }
-
         public bool IsRequestCompleted { get; set; }
```
Tratcher: This was replaced by HandledRequest, no?

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationMiddleware.cs
PinpointTownes: Removing the `AccessTokenFormat`/`TicketDataFormat` stuff also means that people wanting to use opaque tokens won't be able to use the new OAuth2 bearer middleware without writing a custom `SecurityTokenHandler`. Looks like a too much aggressive change.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -0,0 +1,621 @@
...
+                }
+
+                var protocolValidationContext = new OpenIdConnectProtocolValidationContext
+                {
+                    AuthorizationCode = openIdConnectMessage.Code,
+                    Nonce = nonce,
+                };
+
+                Options.ProtocolValidator.Validate(jwt, protocolValidationContext);
+                if (openIdConnectMessage.Code != null)
```
PinpointTownes: Basic access token retrieval has been a major demand. Is this something you could consider?

PinpointTownes:
My 2 cents: the different mentions of OpenID Connect types/concepts/logic in the OAuth2 bearer middleware make these changes very confusing. Maybe it'd better to use a neutral name like `BearerAuthenticationMiddleware` or `TokenAuthenticationMiddleware` and extract OAuth2/OIDC specific stuff to different projects?

src/Microsoft.AspNet.Security.OAuth/Notifications/OAuthBearerAuthenticationNotifications.cs
```diff
@@ -3,64 +3,61 @@
 
 using System;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Security.Notifications;
 
+/// <summary>
```
Praburaj: Is this XML comment necessary for a namespace?

src/Microsoft.AspNet.Security.OAuth/Notifications/OAuthBearerAuthenticationNotifications.cs
```diff
@@ -3,64 +3,61 @@
...
 namespace Microsoft.AspNet.Security.OAuth
 {
     /// <summary>
     /// OAuth bearer token middleware provider
     /// </summary>
-    public class OAuthBearerAuthenticationNotifications : IOAuthBearerAuthenticationNotifications
+    public class OAuthBearerAuthenticationNotifications
     {
         /// <summary>
         /// Initializes a new instance of the <see cref="OAuthBearerAuthenticationProvider"/> class
```
Praburaj: `OAuthBearerAuthenticationProvider` => `OAuthBearerAuthenticationNotifications`

src/Microsoft.AspNet.Security.OAuth/project.json
```diff
@@ -2,19 +2,47 @@
...
+            "frameworkAssemblies": { 
+                 "System.Net.Http.WebRequest": "", 
+                 "System.Net.Http": "" 
+             } 
         },
         "aspnetcore50": {
             "dependencies": {
-                "System.Net.Http.WinHttpHandler": "4.0.0-beta-*"
+                "System.Collections": "4.0.10-beta-*",
+                "System.ComponentModel": "4.0.0-beta-*",
```
Praburaj: I think there is some scope to clean up these dependencies.

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationHandler.cs
```diff
@@ -24,87 +35,167 @@ protected override AuthenticationTicket AuthenticateCore()
...
+
+				// Give application opportunity to find from a different location, adjust, or reject token
+				var messageReceivedNotification =
+					new MessageReceivedNotification<OAuthRequestTokenContext, OAuthBearerAuthenticationOptions>(Context, Options)
+					{
+						ProtocolMessage = requestTokenContext,
+					};
+
+				// notification can set the token
+				await Options.Notifications.MessageReceived(messageReceivedNotification);
```
Praburaj: Instead of raising `MessageReceived` event for every request in the pipeline it would be good if we raise this only if we know there is an `Authorization` header with value starting `Bearer`. Otherwise we will end up calling this too frequently. In the previous implementation we call this notification like that. In our katana's `OpenIDC` or `WSFed` implementation we raise the event only if the request matches several things like : 

1. Url's path is same as callback (there is no equivalent here). 
2. request is `POST` and containing a specific encoding etc.

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationHandler.cs
```diff
@@ -24,87 +35,167 @@ protected override AuthenticationTicket AuthenticateCore()
...
+				{
+					return messageReceivedNotification.AuthenticationTicket;
+				}
+
+				if (messageReceivedNotification.Skipped)
+				{
+					return null;
+				}
+
+				if (string.IsNullOrEmpty(requestTokenContext.Token))
```
Praburaj: Looks like `requestTokenContext.Token` is always going to be null as we initialized the Token value as `null` in the `new OAuthRequestTokenContext(Context, null)` instantiation?

Praburaj: Instead we probably need to check for `Authorization` header `!= null` ?

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationHandler.cs
```diff
@@ -24,87 +35,167 @@ protected override AuthenticationTicket AuthenticateCore()
...
-                    return null;
-                }
+				var validationParameters = Options.TokenValidationParameters.Clone();
+				if (validationParameters.ValidIssuer == null && !string.IsNullOrWhiteSpace(_configuration.Issuer))
+				{
+					validationParameters.ValidIssuer = _configuration.Issuer;
+				}
+				else
+				{
+					IEnumerable<string> issuers = new[] { _configuration.Issuer };
```
Praburaj: minor `var`.

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationHandler.cs
PinpointTownes: > Instead of raising MessageReceived event for every request in the pipeline it would be good if we raise this only if we know there is an Authorization header with value starting Bearer.

Sadly, this would prevent the end dev from changing the location the access token should be retrieved from (e.g an `access_token` parameter in the query string).

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationHandler.cs
PinpointTownes: AFAICT, the access token is expected to be changeable from the `MessageReceived` notification.

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationHandler.cs
```diff
@@ -24,87 +35,167 @@ protected override AuthenticationTicket AuthenticateCore()
...
+					validationParameters.ValidIssuers = (validationParameters.ValidIssuers == null ? issuers : validationParameters.ValidIssuers.Concat(issuers));
+				}
+
+				validationParameters.IssuerSigningKeys = (validationParameters.IssuerSigningKeys == null ? _configuration.SigningKeys : validationParameters.IssuerSigningKeys.Concat(_configuration.SigningKeys));
+				SecurityToken validatedToken;
+				foreach (var validator in Options.SecurityTokenValidators)
+				{
+					if (validator.CanReadToken(requestTokenContext.Token))
+					{
+
```
Praburaj: minor : use `var` in all these lines.

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationHandler.cs
```diff
@@ -24,87 +35,167 @@ protected override AuthenticationTicket AuthenticateCore()
...
+						ClaimsPrincipal principal = Options.SecurityTokenValidators.First().ValidateToken(requestTokenContext.Token, validationParameters, out validatedToken);
+						ClaimsIdentity claimsIdentity = principal.Identity as ClaimsIdentity;
+						AuthenticationTicket ticket = new AuthenticationTicket(claimsIdentity, new AuthenticationProperties());
+						var securityTokenValidatedNotification = new SecurityTokenValidatedNotification<OAuthRequestTokenContext, OAuthBearerAuthenticationOptions>(Context, Options)
+						{
+							ProtocolMessage = requestTokenContext,
+							AuthenticationTicket = ticket
+						};
+
+						if (securityTokenReceivedNotification.HandledResponse)
```
Praburaj: Do we need to raise `SecurityTokenValidated` here and check if handled response on `securityTokenValidatedNotification`?

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationHandler.cs
```diff
@@ -24,87 +35,167 @@ protected override AuthenticationTicket AuthenticateCore()
...
+				}
+				else
+				{
+					IEnumerable<string> issuers = new[] { _configuration.Issuer };
+					validationParameters.ValidIssuers = (validationParameters.ValidIssuers == null ? issuers : validationParameters.ValidIssuers.Concat(issuers));
+				}
+
+				validationParameters.IssuerSigningKeys = (validationParameters.IssuerSigningKeys == null ? _configuration.SigningKeys : validationParameters.IssuerSigningKeys.Concat(_configuration.SigningKeys));
+				SecurityToken validatedToken;
+				foreach (var validator in Options.SecurityTokenValidators)
```
Praburaj: foreach loop made me think we are creating a separate principal for every validator. Instead will it be better to say 

```
var validator = Options.SecurityTokenValidators.Where(v => v.CanReadToken(requestTokenContext.Token)).FirstOrDefault();
```

And do the validation and return from here.

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationHandler.cs
```diff
@@ -24,87 +35,167 @@ protected override AuthenticationTicket AuthenticateCore()
...
+            {
+				// We can't await inside a catch block, capture and handle outside.
+				authFailedEx = ExceptionDispatchInfo.Capture(ex);
+            }
 
-                // Validate expiration time if present
-                DateTimeOffset currentUtc = Options.SystemClock.UtcNow;
+			if (authFailedEx != null)
+			{
+				_logger.WriteError("Exception occurred while processing message: '" + authFailedEx.ToString());
```
Praburaj: Closing `'` for the message.

src/Microsoft.AspNet.Security.OpenIDConnect/Project.json
```diff
@@ -0,0 +1,48 @@
...
+    "frameworks": {
+        "aspnet50": {
+            "frameworkAssemblies": {
+                "System.Net.Http": "",
+                "System.Net.Http.WebRequest": ""
+            }
+        },
+        "aspnetcore50": {
+            "dependencies": {
+                "System.Collections": "4.0.10-beta-*",
```
Praburaj: I think there is scope to clean up some of these transitive dependencies.

src/Microsoft.AspNet.Security.OpenIDConnect/Project.json
```diff
@@ -0,0 +1,48 @@
+{
+    "version": "1.0.0-*",
+    "dependencies": {
+        "Microsoft.AspNet.Http.Extensions": "1.0.0-*",
```
Praburaj: Sort these dependencies.

src/Microsoft.AspNet.Security/Notifications/AuthenticationFailedNotification.cs
```diff
@@ -2,18 +2,19 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 
+using Microsoft.AspNet.Http;
```
Praburaj: Sort namespaces.

src/Microsoft.AspNet.Security/Notifications/NotificationResultState.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
```
Praburaj: License header missing.

src/Microsoft.AspNet.Security/Notifications/SecurityTokenReceivedNotification.cs
```diff
@@ -2,15 +2,19 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 
+using Microsoft.AspNet.Http;
```
Praburaj: Sort namespaces.

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationHandler.cs
Praburaj: Got it.

samples/OpenIDConnectSample/Startup.cs
```diff
@@ -0,0 +1,57 @@
...
+    public class Startup
+    {
+        public void Configure(IApplicationBuilder app)
+        {
+            app.UseServices(services =>
+            {
+                services.AddDataProtection();
+                services.Configure<ExternalAuthenticationOptions>(options =>
+                {
+                    options.SignInAsAuthenticationType = OpenIdConnectAuthenticationDefaults.AuthenticationType;
```
Tratcher: This is the wrong default value. It should be the value from CookieAuth.

samples/OpenIDConnectSample/Startup.cs
```diff
@@ -0,0 +1,57 @@
...
+                });
+
+            });
+
+            app.UseCookieAuthentication(options =>
+            {
+                options.AuthenticationType = OpenIdConnectAuthenticationDefaults.AuthenticationType;
```
Tratcher: This is backwards.

samples/OpenIDConnectSample/Startup.cs
```diff
@@ -0,0 +1,57 @@
...
+            });
+
+            app.UseOpenIdConnectAuthentication(options =>
+                {
+                    options.ClientId = "fe78e0b4-6fe7-47e6-812c-fb75cee266a4";
+                    options.Authority = "https://login.windows.net/cyrano.onmicrosoft.com";
+                    options.RedirectUri = "http://localhost:42023";
+                    options.SignInAsAuthenticationType = OpenIdConnectAuthenticationDefaults.AuthenticationType;
```
Tratcher: This is backwards.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -0,0 +1,337 @@
...
+        public string SignInAsAuthenticationType
+        {
+            get { return TokenValidationParameters.AuthenticationType; }
+            set { TokenValidationParameters.AuthenticationType = value; }
+        }
+
+        /// <summary>
+        /// Gets or sets the type used to secure data handled by the middleware.
+        /// </summary>
+        public ISecureDataFormat<AuthenticationProperties> StateDataFormat
```
Tratcher: Would `NonceDataFormat` make more sense?

src/Microsoft.AspNet.Security/AuthenticationTicket.cs
blowdart: And why ClaimsPrincipal and not IClaimsPrincipal?

Tratcher:
General Bearer token question summarized from earlier discussion:

The prior bearer token middleware was quite simple, just looking at the header and then applying the any interpreter you provided. JWT was plugged into this and AAD and ADFS were layered on top of JWT.

With the new middleware all of these layers have been merged to make things like AAD & ADFS metadata easier to work with.  That's cool, but it's not clear if the the original simple scenario can still be implemented. Can you please provide a sample showing how to use the new middleware without metadata, JWT, etc., but only a custom token reader?

This merger also pulls in several dependencies that I think are inappropriate for the other middleware in this package, the (interactive) generic OAuth middleware used by all of the social auth. I think we'll need to break up the packaging to avoid crossing dependencies. This will require either moving OAuthBearer to its own package, or re-establishing some of the old layering so the added dependencies are pulled in via a higher level package.

src/Microsoft.AspNet.Security/AuthenticationTicket.cs
PinpointTownes: Hum, you mean the old `IClaimsPrincipal` used by WIF (`Microsoft.IdentityModel`)?
The `ClaimsPrincipal` from `mscorlib` doesn't implement this interface, that doesn't exist anymore in the .NET 4.5 port.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: You mean the new `StringDataFormat`?

Tratcher: Woops, yes, I meant to say that the `StringDataFormat` property might be better renamed to `NonceDataFormat`.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -0,0 +1,337 @@
...
+            get
+            {
+                return _backchannelTimeout;
+            }
+
+            set
+            {
+                if (value <= TimeSpan.Zero)
+                {
+                    throw new ArgumentOutOfRangeException("BackchannelTimeout", value, Resources.ArgsException_BackchallelLessThanZero);
```
PinpointTownes: It's time to use `nameof` massively :mask:

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: Yeah, it's definitely a better name.

src/Microsoft.AspNet.Security.OpenIDConnect/NonceCache.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Security.OpenIdConnect
+{
+    public interface INonceCache
```
PinpointTownes: Wouldn't the MSFT OIDC extensions package be a better place for that generic stuff?

Tratcher: This is the OIDC package. Or were you referring to System.IdentityModel.Tokens or Microsoft.IdentityModel.Protocol.Extensions?

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: Out of curiosity, why adding a null check here?
The middleware's constructor handles this specific case by generating a default data format using the data protection stuff.

src/Microsoft.AspNet.Security.OpenIDConnect/NonceCache.cs
PinpointTownes: Ahah thanks, I was indeed referring to `Microsoft.IdentityModel.Protocol.Extensions` but I couldn't remember the exact name :smile:

`System.IdentityModel.Jwt` could also be an option, but this interface is not directly specific to JWT tokens.

src/Microsoft.AspNet.Security.OpenIDConnect/Notifications/AuthorizationCodeReceivedNotification.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Security.OpenIdConnect;
+using Microsoft.IdentityModel.Protocols;
+using System.Diagnostics.CodeAnalysis;
+using System.IdentityModel.Tokens;
+
+namespace Microsoft.AspNet.Security.Notifications
```
PinpointTownes: The other security providers use the `Microsoft.AspNet.Security`.`[Provider]` pattern:

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Security.Cookies/Notifications/CookieApplyRedirectContext.cs#L9
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Security.OAuth/Notifications/OAuthAuthenticatedContext.cs#L12

Definitely not a big deal, but something you should fix to improve consistency.

src/Microsoft.AspNet.Security.OpenIDConnect/Notifications/AuthorizationCodeReceivedNotification.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    /// <summary>
+    /// This Notification can be used to be informed when an 'AuthorizationCode' is received over the OpenIdConnect protocol.
+    /// </summary>
+    public class AuthorizationCodeReceivedNotification : BaseNotification<OpenIdConnectAuthenticationOptions>
+    {
+        /// <summary>
+        /// Creates a <see cref="AuthorizationCodeReceivedNotification"/>
+        /// </summary>
+        public AuthorizationCodeReceivedNotification(HttpContext context, OpenIdConnectAuthenticationOptions options) : base(context, options)
```
PinpointTownes: Style: you should move `: base(...)` to a new line to make it consistent with the other providers:

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Security.Cookies/Notifications/CookieApplyRedirectContext.cs#L24

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationMiddleware.cs
```diff
@@ -0,0 +1,173 @@
...
+using Microsoft.AspNet.Security.DataProtection;
+using Microsoft.AspNet.Security.Infrastructure;
+using Microsoft.Framework.Logging;
+using Microsoft.Framework.OptionsModel;
+using Microsoft.IdentityModel.Protocols;
+using Microsoft.AspNet.Security.DataHandler.Serializer;
+using Microsoft.AspNet.Security.DataHandler.Encoder;
+using System.Text;
+
+namespace Microsoft.AspNet.Security.OpenIdConnect
```
PinpointTownes: The namespace `OpenIdConnect` is not consistent with the folder name `OpenIDConnect`, you should fix that (I suppose it's inherited from Katana 3)

src/Microsoft.AspNet.Security.OpenIDConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -0,0 +1,580 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved. See License.txt in the project root for license information.
```
PinpointTownes: Like `OpenIdConnectAuthenticationHandler` in Katana 3, this file has a name whose case doesn't correspond to the class name: `OpenidConnectAuthenticationHandler` -> `OpenIdConnectAuthenticationHandler`.

src/Microsoft.AspNet.Security/AuthenticationTicket.cs
PinpointTownes: I agree with @Tratcher, having a `Principal` here (whose associated comment is wrong BTW) seems really weird.

src/Microsoft.AspNet.Security/Infrastructure/AuthenticationTokenReceiveContext.cs
```diff
@@ -26,11 +20,6 @@ public class AuthenticationTokenReceiveContext : BaseContext
 
         public AuthenticationTicket Ticket { get; protected set; }
 
-        public void DeserializeTicket(string protectedData)
```
PinpointTownes: Outch, this is a major breaking change. Why removing it?

src/Microsoft.AspNet.Security.OAuth/Notifications/OAuthBearerAuthenticationNotifications.cs
```diff
@@ -3,64 +3,61 @@
...
+		/// <summary>
+		/// Invoked when a protocol message is first received.
+		/// </summary>
+		public Func<MessageReceivedNotification<OAuthRequestTokenContext, OAuthBearerAuthenticationOptions>, Task> MessageReceived { get; set; }
 
-        /// <summary>
+		/// <summary>
+		/// Invoked with the security token that has been extracted from the protocol message.
+		/// </summary>
+		public Func<SecurityTokenReceivedNotification<OAuthRequestTokenContext, OAuthBearerAuthenticationOptions>, Task> SecurityTokenReceived { get; set; }
```
PinpointTownes: Your reworked generic `SecurityTokenReceivedNotification` is used in a weird way: I'm not sure that `OAuthRequestTokenContext` can semantically be considered as a `ProtocolMessage`.

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationHandler.cs
```diff
@@ -24,87 +35,167 @@ protected override AuthenticationTicket AuthenticateCore()
...
-                }
+				var validationParameters = Options.TokenValidationParameters.Clone();
+				if (validationParameters.ValidIssuer == null && !string.IsNullOrWhiteSpace(_configuration.Issuer))
+				{
+					validationParameters.ValidIssuer = _configuration.Issuer;
+				}
+				else
+				{
+					IEnumerable<string> issuers = new[] { _configuration.Issuer };
+					validationParameters.ValidIssuers = (validationParameters.ValidIssuers == null ? issuers : validationParameters.ValidIssuers.Concat(issuers));
```
PinpointTownes: Sounds like it's time to use the new `?.` stuff offered by Roslyn.

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationHandler.cs
```diff
@@ -24,87 +35,167 @@ protected override AuthenticationTicket AuthenticateCore()
...
+				}
+
+				validationParameters.IssuerSigningKeys = (validationParameters.IssuerSigningKeys == null ? _configuration.SigningKeys : validationParameters.IssuerSigningKeys.Concat(_configuration.SigningKeys));
```
PinpointTownes: Same thing here.

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationOptions.cs
```diff
@@ -33,34 +44,121 @@ public OAuthBearerAuthenticationOptions() : base()
...
+		/// Used to know what the current clock time is when calculating or validating token expiration. When not assigned default is based on
+		/// DateTimeOffset.UtcNow. This is typically needed only for unit testing.
+		/// </summary>
+		public ISystemClock SystemClock { get; set; }
+
+		/// <summary>
+		/// Gets or sets the <see cref="SecurityTokenValidators"/> of <see cref="SecurityTokenHandler"/>s used to read and validate <see cref="SecurityToken"/>s. 
+		/// </summary>
+		/// <exception cref="ArgumentNullException">if 'value' is null.</exception>
+		public ICollection<ISecurityTokenValidator> SecurityTokenValidators
```
PinpointTownes: It's a best practice to make collection properties non-settable:

http://msdn.microsoft.com/en-us/library/ms182327(v=vs.120).aspx
http://msdn.microsoft.com/en-us/library/dn169389%28v=vs.110%29.aspx

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -0,0 +1,337 @@
...
+        /// Gets or sets the 'scope'.
+        /// </summary>
+        public string Scope { get; set; }
+
+        /// <summary>
+        /// Gets or sets the AuthenticationType used when creating the <see cref="System.Security.Claims.ClaimsIdentity"/>.
+        /// </summary>
+        public string SignInAsAuthenticationType
+        {
+            get { return TokenValidationParameters.AuthenticationType; }
```
PinpointTownes: This kind of voodoo magic proved to offer a really bad user experience in the past: if you set it prior to setting `TokenValidationParameters`, the value set to SignInAsAuthenticationType will be silently ignored.

brentschmaltz:
Lets move all comments to #112

PR #112, opened 2015-01-12
Additions for OpenIdConnectMiddleware and OAuthBearer Beta1.

brentschmaltz:
OpenIdConnect and OAuthBearer modifications.


samples/OpenIDConnectSample/Startup.cs
```diff
@@ -0,0 +1,57 @@
...
+                services.Configure<ExternalAuthenticationOptions>(options =>
+                {
+                    options.SignInAsAuthenticationType = OpenIdConnectAuthenticationDefaults.AuthenticationType;
+                });
+
+            });
+
+            app.UseCookieAuthentication(options =>
+            {
+                options.AuthenticationType = OpenIdConnectAuthenticationDefaults.AuthenticationType;
```
Tratcher: These auth types are all setup wrong.

samples/OpenIDConnectSample/project.json
```diff
@@ -0,0 +1,18 @@
+﻿{
+    "dependencies": {
+        "Microsoft.AspNet.Security.Cookies": "1.0.0-*",
```
Tratcher: Nit: Sort.

src/Microsoft.AspNet.Security.OAuth/Notifications/OAuthBearerAuthenticationNotifications.cs
```diff
@@ -1,64 +1,61 @@
...
+            SecurityTokenReceived = notification => Task.FromResult(0);
+            SecurityTokenValidated = notification => Task.FromResult(0);
+
             OnApplyChallenge = context =>
             {
                 context.HttpContext.Response.Headers.AppendValues("WWW-Authenticate", context.Challenge);
                 return Task.FromResult(0);
             };
-        }
+		}
```
Tratcher: Nit: Formatting

src/Microsoft.AspNet.Security.OAuth/OAuthBearerAuthenticationMiddleware.cs
```diff
@@ -2,21 +2,25 @@
...
-    /// created by application code directly, instead it is added by calling the the IAppBuilder UseOAuthBearerAuthentication
-    /// extension method.
-    /// </summary>
-    public class OAuthBearerAuthenticationMiddleware : AuthenticationMiddleware<OAuthBearerAuthenticationOptions>
+	/// <summary>
+	/// Bearer authentication middleware component which is added to an HTTP pipeline. This class is not
+	/// created by application code directly, instead it is added by calling the the IAppBuilder UseOAuthBearerAuthentication
+	/// extension method.
+	/// </summary>
+	public class OAuthBearerAuthenticationMiddleware : AuthenticationMiddleware<OAuthBearerAuthenticationOptions>
```
Tratcher: Nit: formatting.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -0,0 +1,581 @@
...
+using System;
+using System.Globalization;
+using System.IdentityModel.Tokens;
+using System.IO;
+using System.Linq;
+using System.Runtime.ExceptionServices;
+using System.Security.Claims;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.Security.OpenIdConnect
```
Tratcher: File name casing is inconsistent.

src/Microsoft.AspNet.Security.OpenIDConnect/Project.json
```diff
@@ -0,0 +1,48 @@
+{
```
Tratcher: project.json needs to be all lower case.

Tratcher:
The file and folder names across this whole commit use inconsistent casing of Openid, OpenID, or OpenId. Pick one.

Tratcher:
I have several questions about bearer token, ping me when you have a minute to go over it.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -0,0 +1,581 @@
...
+            OpenIdConnectMessage openIdConnectMessage = null;
+
+            // assumption: if the ContentType is "application/x-www-form-urlencoded" it should be safe to read as it is small.
+            if (string.Equals(Request.Method, "POST", StringComparison.OrdinalIgnoreCase)
+              && !string.IsNullOrWhiteSpace(Request.ContentType)
+                // May have media/type; charset=utf-8, allow partial match.
+              && Request.ContentType.StartsWith("application/x-www-form-urlencoded", StringComparison.OrdinalIgnoreCase)
+              && Request.Body.CanRead)
+            {
+                if (!Request.Body.CanSeek)
```
Tratcher: ReadFormAsync now has built in buffering, you don't need this anymore.

src/Microsoft.AspNet.Security.OAuth/OAuthAuthenticationMiddleware.cs
```diff
@@ -1,16 +1,16 @@
...
-using System.Diagnostics.CodeAnalysis;
-using System.Globalization;
-using System.Net.Http;
 using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Security.DataHandler;
 using Microsoft.AspNet.Security.DataProtection;
 using Microsoft.AspNet.Security.Infrastructure;
 using Microsoft.Framework.Logging;
 using Microsoft.Framework.OptionsModel;
+using System;
```
Tratcher: System goes first.

src/Microsoft.AspNet.Security.OAuth/project.json
```diff
@@ -3,7 +3,9 @@
     "description": "ASP.NET 5 middleware that enables an application to support any standard OAuth 2.0 authentication workflow.",
     "dependencies": {
         "Microsoft.AspNet.Security": "1.0.0-*",
-        "Microsoft.AspNet.Security.DataProtection": "1.0.0-*"
+        "Microsoft.AspNet.Security.DataProtection": "1.0.0-*",
```
Tratcher: Revert these dependency changes.

src/Microsoft.AspNet.Security.OAuthBearer/Notifications/OAuthBearerAuthenticationNotifications.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="OAuthBearerAuthenticationProvider"/> class
+        /// </summary>
+        public OAuthBearerAuthenticationNotifications()
+        {
+            AuthenticationFailed = notification => Task.FromResult(0);
+            MessageReceived = notification => Task.FromResult(0);
+            SecurityTokenReceived = notification => Task.FromResult(0);
+            SecurityTokenValidated = notification => Task.FromResult(0);
+		}
```
Tratcher: formatting.

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationExtensions.cs
```diff
@@ -1,10 +1,10 @@
 // Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
-using Microsoft.AspNet.Security.OAuth;
+using Microsoft.AspNet.Security.OAuthBearer;
 using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.OptionsModel;
+using System;
```
Tratcher: System goes first.

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationExtensions.cs
```diff
@@ -1,10 +1,10 @@
...
 
 namespace Microsoft.AspNet.Builder
```
Tratcher: Do we need more scenario specific extensions now?

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationHandler.cs
```diff
@@ -0,0 +1,195 @@
...
+
+                authFailedEx.Throw();
+            }
+
+            return null;
+        }
+
+        protected override void ApplyResponseChallenge()
+        {
+            // N/A
```
Tratcher: Bearer does need to issue a challenge. the default is just `Bearer`, but there should be a notification extension.

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationMiddleware.cs
```diff
@@ -0,0 +1,118 @@
...
+using System.Net.Http;
+
+namespace Microsoft.AspNet.Security.OAuthBearer
+{
+	/// <summary>
+	/// Bearer authentication middleware component which is added to an HTTP pipeline. This class is not
+	/// created by application code directly, instead it is added by calling the the IAppBuilder UseOAuthBearerAuthentication
+	/// extension method.
+	/// </summary>
+	public class OAuthBearerAuthenticationMiddleware : AuthenticationMiddleware<OAuthBearerAuthenticationOptions>
```
Tratcher: Formatting

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationMiddleware.cs
```diff
@@ -0,0 +1,118 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.Security.DataProtection;
+using Microsoft.AspNet.Security.Infrastructure;
+using Microsoft.Framework.Logging;
+using Microsoft.Framework.OptionsModel;
+using Microsoft.IdentityModel.Protocols;
+using System;
```
Tratcher: System goes first, all files.

test/Microsoft.AspNet.Security.Test/project.json
```diff
@@ -18,7 +20,8 @@
     "frameworks": {
         "aspnet50": {
             "dependencies": {
-                "Shouldly": "1.1.1.1"
+                "Shouldly": "1.1.1.1",
+                "System.Security.Claims": "1.0.0-*"
```
Praburaj: I think this can be moved to the main dependencies node.

test/Microsoft.AspNet.Security.Test/OpenIdConnectMiddlewareTests/OpenIdConnectMiddlewareTests.cs
```diff
@@ -0,0 +1,359 @@
...
+using Microsoft.AspNet.Http.Security;
+using Microsoft.AspNet.Security.Cookies;
+using Microsoft.AspNet.Security.DataHandler;
+using Microsoft.AspNet.Security.DataProtection;
+using Microsoft.AspNet.Security.OpenIdConnect;
+using Microsoft.AspNet.TestHost;
+using Microsoft.Framework.DependencyInjection;
+using Newtonsoft.Json;
+using Shouldly;
+using System;
```
Praburaj: sort and system namespaces at the top.

test/Microsoft.AspNet.Security.Test/OpenIdConnectMiddlewareTests/OpenIdConnectMiddlewareTests.cs
```diff
@@ -0,0 +1,359 @@
+﻿using Microsoft.AspNet.Builder;
```
Praburaj: license header missing.

samples/OpenIDConnectSample/project.json
```diff
@@ -0,0 +1,18 @@
...
+    },
+    "frameworks": {
+        "aspnet50": { },
+        "aspnetcore50": { }
+    },
+    "commands": {
+        "web": "Microsoft.AspNet.Hosting server=Microsoft.AspNet.Server.WebListener server.urls=http://localhost:12345",
+        "kestrel": "Microsoft.AspNet.Hosting --server Kestrel --server.urls http://localhost:5004"
+    },    
+    "webroot": "d://wwwroot"
```
Praburaj: Avoid hardcoding of webroot.

src/Microsoft.AspNet.Security.OAuth/project.json
```diff
@@ -3,7 +3,9 @@
...
+        "Microsoft.IdentityModel.Protocol.Extensions": "2.0.0-beta1-*",
```
Praburaj: can this be `2.0.0-*` instead. So that we don't have to change this often.

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationHandler.cs
```diff
@@ -0,0 +1,195 @@
...
+            }
+            catch (Exception ex)
+            {
+                // We can't await inside a catch block, capture and handle outside.
+                authFailedEx = ExceptionDispatchInfo.Capture(ex);
+            }
+
+            if (authFailedEx != null)
+            {
+                _logger.WriteError("Exception occurred while processing message: '" + authFailedEx.ToString());
```
Praburaj: There is a `WriteError` extension that can take in the `exception` object directly. https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Framework.Logging/LoggerExtensions.cs#L333

Tratcher:
Still waiting for the simple bearer token sample that doesn't use metadata or anything, it just plugs in its own token reader.

Tratcher:
:watch:

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationMiddleware.cs
```diff
@@ -0,0 +1,173 @@
...
+    /// </summary>
+    public class OpenIdConnectAuthenticationMiddleware : AuthenticationMiddleware<OpenIdConnectAuthenticationOptions>
+    {
+        private readonly ILogger _logger;
+
+        /// <summary>
+        /// Initializes a <see cref="OpenIdConnectAuthenticationMiddleware"/>
+        /// </summary>
+        /// <param name="next">The next middleware in the OWIN pipeline to invoke</param>
+        /// <param name="app">The OWIN application</param>
```
Praburaj: ~~OWIN~~

PinpointTownes:
You should take a look at the different comments mentioned in your old PR: https://github.com/aspnet/Security/pull/111

I agree with @Tratcher: we really need a basic "token validation" middleware that behaves exactly like the OAuth2 bearer token middleware built in Katana, taking no dependency on the OIDC extensions. 

Making `SecurityTokenHandler` the first class citizen in the base bearer middleware is not necessarily a bad thing, but if you replace `ISecureDataFormat<AuthenticationTicket> AccessTokenFormat` by a collection of security token handlers, providing an adapter between the two worlds is a must (think about the scenario where the access token created by the OIDC server is opaque and uses a data protector). But IMHO, providing an `ISecureDataFormat` implementation delegating the `Create/Receive` operations to your security token handlers collection would be better than using directly `SecurityTokenHandler`.

Personally, I'd rename the existing OAuth2 bearer middleware to `TokenAuthenticationMiddleware` and extract all the new OAuth2/OIDC greatness you plan to add to a separate package. `OAuthBearerAuthenticationMiddleware` has been a very confusing name for most people (visit #owin on JabbR if you're not convinced), so I think it's really time to fix that.

samples/OpenIDConnectSample/project.json
brentschmaltz: I am not sure what you are asking?

src/Microsoft.AspNet.Security.OpenIDConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: Casing has been fixed.

src/Microsoft.AspNet.Security.OpenIDConnect/Project.json
brentschmaltz: Fixed

src/Microsoft.AspNet.Security.OpenIDConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: Removed and changes.

samples/OpenIDConnectSample/Startup.cs
brentschmaltz: Why is that? What are you suggesting they be?

PinpointTownes: `CookieAuthenticationDefaults.AuthenticationType`?

Or you can simply omit setting the property.

samples/OpenIDConnectSample/Startup.cs
```diff
@@ -0,0 +1,56 @@
...
+    public class Startup
+    {
+        public void Configure(IApplicationBuilder app)
+        {
+            app.UseServices(services =>
+            {
+                services.AddDataProtection();
+                services.Configure<ExternalAuthenticationOptions>(options =>
+                {
+                    options.SignInAsAuthenticationType = OpenIdConnectAuthenticationDefaults.AuthenticationType;
```
PinpointTownes: `CookieAuthenticationDefaults.AuthenticationType`?

samples/OpenIDConnectSample/Startup.cs
```diff
@@ -0,0 +1,56 @@
...
+            {
+                options.AuthenticationType = OpenIdConnectAuthenticationDefaults.AuthenticationType;
+            });
+
+            app.UseOpenIdConnectAuthentication(options =>
+                {
+                    options.ClientId = "fe78e0b4-6fe7-47e6-812c-fb75cee266a4";
+                    options.Authority = "https://login.windows.net/cyrano.onmicrosoft.com";
+                    options.RedirectUri = "http://localhost:42023";
+                    options.SignInAsAuthenticationType = OpenIdConnectAuthenticationDefaults.AuthenticationType;
```
PinpointTownes: `CookieAuthenticationDefaults.AuthenticationType`?

Or just remove that, given that you already set the cookies middleware as the default authentication middleware via `ExternalAuthenticationOptions`.

samples/OpenIDConnectSample/Startup.cs
```diff
@@ -0,0 +1,56 @@
...
+                    options.AuthenticationType = OpenIdConnectAuthenticationDefaults.AuthenticationType;
```
PinpointTownes: You can also omit it.

test/Microsoft.AspNet.Security.Test/OAuthBearer/OAuthBearerMiddlewareTests.cs
```diff
@@ -0,0 +1,243 @@
...
+
+        private static TestServer CreateServer(Action<OAuthBearerAuthenticationOptions> configureOptions, Func<HttpContext, bool> handler = null)
+        {
+            return TestServer.Create(app =>
+            {
+                app.UseServices(services =>
+                {
+                    services.AddDataProtection();
+                });
+                app.UseCookieAuthentication(options =>
```
Tratcher: Bearer shouldn't have CookieAuth setup.

test/Microsoft.AspNet.Security.Test/OAuthBearer/OAuthBearerMiddlewareTests.cs
```diff
@@ -0,0 +1,243 @@
...
+                request.Headers.Add("Authorization", authorizationHeader);
+            }
+
+            var transaction = new Transaction
+            {
+                Request = request,
+                Response = await server.CreateClient().SendAsync(request),
+            };
+
+            if (transaction.Response.Headers.Contains("Set-Cookie"))
```
Tratcher: No cookies

samples/OpenIDConnectSample/Startup.cs
brentschmaltz: Not blocking, let's think about that later.

src/Microsoft.AspNet.Security.OAuth/OAuthAuthenticationMiddleware.cs
brentschmaltz: Why?

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationHandler.cs
brentschmaltz: Not sure what you are asking.

test/Microsoft.AspNet.Security.Test/OpenIdConnectMiddlewareTests/OpenIdConnectMiddlewareTests.cs
brentschmaltz: I get we always want system on top, but the default from VisualStudio is as this file is laid out.
Why is system on top important?

samples/OpenIDConnectSample/Startup.cs
brentschmaltz: Interesting pinpoint about a simplier, the idea is that it just works if you pass in the authority pinpoint, we reach out for validation coordinates, keys, issues, etc and produce a ClaimsPrincipal. I am pushing for a consistent model for validation. I made a small change where the notification can set the AuthenticationTicket and return.  Giving you complete freedom to take it all over.

samples/OpenIDConnectSample/project.json
brentschmaltz: lol, I read SNORT, not sort.  Will fix.

src/Microsoft.AspNet.Security.OAuth/project.json
brentschmaltz: Do you mean the S.IM.Tokens etc?

src/Microsoft.AspNet.Security.OAuthBearer/Notifications/OAuthBearerAuthenticationNotifications.cs
brentschmaltz: fixed, stupid tabs.

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationExtensions.cs
brentschmaltz: Sure, found the Option to have "system" first.

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationExtensions.cs
brentschmaltz: We need lots of help here.  This is the beginning of this middleware.

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationMiddleware.cs
brentschmaltz: Yep, found the magic option setting.

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationMiddleware.cs
brentschmaltz: Sir yes sir.

test/Microsoft.AspNet.Security.Test/project.json
brentschmaltz: Shouldly is not available for "aspnetcore50". What guidance does the K team have here?

test/Microsoft.AspNet.Security.Test/OpenIdConnectMiddlewareTests/OpenIdConnectMiddlewareTests.cs
brentschmaltz: not anymore.

samples/OpenIDConnectSample/project.json
brentschmaltz: thanks.

src/Microsoft.AspNet.Security.OAuth/project.json
brentschmaltz: thank was a mistake, thanks

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationHandler.cs
brentschmaltz: Good idea, modified that and oidc.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationMiddleware.cs
brentschmaltz: Yep, changed to ASP.NET

samples/OpenIDConnectSample/Startup.cs
brentschmaltz: I am not entirely sure what to do here. One goal I have is to remove the need to explicitly set this Auth type by adding a property to the AuthenticationTicket. It always seemed add to me that the runtime requires the ClaimsIdentity to have a specific authtype for things to work. I am thinking of ways to re-think this.

Tratcher: Please avoid redesigning auth types in this PR, it's way out of scope. If we're going to tackle that issue we need to do it as its own work item across the whole stack.

test/Microsoft.AspNet.Security.Test/project.json
Tratcher: This test project doesn't support aspnetcore50 so you don't need to list any dependencies in the framework specific section. If we decide to test core later we'll have to deal with the dependencies then.

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationHandler.cs
```diff
@@ -0,0 +1,195 @@
...
+                        return ticket;
+                    }
+                }
+
+                throw new InvalidOperationException("No SecurityTokenValidator available for token: " + token ?? "null");
+            }
+            catch (Exception ex)
+            {
+                // We can't await inside a catch block, capture and handle outside.
+                authFailedEx = ExceptionDispatchInfo.Capture(ex);
```
Tratcher: EDI is no longer needed, you can now call await directly in a catch block.

src/Microsoft.AspNet.Security/AuthenticationTicket.cs
```diff
@@ -29,11 +23,34 @@ public AuthenticationTicket(ClaimsIdentity identity, AuthenticationProperties pr
...
+
+        /// <summary>
         /// Gets the authenticated user identity.
         /// </summary>
         public ClaimsIdentity Identity { get; private set; }
 
         /// <summary>
+        /// Gets the authenticated user identity.
+        /// </summary>
+        public ClaimsPrincipal Principal{ get; private set; }
```
Tratcher: We really don't want two disconnected identity fields here. They either need to be connected or there should only be one.

Nit: spacing.

test/Microsoft.AspNet.Security.Test/project.json
brentschmaltz: I get that, but what is the harm?

samples/OpenIDConnectSample/Startup.cs
brentschmaltz: I backed out some of the changes we will need moving forward. So we can get these in, we will deal with the other issues later.

test/Microsoft.AspNet.Security.Test/OAuthBearer/OAuthBearerMiddlewareTests.cs
brentschmaltz: Probably doesn't need DataProtection either.

NuGet.Config
```diff
@@ -3,5 +3,6 @@
   <packageSources>
     <add key="AspNetVNext" value="https://www.myget.org/F/aspnetvnext/api/v2" />
     <add key="NuGet.org" value="https://nuget.org/api/v2/" />
+    <add key="AzureADNighty" value="http://www.myget.org/F/azureadwebstacknightly"/>
```
Praburaj: wondering if having `/api/v2` is required or not necessary. Or may be for consistency with other feeds.

samples/OpenIDConnectSample/Startup.cs
```diff
@@ -0,0 +1,56 @@
...
+
+                    context.Response.ContentType = "text/plain";
+                    await context.Response.WriteAsync("Hello First timer");
+                    return;
+                }
+
+                context.Response.ContentType = "text/plain";
+                await context.Response.WriteAsync("Hello Authenticated User");
+            });
+
```
Praburaj: nit: Delete empty lines.

samples/OpenIDConnectSample/project.json
```diff
@@ -0,0 +1,18 @@
...
+        "Microsoft.AspNet.Server.IIS": "1.0.0-*",
+        "Microsoft.AspNet.Security.OpenIdConnect": "1.0.0-*",
+        "Microsoft.AspNet.Server.WebListener": "1.0.0-*"
+    },
+    "frameworks": {
+        "aspnet50": { },
+        "aspnetcore50": { }
+    },
+    "commands": {
+        "web": "Microsoft.AspNet.Hosting server=Microsoft.AspNet.Server.WebListener server.urls=http://localhost:12345",
```
Praburaj: nit: All of our other samples use 5001 as port for weblistener. Yeah but does not matter though.

src/Microsoft.AspNet.Security.OAuthBearer/Notifications/OAuthBearerAuthenticationNotifications.cs
```diff
@@ -0,0 +1,50 @@
...
+/// </summary>
+namespace Microsoft.AspNet.Security.OAuthBearer
+{
+    /// <summary>
+    /// OAuth bearer token middleware provider
+    /// </summary>
+    public class OAuthBearerAuthenticationNotifications
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="OAuthBearerAuthenticationProvider"/> class
```
Praburaj: OAuthBearerAuthenticationProvider => OAuthBearerAuthenticationNotifications

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationHandler.cs
```diff
@@ -0,0 +1,195 @@
...
+                    return messageReceivedNotification.AuthenticationTicket;
+                }
+
+                if (messageReceivedNotification.Skipped)
+                {
+                    return null;
+                }
+
+                string authorization = Request.Headers.Get("Authorization");
+                if (authorization.StartsWith("Bearer ", StringComparison.OrdinalIgnoreCase))
```
Praburaj: How do I populate the token from a different location other than the Authorization header? 

`if (string.IsNullOrEmpty(requestTokenContext.Token))` seems to be no more available.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationNotifications.cs
```diff
@@ -0,0 +1,59 @@
...
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Security.Notifications;
+using Microsoft.IdentityModel.Protocols;
+
+namespace Microsoft.AspNet.Security.OpenIdConnect
+{
+    /// <summary>
+    /// Specifies events which the <see cref="OpenIdConnectAuthenticationMiddleware"></see> invokes to enable developer control over the authentication process. />
```
Praburaj: Remove the ending ` />` here.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -0,0 +1,587 @@
...
+                    ProtocolMessage = openIdConnectMessage
+                };
+
+                await Options.Notifications.MessageReceived(messageReceivedNotification);
+                if (messageReceivedNotification.HandledResponse)
+                {
+                    return messageReceivedNotification.AuthenticationTicket;
+                }
+
+				if (messageReceivedNotification.Skipped)
```
Praburaj: format the file.

src/Microsoft.AspNet.Security/AuthenticationTicket.cs
```diff
@@ -1,14 +1,8 @@
...
 
-
-using System;
-using System.Collections.Generic;
-using System.Security.Claims;
 using Microsoft.AspNet.Http.Security;
-using Microsoft.AspNet.HttpFeature.Security;
-using Microsoft.AspNet.PipelineCore.Security;
-using Microsoft.AspNet.Security.Infrastructure;
+using System.Security.Claims;
```
Praburaj: Sort namespaces.

src/Microsoft.AspNet.Security/Notifications/AuthenticationFailedNotification.cs
```diff
@@ -2,18 +2,19 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 
+using Microsoft.AspNet.Http;
```
Praburaj: sort namespaces.

test/Microsoft.AspNet.Security.Test/OAuthBearer/OAuthBearerMiddlewareTests.cs
```diff
@@ -0,0 +1,245 @@
...
+                if (configureOptions != null)
+                {
+                    app.UseOAuthBearerAuthentication(configureOptions);
+                }
+                app.Use(async (context, next) =>
+                {
+                    var req = context.Request;
+                    var res = context.Response;
+                    if (req.Path == new PathString("/oauth"))
+                    {
```
Praburaj: Call `await next()` if `req.Path != new PathString("/oauth")`?

test/Microsoft.AspNet.Security.Test/project.json
Praburaj: Oh.. I thought of saying System.Security.Claims be moved to parent dependencies node. That should work on both clrs as desktop version will type forward.

samples/OpenIDConnectSample/Startup.cs
PinpointTownes: And the sample is working as expected?

If it is, that's really sad: IMO, having multiple authentication handlers configured with the same `AuthenticationType` is really bad and should trigger an exception somewhere in `Microsoft.AspNet.Security` as it can't be a valid scenario (here, the `AuthenticationType` is shared with the OIDC middleware).

IIRC, that was not the case with Katana 3, but an `InvalidOperationException` thrown by LINQ was thrown when you called `AuthenticateAsync` with an `AuthenticationType` shared my multiple handlers.

If @Tratcher agrees with the general concept, I suggest opening a new ticket to track that.

samples/OpenIDConnectSample/Startup.cs
PinpointTownes: I think it's time to have a check in all the providers that use a `SignInAsAuthenticationType` property to make sure that it can't be that same as the `AuthenticationType`: an authentication handler shouldn't call itself to delegate the persistence of the authentication ticket. This kind of cyclic call is definitely not a valid scenario.

@Tratcher your thoughts?

src/Microsoft.AspNet.Security/AuthenticationTicket.cs
PinpointTownes: +1 on killing this property: an authentication ticket should be bound to a single identity.

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationHandler.cs
PinpointTownes: He's probably asking for an `ApplyChallenge` notification: https://github.com/aspnet/Security/pull/112/files#diff-f213bf7e29af9267faad33b69cea4cd2L117

samples/OpenIDConnectSample/Startup.cs
PinpointTownes: @brentschmaltz @Tratcher all these changes are great and we definitely need them (the OIDC configuration discovery is just amazing). But we also need a basic token validation middleware that behaves exactly like the existing OAuth2 bearer middleware and allows full pluggability without relying on complicated and OIDC-specific things. Here's my suggestions:

* Rename the existing OAuth2 bearer middleware to `Microsoft.AspNet.Security.Tokens.TokenAuthenticationMiddleware`: there's definitely nothing OAuth2/OIDC-specific with `OAuthBearerAuthenticationMiddleware` in its current form (almost identical to Katana 3) and this led to a terrible confusion. In Average Joe's mind, it should be clear that this middleware can be used with tokens issued by non-OAuth2 authorization servers: for instance, you must be able to plug in a `ISecureDataFormat` that relies on the data protection stuff or that communicates with a remote server to validate an opaque token that cannot be verified locally. That's also the place where I'd create the `ISecureDataFormat` -> `SecurityTokenHandler` adapter.

* Move your new OIDC-stuff to a new package and a new middleware subclassing `TokenAuthenticationMiddleware`... what about `Microsoft.AspNet.Security.Tokens.OpenIdConnect`?

brentschmaltz: To add some clarify, the code written in the tests under: OAuthBearerMiddlewareTests don't use metadata.  Inside any three of the notifications (message received, token received, token validated) you can take over.

test/Microsoft.AspNet.Security.Test/OAuthBearer/OAuthBearerMiddlewareTests.cs
```diff
@@ -0,0 +1,245 @@
...
+                });
+                app.UseCookieAuthentication(options =>
+                {
+                    options.AuthenticationType = "Bearer";
+                });
+                if (configureOptions != null)
+                {
+                    app.UseOAuthBearerAuthentication(configureOptions);
+                }
+                app.Use(async (context, next) =>
```
Tratcher: I don't understand this test, what does it validate? `next()` will just return 404 and any request to `/oauth` will return 200. Nothing checks the user identity, issues 401s, etc..

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationMiddleware.cs
```diff
@@ -0,0 +1,117 @@
...
+                        }
+
+                        Options.MetadataAddress += ".well-known/openid-configuration";
+                    }
+
+                    HttpClient httpClient = new HttpClient(ResolveHttpMessageHandler(Options));
+                    httpClient.Timeout = Options.BackchannelTimeout;
+                    httpClient.MaxResponseContentBufferSize = 1024 * 1024 * 10; // 10 MB
+
+                    Options.ConfigurationManager = new ConfigurationManager<OpenIdConnectConfiguration>(Options.MetadataAddress, httpClient);
```
Tratcher: Wait, why does Bearer default to OIDC metadata?

PinpointTownes: This kind of use will definitely be a niche market (I'd bet that opaque tokens will at least be used in 95% of the real world scenarios), so it shouldn't be enabled by default (see my remark concerning the split of this middleware).

@brentschmaltz but more importantly, do you expect the access token to be signed using the same certificate or RSA key exposed on the JWKS endpoint and used to verify the `id_token`? For obvious reasons, both shouldn't be signed using the same certificate (or your bearer middleware will end up accepting `id_token`s as valid `access_token`s).

I'll ping you on JabbR.

src/Microsoft.AspNet.Security.OAuth/OAuthAuthenticationMiddleware.cs
```diff
@@ -45,18 +45,22 @@ public class OAuthAuthenticationMiddleware<TOptions, TNotifications> : Authentic
             {
                 throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, "AuthenticationType"));
```
PinpointTownes: `nameof`?

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationMiddleware.cs
Tratcher: I talked to Brent offline and understand how this works now. Note you have to specify Authority or MetadataAddress to even reach this point, both of which are optional and have no defaults. for the opaque token scenario you'd hook into the TVP or the notifications and just deal with it. We may find ways to clean up this escalation between scenarios, but I think this is good enough for the initial checkin.

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationMiddleware.cs
```diff
@@ -0,0 +1,117 @@
...
+                Options.TokenValidationParameters.ValidAudience = Options.Audience;
+            }
+
+            if (Options.ConfigurationManager == null)
+            {
+                if (Options.Configuration != null)
+                {
+                    Options.ConfigurationManager = new StaticConfigurationManager<OpenIdConnectConfiguration>(Options.Configuration);
+                }
+                else if (!(string.IsNullOrWhiteSpace(Options.MetadataAddress) && string.IsNullOrWhiteSpace(Options.Authority)))
```
PinpointTownes: This condition is a bit hard to read.

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationMiddleware.cs
PinpointTownes: This seems far more complicated than simply providing an `ISecureDataFormat` or just relying on the default one (that uses the data protection stuff).

And please keep in mind that the access token is usually opaque: http://tools.ietf.org/html/rfc6749#section-1.4

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationOptions.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// Used to know what the current clock time is when calculating or validating token expiration. When not assigned default is based on
+        /// DateTimeOffset.UtcNow. This is typically needed only for unit testing.
+        /// </summary>
+        public ISystemClock SystemClock { get; set; }
+
+        /// <summary>
+        /// Gets or sets the <see cref="SecurityTokenValidators"/> for validating tokens.
+        /// </summary>
+        /// <exception cref="ArgumentNullException">if 'value' is null.</exception>
+        public ICollection<ISecurityTokenValidator> SecurityTokenValidators
```
PinpointTownes: It's better when collections have no public setter (see my remark on the other PR)

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationOptions.cs
```diff
@@ -0,0 +1,159 @@
...
+        public TokenValidationParameters TokenValidationParameters
+        {
+            get
+            {
+                return _tokenValidationParameters;
+            }
+
+            set
+            {
+                if (value == null)
```
PinpointTownes: The other parameters are guarded against null or invalid values in the middleware constructor. It doesn't feel very consistent to check it here.

src/Microsoft.AspNet.Security.OpenIDConnect/Notifications/AuthorizationCodeReceivedNotification.cs
```diff
@@ -0,0 +1,45 @@
...
+        { 
+        }
+
+        /// <summary>
+        /// Gets or sets the 'code'.
+        /// </summary>
+        public string Code { get; set; }
+
+        /// <summary>
+        /// Gets or sets the <see cref="JwtSecurityToken"/> that was received in the id_token + code OpenIdConnectRequest.
```
PinpointTownes: This comment is not really clear. Also, you should rename it to make sure people will understand it's the security token corresponding to the `id_token` and not a JWT access token.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationMiddleware.cs
```diff
@@ -0,0 +1,173 @@
...
+
+            if (Options.StringDataFormat == null)
+            {
+                var dataProtector = dataProtectionProvider.CreateDataProtector(
+                    typeof(OpenIdConnectAuthenticationMiddleware).FullName,
+                    typeof(string).FullName,
+                    Options.AuthenticationType,
+                    "v1");
+
+                Options.StringDataFormat = new SecureDataFormat<string>(new StringSerializer(), dataProtector, TextEncodings.Base64Url);
```
PinpointTownes: Why not `NonceDataFormat`?

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationMiddleware.cs
```diff
@@ -0,0 +1,173 @@
...
+            }
+
+            if (Options.SecurityTokenValidators == null)
```
PinpointTownes: Allowing null values on `IEnumerable` is generally a bad practice.

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationHandler.cs
Tratcher: I see what you mean. It could use a little more polish on the flow in that scenario. You would certainly use the MessageReceived notification, but in order to have the stray token processed normally you'd have to copy it to the Authorization header. Non-blocking, but needs a little more thought. Perhaps the messageReceivedNotification could optionally give back a property with the token.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationMiddleware.cs
```diff
@@ -0,0 +1,173 @@
...
+            }
+
+            if (Options.Notifications == null)
+            {
+                Options.Notifications = new OpenIdConnectAuthenticationNotifications();
+            }
+
+            if (string.IsNullOrWhiteSpace(Options.TokenValidationParameters.ValidAudience) && !string.IsNullOrWhiteSpace(Options.ClientId))
+            {
+                Options.TokenValidationParameters.ValidAudience = Options.ClientId;
```
PinpointTownes: Hum... so you expect that a resource server will be a client?

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationHandler.cs
PinpointTownes: Why not reusing the Katana 3 logic: extracting the access token from the `Authorization` header, triggering the notification, allowing the end dev to change it using another location (e.g from the query string) and guarding the token against null values?

NuGet.Config
brentschmaltz: I thought about that, but am hoping I can get a core50, 50 and 4.5x version together.

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationHandler.cs
brentschmaltz: @Praburaj: you don't have to add the token to the header, the message received notification doesn't assume any token or header.
@Tratcher: I like the idea of returning a token.
@PinpointTownes I wanted to leave options completely open. token may or may not be in a header we understand. that is why I choose 3 notifications.

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationHandler.cs
```diff
@@ -0,0 +1,200 @@
...
+                {
+                    return messageReceivedNotification.AuthenticationTicket;
+                }
+
+                if (messageReceivedNotification.Skipped)
+                {
+                    return null;
+                }
+
+                string authorization = Request.Headers.Get("Authorization");
```
Praburaj: There is going to be a shortly typed header for this shortly. @Tratcher

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationMiddleware.cs
brentschmaltz: I made changes so you don't have to specify or use Authority, MetadataAddress or ConfigManager.
It's not the same or more complicate than ISecureDataFormat, but it is the same as WsFed, OpenIdConnect. For many customers who are receiving AccessTokens from a known source or the same as WsFed,OIDC this is a big win.

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationHandler.cs
PinpointTownes: @brentschmaltz sure, but even if you manually set the token in the first notification...
* The rest of your code always uses `token` and not the property attached to the context.
* The value is always extracted/replaced by the Authorization header in line 65.

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationOptions.cs
brentschmaltz: I agree, I started that change but it requires a bit more work that I wanted to do for this pull. For example, we have to scan the list looking for types.

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationOptions.cs
brentschmaltz: for the validators, I think returning a collection is the answer.  for TVP, I am not so sure I ever want it null.

PinpointTownes: Then why not removing the setter and giving it a default value?

```csharp
public TokenValidationParameters TokenValidationParameters { get; } = new TokenValidationParameters();
```

PinpointTownes: This would also fix the issue I reported in the other PR (when you set `SignInAsAuthenticationType` before TVP)

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationMiddleware.cs
brentschmaltz: I am planning to use it for the 'state' parameter.  Currently we wrap the 'state' inside an AuthenticationProperties structure. This add size and complications when fishing out specific items.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationMiddleware.cs
brentschmaltz: agree as above, planning on making the change to return an empty collection. Just ran out of time for this set of changes.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationNotifications.cs
brentschmaltz: agreed.

src/Microsoft.AspNet.Security/AuthenticationTicket.cs
brentschmaltz: sure

src/Microsoft.AspNet.Security/AuthenticationTicket.cs
brentschmaltz: nope. it should not be bound to a single identity. SecurityTokenHandler returns a ClaimsPrincipal.  We shouldn't drop that. We need changes in other places in the stack so custom identities and principals can exist and be serialized. That is a task for post Beta3.

samples/OpenIDConnectSample/Startup.cs
brentschmaltz: I am actually trying to get rid of SignInAsAuthenticationType. Wouldn't the existence of an authentication ticket be sufficient?

src/Microsoft.AspNet.Security.OpenIDConnect/Notifications/AuthorizationCodeReceivedNotification.cs
brentschmaltz: the comment includes the phrase "code + id_token".  That discounts AccessToken.
what would you like it to say?

test/Microsoft.AspNet.Security.Test/OAuthBearer/OAuthBearerMiddlewareTests.cs
brentschmaltz: Tests plugging in a custom token validator, for say a SAML token.

test/Microsoft.AspNet.Security.Test/project.json
brentschmaltz: I wasn't sure that was in place yet.  I will take that on post beta3.

samples/OpenIDConnectSample/Startup.cs
PinpointTownes: Huh? Why would you remove it?

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationMiddleware.cs
PinpointTownes: Then we need a sample to demonstrate how you can configure the OAuth2 bearer middleware to use the data protector like in Katana 3: https://github.com/AspNet-OpenIdConnect-Server/Owin.Security.OpenIdConnect.Server/blob/vNext/samples/Mvc/Mvc.Server/Startup.cs#L69

src/Microsoft.AspNet.Security.OpenIDConnect/Notifications/AuthorizationCodeReceivedNotification.cs
PinpointTownes: But this property will also be filled if I use the `code id_token token` `response_type`, so no, this doesn't discount `token`.

Why not simply replacing it by:

```
Gets or sets the <see cref="JwtSecurityToken"/> corresponding to the id_token parameter of the OpenID Connect request.
```

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationMiddleware.cs
PinpointTownes: Why not (as long as we're able to easily add custom item to the `state` dictionary)

src/Microsoft.AspNet.Security.OpenIDConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -0,0 +1,587 @@
...
+            ApplyResponseChallengeAsync().GetAwaiter().GetResult();
+        }
+
+        /// <summary>
+        /// Responds to a 401 Challenge sends an OpenIdConnect message to the 'identity authority' to obtain an identity.
+        /// </summary>
+        /// <returns></returns>
+        protected override async Task ApplyResponseChallengeAsync()
+        {
+            if ((Response.StatusCode != 401) || (ChallengeContext == null))
```
PinpointTownes: Like parenthesis? :smile:

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -0,0 +1,337 @@
...
+        public string SignInAsAuthenticationType
+        {
+            get { return TokenValidationParameters.AuthenticationType; }
+            set { TokenValidationParameters.AuthenticationType = value; }
+        }
+
+        /// <summary>
+        /// Gets or sets the type used to secure data handled by the middleware.
+        /// </summary>
+        public ISecureDataFormat<AuthenticationProperties> StateDataFormat
```
PinpointTownes: You should make it an auto-property and let the middleware handle the null case.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -0,0 +1,337 @@
...
+                }
+
+                _stateDataFormat = value;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the type used to secure strings used by the middleware.
+         // </summary>
+        public ISecureDataFormat<string> StringDataFormat
```
PinpointTownes: Same remark.

src/Microsoft.AspNet.Security/AuthenticationTicket.cs
PinpointTownes: In .NET 4.5, `ValidateToken` returns a list of identities, but in the different implementations, it's always composed of a single identity. Is returning a list of identities (breaking the idea that a token = an identity) really useful?

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationMiddleware.cs
PinpointTownes: Oooops, it seems my previous comment disappeared:

> @brentschmaltz but more importantly, do you expect the access token to be signed using the same certificate or RSA key exposed on the JWKS endpoint and used to verify the id_token? For obvious reasons, both shouldn't be signed using the same certificate (or your bearer middleware will end up accepting id_tokens as valid access_tokens).

src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationHandler.cs
Tratcher: No, we decided the Auth header was non-interesting for strong types.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenidConnectAuthenticationHandler.cs
Tratcher: Why is the buffering still showing up here?

samples/OpenIDConnectSample/Startup.cs
Tratcher: We added some validation in the new stack for things like challenging for a missing auth type, but there's not currently any checks for duplicates. Duplicates are not supported and will cause strange side-effects with APIs like SignIn and Challenge.

The correct config is:
ExternalAuthenticationOptions - Use CookieAuthenticationDefaults.AuthenticationType
UseCookieAuthentication - Don't set anything, let it use the default
UseOpenIdConnectAuthentication - Don't set SignInAsAuthenticationType or AuthenticationType, let it use the defaults

src/Microsoft.AspNet.Security/AuthenticationTicket.cs
Tratcher: FYI: I see you've handled this in InvokeReplyPath, but it's not handled in AuthenticationHandler.BaseInitializeAsync or Authenticate. Because of the flow of this speific OIDC middleware most people won't hit those code paths, but long term that's a good example of why we shouldn't have two different identity properties on the ticket.

src/Microsoft.AspNet.Security.OpenIDConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: That's because I didn't know about that new feature.  I thought you were speaking to just using the async.

samples/OpenIDConnectSample/Startup.cs
brentschmaltz: Conceptually this drives the ClaimsIdentity AuthenticationType, that shouldn't be linked.

PinpointTownes: Not really: `SignInAsAuthenticationType` is the way an authentication handler delegates to another one the persistence of the resulting identity.

An example should be clearer: when the Google middleware creates its own identity using the profile endpoint, it calls the authentication handler corresponding to `SignInAsAuthenticationType` - a cookie handler in most cases - and offers it the opportunity to store the identity using its own `AuthenticationType`.

Tratcher:
Merged.

Issue #128, opened 2015-01-26
OpenIdConnectAuthenticationMiddleware initialization issue


Desislav:
Hello,

I was looking at the OpenIdConnectAuthenticationMiddleware and i found some strange lines of code inside its constructor: 

```C#
else if (!(string.IsNullOrWhiteSpace(Options.MetadataAddress) && string.IsNullOrWhiteSpace(Options.Authority)))
{
    //If the previous if statement is true how this condition can be ever satisfied??
   if (string.IsNullOrWhiteSpace(Options.MetadataAddress) && !string.IsNullOrWhiteSpace(Options.Authority))
   {
       Options.MetadataAddress = Options.Authority;
       if (!Options.MetadataAddress.EndsWith("/", StringComparison.Ordinal))
       {
             Options.MetadataAddress += "/";
       }
       Options.MetadataAddress += ".well-known/openid-configuration";
   }

   HttpClient httpClient = new HttpClient(ResolveHttpMessageHandler(Options));
   httpClient.Timeout = Options.BackchannelTimeout;
   httpClient.MaxResponseContentBufferSize = 1024 * 1024 * 10; // 10 MB
   Options.ConfigurationManager = new ConfigurationManager<OpenIdConnectConfiguration>(Options.MetadataAddress, httpClient);
 }
```

Tratcher:
I agree it's not the clearest logic, but I think it's technically correct. Let me try writing it in English to see if it makes any more sense.

First:
If the user has specified MetadataAddress and/or Authority
And then:
If the user did not specify metadata address, but did specify Authority

So specifying MetadataAddress gets you into the first if but skips the second one, where specifying Authority gets you into both.

The first statement could also be written as:
```C#
else if (!string.IsNullOrWhiteSpace(Options.MetadataAddress) || !string.IsNullOrWhiteSpace(Options.Authority))
```
I don't think the second if can be simplified.

Desislav:
Oh i missed the first bracket after the "!" operator. Sorry about that.  Actually your second aproach is what i had in mind.

Anyway what i meant is that the second aproach that you wrote is way more better in terms of readability.

PinpointTownes:
Mouahahaha, I told you this condition was too hard to read: https://github.com/aspnet/Security/pull/112#discussion_r22963831 :laughing:

PR #129, opened 2015-01-26
OpneIdConnectMiddleWare ConfirationManager initialization - issue #128

Desislav:
Fixing the Options.ConfigurationManager initialization inside the OpeIdConnectAuthenticationMiddleWare constructor. The condition:

```C#
if (string.IsNullOrWhiteSpace(Options.MetadataAddress) && !string.IsNullOrWhiteSpace(Options.Authority))
```
inside the middle ware constructor will never been satisfied so it must be fixed.


Tratcher:
The new logic is incorrect. See my comments on the original bug.

Issue #132, opened 2015-01-26
[Exploratory Testing] Add Operation based extension methods/Overloads for the Authorization service and Options.


harshgMSFT:
Today there is no simple way of registering requirements based on operations and then authorizing based on whether a particular user satisfies the requirement for an operation on a resource.
For instance, an admin(The identity) is allowed to edit (The operation based requirement) the price of an album (The resource).
The suggested experience should be something like : 
This will add a policy called ApiOperations with an edit requirement on the Album resource.
```c#
options.AddPolicy("ApiOperations", policyBuilder => policyBuilder.AddRequirement<Album>("Edit"));
OR
options.AddPolicy("ApiOperations", policyBuilder => policyBuilder.AddRequirement<Album>(new OperationRequirement { OpeartionKey = "Edit" }));
```
On calling Authorize on the Authorization service like so
```c#
_authorizationService.AuthorizeAsync("ApiOperations", context, album);
```
which finally calls into handler which has the following method
```c#
 public override Task<bool> CheckAsync<TResource, TRequirement>(AuthorizationContext context, TRequirement requirement)
        {     
        }
```
This item needs more thought and design. Some of the other thoughts were to skip creation of the policy altogether and create it as an anonymous policy.

HaoK:
I chatted with @lodejard and here's what where we ended up for current proposal for the next iteration:

cc @blowdart @yishaigalatzer 

Updated service interface:
```
    public interface IAuthorizationService
        Task<bool> AuthorizeAsync(ClaimsPrincipal user, object resource, params IAuthorizationRequirement[] requirements);

        Task<bool> AuthorizeAsync(ClaimsPrincipal user, object resource, string policyName);

        public static Task<bool> AuthorizeAsync([NotNull] this IAuthorizationService service, ClaimsPrincipal user, object resource, [NotNull] AuthorizationPolicy policy)
```

Example of what operations would look like
```
public static class Operations {
    public static OperationRequirement Edit = new OperationRequirement("Edit");
    public static OperationRequirement Create = new OperationRequirement("Create");
    public static OperationRequirement Delete = new OperationRequirement("Delete");
}

public class ExpenseReportAuthorizationHandler : AuthorizationHandler<OperationRequirement, ExpenseReport> {
    public void HandleAsync(AuthorizationContext context, OperationRequirement req, ExpenseReport resource) 
    {
        if (Repo.CanDo(req.Name, resource)) {
            context.Succeed(req);
        }
    }
}

[...]

Issue #11, opened 2014-05-01
Some methods in IAuthorizationService should be moved to extension methods


victorhurdugaci:
As @Eilon suggested, the overloads without a resource should be extension methods rather than methods on the interface. In most of the cases, whoever implements this service will just call the overload with the resource parameter with a "null" value.

~~~C#
namespace Microsoft.AspNet.Security.Authorization
{
    public interface IAuthorizationService
    {
        bool Authorize(IEnumerable<Claim> claims, ClaimsPrincipal user);
        bool Authorize(IEnumerable<Claim> claims, ClaimsPrincipal user, object resource);
        Task<bool> AuthorizeAsync(IEnumerable<Claim> claims, ClaimsPrincipal user);
        Task<bool> AuthorizeAsync(IEnumerable<Claim> claims, ClaimsPrincipal user, object resource);
    }
}
~~~

HaoK:
While this was talking about the old iteration of Auth service, it still applies to the new overloads we are going to add, so I'll keep this bug around until the next iteration moves some of the new methods to extension methods

PR #135, opened 2015-01-27
Exploratory testing feed back iteration

HaoK:
- Core Auth API now takes list of IAuthorizationRequirements, or policy name.  
- Overload that takes AuthorizationPolicy instance moved to extension method.
- Remove HttpContext from API and replace with ClaimsPrincipal instead
- Add Operation requirement

Fixes https://github.com/aspnet/Security/issues/132
Fixes https://github.com/aspnet/Security/issues/116
Fixes https://github.com/aspnet/Security/issues/11

cc @lodejard @blowdart @yishaigalatzer @harshgMSFT


HaoK:
Added sync overloads, and switched default async implementations to just call the sync version by default.  Only the sync version of Handle is abstract

HaoK:
Fixes https://github.com/aspnet/Security/issues/117

HaoK:
Replaced by https://github.com/aspnet/Security/pull/151

PR #137, opened 2015-01-27
ChallengeContext will be null with [Authorize] attribute

brentschmaltz:
OpenIdConnect set Ticket.Principal, get identity from there.

### DIFF
--- a/src/Microsoft.AspNet.Security.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Security.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
@@ -91,6 +91,7 @@ namespace Microsoft.AspNet.Security.OpenIdConnect
                 {
                     ProtocolMessage = openIdConnectMessage
                 };
+
                 await Options.Notifications.RedirectToIdentityProvider(notification);
 
                 if (!notification.HandledResponse)
@@ -100,6 +101,7 @@ namespace Microsoft.AspNet.Security.OpenIdConnect
                     {
                         _logger.WriteWarning("The logout redirect URI is malformed: " + redirectUri);
                     }
+
                     Response.Redirect(redirectUri);
                 }
             }
@@ -116,7 +118,13 @@ namespace Microsoft.AspNet.Security.OpenIdConnect
         /// <returns></returns>
         protected override async Task ApplyResponseChallengeAsync()
         {
-            if ((Response.StatusCode != 401) || (ChallengeContext == null))
+            if (Response.StatusCode != 401)
+            {
+                return;
+            }
+
+            // Active middleware should redirect on 401 even if there wasn't an explicit challenge.
+            if (ChallengeContext == null && Options.AuthenticationMode == AuthenticationMode.Passive)
             {
                 return;
             }
@@ -124,7 +132,16 @@ namespace Microsoft.AspNet.Security.OpenIdConnect
             // order for redirect_uri
             // 1. challenge.Properties.RedirectUri
             // 2. CurrentUri
-            AuthenticationProperties properties = new AuthenticationProperties(ChallengeContext.Properties);
+            AuthenticationProperties properties;
+            if (ChallengeContext == null)
+            {
+                properties = new AuthenticationProperties();
+            }
+            else
+            {
+                properties = new AuthenticationProperties(ChallengeContext.Properties);
+            }
+
             if (string.IsNullOrEmpty(properties.RedirectUri))
             {
                 properties.RedirectUri = CurrentUri;
@@ -154,7 +171,6 @@ namespace Microsoft.AspNet.Security.OpenIdConnect
                 State = OpenIdConnectAuthenticationDefaults.AuthenticationPropertiesKey + "=" + Uri.EscapeDataString(Options.StateDataFormat.Protect(properties))
             };
 
-            // TODO - brentschmaltz, if INonceCache is set should we even consider if ProtocolValidator is set?
             if (Options.ProtocolValidator.RequireNonce)
             {
                 openIdConnectMessage.Nonce = Options.ProtocolValidator.GenerateNonce();
@@ -179,7 +195,7 @@ namespace Microsoft.AspNet.Security.OpenIdConnect
                 string redirectUri = notification.ProtocolMessage.CreateAuthenticationRequestUrl();
                 if (!Uri.IsWellFormedUriString(redirectUri, UriKind.Absolute))
                 {
-                    _logger.WriteWarning("The authenticate redirect URI is malformed: " + redirectUri);
+                    _logger.WriteWarning("Uri.IsWellFormedUriString(redirectUri, UriKind.Absolute) returned 'false', redirectUri is: " + (redirectUri ?? "null"));
                 }
 
                 Response.Redirect(redirectUri);
--- a/src/Microsoft.AspNet.Security.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Security.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
@@ -99,7 +99,7 @@ namespace Microsoft.AspNet.Security.OpenIdConnect
                     string redirectUri = notification.ProtocolMessage.CreateLogoutRequestUrl();
                     if (!Uri.IsWellFormedUriString(redirectUri, UriKind.Absolute))
                     {
-                        _logger.WriteWarning("The logout redirect URI is malformed: " + redirectUri);
+                        _logger.WriteWarning("The logout redirect URI is malformed: {0}", (redirectUri ?? "null"));
                     }
 
                     Response.Redirect(redirectUri);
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Security/Infrastructure/AuthenticationHandler.cs
```diff
@@ -77,9 +77,12 @@ internal AuthenticationOptions BaseOptions
             if (BaseOptions.AuthenticationMode == AuthenticationMode.Active)
             {
                 AuthenticationTicket ticket = await AuthenticateAsync();
-                if (ticket != null && ticket.Identity != null)
+                if (ticket != null)
                 {
-                    SecurityHelper.AddUserIdentity(Context, ticket.Identity);
+                    if ( ticket.Identity != null)
+                        SecurityHelper.AddUserIdentity(Context, ticket.Identity);
```
Tratcher: I think this duplication needs to be resolved inside of AuthTicket, we don't want every consumer to have to check both properties.

Tratcher:
@Eilon @muratg Triage for beta3. OIDC does not currently work with the Authorize attribute because the attribute only sets 401, it does not call Challenge.

src/Microsoft.AspNet.Security.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -179,7 +195,7 @@ protected override void ApplyResponseChallenge()
                 string redirectUri = notification.ProtocolMessage.CreateAuthenticationRequestUrl();
                 if (!Uri.IsWellFormedUriString(redirectUri, UriKind.Absolute))
                 {
-                    _logger.WriteWarning("The authenticate redirect URI is malformed: " + redirectUri);
+                    _logger.WriteWarning("Uri.IsWellFormedUriString(redirectUri, UriKind.Absolute) returned 'false', redirectUri is: " + (redirectUri ?? "null"));
```
Praburaj: use `_logger.WriteWarning("{0}", redirectUri ?? "null")` overload of logger to substitute parameters.

src/Microsoft.AspNet.Security/Infrastructure/AuthenticationHandler.cs
```diff
@@ -77,9 +77,12 @@ internal AuthenticationOptions BaseOptions
             if (BaseOptions.AuthenticationMode == AuthenticationMode.Active)
             {
                 AuthenticationTicket ticket = await AuthenticateAsync();
-                if (ticket != null && ticket.Identity != null)
+                if (ticket != null)
                 {
-                    SecurityHelper.AddUserIdentity(Context, ticket.Identity);
+                    if ( ticket.Identity != null)
```
Praburaj: Format the file.

src/Microsoft.AspNet.Security.OpenIdConnect/Resources.resx
```diff
@@ -0,0 +1,132 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<root>
```
Praburaj: :+1:

src/Microsoft.AspNet.Security.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
Praburaj: We write the same warning in `ApplyResponseGrantAsync` method in the same file.

src/Microsoft.AspNet.Security.OpenIdConnect/Resources.resx
```diff
@@ -0,0 +1,132 @@
...
+    <value>BackchannelTimeout cannot be less or equal to TimeSpan.Zero.</value>
+  </data>
+  <data name="Exception_OpenIdConnectMessageError" xml:space="preserve">
+    <value>"OpenIdConnectMessage.Error was not null, indicating an error. Error: '{0}'. Error_Description (may be empty): '{1}'. Error_Uri (may be empty): '{2}'."</value>
+  </data>
+  <data name="Exception_RedirectUri_LogoutQueryString_IsNotWellFormed" xml:space="preserve">
+    <value>OIDC_20001: The query string for Logout is not a well formed URI. The runtime cannot redirect. Redirect uri: '{0}'.</value>
+  </data>
+  <data name="Exception_ValidatorHandlerMismatch" xml:space="preserve">
+    <value>An ICertificateValidator cannot be specified at the same time as an HttpMessageHandler unless it is a WebRequestHandler.</value>
```
Eilon: String resources should generally be kept to localizable content only. So, type names should be passed in programmatically through `string.Format()`, ideally using `nameof(...)`.

src/Microsoft.AspNet.Security/Infrastructure/AuthenticationHandler.cs
brentschmaltz: OIDC sets the Principal property.
Since changes w.r.t. that property are still in the early development, I agree to change OIDC until we have a comprehensive fix.

src/Microsoft.AspNet.Security/Infrastructure/AuthenticationHandler.cs
brentschmaltz: See comment to Chris above, we will defer this change for now.

src/Microsoft.AspNet.Security.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: Fixed both.

src/Microsoft.AspNet.Security.OpenIdConnect/Resources.resx
brentschmaltz: I think are asking for ICertificateValidator to be nameof(ICertificateValidator), but I am not sure.
Is there an example of the nameof(...) I can reference?

Praburaj: an example here : https://github.com/aspnet/FileSystem/blob/56c87db7deaf657bde5bb6a7f3b16034aaf3cf28/src/Microsoft.AspNet.FileProviders/EmbeddedFileProvider.cs#L201

```
                throw new InvalidOperationException(string.Format("{0} does not support {1}.", nameof(EmbeddedFileProvider), nameof(WriteContent)));

```

Eilon: Yeah the idea is that instead of hard-coding a member name, you can programmatically pass it in. This makes localization easier (because localizers never see the unlocalizable text), as well as ensures that if the type name / member name changes we'll get a compiler error.

Here's an example:
https://github.com/aspnet/Mvc/blob/12c2759cec61674eb00f60a66a4df73b1af0e309/src/Microsoft.AspNet.Mvc.Core/Filters/ControllerActionFilter.cs#L27-L29

That example happens to use our resource formatting helpers where we code-gen a helper method that calls `string.Format()` for you.

src/Microsoft.AspNet.Security/Infrastructure/AuthenticationHandler.cs
Praburaj: Should this file be excluded from the PR?

src/Microsoft.AspNet.Security.OpenIdConnect/Resources.resx
```diff
@@ -0,0 +1,132 @@
...
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="ArgsException_BackchallelLessThanZero" xml:space="preserve">
+    <value>BackchannelTimeout cannot be less or equal to TimeSpan.Zero.</value>
+  </data>
+  <data name="Exception_OpenIdConnectMessageError" xml:space="preserve">
+    <value>"OpenIdConnectMessage.Error was not null, indicating an error. Error: '{0}'. Error_Description (may be empty): '{1}'. Error_Uri (may be empty): '{2}'."</value>
+  </data>
+  <data name="Exception_RedirectUri_LogoutQueryString_IsNotWellFormed" xml:space="preserve">
+    <value>OIDC_20001: The query string for Logout is not a well formed URI. The runtime cannot redirect. Redirect uri: '{0}'.</value>
```
blowdart: Niggle - why are we embedding error codes again? This isn't 1995 any more. If an error message is not explicit enough to make sense without having an error code then the message is no good.

Eilon: Oh I assumed that was something standard. If it's just something we made up then please remove it.

src/Microsoft.AspNet.Security/Infrastructure/AuthenticationHandler.cs
blowdart: Yea, I just noticed this. No no no no no. You can't just slap a Principal and an Identity property to AuthTicket. This needs reverting before Beta 3 please.

Praburaj: @blowdart - ClaimsPrincipal was introduced in `AuthenticationTicket` with OIDC - https://github.com/aspnet/Security/commit/49e66f03110b486f8dd9959dd2cee802d676c28f. @blowdart mentioned in a conversation having both `Principal` and `Identity` on `AuthenticationTicket` is going to confuse users. So until we decide on whether it is `Principal` or `Identity` to be in `AuthenticationTicket` he suggested to remove the `Principal` property from `AuthenticationTicket` for `beta3`. 

@brentschmaltz @Eilon @Tratcher

blowdart: @brentschmaltz Is there a reason you needed a full blown principal?

PinpointTownes: @blowdart https://github.com/aspnet/Security/pull/112#discussion_r22967092

PR #138, opened 2015-01-28
Added Spotify authentication

jeroenslor:
Added Spotify (oauth) authentication including test and sample code


Eilon:
Hi @jeroenslor I think the focus for this repo is to just have a few very common authentication providers that are most commonly used on web site. I think a Spotify provider would be great to ship as your own NuGet package.

Tratcher:
There's also a simpler way to implement providers now, see:
https://github.com/aspnet/Security/blob/dev/samples/SocialSample/Startup.cs#L116
50 lines vs 15 files.

jeroenslor:
Hi @Eilon Ok I will create my own nuget, thanks for the quick feedback. Is there any other provider I can help you with that is commonly used, maybe linkedin?

jeroenslor:
@Tratcher yeah my first attempt was to only create an extensions method that hides the code you just showed but it turned out that Spotify had some extra auth options. To support that I had to create a custom options class and override a methode on the handers see: https://github.com/jeroenslor/Security/blob/SpotifyAuth/src/Microsoft.AspNet.Security.Spotify/SpotifyAuthenticationHandler.cs#L72

I do really like the pattern you guys used with the notifications, it's pretty flexibel and avoids extra implementations!

Eilon:
@jeroenslor I don't think there are currently any other auth providers that we are looking to add outside of Active Directory related scenarios, but for those we have works in progress.

Tratcher:
FYI: Here is a related community project for Katana. I hear they plan something similar for ASP.NET 5.
https://github.com/owin-middleware/OwinOAuthProviders/tree/master/Owin.Security.Providers

PR #143, opened 2015-02-02
Changed package versions to fixed ones

kichalla:
@suhasj 

Just FYI...the Xunit packages are not available on public Nuget feed and we are figuring out if we need to create a separate feed for external packages, but other than that please take a look at the changes to other packages.


msotclas:
Hi __@kichalla__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>
        In order for us to evaluate and accept your PR, we ask that you sign a contribution license agreement. It's all electronic and will take just minutes. I promise there's no faxing. https://cla2.msopentech.com.
    </span>

TTYL, MSOTBOT;

suhasj:
Adding @Praburaj

suhasj:
@kichalla why are we updating this ? If add the release versions, we would have to update it for each release right ?

kichalla:
@suhasj : yes, we want to lock on the versions for at least the released packages.

Issue #48, opened 2014-08-25
Define central claims transformation point


Tratcher:

brentschmaltz:
I agree we need a single point to evaluate what the 'user' should look like when all of the AuthenticationMiddleware (s) have completed. Since the ISecurityTokenValidator returns a ClaimsPrincipal and is configurable by the user, we need to maintain the object the user returned.

The AuthenticationHandler returns an AuthenticationTicket, if this is preserved and accessible then this would allow complete extensibility.

HaoK:
Discussed as part of the AuthN changes, going to go with a ClaimsTransformationMiddleware which uses a ClaimsTransformationOptions to:

1: Apply transformation to the HttpContext.User as part of the pipeline
2. Also hooks any calls to HttpContext.AuthenticateAsync to apply transformation before returning the resultant identity

PR #146, opened 2015-02-02
Add a basic claims xformation middleware

HaoK:
Fixes https://github.com/aspnet/Security/issues/48

cc @blowdart @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Security/ClaimsTransformationExtensions.cs
+++ b/src/Microsoft.AspNet.Security/ClaimsTransformationExtensions.cs
@@ -1,10 +1,9 @@
 // Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using Microsoft.Framework.DependencyInjection;
-using Microsoft.Framework.OptionsModel;
 using System;
 using Microsoft.AspNet.Security;
+using Microsoft.Framework.DependencyInjection;
 
 namespace Microsoft.AspNet.Builder
 {
@@ -19,11 +18,10 @@ namespace Microsoft.AspNet.Builder
         }
 
         /// <summary>
-        /// Adds a cookie-based authentication middleware to your web application pipeline.
+        /// Adds a claims transformation middleware to your web application pipeline.
         /// </summary>
         /// <param name="app">The IApplicationBuilder passed to your configuration method</param>
         /// <param name="configureOptions">Used to configure the options for the middleware</param>
-        /// <param name="optionsName">The name of the options class that controls the middleware behavior, null will use the default options</param>
         /// <returns>The original app parameter</returns>
         public static IApplicationBuilder UseClaimsTransformation([NotNull] this IApplicationBuilder app, Action<ClaimsTransformationOptions> configureOptions = null)
         {
--- a/src/Microsoft.AspNet.Security/ClaimsTransformationExtensions.cs
+++ b/src/Microsoft.AspNet.Security/ClaimsTransformationExtensions.cs
@@ -1,0 +1,33 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Framework.DependencyInjection;
+using Microsoft.Framework.OptionsModel;
+using System;
+using Microsoft.AspNet.Security;
+
+namespace Microsoft.AspNet.Builder
+{
+    /// <summary>
+    /// Extension methods provided by the claims transformation middleware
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.Security/ClaimsTransformationMiddleware.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        // REVIEW: should this use auto request services like the other auth middlewares?
+        public async Task Invoke(HttpContext context)
+        {
+
+            if (Options.TransformAsync != null)
+            {
+                context.User = await Options.TransformAsync(context.User);
+            }
+            await _next(context);
```
Tratcher: This won't run for passive auth middleware when they are invoked via `HttpContext.Authenticate(...)`. Do you need to wire into that flow as well?

src/Microsoft.AspNet.Security/ClaimsTransformationMiddleware.cs
```diff
@@ -0,0 +1,40 @@
...
+            }
+        }
+
+        public ClaimsTransformationOptions Options { get; set; }
+
+        // REVIEW: should this use auto request services like the other auth middlewares?
+        public async Task Invoke(HttpContext context)
+        {
+
+            if (Options.TransformAsync != null)
```
Tratcher: Why would this ever be null? Just validate it in the constructor.

src/Microsoft.AspNet.Security/ClaimsTransformationMiddleware.cs
HaoK: Good point maybe this actually should start as part of the Mvc authorize filter for now.  Since that is currently what's building the effective principal for the Mvc request.  

Or alternatively instead of this being a separate middleware. Maybe this just should be a part of authentication options and the auth middleware can use this as part of generating the ticket for an auth type?

Tratcher: Most middleware already have a hook for tweaking the identity. I understood this one was so you could centralize that logic rather than setting it in each middleware.

How well defined are your scenarios? I'd be curious to see a real sample.

HaoK: Yeah centralized logic is key. So something like what we did with external sign in options might work. Define the options for claims transformation so you configure that once. And all the middlewares just consume it. That doesn't feel too bad.

HaoK: So run xformation only once after we have the effective principal but before we check the authorization policy?  As things stand today the only place we could do that then is inside the authorize filter itself. Could just have the authorize filter take claimsTransformationOptions via ctor and call it as part of the new authZ logic?  Basically we would only be defining the registration part with the options in security. And Mvc and signalr would need to apply the xform themselves as appropriate.

leastprivilege: And with authorization you mean authentication, right? ;)

yea +1 - each authN middleware contributes an identity to the principal - claims transformation should then see the final principal. once.

src/Microsoft.AspNet.Security/ClaimsTransformationExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.Framework.DependencyInjection;
+
+namespace Microsoft.AspNet.Builder
+{
+    /// <summary>
+    /// Extension methods provided by the claims transformation middleware
+    /// </summary>
+    public static class ClaimsTransformationExtensions
+    {
+        public static IServiceCollection ConfigureClaimsTransformation([NotNull] this IServiceCollection services, [NotNull] Action<ClaimsTransformationOptions> configure)
```
Praburaj: Doc comment?

src/Microsoft.AspNet.Security/ClaimsTransformationMiddleware.cs
```diff
@@ -0,0 +1,40 @@
...
+                configureOptions(Options);
+            }
+        }
+
+        public ClaimsTransformationOptions Options { get; set; }
+
+        // REVIEW: should this use auto request services like the other auth middlewares?
+        public async Task Invoke(HttpContext context)
+        {
+
```
Praburaj: nit: Delete the empty line.

src/Microsoft.AspNet.Security/ClaimsTransformationMiddleware.cs
blowdart: And once it's done. That's it. Identity cannot be changed, especially within authorize rules (cc @loudej)

HaoK:
Will open a new PR after we discuss what we want to do

PR #155, opened 2015-02-13
Initial prototype of AuthN changes

HaoK:
- AuthenticationType -> AuthenticationScheme
- ClaimsIdentity -> ClaimsPrincipal
- Remove ClaimsIdentity from AuthenticationTicket and serialize the ClaimsPrincipal instead

There are 6 cookie unit tests that are still failing (I must have missed/broke something somewhere):
CookieMiddlewareTests.CookieContainsIdentity is the primary failure

cc @lodejard @Praburaj @Tratcher @blowdart


Tratcher:
@brentschmaltz

samples/OpenIdConnectSample/Startup.cs
```diff
@@ -17,7 +17,7 @@ public void Configure(IApplicationBuilder app)
                 services.AddDataProtection();
                 services.Configure<ExternalAuthenticationOptions>(options =>
                 {
-                    options.SignInAsAuthenticationType = CookieAuthenticationDefaults.AuthenticationType;
+                    options.SignInAsAuthenticationScheme = CookieAuthenticationDefaults.AuthenticationScheme;
```
Tratcher: SignInWithAuthenticationScheme?

samples/SocialSample/Startup.cs
```diff
@@ -136,7 +136,7 @@ k web
                         JObject user = JObject.Parse(text);
 
                         var identity = new ClaimsIdentity(
-                            context.Options.AuthenticationType,
+                            context.Options.AuthenticationScheme,
```
Tratcher: "OAuth2-Github", not the auth scheme.

samples/SocialSample/Startup.cs
```diff
@@ -184,9 +184,9 @@ k web
                     context.Response.ContentType = "text/html";
                     await context.Response.WriteAsync("<html><body>");
                     await context.Response.WriteAsync("Choose an authentication type: <br>");
-                    foreach (var type in context.GetAuthenticationTypes())
+                    foreach (var type in context.GetAuthenticationSchemes())
                     {
-                        await context.Response.WriteAsync("<a href=\"?authtype=" + type.AuthenticationType + "\">" + (type.Caption ?? "(suppressed)") + "</a><br>");
+                        await context.Response.WriteAsync("<a href=\"?authtype=" + type.AuthenticationScheme + "\">" + (type.Caption ?? "(suppressed)") + "</a><br>");
```
Tratcher: authtype -> authscheme

samples/OpenIdConnectSample/Startup.cs
HaoK: SignInScheme?  Much shorter without much loss of clarity imo

src/Microsoft.AspNet.Security/Infrastructure/AuthenticationHandler.cs
```diff
@@ -307,7 +311,18 @@ protected virtual void ApplyResponseCore()
         protected virtual async Task ApplyResponseCoreAsync()
         {
             await ApplyResponseGrantAsync();
-            await ApplyResponseChallengeAsync();
+
+            // If authenticate was called and the the status is still 401, authZ failed so set 403 and stop
+            // REVIEW: Does this need to ensure that there's the 401 is challenge for this auth type?
```
Tratcher: Actually, per our discussion, this should move back into ApplyResponseChallenge because CookieAuth is going to redirect to an AccessDenied page instead.

samples/SocialSample/Startup.cs
HaoK: Should we add a new option like ClaimIssuer to AuthenticationOptions and set defaults appropriately instead of just hardcoding?

src/Microsoft.AspNet.Security/Infrastructure/AuthenticationHandler.cs
HaoK: Yeah I'm going to fix 401-403 that in a separate PR, I'll push the first set of AuthZ changes once the matching MVC PR is signed off.  Then a PR once everything is working again with the renames.  Then finally 401-403 and claims xformation will be their own hopefully smaller PRs.

samples/SocialSample/Startup.cs
Tratcher: Sure, and use it in all of the `new Claim`'s below.

src/Microsoft.AspNet.Security/Infrastructure/AuthenticationHandler.cs
Tratcher: Ok, then cut the changes from this PR.

HaoK: Yeah I generated the PR from the wrong base branch, meant to generate from the AuthZ branch.  Should go away once I push the AuthZ changes and rebase to dev

src/Microsoft.AspNet.Security.Cookies/Notifications/CookieValidateIdentityContext.cs
```diff
@@ -49,7 +45,7 @@ public CookieValidateIdentityContext([NotNull] HttpContext context, [NotNull] Au
         /// <param name="identity">The identity used as the replacement</param>
         public void ReplaceIdentity(IIdentity identity)
```
Tratcher: `ReplacePrincipal(IPrincipal principal)`?

src/Microsoft.AspNet.Security.OAuth/OAuthAuthenticationHandler.cs
```diff
@@ -52,26 +52,21 @@ public OAuthAuthenticationHandler(HttpClient backchannel, ILogger logger)
...
-                SignInAsAuthenticationType = Options.SignInAsAuthenticationType,
+                SignInAsAuthenticationScheme = Options.SignInAsAuthenticationScheme,
                 RedirectUri = ticket.Properties.RedirectUri,
             };
             ticket.Properties.RedirectUri = null;
 
             await Options.Notifications.ReturnEndpoint(context);
 
-            if (context.SignInAsAuthenticationType != null && context.Identity != null)
+            if (context.SignInAsAuthenticationScheme != null && context.Principal != null)
```
Tratcher: We need to confirm if we even allow SignInAsAuthenticationScheme to be null.

HaoK: If its null, we sign in with the current AuthenticationScheme, and anyone who later calls AuthenticateAsync with that scheme gets the principal no?

src/Microsoft.AspNet.Security.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -232,9 +232,9 @@ public OpenIdConnectProtocolValidator ProtocolValidator
         public string Scope { get; set; }
 
         /// <summary>
-        /// Gets or sets the AuthenticationType used when creating the <see cref="System.Security.Claims.ClaimsIdentity"/>.
+        /// Gets or sets the AuthenticationScheme used when creating the <see cref="System.Security.Claims.ClaimsIdentity"/>.
         /// </summary>
-        public string SignInAsAuthenticationType
+        public string SignInAsAuthenticationScheme
         {
             get { return TokenValidationParameters.AuthenticationType; }
```
Tratcher: @brentschmaltz We don't need to do this anymore, right?

samples/SocialSample/Startup.cs
blowdart: But hardcode the one for the "scheme" claim.

src/Microsoft.AspNet.Security/DataHandler/Serializer/TicketSerializer.cs
```diff
@@ -38,18 +38,23 @@ public virtual AuthenticationTicket Deserialize(byte[] data)
...
         {
             writer.Write(FormatVersion);
-            ClaimsIdentity identity = model.Identity;
-            writer.Write(identity.AuthenticationType);
-            WriteWithDefault(writer, identity.NameClaimType, DefaultValues.NameClaimType);
-            WriteWithDefault(writer, identity.RoleClaimType, DefaultValues.RoleClaimType);
-            writer.Write(identity.Claims.Count());
-            foreach (var claim in identity.Claims)
+            writer.Write(model.AuthenticationScheme);
+            var principal = model.Principal;
```
Tratcher: @brentschmaltz you've redone the ClaimsPrincipal serializer so we don't need to do this ourselves anymore, correct?

davidfowl: Don't we still need something?

samples/SocialSample/Startup.cs
HaoK: Need to actually add a scheme claim first :)  Should we always add an identity with authtype = "AspNet.AuthenticationScheme", and a single claim with type = AuthenticationScheme, "schemeName"?

samples/OpenIdConnectSample/Startup.cs
Tratcher: *shrug* I think the verbose version is fine, it won't come up in user code much.

src/Microsoft.AspNet.Security.OAuth/OAuthAuthenticationHandler.cs
Tratcher: Nobody can call AuthenticateAsync later, we're operating on a restricted callback path here and require a redirect to get back to the app.

HaoK: Ok sounds like we should throw/blow up if SignInScheme is null or empty early on then?

src/Microsoft.AspNet.Security/DataHandler/Serializer/TicketSerializer.cs
Tratcher: I think we can replace most of this with Brent's code.

src/Microsoft.AspNet.Security.OAuth/OAuthAuthenticationHandler.cs
Tratcher: It should already be checked in the constructor.

samples/SocialSample/Startup.cs
blowdart: Wait, I though the scheme/source claim was going to go into the identity the middleware constructed?

HaoK: Right now, it is only in the metadata that comes with the principal, so you can get out the Scheme from the authentication description, but its not contained in the actual principal/identities necessarily, some of our middlewares are using authType == authScheme, but that isn't a requirement.

blowdart: But a principal can contain multiple identities, from multiple schemes (Think cookie & bearer). So we may want to do it as a claim on the identity itself?

src/Microsoft.AspNet.Security.Twitter/TwitterAuthenticationHandler.cs
```diff
@@ -115,12 +115,12 @@ protected override AuthenticationTicket AuthenticateCore()
...
                 await Options.Notifications.Authenticated(context);
 
-                return new AuthenticationTicket(context.Identity, context.Properties);
+                return new AuthenticationTicket(context.Identity, context.Properties, Options.AuthenticationScheme);
             }
             catch (Exception ex)
             {
                 _logger.WriteError("Authentication failed", ex);
-                return new AuthenticationTicket(null, properties);
+                return new AuthenticationTicket((ClaimsPrincipal)null, properties, Options.AuthenticationScheme);
```
Praburaj: Can we add an overload of AuthenticationTicket with just `properties` and `scheme` that initializes principal internally to null as seem to use this often?

src/Microsoft.AspNet.Security.Google/GoogleAuthenticationHandler.cs
```diff
@@ -34,45 +34,45 @@ public GoogleAuthenticationHandler(HttpClient httpClient, ILogger logger)
 
             var context = new GoogleAuthenticatedContext(Context, Options, user, tokens);
             context.Identity = new ClaimsIdentity(
```
Praburaj: Instead of creating a `ClaimsIdentity` and wrapping it with a `ClaimsPrincipal` in several places like `AuthenticationTicket` can we also new up a `Principal` in the middleware itself so that we have a consistent implementation. All user facing contexts now contain Principal anyway. I think Wilson produces `Principal` now.

samples/SocialSample/Startup.cs
HaoK: Ah yeah so when we smush multiple principals together that will be a problem. So we could add a special claim to every claims idenitity inside of the principal that marks the scheme that generated it

src/Microsoft.AspNet.Security.Google/GoogleAuthenticationHandler.cs
blowdart: Context does need to move to principal really.

samples/SocialSample/Startup.cs
blowdart: We never smush principals though - just append to the existing one on context (and remove identity on context)

HaoK: Unless we are changing Httpcontext.user to be a bunch of principals. We have to smush eventually

blowdart: A single principal, multiple identities in it. Like it used to be in old asp.net :)

src/Microsoft.AspNet.Security.Google/GoogleAuthenticationHandler.cs
Praburaj: So revert change made to `CookieValidateIdentityContext.cs`?

src/Microsoft.AspNet.Security.Twitter/TwitterAuthenticationHandler.cs
HaoK: Sure, I will add this to clean things up

HaoK:
Updated PR replacing ClaimsIdentity with ClaimsPrincipal everywhere in all the various auth context's.

The code reading the ticket back from the cookie is broken somewhere, but looks like the social auth tests now all pass.  I can't debug in VS atm, will update PR again once once all unit tests are passing

samples/SocialSample/Startup.cs
brentschmaltz: Do we have the need to differentiate which parts of the ClaimsPrincipal get persisted?

We can't add arbitrary claims to an identity, this has the potential to change (amongst other things): ClaimsPrincipal.IsInRole("foo"). Think - can I add an arbitrary sid to an NTToken?

brentschmaltz: ClaimsIdentity has a constructor that takes just 'AuthencationType' no reason to pass defaults for Role and Name 
public ClaimsIdentity(string authenticationType)

brentschmaltz: The design to store the 'SessionID' in a claim in a principal seems a bit over engineered and could be simpler.

brentschmaltz: Does the design for cookies work in a heterogeneous environment where some of the web servers are running version 1 and some version 2?  This is important when users are upgrading over time.

blowdart: > We can't add arbitrary claims to an identity

Of course we can - that's exactly what claims transformation *is*. Plus, by setting the issuer to an internal value the user can tell this is added by us. You SID example is disingenuous as we're not adding SIDs. Or Roles.

blowdart: Cookie persistence is a matter for the cookie middleware, and a line comment on a sample.

If you are asking about protection then yes, it has scope for crypto agility which in turn would support upgrades.

src/Microsoft.AspNet.Security/DataHandler/Serializer/TicketSerializer.cs
brentschmaltz: We can use: ClaimsPrincipal.WriteTo(...).

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
```diff
@@ -17,18 +17,20 @@ namespace Microsoft.AspNet.Security.Infrastructure
     public static class SecurityHelper
     {
         /// <summary>
-        /// Add an additional ClaimsIdentity to the ClaimsPrincipal
+        /// Add all ClaimsIdenities from an additional ClaimPrincipal to the ClaimsPrincipal
         /// </summary>
         /// <param name="identity"></param>
-        public static void AddUserIdentity([NotNull] HttpContext context, [NotNull] IIdentity identity)
+        public static void AddUserPrincipal([NotNull] HttpContext context, [NotNull] IPrincipal principal)
```
brentschmaltz: Why is this IPrincipal and not ClaimsPrincpal?

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
```diff
@@ -17,18 +17,20 @@ namespace Microsoft.AspNet.Security.Infrastructure
...
         {
-            var newClaimsPrincipal = new ClaimsPrincipal(identity);
+            var newClaimsPrincipal = new ClaimsPrincipal(principal);
```
brentschmaltz: 'new' here is problematic if the user has created a CustomClaimsPrincipal, it just got dropped.

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
```diff
@@ -17,18 +17,20 @@ namespace Microsoft.AspNet.Security.Infrastructure
...
 
             ClaimsPrincipal existingPrincipal = context.User;
             if (existingPrincipal != null)
             {
                 foreach (var existingClaimsIdentity in existingPrincipal.Identities)
                 {
+                    // REVIEW: No longer use auth type for anything, so we could remove this check, except for the default one HttpContext.user creates
+                    // REVIEW: Need to ignore any identities that did not come from an authentication scheme?
                     if (existingClaimsIdentity.IsAuthenticated)
```
brentschmaltz: Why would an authentication handler add ClaimsIdentities that are not valid?

blowdart: Other middleware in the pipeline might have, so why not be safe?

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
```diff
@@ -38,26 +40,26 @@ public static void AddUserIdentity([NotNull] HttpContext context, [NotNull] IIde
...
+        /// <param name="authenticationScheme">The authentication type to look for</param>
+        public static bool LookupSignIn(IEnumerable<ClaimsIdentity> identities, string authenticationScheme, out ClaimsIdentity identity)
         {
             identity = null;
             foreach (var claimsIdentity in identities)
             {
-                if (string.Equals(authenticationType, claimsIdentity.AuthenticationType, StringComparison.Ordinal))
+                if (string.Equals(authenticationScheme, claimsIdentity.AuthenticationType, StringComparison.Ordinal))
                 {
                     identity = claimsIdentity;
```
brentschmaltz: From what I can see, LookupSignIn is called from the AuthHandler as a result of calling HttpContextResponse.SignIn.
I am not seeing how the CookieAH will persist HttpContext.User.

src/Microsoft.AspNet.Security.OAuth/OAuthAuthenticationHandler.cs
```diff
@@ -176,7 +171,7 @@ protected override AuthenticationTicket AuthenticateCore()
                 Properties = properties,
             };
             await Options.Notifications.GetUserInformationAsync(context);
-            return new AuthenticationTicket(context.Identity, context.Properties);
+            return new AuthenticationTicket(context.Principal, context.Properties, Options.AuthenticationScheme);
         }
 
         protected override void ApplyResponseChallenge()
```
brentschmaltz: On a separate note: ExchangeCodeAsync has no extensibility for adding parameters when exchanging a code for an AccessToken. Extensibility is required for a lot of customers. We are currently designing a 'code' flow including token / refresh caching. We will share the design.

Just a heads up to avoid duplication of work. If any other changes are planned for OAuthMW we should have a conversation.

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
brentschmaltz: This requires AH to ensure CI.IsAuthenticated to be true. If an AH adds a CI, it should do so for a valid reason, not willy-nilly.

src/Microsoft.AspNet.Security/DataHandler/Serializer/TicketSerializer.cs
HaoK: WriteTo only appears to exist in Core and not Aspnet50, is this worth an ifdef?

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
HaoK: I'm not sure, I just preserved existing behavior, maybe this can be changed to ClaimsIdentity now, @Tratcher thoughts?

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
HaoK: We could flip this logic around if we take a ClaimsPrincipal, append all the old identities from the previous ClaimsPrincipal to the new Principal passed in...

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
HaoK: I think the main reason for the IsAuthenticated check, is to not keep the default empty ClaimsPrincipal which has a single empty identity which is used as the anonymous/default user on HttpContext

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
blowdart: Yup, swap it.

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
HaoK: Transformations are not part of this PR, that will go in after the initial stuff all works again

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
blowdart: So this boils down to where the principal gets created, and what auth does. Personally I'd have a principal created at the beginning of the request and then we add identities to it, making it non-anonymous. So we'd accept a claimsidentity here into this function and then just add it onto the context.User principal.

HaoK: AuthenticationSchemes generate entire ClaimsPrincipals, not Identities anymore, this matters for when there are more than one Active Scheme, need to club the principal identities together for both schemes

src/Microsoft.AspNet.Security/DataHandler/Serializer/TicketSerializer.cs
brentschmaltz: WriteTo will be in desktop soon.

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
brentschmaltz: Does HttpContext.User have a default ClaimsPrincipal? This may be driving the whole check about existing principal. In the 90% case, a single AH will create a single ClaimsPrincipal. WsFed, OIDC follow this model.

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
brentschmaltz: I suspected so. 
In that can we:
1. Can the runtime handle HttpContext.User == null.
2. Set HttpContext.User to a singleton and use object.RefEquals to check for default.

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
brentschmaltz: OK, but it is related to how the Authenticated CP is put in the session. LookupSignIn sets the outparam to a single identity. My research indicates that results in a single identity being persisted.  Is this correct?

samples/SocialSample/Startup.cs
brentschmaltz: Barry, It is not disingenuous. The runtime should not be adding anything to the ClaimsPrincipal after the AH has completed it's work. How a ClaimsIdentity is constructed has a direct relationship to ClaimsPrincipal.IsInRole being true or false.

Perhaps, I am not understanding the relationship between cookie versions and claims transformation. What I am looking for is a guarantee that a V1 host can understand a V2 cookie (assuming the same key and algorithm is used for protection).

HaoK: In K, we aren't really using IsInRole anymore, our AuthZ policies will look for the presence of any claim with type ClaimsTypeRole and the value of the role name from a union claims from a specified set of authentication schemes.

src/Microsoft.AspNet.Security/DataHandler/Serializer/TicketSerializer.cs
HaoK: Filed https://github.com/aspnet/Security/issues/158 to track switching to WriteTo once its available in ASPNET50 and CORE50

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
HaoK: The runtime wants to guarantee that HttpContext.User is not null.

How about a variant of option 2 where we set the default user to a special type?  public class DefaultUserPrincipal : ClaimsPrincipal.  I am hesitant about using a singleton because anyone can muck with HttpContext.User and add claims to it etc

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
HaoK: Actually looks like I messed up the LookupSignIn changes, I need to fix that and I'll update the PR, its supposed to be accepting a ClaimsPrincipal, I actually nuked it accidentally from the code path entirely, which might explain why the principals are missing from the cookie tests :)

HaoK:
Updated PR, the concept of Accept doesn't look needed until we add back the ability to authenticate for multiple schemes at once (but this complexity might not be needed anymore).

Issue with cookies was actually in HttpAbstractions, wrapping an IPrincipal in a ClaimsPrincipal does not do what I expected...

https://github.com/aspnet/HttpAbstractions/commit/284c885e22965ef6d7c91fdcde89eee742db5ff7

HaoK:
Only remaining failure is:
GoogleMiddlewareTests.AuthenticatedEventCanGetRefreshToken

HaoK:
Fixed last google test failure, verification code in the test was only looking at the first identity which was causing the failure

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
blowdart: That's what we used to have, a default "Anonymous" Principal was created when a request began.

Checking IsAuthenticated would then be the right check, rather than a RefEquals.

Praburaj:
Just curious if tried MusicStore with these changes to see how this plays E2E. Tests simulate all of the security flows e2e. Will be good to try them.

HaoK:
Music store is at the bottom, MVC has to be updated before Music store can be...  Doing MVC next

src/Microsoft.AspNet.Security/DataHandler/Serializer/TicketSerializer.cs
lodejard: Don't think you need to write the scheme out as information. The use of data protection purposes prevents serialized information going out on one scheme coming back with another.

samples/SocialSample/Startup.cs
brentschmaltz: Fine, I still maintain that the runtime should not add any claims for it's own housekeeping.

src/Microsoft.AspNet.Security/Infrastructure/SecurityHelper.cs
brentschmaltz: It just seems odd that the runtime would need to filter the ClaimsIdentities that were created by a middleware for its own accounting logic.

blowdart: There's nothing odd about it.

Consider an api scenario called from an authenticated web site. The API wants a bearer token, but the cookie flows as well, we can't stop that. So the cookie middleware creates an identity and the bearer token middleware creates an identity. We need to know which identity came from what scheme so we can limit webapi authorization to those schemes

And as you told us we can't use AuthenticationType for this we're stuck adding a claim unless you have an alternative.

HaoK:
merged 775eb5ece452e10a27aa166f52b58a6c0b615d18

PR #162, opened 2015-02-23
Authentication -> Authorization rename/split

HaoK:
Remove Security sub namespace and rename to Auth
Split Authorization into its own assembly


HaoK:
cc @Praburaj @Tratcher

HaoK:
Merged 775eb5ece452e10a27aa166f52b58a6c0b615d18

PR #164, opened 2015-02-24
Move Active Authentication Mode into AutomaticAuthenticationHandler

HaoK:
- Rename AuthenticationMode to AutomaticAuthentication bool on AutomaticAuthenticationOptions
- Split old active behavior into a new base handler class : `AutomaticAuthenticationHandler` which derives from `AuthenticationHandler<AutomaticAuthenticationOptions>`
- Remove SecurityHelper LookupChallenge and added two new virtuals to handle the old active authentication checks (which work when no scheme is specified)
- Also attempted to address 401-403 with this change (needs more unit tests here)

cc @lodejard @Praburaj @blowdart @Tratcher


Security.sln
```diff
@@ -60,6 +60,8 @@ Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.AspNet.Authorizat
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.AspNet.Authorization", "src\Microsoft.AspNet.Authorization\Microsoft.AspNet.Authorization.kproj", "{6AB3E514-5894-4131-9399-DC7D5284ADDB}"
 EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.Framework.WebEncoders", "..\HttpAbstractions\src\Microsoft.Framework.WebEncoders\Microsoft.Framework.WebEncoders.kproj", "{DD2CE416-765E-4000-A03E-C2FF165DA1B6}"
```
Tratcher: I assume you didn't mean to check in this relative reference.

HaoK: Yeah I'll revert the global.json and revert all the sln changes as part of the final squash/merge, they won't make it back to dev

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -386,7 +381,16 @@ public virtual void Challenge(IChallengeContext context)
         /// <returns></returns>
         protected virtual Task ApplyResponseChallengeAsync()
         {
-            ApplyResponseChallenge();
+            // If authenticate was called and the the status is still 401, authZ failed so set 403 and stop
+            // REVIEW: Does this need to ensure that there's the 401 is challenge for this auth type?
+            if (ShouldConvertChallengeToForbidden())
```
Tratcher: All of this logic should be moved to AutomaticAuthenticationHandler, that's the only place where it should ever be true.

Tratcher: Actually now that I remember some of our earlier conversations, this was supposed to move back down into CookieAuth & BearerAuth's ApplyResponseChallenge because CookieAuth doesn't want to issue 403s, it wants to redirect to an AccessDenied page similar to how it handles challenges.

HaoK: Okay I'll move the logic and also add a test verifying that Cookie/Bearer don't turn 401's into 403s if they weren't asked to produce a ticket

HaoK:
Updated PR, moved ShouldConvertChallengeToForbidden to AutomaticHandler, added a test for cookie/bearer to ensure they don't touch 401's if they weren't authenticated.

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -334,7 +314,7 @@ protected virtual Task ApplyResponseGrantAsync()
 
         public virtual void SignIn(ISignInContext context)
         {
-            SignInIdentityContext = new SignInIdentityContext(context.Principal, new AuthenticationProperties(context.Properties));
+            SignInContext = new SignInContext(context.Principal, new AuthenticationProperties(context.Properties));
```
Tratcher: Shouldn't SignIn and SignOut also check for an AuthScheme match before doing anything here?

src/Microsoft.AspNet.Authentication/AutomaticAuthenticationHandler.cs
```diff
@@ -0,0 +1,98 @@
...
+                (authenticationSchemes == null || !authenticationSchemes.Any());
+        }
+
+        /// <summary>
+        /// Override this method to deal with 401 challenge concerns, if an authentication scheme in question
+        /// deals an authentication interaction as part of it's request flow. (like adding a response header, or
+        /// changing the 401 result to 302 of a login page or external sign-in location.)
+        /// </summary>
+        /// <returns></returns>
+        protected override Task ApplyResponseChallengeAsync()
```
Tratcher: This method is overridden in both Cookies and Bearer, so I don't think there's much use in even implementing it here.

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -329,8 +328,8 @@ protected override void ApplyResponseChallenge()
                 return;
             }
```
Tratcher: We need to implement the Access Denied scenario here as a mirror of Options.LoginPath.

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
HaoK: Doh yeah. I will add some tests verifying the negative case too

src/Microsoft.AspNet.Authentication/AutomaticAuthenticationHandler.cs
HaoK: Well its a bit confusing, cookies is overriding the sync version only, it would have to override the async version as well then, so this method is whats being used for cookies to handle the 401-403

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
HaoK: Filed https://github.com/aspnet/Security/issues/166 to deal with this in a different PR

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
HaoK: Added tests verifying that exceptions are thrown (rather than sign in/sign out happening) when wrong auth type is specified

HaoK:
Merged 775eb5ece452e10a27aa166f52b58a6c0b615d18

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -566,7 +566,7 @@ public override Task<bool> InvokeAsync()
             {
                 if (ticket.Principal != null)
                 {
-                    Request.HttpContext.Response.SignIn(ticket.Properties, ticket.Principal.Identities);
+                    Request.HttpContext.Response.SignIn(ticket.AuthenticationScheme, ticket.Principal, ticket.Properties);
```
PinpointTownes: An invalid scheme is used here and makes the OIDC client middleware totally buggy because it cannot delegate the principal persistence to the cookie middleware.

Use __Option.SignInScheme__ instead.

HaoK: Ah thanks, saved me from having to track this down today!

PinpointTownes: I opened a ticket for that (https://github.com/aspnet/Security/issues/167) and submitted a PR (https://github.com/aspnet/Security/pull/168) that fixes it :smile:

Issue #167, opened 2015-03-03
Incorrect handler delegation in OpenIdConnectAuthenticationHandler.InvokeReplyPathAsync


PinpointTownes:
Looks like your new monster has been released too quickly in the wild :smile: :laughing: 

An invalid scheme is used in `OpenIdConnectAuthenticationHandler.InvokeReplyPathAsync` and makes the OIDC client middleware totally buggy because it cannot delegate the principal persistence to the cookie middleware. You should use `Option.SignInScheme` instead.

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs#L569

(you should also fix the file name)

HaoK:
Thanks fix is pushed, I'll fix the file name in a different commit 0577454f13f6ec0e82a424225729ccb57e2911ac

PinpointTownes:
Sadly, your fix is incomplete: it doesn't support setting the `SignInScheme` property via `ExternalAuthenticationOptions.SignInScheme`.

See https://github.com/aspnet/Security/pull/168/files#diff-960390b7edd1b3a479fbd7fcfd07fbe8R48

HaoK:
Yeah I know, I'm just trying to get MusicStore unblocked for now, the test coverage for OpenidConnect is...fairly lacking right now obviously, once all the existing tests are fixed, I'll look into adding some more coverage so we can catch this sort of failure before checkin...

PinpointTownes:
You should maybe re-open this ticket to make sure this new issue is correctly tracked.

HaoK:
Merged 4a2a742ad5e52b2f9e0529b8c6f5f9decd3e80f0

Issue #144, opened 2015-02-02
OpenIdConnect flow does not set the SignInAsAuthenticationType cookie


Praburaj:
Checkin : https://github.com/aspnet/Security/commit/3483842ab76d4198fc230c7a1905b0ada9b4bf89#commitcomment-9535319 seems to break openIdconnect flow as it does not set the external cookie. 

Test failure here : https://github.com/aspnet/MusicStore/commit/7783c368569fbaf267121eff8eedefbd05cb974d is due to this.

PinpointTownes:
On a side note, `OpenIdConnectAuthenticationMiddleware` doesn't use a correct value for `TokenValidationParameters.AuthenticationType`: it should use `Options.AuthenticationType` instead of `externalOptions.Options.SignInAsAuthenticationType` to make sure that the `ClaimsIdentity` returned by `AuthenticateCoreAsync` has the same `AuthenticationType` as the middleware itself.

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Security.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs#L51

Praburaj:
@brentschmaltz

PR #142, opened 2015-02-02
Fix #140 and rework the different OAuth2 Bearer middleware tests

PinpointTownes:
The PR is a bit noisy but I had to rework large parts of various tests, that were not really conclusive.

/cc @brentschmaltz @Tratcher

Issue #154, opened 2015-02-13
Revisit AuthN (switch to AuthenticationScheme/ClaimsPrincipal)


HaoK:
Summary of plan:
- AuthenticationType will be renamed AuthenticationScheme which represents a particular authentication middleware.
- Auth Middleware will now produce a ClaimsPrincipal instead of a ClaimsIdentity with authentication type set
- Active/Passive removed from base auth middlware and opt in functionality pushed into a new base class dervied from AuthMiddleware that has an Automatic = true|false, Cookie(true by default)/Bearer(false by default) will opt in.  Concept is removed from all social middlewares (always passive)

leastprivilege:
Why the move to ClaimsPrincipal? What about multiple authN middlewares - e.g. cookies and client certificates?

HaoK:
The behavior is going to be pretty similar as today for multiple authN middlewares, each one generates a ClaimsPrincipal, and at the end, HttpContext.User will be the aggregate of all of the authN principals

The big change is to allow Auth middlewares to produce more than one ClaimsIdentity and leave the authenticationType on the identities untouched, authenticationScheme will be how you refer to a specific auth middleware, instead of authenticationType like before.

cc @blowdart @Tratcher @lodejard

leastprivilege:
So the claims transformation will see a list of ClaimsPrincipal and be able to normalize to one?

sounds good.

lodejard:
Yeah, key difference will be the `identity.AuthenticationType` strings must no longer be used to determine which authn mechanism delivered the identity to the current request. Those AT strings will now be considered arbitrary data, serialized back and forth with the cookie, bearer token, etc. 


A new string named AuthenticationScheme will be used to associate a set of identities with the-way-they-arrived.

So like Hao said, in a nutshell, a cookie or bearer token will be able to contain multiple identities, and the identities' AuthenticationType may be assigned to be any value.

PinpointTownes:
> Concept is removed from all social middlewares (always passive)

So we won't be able to automatically redirect users to an IDP without relying on an explicit `Challenge`? This was particularly useful on applications using a single external provider (now, we'll need a controller or a middleware to trigger the challenge stuff).

> A new string named AuthenticationScheme will be used to associate a set of identities with the-way-they-arrived.

Is it going to be included as a claim?

HaoK:
Merged 775eb5ece452e10a27aa166f52b58a6c0b615d18

PinpointTownes:
[...]

PR #168, opened 2015-03-03
Fix #140, #167, address a comment in #144 and rework the bearer middleware tests

### LINKED ISSUE #140: OAuthBearerAuthenticationNotifications.SecurityTokenValidated is never triggered
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Security.OAuthBearer/OAuthBearerAuthenticationHandler.cs#L134-L148

/cc @brentschmaltz @Tratcher

PinpointTownes:
The PR is a bit noisy but I had to rework large parts of various tests, that were not really conclusive.

(GitHub accidentally closed my previous PR when I rebased it (https://github.com/aspnet/Security/pull/142) so I had to re-submit it)


msotclas:
Hi __@PinpointTownes__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>You've already signed the contribution license agreement. Thanks!</span>
        <p>The agreement was validated by Microsoft Open Technologies, Inc. and real humans are currently evaluating your PR.</p>

TTYL, MSOTBOT;

PinpointTownes:
I also included fixes for https://github.com/aspnet/Security/issues/144#issuecomment-72757076 and  https://github.com/aspnet/Security/issues/167

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -566,7 +565,7 @@ public override Task<bool> InvokeAsync()
             {
                 if (ticket.Principal != null)
                 {
-                    Request.HttpContext.Response.SignIn(ticket.AuthenticationScheme, ticket.Principal, ticket.Properties);
+                    Request.HttpContext.Response.SignIn(Options.SignInScheme, ticket.Principal, ticket.Properties);
```
Praburaj: I think this change is already in. 

/CC @HaoK

HaoK: Only this one line is in, so yeah a rebase probably would make this diff go away, this PR has more than that fix though, I am not familiar enough with this code to determine whether these other changes are appropriate

PinpointTownes: @Praburaj rebased, thanks :dancers:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
```diff
@@ -46,9 +45,14 @@ public class OpenIdConnectAuthenticationMiddleware : AuthenticationMiddleware<Op
         {
             _logger = loggerFactory.CreateLogger<OpenIdConnectAuthenticationMiddleware>();
 
+            if (string.IsNullOrEmpty(Options.SignInScheme))
+            {
+                Options.SignInScheme = externalOptions.Options.SignInScheme;
+            }
```
Praburaj: Should you throw if `SignInScheme` is null or empty. I see this check in some middlewares and some lack. 

https://github.com/aspnet/Security/blob/e3b457bac7693818786996a9f5acaef40b02ed1c/src/Microsoft.AspNet.Authentication.Twitter/TwitterAuthenticationMiddleware.cs#L74

PinpointTownes: IMHO, the persistence delegation should be optional so that the end dev can implement his own logic, just like the `CallbackPath` is not mandatory.

That said, I agree that this kind of inconsistency is bad and should be fixed. https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Twitter/TwitterAuthenticationHandler.cs#L206 seems to indicate that we should remove the null check you mentioned and make `SignInScheme` optional.

test/Microsoft.AspNet.Authentication.Test/OAuthBearer/OAuthBearerMiddlewareTests.cs
```diff
@@ -187,43 +254,30 @@ public BlobTokenValidator(string authenticationScheme)
...
-                get
-                {
-                    return 2*2*1024;
-                }
-
-                set
-                {
-                    throw new NotImplementedException();
-                }
+                get { return 2*2*1024; }
```
Praburaj: Format the file.

Praburaj:
Changes look good to me. @tushargupta51 - could you also review?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
HaoK: So, we are talking about SignInScheme that lives on ExternalAuthenticationOptions correct?  This should be optional always, since its only useful when forwarding a "passive" (new terminology is Automatic = false), auth middleware's ticket to something automatic like bearer/cookie for persistence.  We probably should rename this to convey the usage.  Perhaps something like SignInForwardingOptions && SignInScheme maybe?  @lodejard what do you think?

test/Microsoft.AspNet.Authentication.Test/OAuthBearer/OAuthBearerMiddlewareTests.cs
PinpointTownes: Sure. I removed this value and replaced it with a NIE as it's actually not even used.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
PinpointTownes: Actually, this specific `SignInScheme` lives on `OpenIdConnectAuthenticationOptions` and is just a wrapper around TokenValidationParameters.AuthenticationType. But of course, it serves the same purpose.

That said, I'm not sure that the passive/active/automatic/manual question is pertinent here, as it can be used in both modes (and anyway, this middleware doesn't inherit from `AutomaticAuthenticationHandler`).

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
```diff
@@ -46,9 +45,14 @@ public class OpenIdConnectAuthenticationMiddleware : AuthenticationMiddleware<Op
         {
             _logger = loggerFactory.CreateLogger<OpenIdConnectAuthenticationMiddleware>();
 
+            if (string.IsNullOrEmpty(Options.SignInScheme))
```
brentschmaltz: What is SignInScheme used for?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
brentschmaltz: I asked above, so I will ask again here.
What is SignInScheme used for?

PinpointTownes: It's _SignInAsAuthenticationType_'s new name: it's used to delegate signin persistence to another middleware (classically, to a cookie middleware)

HaoK: Can we consider sharing the same options instead of introducing a similar flag?  i.e. OIDC can use ExternalAuthenticationOptions instead of looking at its similarly named (and used flag).  Hopefully this will make it clear that its used for exactly the same purpose as the other one...

HaoK: We can also rename ExternalAuthenticationOptions to something better as part of this consolidation if we can come up with a better name

PinpointTownes: If we decide to remove `SignInScheme` from the OIDC options to favor the global `ExternalAuthenticationOptions`, we must be aware that it will be far less granular: if we have 2 OIDC middleware in a pipeline, we won't be able to define 2 different cookie middleware for these instances.

We could also make `OpenIdConnectAuthenticationOptions` inherit from `ExternalAuthenticationOptions`.

HaoK: Well if that really is important, we can, it just requires an additional argument corresponding to the named instance of the ExternalAuthenticationOptions to use, typically there will be only one, but for the multiple middleware scenario, names need to specified to be able to resolve the specific instance of the ExternalAuthenticationOptions.  But given that this scenario doesn't work today with the other social middlewares, I'm not sure how important it is for OIDC

PinpointTownes: This would require two (or three if you have a global configuration) `services.Configure<ExternalAuthenticationOptions>` calls + the OIDC options themselves and I'm not sure it would be easier from a UX perspective.

HaoK: So is it more likely to have multiple OIDC middleware in the pipeline, as compared to Google/Facebook/Twitter?

PinpointTownes: More likely? In most cases, I'd say no. But anyway, the same remarks apply to the OAuth2 providers, that also expose a `SignInScheme` property.

HaoK: Okay for now lets just leave them where they are, I filed https://github.com/aspnet/Security/issues/174 to sort this out later

test/Microsoft.AspNet.Authentication.Test/OAuthBearer/OAuthBearerMiddlewareTests.cs
```diff
@@ -27,6 +26,8 @@ public class OAuthBearerMiddlewareTests
         {
             var server = CreateServer(options =>
             {
+                options.AutomaticAuthentication = true;
```
PinpointTownes: Given that the OAuth2 bearer middleware is now passive by default, setting this property to true is now mandatory. IMHO, it's not terribly intuitive and it should be automatic by default, just like it was in Katana.

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs

http://katanaproject.codeplex.com/SourceControl/latest#src/Microsoft.Owin.Security/AuthenticationOptions.cs

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
PinpointTownes: Great :sunglasses:

tushargupta51:
The changes to the bearer middleware tests looks good.

test/Microsoft.AspNet.Authentication.Test/OAuthBearer/OAuthBearerMiddlewareTests.cs
HaoK: @blowdart do you remember why we decided to make bearer passive by default unlike Cookies which is active/automatic by default?

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationMiddleware.cs
```diff
@@ -4,8 +4,6 @@
 using System;
 using Microsoft.AspNet.Authentication.Cookies.Infrastructure;
 using Microsoft.AspNet.Authentication.DataHandler;
-using Microsoft.AspNet.Authentication.Cookies.Infrastructure;
```
PinpointTownes: Totally unrelated changes, but these 2 lines were causing compilation warnings.

PinpointTownes: Fixed a few minutes ago by @HaoK: https://github.com/aspnet/Security/commit/e2bb76280f07404f23ecbe8f4b2481082692abc2

test/Microsoft.AspNet.Authentication.Test/OAuthBearer/OAuthBearerMiddlewareTests.cs
blowdart: Nope. Was it a @lodejard suggestion?

HaoK: Yeah, guess I don't have any context into why we decided for this to be passive by default, maybe for the scenario where you have both cookies and bearer?

PinpointTownes: In this rather uncommon scenario (at least, I hope), I don't see why cookie authentication should be automatic by default as it is far more prone to XSRF attacks than bearer authentication.

blowdart: Well I'm not sure we want to get into "If X is present Y should not be automatic." That's just begging for pain.

If you have both cookie and bearer well the principle is going to have two identities on it, which is what we want anyway right?

PinpointTownes: Sure, that was just a general consideration regarding the default values for cookies (automatic) and bearer (manual), that I suggest to align if there's no strong reason to keep bearer manual by default.

HaoK: Well the behavior we are trying to decide is whether you need to ask for the bearer token via Authenticate, or whether its there automatically even if you don't ask for it (via policy for example)  We were going to make Cookie passive too, but that broke down quickly

PinpointTownes: Passive mode is great when you have an easy and clear way to trigger authentication (like the `HostAuthenticationFilter/Attribute` in Web API OWIN). Currently, the only OTB solution is through the authorization stuff.

Making it manual by default is probably going to confuse users, but that's just my 2 cents.

blowdart: Yea, and I'm still of the opinion everything, including social should end up as active :)

HaoK: yes that is the idea, if you want the bearer token to show up on the user, you most likely have an authZ attribute which asked for it.

PinpointTownes: "most likely" is the keyword here.

That's probably not the best place to discuss about that, but I've written tons of methods where authentication was not mandatory but if the user was authenticated, the payload was different (for instance, a profile endpoint: if the request is anonymous, you get a minimal representation. If you're connected but not the user corresponding to the profile, you get a intermediate representation... and you get the full profile it it's yours)

@blowdart I agree. Even if we could argue that social providers should be configured to manual mode, you should have a way to force the automatic/active mode, just like you could with Katana 3.
https://github.com/aspnet/Security/issues/154#issuecomment-77431018

PinpointTownes: @lodejard ping:

> do you remember why we decided to make bearer passive by default unlike Cookies which is active/automatic by default?

PinpointTownes:
Rebased.

test/Microsoft.AspNet.Authentication.Test/OAuthBearer/OAuthBearerMiddlewareTests.cs
PinpointTownes: Now that a PR has been submitted to reunify `AuthenticationHandler` and `AutomaticAuthenticationHandler`, I suggest moving this discussion there: https://github.com/aspnet/Security/pull/189

PinpointTownes:
Squashed to incorporate feedback in the appropriate commits.

PinpointTownes:
Conflict fixed.

@HaoK @tushargupta51 @brentschmaltz any chance you could have a final look and merge it? :smile:

HaoK:
I'll try to get to this after my hosting changes are done...

PinpointTownes:
@HaoK thanks. Don't hesitate to ping me if there are conflicts that need to be fixed.

PinpointTownes:
I took the liberty to fix `OpenIdConnectAuthenticationOptions.SignInScheme`.

@HaoK should we also update `OAuthAuthenticationOptions`, `TwitterAuthenticationOptions` and `ExternalAuthenticationOptions` to align the documentation?

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs#L104
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Twitter/TwitterAuthenticationOptions.cs#L94
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication/ExternalAuthenticationOptions.cs#L11

HaoK:
Yeah looks good, go ahead and normalize the doc comments to be consistent, I'll merge this in soon

HaoK:
I'm not suggesting you make this change in this PR, but what do you think about maybe ExternalAuthenticationOptions -> SignInSchemeOptions is that better or worse?  I'm not sure what's a good property name if we did that, Scheme or AuthenticationScheme perhaps?

PinpointTownes:
Personally, I think I'd simply remove `ExternalAuthenticationOptions` and rely exclusively on `OAuthAuthenticationOptions`/`OpenIdConnectAuthenticationOptions`/`TwitterAuthenticationOptions` to set up `SignInScheme`: I'm convinced it would make this "delegated persistence" stuff more evident and easier to understand for end developers.

We could also set a default value (`Cookies` to match the default scheme for the cookies middleware? But IIRC, that's not the default scheme used by aspnet/Identity) for these properties, so people wouldn't have to set an explicit value for basic `cookies + social provider` scenarios..

HaoK:
Yeah so the idea is we want to easily configure all of the auth middlewares via one central services.Configure(options => options.SignInScheme = "Cookie").  Otherwise any app that's using Google/Facebook/Twitter/MAuth would need to do this for every auth middleware mechanically (and the error story if you misconfigure this is pretty poor)

HaoK:
The intent is basically to be the centralized default SignInScheme, which might not be a bad name: SignInOptions.DefaultSignInScheme, as all auth middlewares would use this as the default scheme for their SignInScheme

PinpointTownes:
Indeed, removing `ExternalAuthenticationOptions` would make global `SignInScheme` more painful to set up as you'd have to configure it for each authentication middleware :smile: 

Another approach: the OIDC/OAuth2/Twitter options could inherit from `ExternalAuthenticationOptions` and stop exposing a custom `SignInScheme`. The default scheme could be configured via `services.Configure<ExternalAuthenticationOptions>(options => options.SignInScheme = "Cookie")` as you suggested, and it could be replaced by `services.Configure<OpenIdConnectAuthenticationOptions>(options => options.SignInScheme = "Other")`.

That said, I'm not sure it would work as-is with the current aspnet/Options bit. In this case, would `OpenIdConnectAuthenticationOptions` automagically be created with the settings applied to `ExternalAuthenticationOptions` it inherits from?

`SignInOptions.DefaultSignInScheme` sounds like a good name, though it's not obvious it would only be supported by OIDC/OAuth2/Twitter providers.

HaoK:
Yeah configuration options doesn't really look at inheritance, you are configuring a singleton per TOptions, so you'd be configuring only the base options, which if noone is asking for, will be ignored.  So that doesn't work unfortunately.

Yeah perhaps we still need something that sounds like External/Delegating/NonAutomatic something that conveys that this is for the non automatic auth middlewares to use.

[GoodNameForNonAutomatic]AuthenticationOptions.DefaultSignInScheme i think is the correct name, just need to find a replacement for External

HaoK:
Perhaps we could just use a grab all SharedAuthenticationOptions.DefaultSignInScheme?

@Tratcher what do you think about renaming ExternalAuthenticationOptions to be something more general?

Tratcher:
@HaoK Sounds fine. It doesn't have any other fields.

brentschmaltz:
@HaoK Help me understand what SignInScheme is used for and if it is completely implemented?

HaoK:
@brentschmaltz refer to @PinpointTownes update to the doc comments, its fairly clear, nothing has changed in its meaning or intent since Katana, we just renamed it from SignInAsAuthenticationType

https://github.com/PinpointTownes/Security/commit/39a970c8f25b685e462f165aa768c0c929cbbad3

PinpointTownes:
@brentschmaltz if you think it's not clear enough, creating a wiki page with various code samples and copying the URL in the documentation might be a good option.

HaoK:
@PinpointTownes Do you want to squash your commits down to one or two?  If I do it, the changes will look like mine instead of yours.

PinpointTownes:
@HaoK done. I hope the commits will be descriptive enough :smile:

PR #177, opened 2015-03-07
Reading AuthenticationProperties from SignOutContext

Praburaj:
This will enable users to set a specific redirect uri and call signout.

This will depend on https://github.com/aspnet/HttpAbstractions/pull/218

@brentschmaltz @Tratcher @HaoK

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
@@ -77,7 +77,7 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
                 // 1. properties.Redirect
                 // 2. Options.Wreply
                 var properties = new AuthenticationProperties(signout.Properties);
-                if (properties != null && !string.IsNullOrEmpty(properties.RedirectUri))
+                if (!string.IsNullOrEmpty(properties.RedirectUri))
                 {
                     openIdConnectMessage.PostLogoutRedirectUri = properties.RedirectUri;
                 }
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
@@ -8,10 +8,9 @@ using System.IO;
 using System.Linq;
 using System.Security.Claims;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Authentication.Notifications;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Authentication;
-using Microsoft.AspNet.Authentication;
-using Microsoft.AspNet.Authentication.Notifications;
 using Microsoft.Framework.Logging;
 using Microsoft.IdentityModel.Protocols;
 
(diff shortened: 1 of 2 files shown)


HaoK:
Change looks good, is there any way we can add a test verifying the scenario/what this is supposed to fix?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -77,7 +76,7 @@ protected override void ApplyResponseGrant()
                 // Set End_Session_Endpoint in order:
                 // 1. properties.Redirect
                 // 2. Options.Wreply
-                AuthenticationProperties properties = new AuthenticationProperties();
+                var properties = new AuthenticationProperties(signout.Properties);
                 if (properties != null && !string.IsNullOrEmpty(properties.RedirectUri))
```
PinpointTownes: You can remove the `properties` null check, it's obviously useless :smile:

Praburaj: You're right. I'll remove that.

Praburaj:
@HaoK sure I can add a test. Let me update the PR.

Praburaj:
@HaoK Added test.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectMiddlewareTests.cs
```diff
@@ -154,6 +154,22 @@ public class OpenIdConnectMiddlewareTests
         }
 
         [Fact]
+        public async Task SignOutWith_Specifi_RedirectUri_From_Authentication_Properites()
```
HaoK: nit: Specifi[c] typo

HaoK:
looks good to me :shipit:

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectMiddlewareTests.cs
Praburaj: Fixed.

muratg:
:shipit:

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectMiddlewareTests.cs
```diff
@@ -154,6 +154,22 @@ public class OpenIdConnectMiddlewareTests
...
+        {
+            var stateFormat = new PropertiesDataFormat(new EphemeralDataProtectionProvider().CreateProtector("OpenIdConnectTest"));
```
PinpointTownes: Unused variable?

Praburaj: Thanks. Removed this var from a bunch of tests.

brentschmaltz:
This is a needed feature, and it helps. I think it could rate higher on the usability scale if users could set named properties like we do on the OpenIdConnectMessage.

Praburaj:
Merged : https://github.com/aspnet/Security/commit/b7c8af8503831ba4becaa93b9453510b81c4d30e

PR #183, opened 2015-03-11
Design PR - Prototype ClaimsTransformation (non automatic flow)

HaoK:
Addresses https://github.com/aspnet/Security/issues/48 second part:

Also hook any calls to HttpContext.AuthenticateAsync to apply transformation before returning the resultant identity

This is a bit messy/fragile as currently each AuthMiddleware has to manually call ApplyClaimsTransformation, thoughts on how best to force this to happen?  If the ClaimsTransformation options were moved up into Http.Authentication, HttpContext could do it as the last step before returning any results from Authenticate instead of relying on the implementations of the handler to remember to apply the xform...  

cc @Praburaj @Tratcher @loudej @brentschmaltz

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,3 +1,3 @@
 {
-    "sources": ["src"]
+    "sources": ["src", "C:\\Github\\HttpAbstractions\\src" ]
 }
--- a/src/Microsoft.AspNet.Authentication/ClaimsTransformationAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication/ClaimsTransformationAuthenticationHandler.cs
@@ -1,0 +1,105 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Authentication;
+using Microsoft.AspNet.Http.Core.Authentication;
+
+namespace Microsoft.AspNet.Authentication
+{
+    /// <summary>
+    /// Handler that applies ClaimsTransformation to authentication
+    /// </summary>
+    public class ClaimsTransformationAuthenticationHandler : IAuthenticationHandler
+    {
+        private readonly Func<ClaimsPrincipal, ClaimsPrincipal> _transform;
+
+        public ClaimsTransformationAuthenticationHandler(Func<ClaimsPrincipal, ClaimsPrincipal> transform)
+        {
+            _transform = transform;
+        }
+
+        public IAuthenticationHandler PriorHandler { get; set; }
+
+        private void ApplyTransform(IAuthenticateContext context)
+        {
+            if (_transform != null)
+            {
+                var authContext = context as AuthenticateContext;
+                if (context != null)
+                {
+                    authContext.Result = new AuthenticationResult(
+                        _transform.Invoke(authContext.Result.Principal),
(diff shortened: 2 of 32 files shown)


HaoK:
@davidfowl suggested swapping in a claims transformation decorated HttpAuthenticationFeature, I will try that next

HaoK:
Updated with feedback from @lodejard 

Transformation now done via ClaimsTransformation IAuthenticationHandler.  

There's an issue where Automatic authentication schemes like Cookies will apply the transformation twice (once on the automatic authenticate, and once when the Xform middleware runs)

global.json
```diff
@@ -1,3 +1,3 @@
 {
-    "sources": ["src"]
+    "sources": ["src", "C:\\Github\\HttpAbstractions\\src" ]
```
davidfowl: oops :smile:

HaoK:
cc @brentschmaltz

src/Microsoft.AspNet.Authentication/ClaimsTransformationAuthenticationHandler.cs
```diff
@@ -0,0 +1,106 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
```
brentschmaltz: My first reaction is to question why this is modeled as an AuthenticationMiddleware. Most of the functionality of the AMW doesn't seem to be usefull. 

What about a delegate on HttpContext? It would be simpler, if all the user wants to do is perform a transformation, they must now understand the AH model.

HaoK: The end user doesn't really need to understand the implementation. The usage is to simply call configureClaimsTranformation to plug in their xform with the understanding that it will be invoked whenever authenticate is called. And once on HttpContext.User.

brentschmaltz: What is the code the user would have to write to set it up?
All we seem to be doing is exposing a Delegate that performs claims transformation.  ClaimsPrincipal in, ClaimsPrincipal out.

Why does that require a whole MiddleWare?  It has to be maintained, tested, etc.
Where would the middleware go in the layering of middlewares?

HaoK: User code is pretty straight forward `services.ConfigureClaimsTransformation(Func<ClaimsPrincipal, ClaimsPrincipal>)`  

The xform middleware goes in the pipeline after all of the other auth middlewares.  Unfortunately understanding middlewares is pretty much required in ASPNET5 now.

brentschmaltz: >>> Unfortunately understanding middlewares is pretty much required in ASPNET5 now.
That is actually a shame.

But I have to disagree with this fix, it is way too complicated.
For a user that wants to use a single OIDC provider all they should need to know about the middleware is app.UseOIDC(clientId, authority).  Nothing else.

I am currently fixing things so they don't even have to think about 'CookieMiddleWare'.
This change moves us away from the simple model and forces the user to know about types and code not associated with what they are trying to do.

PinpointTownes: > I am currently fixing things so they don't even have to think about 'CookieMiddleWare'.
This change moves us away from the simple model and forces the user to know about types and code not associated with what they are trying to do.

Smells oil out there :smile:

HaoK:
Merged 14d1b467c69dd6f836f5683b6e1545b7f3f859ee

PR #202, opened 2015-03-31
logging, added tests, fixed code only flow.

brentschmaltz:
still lots of additional testing to come.


src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -115,28 +115,59 @@ protected override void ApplyResponseChallenge()
...
         protected override async Task ApplyResponseChallengeAsync()
         {
+            bool logLevelError = _logger.IsEnabled(LogLevel.Error);
+            bool logLevelDebug = _logger.IsEnabled(LogLevel.Debug);
+            bool logLevelInformational = _logger.IsEnabled(LogLevel.Information);
+            bool logLevelWarning = _logger.IsEnabled(LogLevel.Warning);
+
+            if (logLevelDebug)
+            {
+                if (logLevelDebug)
```
HaoK: Duplicated debug check?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -115,28 +115,59 @@ protected override void ApplyResponseChallenge()
...
+                {
+                    _logger.LogDebug(string.Format(CultureInfo.CurrentCulture, Resources.OIDCH_0026_ApplyResponseChallengeAsync, this.GetType().ToString()));
+                }
+            }
+
             if (ShouldConvertChallengeToForbidden())
             {
+                if (logLevelDebug)
```
HaoK: I don't think the general recommendation is to check log level, the logger should take care of this, only when doing computationally heavy log messages is explicit guard necessary

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -178,33 +234,61 @@ protected override void ApplyResponseChallenge()
...
             {
-                openIdConnectMessage.Nonce = Options.ProtocolValidator.GenerateNonce();
+                message.Nonce = Options.ProtocolValidator.GenerateNonce();
                 if (Options.NonceCache != null)
                 {
-                    Options.NonceCache.AddNonce(openIdConnectMessage.Nonce);
+                    if (!Options.NonceCache.TryAddNonce(message.Nonce))
+                    {
+                        string msg = string.Format(CultureInfo.CurrentCulture, Resources.OIDCH_0033_TryAddNonceFailed, message.Nonce);
+                        if (logLevelError)
```
HaoK: This guard has no perf benefit as the msg generation is done outside the guard

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectIdentityFlowTests.cs
```diff
@@ -0,0 +1,640 @@
...
+        public void Authenticated(ClaimsPrincipal principal, IDictionary<string, string> properties, IDictionary<string, object> description)
+        {
+        }
+
+        public void NotAuthenticated()
+        {
+        }
+    }
+
+    public class OpenIdConnectAuthenticationHandlerPublic : OpenIdConnectAuthenticationHandler
```
HaoK: nit: consider TestOpenIdConnectAuthHandler instead of Public suffix?

HaoK:
yay for tests, only major thing worth considering is removing all of the log level guards for readability, as its really not needed given how straight forward your log message generation is.

brentschmaltz:
This is just the beginning for tests. I was trying to figure out a pattern.

About the annoying external log level check, it's a perf thing, stops the context switch. I plan on lots and lots of logging.

Can you live with it?

Sent from my Windows Phone
________________________________

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectIdentityFlowTests.cs
brentschmaltz: Good idea.

Sent from my Windows Phone
________________________________

PinpointTownes:
@brentschmaltz honestly, these changes make the code far harder to read (it's really a PITA :smile:)

If you're particularly concerned about performance, why not using the logger extensions that take a format string and multiple arguments instead of manually calling `string.Format`?

https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Framework.Logging.Interfaces/LoggerExtensions.cs#L52

BTW, @GrabYourPitchforks has done great things with string interpolation in aspnet/DataProtection, and it really makes logging clearer: https://github.com/aspnet/DataProtection/commit/4365b531d8a4ec319ff26a1e661a6991efbd140f :smile:

GrabYourPitchforks:
@PinpointTownes Thanks for the kind words! Unfortunately one of the downsides to what I did in DataProtection is that logging messages are hardcoded literal strings and thus are non-localized. The code Brent submitted here is more future-proof once localization work is performed.

Calling the logger extensions that take format strings is certainly a good idea. Not only is performance improved, it also gives logging sinks a crack at the raw values without requiring them to parse the string manually.

rynowak:
re: logging - the state of the art is like this: https://github.com/aspnet/Routing/pull/172/files#diff-4997dfe8056272b794ccb3bcdfe214d9R117

Our logging framework understands the key-values - so that loggers that work with key-value-pairs can have them, and loggers that don't still get a new message. /cc @loudej in case he wants to comment.

PinpointTownes:
@GrabYourPitchforks interesting, I didn't know that the ASP.NET team wanted to localize tracing messages :smile:
Do you know whether tracing components will have their own culture or simply use `CultureInfo.CurrentCulture`?

I wonder if an i18n + string interpolation mix wouldn't be the best of both worlds: you'd simply call `logger.LogWarning($"Request did not match the route '{RouteName}' with template '{RouteTemplate}'.")` and it would internally retrieve the localized format string corresponding to "Request did not match the route '{0}' with template '{1}'." and call `FormattableString.ToString()` if needed.

HaoK:
Actually we decided against localizing log messages at this time since we didn't want admins to potentially know all their users languages. Log Messages should always be in English at this time.

GrabYourPitchforks:
Just to clarify, the existing ASP.NET runtime and other .NET Framework components localize log messages using the system installed culture, not the invoking thread's culture. That way sysadmins always see log messages in their own language. Presumably this will be the behavior in the future, but it's just not ready at this time.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: I will fix that.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: I like to do it this way as there are a log of logs. Each one is a function call which is a small performance ding.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: We are going to throw anyway so any perf here is small, but I like to be consistent.

brentschmaltz:
I always use CurrentCulture for the localization reason, just a habit. I actually like to use string.cat as it is much faster than string.format. If you are not convinced, run some numbers on loops of: string.Format vs.
str1 + str2.

About the 'loglevel' check, I favor that even though it's a bit messy, macros remember them made that style neater, because if you have logging I want to minimize the perf hit. Something I like to do.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectIdentityFlowTests.cs
brentschmaltz: When I saw this I was thinking you meant to rename the entire file OpenIdConnectAuthenticationHandlerTests, which I think is more accurate. I prefer OpenIdConnectAuthenticaitonHandlerTest over TestOpenIdConnectAuthenticationHandler as the prefix makes finding things in sorted lists slower cause everything starts with T...

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
Eilon: @brentschmaltz the performance should be measured before making such a claim. We've already done work to ensure this is a very efficient operation. If you have evidence to the contrary, please file a bug in the https://github.com/aspnet/Logging/issues/ repo so we can address for everyone. In the meantime, please avoid the extra checks.

Eilon:
BTW log level guidance is available here: https://github.com/aspnet/Logging/wiki/Guidelines#log-levels

Regarding string concatenation, that doesn't work with resource strings and so should never be used for anything that goes in a resx file (regardless of whether it's ultimately localized).

And regarding the culture, there's no matter of taste or style, it's a correctness issue, per what @GrabYourPitchforks mentioned.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: I always do measure before speaking :-). Following are the results of checking a bool vs. calling into a method. I had to bump it up so high because at 500,000 I was getting '0' for the 'bool' check.

Checking bools: '00:00:00.0155706'
Iterations: '1000000'
Calling method: '00:00:01.1094496'
Iterations: '1000000'

That said, I will back it out if everyone thinks it's not worth it.

brentschmaltz:
clarifying commnet on what @GrabYourPitchforks said, using String.Format(CultureInfo.CurrentCulture, ...) is correct or NOT correct?

Eilon:
CurrentCulture is incorrect. I think you want InvariantCulture.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
Eilon: I think we've actually just done some optimizations to Logging to make that operation more performant. Also, the percentage of time in a real request spent doing these checks is negligible, so the delta is presumably insignificant either way.

brentschmaltz:
Yes, it is not a style thing but a correctness thing. I have always used InvariantCulture when the data was never presented to the user and CurrentCulture when the data will be displayed to the user, such as a log entry. Are you saying that is not correct and we should always use InvariantCulture?

Tratcher:
The problem is that there are two users, the remote client and the local admin. CurrentCulture represents the remote client, correct?

brentschmaltz:
Welcome back Tratcher :-)

I thought CurrentCulture represents the culture of the current thread. For logs, I am tailoring them for the dev who uses them for debugging. I know GrabYourPitchforks spoke about the system admin, but I was leaning towards the dev.  
If we use InvariantCulture, we are essentially using English culture for all logs when they are localized.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: I am fine with backing of the bool check.

Eilon:
Right, the CurrentCulture will be set to whatever the person using the site/service/API wanted. I.e. someone accessing an API in France could conceivably get French messages. But the developer should always get a single culture. Currently we've said that's the InvariantCulture, but in the future we might want a way for the developer to configure in which culture they want to see all their log messages.

brentschmaltz:
OK, do you want to introduce extensibility for 'culture' now or just go with InvariantCulture.
If we go later, we can always set the default to InvariantCulture.

Eilon:
Right, the dev doing debugging certainly doesn't want to use the culture of the person visiting the site.

I wouldn't do anything special here: just use InvariantCulture.

PinpointTownes:
> I wouldn't do anything special here: just use InvariantCulture.

Or better, don't use `string.Format` at all and let aspnet/Logging handle it internally? :smile:

brentschmaltz:
@PinpointTownes yep, I am going to investigate just that, let the logger do the string format.  Then we just need to modify the logger.

Eilon:
Oh right yeah that's best. Do as little work as possible here, let the logging system do it.

brentschmaltz:
Ding Ding Ding, we have a winner :-)

PinpointTownes:
Yeahhhh. What's the prize? :smile: 

(I'll start flooding with feedback/questions when the bool checks are reverted :smiling_imp:)

src/Microsoft.AspNet.Authentication.OpenIdConnect/INonceCache.cs
```diff
@@ -5,8 +5,10 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
 {
     public interface INonceCache
```
PinpointTownes: Have you considered replacing this interface by `IDistributedCache`?

https://github.com/aspnet/Caching/blob/dev/src/Microsoft.Framework.Caching.Distributed/IDistributedCache.cs

src/Microsoft.AspNet.Authentication.OpenIdConnect/INonceCache.cs
```diff
@@ -5,8 +5,10 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
...
     {
-        string AddNonce(string nonce);
+        bool TryAddNonce(string nonce);
+
         bool TryRemoveNonce(string nonce);
+
         bool HasNonce(string nonce);
```
PinpointTownes: This method doesn't seem to be used. Is it useful for external code?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -129,6 +128,14 @@ public string Caption
         }
 
         /// <summary>
+        /// An optional constrained path on which to process the authentication callback.
+        /// If not provided and RedirectUri is available, this value will be generated from RedirectUri.
+        /// </summary>
+        /// <remarks>If you set this value, then the <see cref="OpenIdConnectAuthenticationHandler"/> will only listen for posts at this address. 
+        /// If the IdentityProvider does not post to this address, you may end up in a 401 -> IdentityProvider -> Client -> 401 -> ...</remarks>
```
PinpointTownes: Heh? :smile:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -218,11 +219,22 @@ public OpenIdConnectProtocolValidator ProtocolValidator
...
+
+        /// <summary>
         /// Gets or sets the 'resource'.
         /// </summary>
         public string Resource { get; set; }
 
         /// <summary>
+        /// Gets or sets the 'response_mode'.
+        /// </summary>
+        public string ResponseMode { get; set; }
```
PinpointTownes: It's great to bring back `ResponseMode`, but `response_mode=query` support doesn't seem to be added to this PR.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -233,10 +245,12 @@ public OpenIdConnectProtocolValidator ProtocolValidator
         public string Scope { get; set; }
 
         /// <summary>
-        /// Gets or sets the AuthenticationScheme used when creating the <see cref="System.Security.Claims.ClaimsIdentity"/>.
+        /// Gets or sets the SignInScheme which will be used to set the <see cref="System.Security.Claims.ClaimsIdentity.AuthenticationType"/>.
```
PinpointTownes: This comment is not clear and should mention that `SignInScheme` actually corresponds to the final authentication middleware that will persist the identity (usually a cookie middleware).

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -98,7 +98,7 @@ protected override void ApplyResponseGrant()
                     string redirectUri = notification.ProtocolMessage.CreateLogoutRequestUrl();
                     if (!Uri.IsWellFormedUriString(redirectUri, UriKind.Absolute))
                     {
-                        _logger.LogWarning("The logout redirect URI is malformed: {0}", (redirectUri ?? "null"));
+                        _logger.LogWarning(Resources.OIDCH_0051_RedirectUriLogoutIsNotWellFormed, (redirectUri ?? "null"));
```
PinpointTownes: I think it would be better to simply pass `redirectUri` and let the logging block handle the default value.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -115,28 +115,38 @@ protected override void ApplyResponseChallenge()
         /// Responds to a 401 Challenge. Sends an OpenIdConnect message to the 'identity authority' to obtain an identity.
         /// </summary>
         /// <returns></returns>
+        /// <remarks>Uses log id's OIDCH-0026 - OIDCH-0050, next num: 37</remarks>
         protected override async Task ApplyResponseChallengeAsync()
         {
+            if (_logger.IsEnabled(LogLevel.Debug))
+            {
+                _logger.LogDebug(format: Resources.OIDCH_0026_ApplyResponseChallengeAsync, args: this.GetType().ToString());
```
PinpointTownes: Is `GetType().ToString()` that slow?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -235,180 +276,215 @@ protected override AuthenticationTicket AuthenticateCore()
             {
                 IFormCollection form = await Request.ReadFormAsync();
```
PinpointTownes: @Tratcher is this still useful? Isn't `Request.ReadFormAsync` supposed to do some internal caching?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -235,180 +276,215 @@ protected override AuthenticationTicket AuthenticateCore()
...
                 Request.Body.Seek(0, SeekOrigin.Begin);
-
-                openIdConnectMessage = new OpenIdConnectMessage(form);
+                message = new OpenIdConnectMessage(form);
             }
 
-            if (openIdConnectMessage == null)
+            if (message == null)
```
PinpointTownes: If you plan to bring back `response_mode`, you should add `GET` support.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -235,180 +276,215 @@ protected override AuthenticationTicket AuthenticateCore()
...
-                    throw new OpenIdConnectProtocolException(
-                        string.Format(CultureInfo.InvariantCulture,
-                                      openIdConnectMessage.Error,
-                                      Resources.Exception_OpenIdConnectMessageError, openIdConnectMessage.ErrorDescription ?? string.Empty, openIdConnectMessage.ErrorUri ?? string.Empty));
-                }
-
-                // code is only accepted with id_token, in this version, hence check for code is inside this if
-                // OpenIdConnect protocol allows a Code to be received without the id_token
-                if (string.IsNullOrWhiteSpace(openIdConnectMessage.IdToken))
+                AuthenticationProperties properties = GetPropertiesFromState(message.State);
```
PinpointTownes: `var`?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -147,15 +157,27 @@ protected override void ApplyResponseChallenge()
                 properties = new AuthenticationProperties(ChallengeContext.Properties);
             }
 
-            if (string.IsNullOrEmpty(properties.RedirectUri))
+            string redirectToUse = null;
```
PinpointTownes: Hum, all this stuff seems really odd: remember that `properties.RedirectUri` corresponds to the final location the user agent will be redirect to after the whole authentication/authorization process, not the `redirect_uri` (as defined by the OIDC specs) used for the callback part.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -484,39 +565,42 @@ private void RememberNonce(string nonce)
...
+        private string ReadNonceCookie(string nonce)
         {
-            if (nonceExpectedValue == null)
+            if (nonce == null)
             {
                 return null;
             }
 
-            foreach (var nonceKey in Request.Cookies.Keys)
+            else
```
PinpointTownes: This `else` is obviously useless and adds unnecessary nesting :sob:

PinpointTownes:
A general remark: IMHO, you should avoid using named parameters for logging for at least two reasons:
* It's really ugly and unnecessary :baby_chick: 
* You'll have to submit a new PR to use the new extensions offered by this PR: https://github.com/aspnet/Logging/pull/101/files

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -163,14 +185,15 @@ protected override void ApplyResponseChallenge()
                 _configuration = await Options.ConfigurationManager.GetConfigurationAsync(Context.RequestAborted);
             }
 
-            OpenIdConnectMessage openIdConnectMessage = new OpenIdConnectMessage
+            var message = new OpenIdConnectMessage
             {
                 ClientId = Options.ClientId,
                 IssuerAddress = _configuration == null ? string.Empty : (_configuration.AuthorizationEndpoint ?? string.Empty),
```
PinpointTownes: `_configuration?.AuthorizationEndpoint ?? string.Empty`

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -235,180 +276,215 @@ protected override AuthenticationTicket AuthenticateCore()
...
+                    TokenValidationParameters validationParameters = Options.TokenValidationParameters.Clone();
+                    if (_configuration != null)
+                    {
+                        if (string.IsNullOrWhiteSpace(validationParameters.ValidIssuer))
+                        {
+                            validationParameters.ValidIssuer = _configuration.Issuer;
+                        }
+                        else if (!string.IsNullOrWhiteSpace(_configuration.Issuer))
+                        {
+                            validationParameters.ValidIssuers = (validationParameters.ValidIssuers == null ? new[] { _configuration.Issuer } : validationParameters.ValidIssuers.Concat(new[] { _configuration.Issuer }));
```
PinpointTownes: `validationParameters.ValidIssuers?.Concat(new[] { _configuration.Issuer }) ?? new[] { _configuration.Issuer }`

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -235,180 +276,215 @@ protected override AuthenticationTicket AuthenticateCore()
...
+                        }
 
-                AuthenticationTicket ticket;
-                SecurityToken validatedToken = null;
-                ClaimsPrincipal principal = null;
-                JwtSecurityToken jwt = null;
+                        validationParameters.IssuerSigningKeys = (validationParameters.IssuerSigningKeys == null ? _configuration.SigningKeys : validationParameters.IssuerSigningKeys.Concat(_configuration.SigningKeys));
```
PinpointTownes: `validationParameters.IssuerSigningKeys?.Concat(_configuration.SigningKeys) ?? _configuration.SigningKeys`

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -235,180 +276,215 @@ protected override AuthenticationTicket AuthenticateCore()
...
-                        jwt = validatedToken as JwtSecurityToken;
-                        if (jwt == null)
+                        if (validator.CanReadToken(message.IdToken))
                         {
-                            throw new InvalidOperationException("Validated Security Token must be a JwtSecurityToken was: " + (validatedToken == null ? "null" : validatedToken.GetType().ToString()));
+                            principal = validator.ValidateToken(message.IdToken, validationParameters, out validatedToken);
+                            jwt = validatedToken as JwtSecurityToken;
+                            if (jwt == null)
+                            {
+                                _logger.LogError(format: Resources.OIDCH_0010_ValidatedSecurityTokenNotJwt, args: (validatedToken == null ? "null" : validatedToken.GetType().ToString()));
```
PinpointTownes: `validatedToken?.GetType().ToString()`

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -235,180 +276,215 @@ protected override AuthenticationTicket AuthenticateCore()
...
+                                throw new InvalidOperationException(string.Format(CultureInfo.InvariantCulture, Resources.OIDCH_0010_ValidatedSecurityTokenNotJwt, (validatedToken == null ? "null" : validatedToken.GetType().ToString())));
```
PinpointTownes: Same remark.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
Tratcher: This code is fine. Yes ReadFormAsync caches the results internally, but he's not really caching here, he's just making each individual line easy to read by using intermediate variables.

PinpointTownes: My remark was rather about the `Request.Body.Seek(0, SeekOrigin.Begin);` call, which is not necessary if `Request.ReadFormAsync` caches the result and doesn't need to read the request stream again (which wouldn't work for streamed requests).

Tratcher: Ah. While we have formalized the request buffering and caching behavior, we haven't officially decided who is responsible for re-winding the request stream after it's read and buffered. Even if ReadFormAsync caches the parsed form, that doesn't help another API like Nancy that may try to read the stream directly. I think this line should move into ReadFormAsync, but that hasn't happened yet.

src/Microsoft.AspNet.Authentication.OpenIdConnect/INonceCache.cs
brentschmaltz: I haven't. Feel free to open an issue.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: It's OK to have this a user can write their own handler.

src/Microsoft.AspNet.Authentication.OpenIdConnect/INonceCache.cs
brentschmaltz: It is used in the handler.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: I am not sure what you are asking?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: I want the comment short and not link it to another middleware that may or may not be present.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: That assumes all loggers will check for null before formatting. I choose not to make that assumption.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: I choose to wrap any strings creations.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: interesting syntax.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: When that happens, we can then make the change.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: We are not planning to add GET support for this handler, but users can write their own handler.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: not a necessary change.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: I find code like that hard to debug.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: I suppose it's a style thing.  Is there any bug or pref advantage?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: I like that syntax.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: That is true, will adjust.

src/Microsoft.AspNet.Authentication.OpenIdConnect/INonceCache.cs
PinpointTownes: I must be blind, I can't find it :see_no_evil:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: I'm just saying that this remark is not totally clear :smile:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: If they write their own handler, they also can write their own options.

This property had been removed for clarity: http://katanaproject.codeplex.com/workitem/313

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: We should discuss this if it is not clear. One way that properties.RedirectUri gets set is when a user issues a challenge, as opposed to hitting the [Authorize] attribute. In that case, users have expressed the need and desire to be able to set the redirect_uri from the challenge overriding the value on Options. If you think we should distinguish between the two then we should perhaps introduce a new value to capture this intent and keep the two separate.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: In most cases, it will be present, because the OIDC (just like the other social middleware) doesn't issue any cookie: persistence is always delegated to another middleware.

And BTW, you SHOULDN'T set `TokenValidationParameters.AuthenticationType` using `SignInScheme` but with `AuthenticationScheme`. See my PR: https://github.com/aspnet/Security/pull/168/files#diff-960390b7edd1b3a479fbd7fcfd07fbe8R55

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
PinpointTownes: I'm pretty sure that's how (semantic) logging should work.

@Eilon any thoughts? :smile:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
PinpointTownes: Sure. But on the other hand, it would make it compliant with the latest guidelines :smile: 

>The var keyword is to be used as much as the compiler will allow.

https://github.com/aspnet/Home/wiki/Engineering-guidelines#usage-of-the-var-keyword

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
PinpointTownes: Nope, it's just clearer (this syntax is now used massively in aspnet/dnx)

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
PinpointTownes: I'm not sure why one would override the `redirect_uri` parameter from the challenge context (it can be changed from `RedirectToIdentityProviderNotification`), but if you specify an explicit value for `Options.RedirectUri`, the OIDC handler will only listen on this path and ignore the values you could use in `AuthenticationProperties`, no?

src/Microsoft.AspNet.Authentication.OpenIdConnect/INonceCache.cs
PinpointTownes: Done: https://github.com/aspnet/Security/issues/212

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
HaoK: yeah there's no point introducing a property if you don't use it immediately, as pinpoint says, they can always add this if they want it themselves

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
HaoK: You don't have to mention specifically cookie middleware, but it would be good to have more detail given that this is one of the more critical properties which will break everything if misconfigured

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
HaoK: I don't think we special case null strings when logging anywhere else, so again, consistency should win in this case

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
HaoK: I thought we agreed to get rid of all enabled checks?

HaoK:
+1 to getting rid of named arguments in logging, it adds no value and just makes the logging calls look longer and weirder.  

I share your pain about following shared coding/style guidelines (my personal pet peeve is the hated { living on its own line).  But given where this code lives, it needs to look/feel like all of repros.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
Eilon: Yeah there should be no need. Ultimately it's typically called by string.Format, which is totally fine for nulls. The rule of thumb is that if we don't do things anywhere else, no need to do them here either.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: So they don't have to hook the notification.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: fine, if and when it crashes, I will send them your way.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: When we figure out what to say, we will make a change.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
Eilon: If it crashes here it probably means there's a bug elsewhere in the system that we need to fix.

src/Microsoft.AspNet.Authentication.OpenIdConnect/INonceCache.cs
brentschmaltz: if (!Options.NonceCache.TryRemoveNonce(nonce))

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: Then make a suggestion.

src/Microsoft.AspNet.Authentication.OpenIdConnect/INonceCache.cs
PinpointTownes: My question was about HasNonce, not TryRemoveNonce :sunglasses:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: Actually I am on the fence on this, but I am tired of discussing it, so I will remove it.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: Actually we didn't. If we feel justified it is ok to wrap it.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: I have not issue with people doing this, but I don't see any reason to change right now.

brentschmaltz:
Where is the code/style guideline published?

Eilon:
@brentschmaltz https://github.com/aspnet/Home/wiki/Engineering-guidelines

brentschmaltz:
I don't see anything about these types of issues: 
 validatedToken?.GetType().ToString() 
vs.
 (validatedToken == null ? "null" : validatedToken.GetType().ToString()

HaoK:
As pinpoint already pointed out, everyone else is using the ? operator heavily, its just as clear and more compact.

validatedToken?.GetType().ToString()

Eilon:
The team overall has decided to embrace C# 6 features rather heavily. It tends to make code more terse and generally easier to read (once you get used to it :smile: ).

brentschmaltz:
This is fine commentary, but the coding guidelines should reflect the required style.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: I agree with there may be a bug somewhere, which we may or may not have caught before shipping, which may or may not cause an issue or failure. I have found that developing with a defensive attitude helps with the user experience and helps avoid low grade errors. At the end of the day customer experience rules.

src/Microsoft.AspNet.Authentication.OpenIdConnect/INonceCache.cs
brentschmaltz: Yes you are correct, its TryAddNonce.  AddNonce can be removed.
Thanks.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: @PinpointTownes in review, I think it is confusing. I scaled back to just the 'local' redirect. I think part of the confusion is that Option.RedirectUri has the same name as AuthenticationProperties.RedirectUri but they are used for different purposes. Let's deal with dynamic 'redirect_uri' separately.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -115,28 +115,37 @@ protected override void ApplyResponseChallenge()
         /// Responds to a 401 Challenge. Sends an OpenIdConnect message to the 'identity authority' to obtain an identity.
         /// </summary>
         /// <returns></returns>
+        /// <remarks>Uses log id's OIDCH-0026 - OIDCH-0050, next num: 37</remarks>
```
PinpointTownes: Is it useful to expose this remark in the public documentation?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -147,12 +156,22 @@ protected override void ApplyResponseChallenge()
                 properties = new AuthenticationProperties(ChallengeContext.Properties);
             }
 
-            if (string.IsNullOrEmpty(properties.RedirectUri))
+            if (!string.IsNullOrWhiteSpace(properties.RedirectUri))
             {
+                _logger.LogDebug(format: Resources.OIDCH_0030_Using_Properties_RedirectUri, args: properties.RedirectUri);
+            }
+            else if (Options.DefaultToCurrentUriOnRedirect)
```
PinpointTownes: Is `DefaultToCurrentUriOnRedirect` still useful? I'd personally remove it and bring back the old behavior (i.e always use `properties.RedirectUri` when present).

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
PinpointTownes: Sad decision :sob:

GET support was probably the most wanted feature concerning the OIDC client middleware (at least on JabbR). You must keep in mind that `form_post` is not supported everywhere (e.g Google or Ping Identity with some `response_type`).

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -235,180 +273,215 @@ protected override AuthenticationTicket AuthenticateCore()
...
                     {
-                        validationParameters.ValidIssuers = (validationParameters.ValidIssuers == null ? new[] { _configuration.Issuer } : validationParameters.ValidIssuers.Concat(new[] { _configuration.Issuer }));
+                        _logger.LogInformation(message: Resources.OIDCH_0009_SecurityTokenReceivedNotificationSkipped);
+                        return null;
                     }
 
-                    validationParameters.IssuerSigningKeys = (validationParameters.IssuerSigningKeys == null ? _configuration.SigningKeys : validationParameters.IssuerSigningKeys.Concat(_configuration.SigningKeys));
-                }
+                    // Copy and augment to avoid cross request race conditions for updated configurations.
+                    TokenValidationParameters validationParameters = Options.TokenValidationParameters.Clone();
```
PinpointTownes: `var`?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -235,180 +273,215 @@ protected override AuthenticationTicket AuthenticateCore()
...
-                        jwt = validatedToken as JwtSecurityToken;
-                        if (jwt == null)
+                        if (validator.CanReadToken(message.IdToken))
                         {
-                            throw new InvalidOperationException("Validated Security Token must be a JwtSecurityToken was: " + (validatedToken == null ? "null" : validatedToken.GetType().ToString()));
+                            principal = validator.ValidateToken(message.IdToken, validationParameters, out validatedToken);
+                            jwt = validatedToken as JwtSecurityToken;
+                            if (jwt == null)
+                            {
+                                _logger.LogError(format: Resources.OIDCH_0010_ValidatedSecurityTokenNotJwt, args: validatedToken?.GetType().ToString() ?? null);
```
PinpointTownes: Adding ` ?? null` is not useful, it will still return `null` :smile:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -235,180 +273,215 @@ protected override AuthenticationTicket AuthenticateCore()
...
+                                throw new InvalidOperationException(string.Format(CultureInfo.InvariantCulture, Resources.OIDCH_0010_ValidatedSecurityTokenNotJwt, validatedToken?.GetType().ToString() ?? null));
```
PinpointTownes: Same remark.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -235,180 +273,215 @@ protected override AuthenticationTicket AuthenticateCore()
...
-                if (Options.UseTokenLifetime)
-                {
-                    // Override any session persistence to match the token lifetime.
-                    DateTime issued = validatedToken.ValidFrom;
-                    if (issued != DateTime.MinValue)
+                    // Rename?
+                    if (Options.UseTokenLifetime)
                     {
-                        ticket.Properties.IssuedUtc = issued;
+                        DateTime issued = validatedToken.ValidFrom;
```
PinpointTownes: Not important, but you can use var here too.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -235,180 +273,215 @@ protected override AuthenticationTicket AuthenticateCore()
...
+                        if (issued != DateTime.MinValue)
+                        {
+                            ticket.Properties.IssuedUtc = issued;
+                        }
+
+                        DateTime expires = validatedToken.ValidTo;
```
PinpointTownes: Same thing here.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -235,180 +273,215 @@ protected override AuthenticationTicket AuthenticateCore()
...
+                    };
 
-                if (securityTokenValidatedNotification.Skipped)
-                {
-                    return null;
+                    Options.ProtocolValidator.Validate(jwt, protocolValidationContext);
                 }
 
-                var protocolValidationContext = new OpenIdConnectProtocolValidationContext
+                if (message.Code != null)
```
PinpointTownes: Have you considered adding a built-in HTTP client to automagically retrieve an access token using the authorization code? It would be really useful.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -218,11 +217,22 @@ public OpenIdConnectProtocolValidator ProtocolValidator
...
+
+        /// <summary>
         /// Gets or sets the 'resource'.
         /// </summary>
         public string Resource { get; set; }
 
         /// <summary>
+        /// Gets or sets the 'response_mode'.
+        /// </summary>
+        public string ResponseMode { get; private set; }
```
PinpointTownes: I'm definitely not sure that exposing `ResponseMode` is really useful, specially since you can't change it (and BTW, the comment is wrong, you can't set it).

brentschmaltz: No, but you can get it. If you want it changed, open an issue.

brentschmaltz:
@PinpointTownes @HaoK I missed how much you dislike the arg names I will change that.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: Yes that feature is next, we have a working prototype, work on that is blocked until this PR is complete.

PinpointTownes: Wonderful :+1:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: If properties.RedirectUri is present, it will be used.

PinpointTownes: Only if `DefaultToCurrentUriOnRedirect` is `true`. IMHO, allowing to disable `properties.RedirectUri` from the options is not really useful.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -233,10 +243,12 @@ public OpenIdConnectProtocolValidator ProtocolValidator
...
 
         /// <summary>
-        /// Gets or sets the AuthenticationScheme used when creating the <see cref="System.Security.Claims.ClaimsIdentity"/>.
+        /// Gets or sets the SignInScheme which will be used to set the <see cref="System.Security.Claims.ClaimsIdentity.AuthenticationType"/>.
         /// </summary>
         public string SignInScheme
         {
+            // [brentschmaltz] - it is not clear to me why we need both AuthenticationScheme and SignInScheme.
+            // and, why this maps to AuthenticationType.
             get { return TokenValidationParameters.AuthenticationType; }
```
PinpointTownes: I fixed `SignInScheme` in my PR: https://github.com/PinpointTownes/Security/commit/d959eab910a188f9b2cac4b425572fb43decc5ca

brentschmaltz:
@Eilon @HaoK @PinpointTownes We have spent a lot of time on style and related items but not bugs, I include a 'codestyle' project in Wilson so that developers can easily confirm to Wilsons style. Can we do the same here?

Eilon:
@brentschmaltz unfortunately StyleCop doesn't support C# 6. We worked on trying to add support for it but it was too complex.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: I am not ignoring this, but there are some other issues in the runtime that when addressed I am hoping will make this moot.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: I know, but GET has known weaknesses that our security protocol experts have advised us against implementing. Is there any advantage to GET?

PinpointTownes: An advantage? Well, only one: it's supported everywhere :smile: 

That would be great if the experts you mentioned could share more information about the weaknesses `response_mode=query` has/may have. Specially since it's heavily used by the OAuth2 (Google/Facebook) and OAuth1 social providers.

PinpointTownes: If you think GET is too weak, we might add built-in support but disable it by default.

brentschmaltz:
@HaoK @Eilon @PinpointTownes I would like to merge this in, any pushback?

PinpointTownes:
@brentschmaltz not particularly. I'll clone and play with the final bits in the next hours :video_game:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
```diff
@@ -115,28 +115,37 @@ protected override void ApplyResponseChallenge()
...
             if (ShouldConvertChallengeToForbidden())
             {
+                _logger.LogDebug(Resources.OIDCH_0027_401_ConvertedTo_403);
                 Response.StatusCode = 403;
                 return;
             }
 
             if (Response.StatusCode != 401)
             {
+                _logger.LogDebug(Resources.OIDCH_0028_StatusCodeNot401, Response.StatusCode.ToString());
```
davidfowl: Why are you doing .ToString everywhere?

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
```diff
@@ -0,0 +1,647 @@
...
+            ILoggerFactory loggerFactory,
+            IOptions<OpenIdConnectAuthenticationOptions> options,
+            ConfigureOptions<OpenIdConnectAuthenticationOptions> configureOptions
+            )
+        : base(next, dataProtectionProvider, loggerFactory, options, configureOptions)
+        {
+            DataProtectionProvider = dataProtectionProvider;
+            OptionsPublic = options.Options;
+            SecureDataFormat = Options.StateDataFormat;
+            THIS = this;
```
Eilon: Hmm some of this test code looks suspicious to me. We can't have singletons in tests because tests run in parallel.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
```diff
@@ -0,0 +1,647 @@
...
+            logger.IsEnabled(LogLevel.Critical).ShouldBe<bool>(true);
+            logger.IsEnabled(LogLevel.Debug).ShouldBe<bool>(false);
+            logger.IsEnabled(LogLevel.Error).ShouldBe<bool>(false);
+            logger.IsEnabled(LogLevel.Information).ShouldBe<bool>(false);
+            logger.IsEnabled(LogLevel.Verbose).ShouldBe<bool>(false);
+            logger.IsEnabled(LogLevel.Warning).ShouldBe<bool>(false);
+        }
+
+        [Fact]
+        public async Task AuthenticateCore()
```
Eilon: 1. Should use a `[Theory]` for this to test multiple variations
2. Test names should say what they're doing and what they're expecting. E.g. `AuthenticateCore_WithXyz_DoesSuchAndSuch`.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
```diff
@@ -0,0 +1,647 @@
...
+            {
+                services.AddDataProtection();
+            });
+        }
+    }
+
+    /// <summary>
+    /// Extension specifies <see cref="OpenIdConnectAuthenticationMiddlewarePublic"/> as the middleware.
+    /// </summary>
+    public static class OpenIdConnectAuthenticationExtensions
```
Eilon: These unit tests seem to need quite a lot of helper classes. Is there a design change we can make to the runtime types to make them easier to test? That usually leads to better design anyway.

brentschmaltz:
good, then I will merge it in.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
brentschmaltz: Probably, but I would have to think about it. Is that blocking for this merge?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: Does that cause a bug?

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
brentschmaltz: This test is atomic.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
Eilon: It will cause the data type to change (obviously), which would affect loggers that store strongly-typed data. E.g. some loggers will store int's as int's, which would allow for richer querying of logs (e.g. `select * from logs where log.statusCode >= 400 && log.statusCode < 500`).

Eilon: (As such it's not strictly *wrong*, but its making the data potentially less useful in some logging systems.)

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
brentschmaltz: I agree using [Theory] is a good way to go. I have a lot of tests to write, this is just the beginning.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
Eilon: Eh, this in particular isn't blocking, but should be cleaned up soon.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
Eilon: Yeah I think this is find for starters. Sometimes it's easier to factor it once there are more tests because the patterns are clearer.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
Eilon: Discussing this via email.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
```diff
@@ -0,0 +1,766 @@
...
+        /// </summary>
+        /// <param name="name">configuration to return</param>
+        /// <returns></returns>
+        public OpenIdConnectAuthenticationOptions GetNamedOptions(string name)
+        {
+            return _options;
+        }
+    }
+
+    public class CustomConfigureOptions : ConfigureOptions<OpenIdConnectAuthenticationOptions>
```
HaoK: Why do you need a custom configure options that doesn't do anything different than the base?

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
```diff
@@ -0,0 +1,766 @@
...
+                {
+                    AuthorizationCodeReceived = (notification) =>
+                    {
+                        notification.SkipToNextMiddleware();
+                        return Task.FromResult<object>(null);
+                    }
+                };
+        }
+
+        private static void DefaultOptions(OpenIdConnectAuthenticationOptions options)
```
HaoK: This should be called something more like SetDefaults

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
```diff
@@ -0,0 +1,766 @@
...
+        /// <param name="options">Options which control the processing of the OpenIdConnect protocol and token validation.</param>
+        /// <param name="loggerFactory">custom loggerFactory</param>
+        /// <returns>The application builder</returns>
+        public static IApplicationBuilder UseCustomOpenIdConnectAuthentication(this IApplicationBuilder app, IOptions<OpenIdConnectAuthenticationOptions> options, ILoggerFactory loggerFactory)
+        {
+            return app.UseMiddleware<CustomOpenIdConnectAuthenticationMiddleware>(options, loggerFactory);
+        }
+    }
+
+    public class OpenIdConnectAuthenticationContext : IAuthenticateContext
```
HaoK: What's using this class?  I can't find any other references to this

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
```diff
@@ -0,0 +1,766 @@
...
+
+        public void NotAuthenticated()
+        {
+        }
+    }
+
+    /// <summary>
+    /// Provides a Facade over IOptions
+    /// </summary>
+    public class Options : IOptions<OpenIdConnectAuthenticationOptions>
```
HaoK: Why do you need this?  You should be able to use the real options, or you could use Moq

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -233,12 +243,15 @@ public OpenIdConnectProtocolValidator ProtocolValidator
...
+        /// Gets or sets the SignInScheme which will be used to set the <see cref="System.Security.Claims.ClaimsIdentity.AuthenticationType"/>.
         /// </summary>
-        public string SignInScheme { get; set; }
+        public string SignInScheme
+        {
+            // [brentschmaltz] - it is not clear to me why we need both AuthenticationScheme and SignInScheme.
+            // and, why this maps to AuthenticationType.
+            get { return TokenValidationParameters.AuthenticationType; }
+            set { TokenValidationParameters.AuthenticationType = value; }
+        }
```
HaoK: @PinpointTownes didn't you just change this?  I think this PR might need to be rebased against dev

PinpointTownes: @HaoK absolutely: https://github.com/aspnet/Security/commit/5e03a6c1ad2d94a1e6c4757e0931d199761942a8#diff-a32201aa730db12f1448f183d71510adR241

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
Eilon: BTW recall that Moq doesn't work on CoreCLR, so if these tests should run there, sometimes it's best to just hand-write mocks (if needed).

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: @PinpointTownes contact Mike Jones, one of the authors of the JSON web spec.  He can help you out with reasoning about 'GET'

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
brentschmaltz: The reason is I want to share the Action<OpenIdConnectAuthenticationOptions> for ConfigOptions and IOptions.  Both paths through the runtime need to produce the same results.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: Nice catch, I will fix it.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
brentschmaltz: I will be using it when I get to the Authenticate part.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
brentschmaltz: Good Question. Extensibility.
I want to make sure that my class is called.
Currently, name isn't used, future tests will start making use of the name.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
brentschmaltz: That is good reasoning, I will change it.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
PinpointTownes: Security considerations are already listed on the OAuth2 multiple response_mode specs (http://openid.net/specs/oauth-v2-multiple-response-types-1_0.html):

> There are security implications to encoding response values in the query string. The HTTP Referer header includes query parameters, and so any values encoded in query parameters will leak to third parties. Thus, while it is safe to encode an Authorization Code as a query parameter when using a Confidential Client (because it can't be used without the Client Secret, which third parties won't have), more sensitive information such as Access Tokens and ID Tokens MUST NOT be encoded in the query string. In no case should a set of Authorization Response parameters whose default Response Mode is the fragment encoding be encoded using the query encoding. 

As mentioned, supporting `response_mode=query` + `response_type=code` is safe. If it was not the case, `query` wouldn't be the default `response_mode` for `response_type=code`. 

Blocking GET callbacks exposing `token` or `id_token` would be enough to make sure safe `response_type`/`response_mode` combinations are used.

(related topic: https://katanaproject.codeplex.com/SourceControl/network/forks/manfredsteyer/OIDCResponseModeQuery/contribution/7009)

/cc @selfissued

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs
```diff
@@ -99,10 +99,7 @@ public string Caption
         public PathString CallbackPath { get; set; }
 
         /// <summary>
-        /// Gets or sets the authentication scheme corresponding to the middleware
-        /// responsible of persisting user's identity after a successful authentication.
-        /// This value typically corresponds to a cookie middleware registered in the Startup class.
-        /// When omitted, <see cref="ExternalAuthenticationOptions.SignInScheme"/> is used as a fallback value.
+        /// Gets or sets the name of another authentication middleware which will be responsible for actually issuing a user <see cref="System.Security.Claims.ClaimsIdentity"/>.
```
PinpointTownes: Huh, why does this PR contain already merged commits? :cry: 

/cc @HaoK

HaoK: Rebasing messes up PRs, there are ways to fix it (force push)

HaoK: Another option is to just create a new PR and branch with squashed changes for final review

PinpointTownes: I'm definitely not a git expert, but it seems that @brentschmaltz did a `git merge` instead of a `git rebase`, which causes these weird issues: https://github.com/brentschmaltz/Security/commit/efaef8952a09c4ec0681db5e0a3c28cb1d96270f

brentschmaltz: I am also not a git expert, it did a pull git went off on its own.
I will tidy it up guys.

brentschmaltz: @PinpointTownes @HaoK it looks like when I did the pull, it hit a merge conflict. The updated versions are showing as 'unstagged'.  I will commit them and push to my fork.

brentschmaltz:
@HaoK Is there anything blocking this pull request?

HaoK:
I don't think there's anything particularly blocking no

brentschmaltz:
@HaoK Can you merge it? I don't think I have the ability to execute the merge.

HaoK:
You've checked in changes before, so you must have permission.  Just you will want to squash all your changes together before you push to dev (when you merge this branch, just do git merge --squash), and then it should just be one change that you push to dev

brentschmaltz:
I till try, I had pushed to Katana, but never aspnet.

HaoK:
Merged 501bd4ff105ee3985219a01d772f737bffc3e563

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
leastprivilege: I don't see a use case for redirect_uri not being set - it is a MANDATORY protocol parameter in OIDC.

PR #213, opened 2015-04-08
React to Challenge changes

HaoK:
cc @Tratcher


Tratcher:
:shipit:

davidfowl:
Why does this fail on appveyor?

HaoK:
How does appveyor work exactly?  This pr depends on a pr from http abstractions as well.

davidfowl:
AppVeyor represents what you get if you clone and build. It isn't using the volatile feed so sweeping changes will cause breaks until coherence passes.

HaoK:
9ce84d39c2c852e418eae13a6a984159343b4162

Issue #118, opened 2015-01-15
Integrate cookie chunking APIs with new strongly typed header values


Tratcher:
See: https://github.com/aspnet/HttpAbstractions/issues/167
Microsoft.AspNet.Security.Cookies.Infrastructure.ChunkingCookieManager could also benifit from using the SetCookieHeaderValue.

Tratcher:
@muratg

PR #216, opened 2015-04-13
Use common cookie header formatters.

Tratcher:
#118 
@muratg @davidfowl


Tratcher:
*ping*
@Eilon

src/Microsoft.AspNet.Authentication.Cookies/Infrastructure/ChunkingCookieManager.cs
```diff
@@ -143,19 +141,16 @@ public void AppendResponseCookie([NotNull] HttpContext context, [NotNull] string
                 quoted = true;
                 value = RemoveQuotes(value);
             }
-            var escapedValue = Uri.EscapeDataString(value);
+            var escapedValue = UrlEncoder.Default.UrlEncode(value);
```
Eilon: Any reason it goes straight to the default instead of acquiring the service?

Tratcher:
Updated to resolve the encoder from services.

Eilon:
:shipit:

PR #222, opened 2015-04-21
Cleanup: Move Logger to base AuthenticationHandler + use more var

HaoK:
- Push Logger to base handler class/Initialize
- Pass to use var consistently


HaoK:
@Tratcher

src/Microsoft.AspNet.Authentication/AuthenticationMiddleware.cs
```diff
@@ -28,6 +30,8 @@ public abstract class AuthenticationMiddleware<TOptions> where TOptions : Authen
             {
                 Options = options.Options;
             }
+            Logger = loggerFactory.CreateLogger(this.GetType().FullName);
```
Tratcher: This gets the name of the derived type, correct?

Tratcher:
Squash and :shipit:

src/Microsoft.AspNet.Authentication/AuthenticationMiddleware.cs
HaoK: Yah it gets the actual type that's being constructed.

HaoK:
30d350da26d5c40e9e160ccb2cd120303da5baf1

Issue #161, opened 2015-02-22
Prefer the new encoders over Uri.EscapeDataString


GrabYourPitchforks:
There are some call sites in this repo that invoke Uri.EscapeDataString. Instead of calling this static API, the caller should make an effort to use the IUrlEncoder that has been configured by the application developer, as it could have more permissive or restrictive logic than that of Uri.EscapeDataString.

If you have access to an IServiceProvider instance, import the `Microsoft.AspNet.WebUtilities.Encoders` namespace (in the `Microsoft.AspNet.Http.Extensions` package) and call the `IServiceProvider.GetUrlEncoder()` extension method

If you do not have access to an IServiceProvider, you can get a default static singleton instance of an `IUrlEncoder` via the `UrlEncoder.Default` static property (namespace `Microsoft.AspNet.WebUtilities.Encoders`, project `Microsoft.AspNet.WebUtilities`).

HaoK:
87c31c5526edee557997f03c7b13d46cb5d1d893

PR #225, opened 2015-04-23
Switch to IUrlEncoder

HaoK:
Fixes https://github.com/aspnet/Security/issues/161

@Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
@@ -10,6 +10,7 @@ using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Authentication;
 using Microsoft.Framework.Internal;
 using Microsoft.Framework.Logging;
+using Microsoft.Framework.WebEncoders;
 
 namespace Microsoft.AspNet.Authentication
 {
@@ -50,6 +51,8 @@ namespace Microsoft.AspNet.Authentication
 
         protected ILogger Logger { get; private set; }
 
+        protected IUrlEncoder UrlEncoder { get; private set; }
+
         internal AuthenticationOptions BaseOptions
         {
             get { return _baseOptions; }
@@ -61,12 +64,13 @@ namespace Microsoft.AspNet.Authentication
 
         public bool Faulted { get; set; }
 
-        protected async Task BaseInitializeAsync([NotNull] AuthenticationOptions options, [NotNull] HttpContext context, [NotNull] ILogger logger)
+        protected async Task BaseInitializeAsync([NotNull] AuthenticationOptions options, [NotNull] HttpContext context, [NotNull] ILogger logger, [NotNull] IUrlEncoder encoder)
         {
             _baseOptions = options;
             Context = context;
             RequestPathBase = Request.PathBase;
             Logger = logger;
+            UrlEncoder = encoder;
 
             RegisterAuthenticationHandler();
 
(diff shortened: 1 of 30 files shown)


src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -61,12 +64,13 @@ internal AuthenticationOptions BaseOptions
...
         public bool Faulted { get; set; }
 
-        protected async Task BaseInitializeAsync([NotNull] AuthenticationOptions options, [NotNull] HttpContext context, [NotNull] ILogger logger)
+        protected async Task BaseInitializeAsync([NotNull] AuthenticationOptions options, [NotNull] HttpContext context, [NotNull] ILogger logger, [NotNull] IUrlEncoder encoder)
         {
             _baseOptions = options;
             Context = context;
             RequestPathBase = Request.PathBase;
             Logger = logger;
+            UrlEncoder = encoder;
```
Tratcher: Was this actually needed in every handler?

Tratcher: I only count 3 handlers that use it, cookies, facebook, and OIDC. That's probably insufficient for putting it in the base class.

HaoK: Also Twitter uses it, so 4.  Feels kind of the logger to me, we want to encourage people to use the UrlEncoder right?  So that said, if you feel super strongly that we should keep the base handler properties as slim as possible, I can remove it.  But generally I don't really see how it hurts if we already have 4 handlers ourselves that use it...

Tratcher: Part of the trouble is who should add it as a service? Look at my last PR that added this to cookies. Most of the samples had to add the service themselves, which seems inappropriate to require for middleware that don't even use it.  If you can improve the service registration then that will help.

HaoK: yeah I don't disagree, but actually you already added that service to all of the samples since cookie requires it, so I didn't really have to do anything extra other than update the bearer one I think.  Perhaps what we need is a dumping ground AddSharedAuthentication extension which we can put things like DataProtection/WebEncoders and just call that consistently.  I don't really like how ConfigureCookieAuthentication is secretly adding web encoder today... that is pretty non standard :)

Tratcher: Try it and let's see what you come up with.

HaoK: Another option we could do, is remove the ability to configure options as part of Use, so force an explicitly AddCookie/Auth for each instance of middleware, the downside of this is with multiple instances of the same middleware, but I guess the overload would be UseXyzAuthentication(string authenticationScheme) and they would just be required to specify that correctly as part of the ConfigureXyzAuthentication(() => { }, authenticationScheme

Tratcher: Not my favorite option. I'd rather go with AddAuthenticationServices()

HaoK: Updated with AddAuthentication helper which adds both DataProtection and WebEncoders

HaoK: We can also just roll AddAuthentication into AddIdentity, so this won't have to be called explicitly except when identity isn't being used.

Tratcher:
:shipit:

HaoK:
87c31c5526edee557997f03c7b13d46cb5d1d893

Issue #214, opened 2015-04-09
[Design] OpenIdConnect needs a definition for agumenting the 'state' parameter on request.


brentschmaltz:
If a users wants to add to the 'state' parameter when sending a request, there is no easy way to do this.

PinpointTownes:
Adding custom properties to `AuthenticationProperties` when calling `HttpResponse.Challenge` sounds rather easy for me. Do you have something particular in mind?

brentschmaltz:
That works fine with Challenge(...). I am thinking about a user that is hooking the RedirectToIdentityProviderNotification. The runtime sets OpenIdConnectMessage.State with protected AuthenticationProperties before the call. How would a user add to the state. I am thinking that it maybe we may want to pass the AuthenticationProperties to the notification and protect afterwards.

Eilon:
This should be done as part of the recent PR that was merged.

PR #233, opened 2015-05-01
Changes for State and test organization

brentschmaltz:
Updated from clean base.

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
@@ -57,7 +57,7 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
                     _configuration = await Options.ConfigurationManager.GetConfigurationAsync(Context.RequestAborted);
                 }
 
-                var openIdConnectMessage = new OpenIdConnectMessage()
+                var message = new OpenIdConnectMessage()
                 {
                     IssuerAddress = _configuration == null ? string.Empty : (_configuration.EndSessionEndpoint ?? string.Empty),
                     RequestType = OpenIdConnectRequestType.LogoutRequest,
@@ -69,21 +69,20 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
                 var properties = new AuthenticationProperties(signout.Properties);
                 if (!string.IsNullOrEmpty(properties.RedirectUri))
                 {
-                    openIdConnectMessage.PostLogoutRedirectUri = properties.RedirectUri;
+                    message.PostLogoutRedirectUri = properties.RedirectUri;
                 }
                 else if (!string.IsNullOrWhiteSpace(Options.PostLogoutRedirectUri))
                 {
-                    openIdConnectMessage.PostLogoutRedirectUri = Options.PostLogoutRedirectUri;
+                    message.PostLogoutRedirectUri = Options.PostLogoutRedirectUri;
                 }
 
-                var notification = new RedirectToIdentityProviderNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions>(Context, Options)
+                var notification = new RedirectToIdentityProviderNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions>(Context, Options, message, properties);
+                if (Options.Notifications.RedirectToIdentityProvider != null)
                 {
-                    ProtocolMessage = openIdConnectMessage
-                };
+                    await Options.Notifications.RedirectToIdentityProvider(notification);
+                }
 
-                await Options.Notifications.RedirectToIdentityProvider(notification);
-
-                if (!notification.HandledResponse)
+                if (!notification.HandledResponse || !notification.Skipped)
                 {
                     var redirectUri = notification.ProtocolMessage.CreateLogoutRequestUrl();
                     if (!Uri.IsWellFormedUriString(redirectUri, UriKind.Absolute))
@@ -174,18 +173,36 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
                 ClientId = Options.ClientId,
                 IssuerAddress = _configuration?.AuthorizationEndpoint ?? string.Empty,
                 RedirectUri = Options.RedirectUri,
-                // [brentschmaltz] - this should be a property on RedirectToIdentityProviderNotification not on the OIDCMessage.
+                // [brentschmaltz] - #215 this should be a property on RedirectToIdentityProviderNotification not on the OIDCMessage.
                 RequestType = OpenIdConnectRequestType.AuthenticationRequest,
                 Resource = Options.Resource,
                 ResponseMode = Options.ResponseMode,
                 ResponseType = Options.ResponseType,
-                Scope = Options.Scope,
-                State = OpenIdConnectAuthenticationDefaults.AuthenticationPropertiesKey + "=" + UrlEncoder.UrlEncode(Options.StateDataFormat.Protect(properties))
+                Scope = Options.Scope
             };
 
             if (Options.ProtocolValidator.RequireNonce)
             {
                 message.Nonce = Options.ProtocolValidator.GenerateNonce();
+            }
+
+            var redirectToIdentityProviderNotification =
+                new RedirectToIdentityProviderNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions>(Context, Options, message, properties);
+
+            await Options.Notifications.RedirectToIdentityProvider(redirectToIdentityProviderNotification);
+            if (redirectToIdentityProviderNotification.HandledResponse)
+            {
+                Logger.LogInformation(Resources.OIDCH_0034_RedirectToIdentityProviderNotificationHandledResponse);
+                return;
+            }
+            else if (redirectToIdentityProviderNotification.Skipped)
+            {
+                Logger.LogInformation(Resources.OIDCH_0035_RedirectToIdentityProviderNotificationSkipped);
+                return;
+            }
+
+            if (!string.IsNullOrWhiteSpace(message.Nonce))
+            {
                 if (Options.NonceCache != null)
                 {
                     if (!Options.NonceCache.TryAddNonce(message.Nonce))
@@ -200,21 +217,15 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
                 }
             }
 
-            var redirectToIdentityProviderNotification = new RedirectToIdentityProviderNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions>(Context, Options)
+            // If 'OpenIdConnectMessage.State' is null, then the user never set it in the notification. Just set the state
+            if (string.IsNullOrWhiteSpace(redirectToIdentityProviderNotification.ProtocolMessage.State))
             {
-                ProtocolMessage = message
-            };
-
-            await Options.Notifications.RedirectToIdentityProvider(redirectToIdentityProviderNotification);
-            if (redirectToIdentityProviderNotification.HandledResponse)
-            {
-                Logger.LogInformation(Resources.OIDCH_0034_RedirectToIdentityProviderNotificationHandledResponse);
-                return;
+                redirectToIdentityProviderNotification.ProtocolMessage.State = OpenIdConnectAuthenticationDefaults.AuthenticationPropertiesKey + "=" + Options.StateDataFormat.Protect(properties);
             }
-            else if (redirectToIdentityProviderNotification.Skipped)
+            // the user did set 'OpenIdConnectMessage.State, add a parameter to the state
+            else
             {
-                Logger.LogInformation(Resources.OIDCH_0035_RedirectToIdentityProviderNotificationSkipped);
-                return;
+                redirectToIdentityProviderNotification.ProtocolMessage.State = OpenIdConnectAuthenticationDefaults.AuthenticationPropertiesKey + "=" + Options.StateDataFormat.Protect(properties) + "&userstate=" + redirectToIdentityProviderNotification.ProtocolMessage.State;
             }
 
             var redirectUri = redirectToIdentityProviderNotification.ProtocolMessage.CreateAuthenticationRequestUrl();
@@ -222,6 +233,8 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
             {
                 Logger.LogWarning(Resources.OIDCH_0036_UriIsNotWellFormed, redirectUri);
             }
+
+            Logger.LogDebug(Resources.OIDCH_0037_RedirectUri, redirectUri);
 
             Response.Redirect(redirectUri);
         }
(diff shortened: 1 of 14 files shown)


src/Microsoft.AspNet.Authentication/Notifications/RedirectToIdentityProviderNotification.cs
```diff
@@ -1,16 +1,64 @@
...
+            AuthenticationProperties = properties;
+        }
+
+        /// <summary>
+        /// Gets or sets the <see cref="{TMessage}"/>.
+        /// </summary>
+        /// <exception cref="ArgumentNullException">if 'value' is null.</exception>
+        public TMessage ProtocolMessage
         {
+            get { return _message; }
```
HaoK: Is the public set really necessary on Message/AuthProperties?  can these not be private and only settable via the ctor?  If the sets are only needed for unit internal set, rather than adding null checks seems better.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationNotifications.cs
```diff
@@ -44,7 +46,19 @@ public OpenIdConnectAuthenticationNotifications()
         /// <summary>
         /// Invoked to manipulate redirects to the identity provider for SignIn, SignOut, or Challenge.
         /// </summary>
-        public Func<RedirectToIdentityProviderNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions>, Task> RedirectToIdentityProvider { get; set; }
+        public Func<RedirectToIdentityProviderNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions>, Task> RedirectToIdentityProvider
```
HaoK: So why is this Func special compared to all of the other Func setters that allow setting to null?  I'm not arguing necessarily that they should be nullable, I'm just pointing out that they should all behave the same

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/DerivedTypes.cs
```diff
@@ -0,0 +1,360 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
```
HaoK: I'm not sure why we need all of these derived classes in this file, seems to be a sign of code smell...

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/DerivedTypes.cs
```diff
@@ -0,0 +1,360 @@
...
+            }
+        }
+
+        protected override AuthenticationHandler<OpenIdConnectAuthenticationOptions> CreateHandler()
+        {
+            return _handler ?? base.CreateHandler();
+        }
+    }
+
+    public class CustomOpenIdConnectMessage : OpenIdConnectMessage
```
HaoK: Why can't we just test the OpenIdConnectMessage directly?

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/DerivedTypes.cs
```diff
@@ -0,0 +1,360 @@
...
+            return _protect;
+        }
+
+        AuthenticationProperties ISecureDataFormat<AuthenticationProperties>.Unprotect(string protectedText)
+        {
+            return _unprotect;
+        }
+    }
+
+    public class CustomConfigureOptions : ConfigureOptions<OpenIdConnectAuthenticationOptions>
```
HaoK: This doesn't seem to do anything, why preemptively introduce this before its needed?

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
```diff
@@ -443,12 +275,12 @@ private static TestServer CreateServer(IOptions<OpenIdConnectAuthenticationOptio
             );
         }
 
-        private static TestServer CreateServer(CustomConfigureOptions configureOptions, ILoggerFactory loggerFactory)
+        private static TestServer CreateServer(CustomConfigureOptions configureOptions, IUrlEncoder encoder, ILoggerFactory loggerFactory, OpenIdConnectAuthenticationHandler handler = null)
```
HaoK: Instead of CustomConfigureOptions, this should really take either `Action<OpenIdOptions>`, or if you want full control over all services to be able to configure everything `Action<IServiceCollection>`.  All of this test code should be setting up the services via DI and replacing them with the appropriate test services, not injecting them via some new parallel custom test code path that is different from the normal code path...

HaoK: Something that looks like this should let you replace everything you need via the service collection

https://github.com/aspnet/Identity/blob/1a74ca6972ac13569bc348672aefd19502e147e3/test/Microsoft.AspNet.Identity.InMemory.Test/FunctionalTest.cs#L157

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/Extensions.cs
```diff
@@ -0,0 +1,41 @@
...
+using Microsoft.Framework.Logging;
+using Microsoft.Framework.OptionsModel;
+using Microsoft.Framework.WebEncoders;
+
+namespace Microsoft.AspNet.Authentication.Tests.OpenIdConnect
+{
+    /// <summary>
+    /// Extension specifies <see cref="CustomOpenIdConnectAuthenticationMiddleware"/> as the middleware.
+    /// </summary>
+    public static class OpenIdConnectAuthenticationExtensions
```
HaoK: You should be testing the real extensions/usage and real middleware classes/handlers, this feels wrong that the tests are all using custom derived versions of everything...

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
```diff
@@ -443,12 +275,12 @@ private static TestServer CreateServer(IOptions<OpenIdConnectAuthenticationOptio
...
         {
             return TestServer.Create(
                 app =>
                 {
-                    app.UseCustomOpenIdConnectAuthentication(configureOptions, loggerFactory);
+                    app.UseCustomOpenIdConnectAuthentication(configureOptions, encoder, loggerFactory, handler);
```
HaoK: This should be configuring the test server exactly the same as real apps, just replacing the services with test services so the middleware picks up the correct encoder.  This doesn't look like the right direction for tests to be going

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -200,21 +217,15 @@ protected override void ApplyResponseChallenge()
...
-                return;
+                redirectToIdentityProviderNotification.ProtocolMessage.State = OpenIdConnectAuthenticationDefaults.AuthenticationPropertiesKey + "=" + Options.StateDataFormat.Protect(properties);
             }
-            else if (redirectToIdentityProviderNotification.Skipped)
+            // the user did set 'OpenIdConnectMessage.State, add a parameter to the state
+            else
             {
-                Logger.LogInformation(Resources.OIDCH_0035_RedirectToIdentityProviderNotificationSkipped);
-                return;
+                redirectToIdentityProviderNotification.ProtocolMessage.State = OpenIdConnectAuthenticationDefaults.AuthenticationPropertiesKey + "=" + Options.StateDataFormat.Protect(properties) + "&userstate=" + redirectToIdentityProviderNotification.ProtocolMessage.State;
```
suhasj: This might have been discussed by why is `UrlEncoder.UrlEncode` not used to encode state data in this change ?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationNotifications.cs
brentschmaltz: Nothing special with this Func. I stopped here to see if folks would like to see this pattern for all the Func's.
My testing showed it is very easy to generate code that will cause a null-ref.

Tratcher: In theory we could use [NotNull] for these instead. It would be less work for us, and it would generate the same output as a compile step.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/DerivedTypes.cs
brentschmaltz: Can you please explain this statement: " sign of code smell...". 
The derived classes allow me to test generation of a wide set of variations.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationNotifications.cs
HaoK: :+1: for [NotNull]

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/DerivedTypes.cs
brentschmaltz: this is for extensibility.  users can set the ProtocolMessage in notifications.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
brentschmaltz: The testing here is at the level of UnitTest / Functional where I want tight control over exactly the flow in each method.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationNotifications.cs
brentschmaltz: This can work in this case, but for the notifications themselves, since the Ticket and Message can be set, we need some stronger validation.

I noticed in FB and Goodle MW and Handlers we don't check for null in public methods.

src/Microsoft.AspNet.Authentication/Notifications/RedirectToIdentityProviderNotification.cs
brentschmaltz: This allows users to read a message and modify it.  If you think about it, this allows some very powerful extensibility.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/DerivedTypes.cs
brentschmaltz: I didn't complete my testing before vacation and saw no reason to remove something I will put right back.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/Extensions.cs
brentschmaltz: The purpose of these tests is at the Unit / Functional level. Once I am satisfied that this level is functioning correctly, I will bump up to the next level.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
brentschmaltz: These are Unit / Functional tests, those will follow.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
brentschmaltz: @suhasj When the message is created by OpenIdConnectMessage.CreateAuthenticationRequestUrl formatting will take place.  If we encode here, it will be duplicated.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationNotifications.cs
Eilon: All public methods need to check for null. If we have some that don't, then that's a bug.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/DerivedTypes.cs
```diff
@@ -0,0 +1,360 @@
...
+
+                _logEntries.Add(logEntry);
+                Debug.WriteLine(logEntry.ToString());
+            }
+        }
+
+        public List<LogEntry> Logs { get { return _logEntries; } }
+    }
+
+    public class CustomLoggerFactory : ILoggerFactory
```
Eilon: I think a test version of this exists in https://github.com/aspnet/Testing/tree/dev/src/Microsoft.Framework.Logging.Testing (TestLogger and NullLogger)

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/DerivedTypes.cs
```diff
@@ -0,0 +1,360 @@
...
+            return _logger;
+        }
+
+        public CustomLogger Logger { get { return _logger; } }
+    }
+
+    /// <summary>
+    /// Allows for controlling methods and capturing and exposing artifacts.
+    /// </summary>
+    public class CustomOpenIdConnectAuthenticationHandler : OpenIdConnectAuthenticationHandler
```
Eilon: Should look into using Moq for these types of mocks. It makes the actual test code a lot more readable because they end up being largely self-contained. Using extra types can make it more difficult to understand a test because it's not always clear what's being customized vs. what's standard behavior.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationNotifications.cs
brentschmaltz: Then we need to take action as you will find a lot of such code.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/DerivedTypes.cs
brentschmaltz: Absolutely when I kick to Functional and Scenario testing. For unit testing I am looking for tight control.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/DerivedTypes.cs
brentschmaltz: I am not actually testing a custom logger, but am using a custom logger to obtain and compare logs to ensure correctness.

brentschmaltz:
@HaoK @Eilon @tushargupta51 What is needed to pull this code?

Tratcher:
https://github.com/aspnet/Security/issues/214

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
Tratcher: We need to parse the state back out for them on the other end.

brentschmaltz:
Reissued as: 
https://github.com/aspnet/Security/pull/301

PR #241, opened 2015-05-05
Using [NotNull] and  'nameof' operator

hishamco:
Using the "nameof" operator instead of magic string will help us if the variable change in the future for whatever the reason

### DIFF
--- a/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationOptions.cs
+++ b/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationOptions.cs
@@ -5,6 +5,7 @@ using System;
 using System.Diagnostics.CodeAnalysis;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Authentication.Cookies.Infrastructure;
+using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Authentication.Cookies
 {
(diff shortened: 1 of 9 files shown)


msotclas:
Hi __@hishamco__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <span>You've already signed the contribution license agreement. Thanks!</span>
        <p>The agreement was validated by Microsoft Open Technologies, Inc. and real humans are currently evaluating your PR.</p>

TTYL, MSOTBOT;

PinpointTownes:
Ahahah, @Eilon: 0, @dnfclas: 1 :smile:

Just a question: won't some of these null checks be replaced by `NotNull` in the near future?

(I know the EF team is massively using `NotNull` on properties, but it doesn't seem a common practice in the other repositories. Is this something you will encourage?)

Eilon:
@PinpointTownes in general yeah we're using `[NotNull]` so maybe it's a good idea to just do that change instead... I think maybe the `dnx` repo can't use it because of circular dependencies, but all the other repos that use `project.json` should...

And yeah, I'm learning how to use the new bot, and this was my test PR :smile:

hishamco:
@Eilon shall I replace them with [NotNull] like what I did in other repos

Eilon:
@hishamco might as well. As long as it was a `null` check, we should use `[NotNull]`.

hishamco:
@Eilon I think there are 3 places that I should use ```[NotNull]``` instead. I think the once that used Resources ```nameof```  will be the better fix. Am I right?

Eilon:
Yeah change to `[NotNull]` wherever appropriate, but for places where it's null-or-empty, or in some of the places here it's using `string.IsNullOrWhiteSpace` it should use a proper resource string with an appropriate message.

hishamco:
@Eilon I updated the PR to use ```[NotNull]``` and ```nameof``` operator

PR #244, opened 2015-05-05
provide authenticationType parameter to get sample work

Spoonin:
ClaimsIdentity needs to be instantiated with authenticationType parameter so it's IsAuthenticated value evaluated to true 
detailed - http://leastprivilege.com/2012/09/24/claimsidentity-isauthenticated-and-authenticationtype-in-net-4-5/


HaoK:
Looks like this has already been fixed, thanks

https://github.com/aspnet/Security/blob/dev/samples/CookieSample/Startup.cs

PR #247, opened 2015-05-06
Add custom requirements for Role/Names

HaoK:
- RequireRole/UserName now use the IsInRole/Name off the identity instead of the claim types
- Also cleaned up the existing handlers and combined them with the requirements to simplify service registration, our standard requirements are now themselves handlers (so we no longer need to register them as handlers)

cc @Tratcher @blowdart


Tratcher:
*shrug* :shipit:

blowdart:
I love that our automated builds test everywhere. Still, looks good to me.

But I think a squirrel is not a proper ship it icon, so

"Make it so Mr Hao."

HaoK:
434d158c7620fd75492ea5d03b6885246745b26c

PR #250, opened 2015-05-11
AuthZ: Replace params overloads with IEnumerable to enable sugar

HaoK:
Enables making resource optional in some sugar overloads

cc @blowdart @divega


HaoK:
Updated PR to also throw with empty AuthZPolicies instead of just always failing authZ

HaoK:
17deab142da56b656bb2cd84dce294ca15135397

Issue #236, opened 2015-05-01
Add standard sugar to enable configuring ClaimsTransformation via Use


HaoK:
Consistency with other authN middlewares

HaoK:
bb2e12a8e611e0f7a2f54ee2fab15373c02d6fca

PR #254, opened 2015-05-12
Add sugar for UseClaimsTransformation

HaoK:
Allows configuration of claims transformation inline with UseClaimsTransformation just like the other Auth middlewares

Fixes https://github.com/aspnet/Security/issues/236

cc @Tratcher @blowdart


Tratcher:
:shipit:

HaoK:
bb2e12a8e611e0f7a2f54ee2fab15373c02d6fca

PR #255, opened 2015-05-12
Tweak  SecurityHelper.AddUserPrincipal logic

HaoK:
Outcome of discussion with @blowdart and @lodejard:

The new merging behavior is the following:
- Respect and preserve the order and all identities of the new ClaimsPrincipal being added.
- Add all existing identities in context.User EXCEPT any that have no claims AND !IsAuthenticated
- Create a new ClaimsPrincipal and copy to it, instead of directly modifying the new principal

cc @davidfowl @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Authentication/SecurityHelper.cs
+++ b/src/Microsoft.AspNet.Authentication/SecurityHelper.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Linq;
 using System.Security.Claims;
 using Microsoft.AspNet.Http;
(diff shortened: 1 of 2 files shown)


Tratcher:
Why does the order matter, and why this order?  I would expect Add(1), Add(2), Add(3) to result in 1,2,3, not 3,2,1 like you're doing here.

HaoK:
The order only matters to the degree that we want the last call to Authenticate (where AutomaticAuthentication = true) ends up being the User.Identity.  We don't care so much about the rest of the ordering (so the easiest was just to add all of the new principal's identities first, and then copy over the rest from the existing principal).

HaoK:
Oh and the motivation surrounding this, was to make it so that HttpContext.User.Identity.IsAuthenticated would return true in the happy path

Tratcher:
:shipit:

Issue #101, opened 2014-11-20
OAuth: Expose the result of the token request to the notification


Tratcher:
When the code is exchanged for a token there may be additional data returned from the provider. Expose the whole `TokenResponse` in the `GetUserInformationAsync/Authenticated` notifications.

Example: Google OpenId -> Oauth2 migrations
https://katanaproject.codeplex.com/workitem/359
https://katanaproject.codeplex.com/SourceControl/network/forks/suhasj/Katana/contribution/7719

Eilon:
@HaoK please work with @PinpointTownes on his PR so we can try to get it in by end of next week.

HaoK:
Merged 9bb8b6114697f14a42e747acd1d1a41e6eed5af1

Issue #47, opened 2014-08-25
Clean up notification extensibility model


Tratcher:
The original batch of social auth middleware (Facebook, Twitter, etc.) all used three different extensibility models for their notifications (formerly Providers).  There was an `I*AuthenticationProvider` interface with methods and a *AuthenticationProvider class defining both virtual methods to be overridden and `On*` settable delegates. The second round (WsFed, ODIC) got rid of the interface and virtual methods, leaving only the settable delegates. Note you could still derive from the notification class, you would just need to set the delegates in your constructor.

HaoK:
@PinpointTownes does your PR address everything here?

PinpointTownes:
@HaoK not really. This ticket was about unifying `I*AuthenticationProvider` and `*AuthenticationNotifications`, but `I*AuthenticationProvider` has since been replaced by `I*AuthenticationNotifications`.

The question is now: do we want to have...
* an `I*AuthenticationNotifications` interface and a default `*AuthenticationNotifications` class exposing `Func` properties like the OAuth2 client middleware https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OAuth/Notifications/IOAuthAuthenticationNotifications.cs and https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OAuth/Notifications/OAuthAuthenticationNotifications.cs
* or a single `*AuthenticationNotifications` class that doesn't have any method and simply exposes `Func` properties, like the OIDC client middleware: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationNotifications.cs

The first approach is definitely less DRY but seems more flexible and allows advanced scenarios like interface interception via Dynamic Proxy.

Tratcher:
I recommend the second approach. It's much cleaner and I have never seen anyone utilize the extensibility provided by the first approach.

PinpointTownes:
@Tratcher here's a sample of subclassing a default `*AuthenticationProvider` class: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/blob/vNext/samples/Mvc/Mvc.Server/Providers/AuthorizationProvider.cs#L11

Overriding a method is much cleaner that assigning a delegate.

anurse:
@HaoK We need to do design on this. @glennc can assign an existing design meeting slot for this.

[...]

PR #257, opened 2015-05-14
Revisit OAuthAuthenticationHandler and add a new SaveTokensAsClaims option

PinpointTownes:
In the current bits, the OAuth2 client middleware automagically saves access and refresh tokens in the `ClaimsPrincipal` it issues, but this logic disappears when you override the default `OnGetUserInformation` notification: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationDefaults.cs#L13

This pull request removes `DefaultOnGetUserInformationAsync`, moves its logic directly to `OAuthAuthenticationHandler` and adds a new `SaveTokens` property to `OAuthAuthenticationOptions` to decide whether tokens should be stored in `ClaimsPrincipal` (and in the final authentication cookie) or not, independently of the `OnGetUserInformation` notification.

This will also make `OAuthAuthenticationHandler` inheritors more DRY, as they won't have to repeat the same snippet again and again: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/src/AspNet.Security.OAuth.GitHub/GitHubAuthenticationHandler.cs#L52-L70

/cc @HaoK @Tratcher


PinpointTownes:
@Eilon please tell me I won't have to sign a third CLA :smile: 

(I submitted a PR two months ago: https://github.com/aspnet/Security/pull/168)

samples/SocialSample/Startup.cs
```diff
@@ -218,7 +215,7 @@ 127.0.0.1 MsSecSample.localhost.this
             // Deny anonymous request beyond this point.
             app.Use(async (context, next) =>
             {
-                if (string.IsNullOrEmpty(context.User.Identity.Name))
+                if (context.User == null || !context.User.Identities.Any(identity => identity.IsAuthenticated))
```
PinpointTownes: Unrelated change, but I fixed it because the "GitHub-AccessToken" sample didn't work properly.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
```diff
@@ -159,14 +160,44 @@ protected override AuthenticationTicket AuthenticateCore()
             return new TokenResponse(oauth2Token);
         }
 
-        protected virtual async Task<AuthenticationTicket> GetUserInformationAsync(AuthenticationProperties properties, TokenResponse tokens)
+        protected virtual async Task<AuthenticationTicket> CreateTicketAsync(AuthenticationProperties properties, TokenResponse tokens)
```
PinpointTownes: Not a "necessary" change, but I believe that `CreateTicketAsync` is a far more appropriate name for this method.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
```diff
@@ -159,14 +160,44 @@ protected override AuthenticationTicket AuthenticateCore()
...
+            {
+                Principal = new ClaimsPrincipal(identity),
+                Properties = properties
             };
-            await Options.Notifications.GetUserInformationAsync(context);
-            return new AuthenticationTicket(context.Principal, context.Properties, Options.AuthenticationScheme);
+            
+            await Options.Notifications.GetUserInformationAsync(notification);
+
+            if (notification.Principal == null)
```
PinpointTownes: @Tratcher should we still allow `OAuthGetUserInformationContext.Principal` to be replaced?

Eilon:
@PinpointTownes you do need to sign a .NET Foundation CLA (we just changed the projects from MS Open Tech), but this is the last CLA we plan to have! Also, this new CLA is much simpler and quicker than the previous CLA. Sorry for the hassle!

PinpointTownes:
@Eilon meh, okay, but "ASP.NET 5" is not listed on this page: https://cla.dotnetfoundation.org/ :cry: 

(btw, it seems that the Contribution page has not been updated: https://github.com/aspnet/Home/blob/release/CONTRIBUTING.md#contributing-code-and-content)

Eilon:
@PinpointTownes sorry that link is out-of-date... I fixed it this morning. You want to use the fully automated system at https://cla2.dotnetfoundation.org/ .

PinpointTownes:
Ah. I had filed the wrong form: https://cla.dotnetfoundation.org/ (the first result on Google...)

@dnfclas thanks, dear bot :smile:

Eilon:
Thanks! BTW I contacted the bot operators to have the CLA sites clarified as to which site is for which. Hopefully they can get that resolved soon!

PinpointTownes:
@Eilon wonderful, thanks!

PinpointTownes:
@HaoK @Tratcher any remark concerning this PR? :cold_sweat:

samples/SocialSample/Startup.cs
Tratcher: context.User is guaranteed never to be null.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
```diff
@@ -159,14 +160,44 @@ protected override AuthenticationTicket AuthenticateCore()
...
+
+            var notification = new OAuthGetUserInformationContext(Context, Options, Backchannel, tokens)
+            {
+                Principal = new ClaimsPrincipal(identity),
+                Properties = properties
             };
-            await Options.Notifications.GetUserInformationAsync(context);
-            return new AuthenticationTicket(context.Principal, context.Properties, Options.AuthenticationScheme);
+            
+            await Options.Notifications.GetUserInformationAsync(notification);
```
Tratcher: Should this also be renamed?

samples/SocialSample/Startup.cs
PinpointTownes: Ah nice, I'll remove the unnecessary check then :smile:

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
PinpointTownes: It's not necessary but yep, it would be ideal. 

How about **CreateTicket**/**OnCreateTicket**?

Tratcher: Yes, OnCreateTicket.

Note that the deriving providers just call OnAuthenticated, they don't call this one directly. It would be nice if we could merge the two.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
Tratcher: Yes.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
```diff
@@ -159,14 +160,44 @@ protected override AuthenticationTicket AuthenticateCore()
             return new TokenResponse(oauth2Token);
         }
 
-        protected virtual async Task<AuthenticationTicket> GetUserInformationAsync(AuthenticationProperties properties, TokenResponse tokens)
+        protected virtual async Task<AuthenticationTicket> CreateTicketAsync(AuthenticationProperties properties, TokenResponse tokens)
         {
-            var context = new OAuthGetUserInformationContext(Context, Options, Backchannel, tokens)
+            var identity = new ClaimsIdentity(Options.AuthenticationScheme);
```
Tratcher: Use Options.ClaimsIssure. Compare https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Facebook/FacebookAuthenticationHandler.cs#L67

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs
```diff
@@ -109,5 +101,12 @@ public string Caption
         /// Gets or sets the type used to secure data handled by the middleware.
         /// </summary>
         public ISecureDataFormat<AuthenticationProperties> StateDataFormat { get; set; }
+
+        /// <summary>
+        /// Defines whether access and refresh tokens should be stored in the <see cref="ClaimsPrincipal"/>
+        /// after a successful authentication. You can set this property to <see cref="false"/> to reduce
+        /// the size of the final authentication cookie.
+        /// </summary>
+        public bool SaveTokens { get; set; } = true;
```
Tratcher: The deriving middleware won't honor this, correct? That seems weird.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
PinpointTownes: The `authenticationType` parameter is not really an issuer, but I'll fix it for consistency :smile:

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs
PinpointTownes: They will if they override `CreateTicketAsync` and call `base.CreateTicketAsync`.
Would you prefer a new `CreateTicketCoreAsync` method so inheritors don't have to call `base.CreateTicketAsync`?

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
PinpointTownes: Okay, then keeping this new null check is probably necessary.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs
Tratcher: They don't call base.CreateTicketAsync today. Take a look and make a suggestion.

PinpointTownes: Yup, I haven't updated the social providers yet ; I just wanted to make sure I was adopting the right design before :smile:

I'll update them in the next commits.

src/Microsoft.AspNet.Authentication.Facebook/FacebookAuthenticationHandler.cs
```diff
@@ -49,8 +49,20 @@ public FacebookAuthenticationHandler(HttpClient httpClient)
...
-        protected override async Task<AuthenticationTicket> GetUserInformationAsync(AuthenticationProperties properties, TokenResponse tokens)
+        protected override async Task<AuthenticationTicket> CreateTicketAsync(AuthenticationProperties properties, TokenResponse tokens)
         {
+            var ticket = await base.CreateTicketAsync(properties, tokens);
+            if (ticket == null)
+            {
+                return null;
+            }
+            
+            var identity = ticket.Principal?.Identity as ClaimsIdentity;
```
PinpointTownes: I'm not a huge fan of that, but `AuthenticationTicket.Principal` is not guaranteed to be non-null and `ClaimsPrincipal.Identity` may be null when no identity has been added to the principal.

@Tratcher should we update `AuthenticationTicket` to avoid these 2 cases or is this null check fine?

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs
PinpointTownes: Done. I'll port this option to the Twitter provider when everything is done.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
PinpointTownes: I'm not sure how to do that yet. `*Authenticated` events expose strongly-typed properties that we couldn't have in the generic `CreateTicket` notification.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
```diff
@@ -159,14 +160,44 @@ protected override AuthenticationTicket AuthenticateCore()
             return new TokenResponse(oauth2Token);
         }
 
-        protected virtual async Task<AuthenticationTicket> GetUserInformationAsync(AuthenticationProperties properties, TokenResponse tokens)
+        protected virtual async Task<AuthenticationTicket> CreateTicketAsync(AuthenticationProperties properties, TokenResponse tokens)
         {
-            var context = new OAuthGetUserInformationContext(Context, Options, Backchannel, tokens)
+            var identity = new ClaimsIdentity(Options.ClaimsIssuer);
```
HaoK: I'm not opposed to changing this back to AuthenticationScheme if we think that's more appropriate.  We don't really use AuthenticationType for anything anymore, other than null or not for IsAuthenticated

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
```diff
@@ -159,14 +160,44 @@ protected override AuthenticationTicket AuthenticateCore()
...
+
+            if (Options.SaveTokens)
             {
-                Properties = properties,
+                identity.AddClaim(new Claim("access_token", tokens.AccessToken, ClaimValueTypes.String, Options.ClaimsIssuer));
```
HaoK: Should these hardcoded claim types move to some Constants/Defaults?  I'm assuming we don't need the more extreme option of making these configurable...

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
HaoK: Since the context is called OAuthCreateTicket, should this method also convey that as well?  CreateOAuthTicketAsync?

HaoK: (and if so, we should pick a consistent naming scheme for both context and method :))

PinpointTownes: Well, the other notifications don't match the methods they are invoked from, so I'm not sure it's really important (`ApplyRedirect` from `ApplyResponseChallenge`, `ReturnEndpoint` from `InvokeReturnPathAsync`).

That said, if you really want me to, I'll change it ;)

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
PinpointTownes: Yup, that's a good idea.

How about `OAuthAuthenticationClaims` or `OAuthAuthenticationConstants`?

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
PinpointTownes: IMHO, `AuthenticationScheme` definitely sounds more appropriate, @Tratcher thinks we should use `Options.ClaimsIssuer` instead.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
HaoK: Well at this point, I was just asking whether that would be better.  Since this seems to be a more general issue, we certainly don't have to do this as part of this change, I'll file an issue for reviewing the names of the Oauth context/methods...

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
```diff
@@ -159,14 +160,44 @@ protected override AuthenticationTicket AuthenticateCore()
...
+
+            var notification = new OAuthCreateTicketContext(Context, Options, Backchannel, tokens)
+            {
+                Principal = new ClaimsPrincipal(identity),
+                Properties = properties
             };
-            await Options.Notifications.GetUserInformationAsync(context);
-            return new AuthenticationTicket(context.Principal, context.Properties, Options.AuthenticationScheme);
+            
+            await Options.Notifications.CreateTicket(notification);
```
HaoK: nit: typically public async methods need to end with Async (not my personal fav, but wearing my @divega hat today)

src/Microsoft.AspNet.Authentication.Facebook/FacebookAuthenticationHandler.cs
Tratcher: this code is OK. AuthTicket has scenarios that need to support a null principal.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs
HaoK: Maybe SaveTokensAsClaims or ExposeTokenClaims to be slightly more clear what this is doing and used for?  I don't think its super obvious what SaveTokens means by itself...

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
PinpointTownes: I agree, but notifications never use the `async` suffix: this method was the only exception, so I preferred removing the suffix for consistency.

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Cookies/Notifications/ICookieAuthenticationNotifications.cs

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationNotifications.cs

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
Tratcher: I don't think we need to specify OAuth in the name of methods on a class already named OAuth, it's just redundant.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
HaoK: I think @Tratcher mostly was arguing consistency since I changed everything to use ClaimsIssuer.  I recall that was something I questioned myself when doing it, so now is a good time to revisit to make sure we pick the most appropriate one to use...

Tratcher: Only because that's what we changed all of the others to. That's a debate I don't want to restart in this PR.  If it's out of date file a different bug.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs
PinpointTownes: @Tratcher any preference?

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
HaoK: Yeah just means we weren't consistent in following design guidelines, leave it for now, I'll clean it up seperately

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
HaoK: I think we pretty much have to go with OAuthAuthenticationDefaults to follow the naming pattern that exists here today.  I'd personally prefer Constants but probably should just be consistent.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs
Tratcher: SaveTokensAsClaims is more descriptive.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
PinpointTownes: Okay. One note though: constants exposed by `*AuthenticationDefaults` always have their counterpart in `*AuthenticationOptions`. Here, there's no option to change the claim names.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
HaoK: The main reason why I brought this up, is because something called "AuthenticationTicket CreateTicketAsync" I would assume is something comes from the base AuthenticationHandler.  Its a not obvious from the name that this is OAuth specific

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs
PinpointTownes: Renamed to `SaveTokensAsClaims`.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
HaoK: You did a good job convincing me that these should be configurable via the options now, if that was your intent :)

PinpointTownes: No, it wasn't :smile: 

Should we create a new setting per claim (`access_token`, `expires_in`, `refresh_token` and `token_type`) or should we use a more elaborated claims map dictionary?

PinpointTownes: Note that the social providers also have their own constant claims, that cannot be changed via options. Should we also make them configurable?

Tratcher: Noooo.... at least not yet.

HaoK: Yeah, in general I did this for identity with a specific setting per claim.  Should just consistently apply that across all the social providers

PinpointTownes: @Tratcher no to what? Making them configurable? Creating a new setting per claim? Using a claim map?

Tratcher: Feature creep... save it for a later PR.

PinpointTownes: That's @HaoK's fault :smile:

HaoK: Yeah well I just comment on stuff as I see it, but Chris is right, this should go in an issue/different PR :)

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs
```diff
@@ -77,7 +69,7 @@ public string Caption
         /// <value>
         /// The back channel timeout.
         /// </value>
-        public TimeSpan BackchannelTimeout { get; set; }
+        public TimeSpan BackchannelTimeout { get; set; } = TimeSpan.FromSeconds(60);
```
PinpointTownes: Unrelated change, but I'd like to make all these options virtual, so inheritors can replace the documentation and set values inline:

`/// <summary>My different summary</summary>
public override TimeSpan BackchannelTimeout { get; set; } = TimeSpan.FromSeconds(120);`

Tratcher: Are there any of these you actually expect to be overridden?

PinpointTownes: Yup, specially `AuthenticationScheme`, `Caption`, `ClaimsIssuer`, `CallbackPath`, `AuthorizationEndpoint`, `TokenEndpoint` and `UserInformationEndpoint`:

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/src/AspNet.Security.OAuth.GitHub/GitHubAuthenticationOptions.cs#L11-L19

HaoK: FxCop typically warns about setting virtual properties in ctors, and we don't fully control the derived classes so...

http://stackoverflow.com/questions/465436/is-it-ok-to-call-virtual-properties-from-the-constructor-of-a-nhibernate-entity

HaoK: So basically if you want to make them virtual, we can't use auto properties, and should set the fields to the defaults in the ctor instead of the property setter

PinpointTownes: Will FxCop even complain if we have no "explicit" constructor? ('cuz everything is set using the new C# 6 syntax)

HaoK: Good point, maybe the new C# property initialization turns into setting the field already, so we get this for free...

PinpointTownes: According to some ILSpying, I'd say that's the case:

```
// Microsoft.AspNet.Authentication.OAuth.OAuthAuthenticationOptions
[DebuggerBrowsable(DebuggerBrowsableState.Never), CompilerGenerated]
private TimeSpan <BackchannelTimeout>k__BackingField = TimeSpan.FromSeconds(60.0);
[DebuggerBrowsable(DebuggerBrowsableState.Never), CompilerGenerated]
private readonly IList<string> <Scope>k__BackingField = new List<string>();
[DebuggerBrowsable(DebuggerBrowsableState.Never), CompilerGenerated]
private bool <SaveTokensAsClaims>k__BackingField = true;
public OAuthAuthenticationOptions()
{
}
```

Tratcher:
Possibly related: https://github.com/aspnet/Security/issues/101

PinpointTownes:
@Tratcher I'm testing a new approach that merges `OnAuthenticated` and `OnCreateTicket`.
Feel free to tell me if you prefer the new design or the old one.

Regarding #101, I'll fix it in a future commit, when we're done with the general design.

PinpointTownes:
@Tratcher so, any preference? https://github.com/PinpointTownes/Security/commit/517125d8b82ac89de1d869e4f88c949d9931032c

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs
```diff
@@ -109,5 +101,12 @@ public string Caption
         /// Gets or sets the type used to secure data handled by the middleware.
         /// </summary>
         public ISecureDataFormat<AuthenticationProperties> StateDataFormat { get; set; }
+
+        /// <summary>
+        /// Defines whether access and refresh tokens should be stored in the <see cref="ClaimsPrincipal"/>
+        /// after a successful authentication. You can set this property to <see cref="false"/> to reduce
+        /// the size of the final authentication cookie.
+        /// </summary>
+        public bool SaveTokensAsClaims { get; set; } = true;
```
Tratcher: Should this really be enabled by default? It's not needed in the default app flow, and we're already having major cookie size problems.

PinpointTownes: I opted for `true` because it was the default behavior of `OnGetUserInformationAsync`, but yeah, it's probably better to disable it by default.

src/Microsoft.AspNet.Authentication.Google/Notifications/GoogleAuthenticationNotifications.cs
```diff
@@ -23,16 +24,33 @@ public GoogleAuthenticationNotifications()
...
+        /// Invoked after the provider successfully authenticates a user.
+        /// </summary>
+        /// <param name="context">Contains information about the login session as well as the user <see cref="ClaimsIdentity"/>.</param>
+        /// <returns>A <see cref="Task"/> representing the completed operation.</returns>
+        public sealed override async Task Authenticated(OAuthAuthenticatedContext context)
+        {
+            var notification = context as GoogleAuthenticatedContext;
+            if (notification == null)
+            {
+                throw new InvalidOperationException("OAuthAuthenticationHandler.CreateTicketAsync must be overridden " +
```
Tratcher: This error is a little strange. How would you realistically cause this?

PinpointTownes: Extremely easily: create your own OAuth2 provider, with your own notifications and your own `OnAuthenticated` (i.e with a subclass of `OAuthAuthenticatedContext`) but don't override `CreateTicketAsync` to invoke the `Authenticated` taking your "subclass context" instead of the default `OAuthAuthenticatedContext` and boom.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs
Tratcher: Actually, would it make sense to set it to true in the base options and false in the derived options? Otherwise the default behavior of the base class is an empty identity, no?

src/Microsoft.AspNet.Authentication.OAuth/Notifications/OAuthAuthenticatedContext.cs
```diff
@@ -39,11 +36,6 @@ public class OAuthAuthenticatedContext : BaseContext<OAuthAuthenticationOptions>
         }
 
         /// <summary>
-        /// Gets the JSON-serialized user.
-        /// </summary>
-        public JObject User { get; protected set; }
```
Tratcher: Why was User taken off the base context?

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs
PinpointTownes: It might be a bit surprising from a configuration perspective, but as long as it is correctly documented, it shouldn't be an issue.

I'll fix that. Am I allowed to use a virtual property? :smile:

src/Microsoft.AspNet.Authentication.Google/Notifications/GoogleAuthenticationNotifications.cs
Tratcher: Ok, but that only make sense if you copy pasted this method into your own notification class. I don't think we should add an exception that only fires for developer copy-paste errors. At most this would be a Debug.Assert.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs
Tratcher: Try it and we'll see :-)

src/Microsoft.AspNet.Authentication.OAuth/Notifications/OAuthAuthenticatedContext.cs
PinpointTownes: Because in the "generic" OAuth2 handler, there's no "automagical" user-info roundtrip, so you can't populate the `User` object :cry: 

(we could maybe keep it in the base event class and throw an exception when accessing the `User` property, but I'm not sure it's the right approach)

samples/SocialSample/Startup.cs
```diff
@@ -130,48 +131,45 @@ 127.0.0.1 MsSecSample.localhost.this
...
+                        var userResponse = await client.SendAsync(userRequest, notification.HttpContext.RequestAborted);
                         userResponse.EnsureSuccessStatusCode();
                         var text = await userResponse.Content.ReadAsStringAsync();
                         var user = JObject.Parse(text);
 
-                        var identity = new ClaimsIdentity(
-                            context.Options.AuthenticationScheme,
-                            ClaimsIdentity.DefaultNameClaimType,
-                            ClaimsIdentity.DefaultRoleClaimType);
+                        var identity = (ClaimsIdentity) notification.Principal.Identity;
```
HaoK: Is this cast needed?  Isn't ClaimsPrincipal.Identity a ClaimsIdentity already?

src/Microsoft.AspNet.Authentication.Facebook/Notifications/FacebookAuthenticationNotifications.cs
```diff
@@ -23,16 +24,33 @@ public FacebookAuthenticationNotifications()
         /// <summary>
         /// Gets or sets the function that is invoked when the Authenticated method is invoked.
         /// </summary>
-        public Func<FacebookAuthenticatedContext, Task> OnAuthenticated { get; set; }
+        public new Func<FacebookAuthenticatedContext, Task> OnAuthenticated { get; set; }
```
HaoK: Why not make the base virtual to allow override instead of new?

samples/SocialSample/Startup.cs
PinpointTownes: Sadly no, it's not (and it's really a shame!)

http://referencesource.microsoft.com/#mscorlib/system/security/claims/ClaimsPrincipal.cs,683

We could probably expose the identity in the notification class, but we would have two properties: one for the principal and one for identity. I'm not a huge fan of that, as they basically represent similar things.

src/Microsoft.AspNet.Authentication.Google/Notifications/GoogleAuthenticationNotifications.cs
HaoK: Is there any way we can consolidate the commonality of Facebook/Google into a new generic `SocialNotifications<TNotification>` so we don't basically have same code for each provider?

samples/SocialSample/Startup.cs
blowdart: Do *not* have an identity and a principal property.

src/Microsoft.AspNet.Authentication.Facebook/Notifications/FacebookAuthenticationNotifications.cs
PinpointTownes: 'cause it's not the same return type (`Func<FacebookAuthenticatedContext, Task>` vs `Func<OAuthAuthenticatedContext, Task>` for the base event)

HaoK: So my gut feeling is that we should try to find some way to share some of this mechanical code that's in all of the social providers, the notifications class for example seems like it would benefit from a generic base class so you don't have to do these ugly new's to make it return the right context.

PinpointTownes: Social providers basically use a subclass of the notifications base class to implement only one notification (that replaces the generic one defined by `OAuthAuthenticationNotifications`): `OnAuthenticated`.

If we can include the access or refresh tokens directly in `OAuthAuthenticatedContext`, we can't add any provider specific property there (like `UserId`, `UserName`, `Email`) as we have no guarantee that these properties will even be returned by the user info endpoint offered by the provider.

Another approach might be to remove `Facebook/Google/MSFT/AuthenticatedContext` and the "specific" `Authenticated` notifications to keep only `OAuthAuthenticatedContext`. But in this case, users would have to deal with the raw `JToken` payload, assuming providers always return JSON payloads, which is definitely not a guarantee.

Tratcher: Interesting, if OAuthAuthenticationNotifications was generic we wouldn't need any of these sub classes at all.

Tratcher: Oh, ug, except that generic becomes viral up to Options and elsewhere.

HaoK: Ok I don't want to cause more feature creep. I will take a look at it this week and see if there's anything we can do to simplify things.

PinpointTownes: @Tratcher yup, that's exactly why I tried to find another solution (aka "the ugly new keyword") :smile:

HaoK: yeah welcome to the hell generic hell that hit identity :)  I'll take a look and see if there's anything we can do, I really think we should git rid of the copy/paste with a few words changed classes if we can...

samples/SocialSample/Startup.cs
PinpointTownes: @blowdart yup, I clearly prefer having a single property, even if we need to have a ugly cast.

src/Microsoft.AspNet.Authentication.Facebook/Notifications/FacebookAuthenticationNotifications.cs
PinpointTownes: @HaoK @Tratcher do you want me to test the generic approach?

Tratcher: Feel free to try it offline, but kill it fast if it looks like it's getting out of hand.

PinpointTownes: Actually, it's not that ugly (specially if we remove the TNotifications generic parameter), but it's not semantically perfect:

If we replace `OAuthAuthenticationNotifications` by `OAuthAuthenticationNotifications<TUser>` and `Func<OAuthAuthenticatedContext, Task> OnAuthenticated` by `Func<OAuthAuthenticatedContext<TUser>, Task> OnAuthenticated`, it implies that we necessarily have a "User" property on `OAuthAuthenticatedContext`, which is not the case if the OAuth2 provider doesn't offer a user-info endpoint. Not to mention that when you don't override the default OAuth2 handler and use the delegate approach, you simply can't fill a user object before the `Authenticated` notification is actually triggered and handled in Startup.cs.

With generics, the "best" approach might be to take a new `TAuthenticatedContext` (hum) generic parameter, change `OnAuthenticated`'s signature to `Func<TAuthenticatedContext, Task> OnAuthenticated`, and expose provider-specific stuff in `*AuthenticatedContext` subclasses, as we do today.

@HaoK @Tratcher your thoughts?

PinpointTownes:
@HaoK @Tratcher new approach using a new generic parameter: https://github.com/PinpointTownes/Security/commit/466b9a59dac810e8223d760bf76a945e3342f988

(the code won't compile as I only updated the Google provider)

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationExtensions.cs
```diff
@@ -19,10 +19,10 @@ public static class OAuthAuthenticationExtensions
...
         /// <returns>The updated <see cref="IApplicationBuilder"/>.</returns>
-        public static IApplicationBuilder UseOAuthAuthentication([NotNull] this IApplicationBuilder app, [NotNull] string authenticationScheme, Action<OAuthAuthenticationOptions<IOAuthAuthenticationNotifications>> configureOptions = null)
+        public static IApplicationBuilder UseOAuthAuthentication([NotNull] this IApplicationBuilder app, [NotNull] string authenticationScheme, Action<OAuthAuthenticationOptions<IOAuthAuthenticationNotifications<OAuthAuthenticatedContext>, OAuthAuthenticatedContext>> configureOptions = null)
         {
-            return app.UseMiddleware<OAuthAuthenticationMiddleware<OAuthAuthenticationOptions<IOAuthAuthenticationNotifications>, IOAuthAuthenticationNotifications>>(
-                new ConfigureOptions<OAuthAuthenticationOptions<IOAuthAuthenticationNotifications>>(options =>
+            return app.UseMiddleware<OAuthAuthenticationMiddleware<OAuthAuthenticationOptions<IOAuthAuthenticationNotifications<OAuthAuthenticatedContext>, OAuthAuthenticatedContext>, IOAuthAuthenticationNotifications<OAuthAuthenticatedContext>, OAuthAuthenticatedContext>>(
```
Tratcher: Fail. Kill it.

blowdart: And burn the body.

HaoK: Need a lot of lipstick for this pig...  Can you add some sugar/empty stub classes to make this less gross?

HaoK: Actually most users never see/write this code, this is hidden inside of the oauth extension method, it wouldn't impact the github usage in the sample at all would it?

Tratcher: And anyone else implementing middleware, right?

Regardless, this is too ugly even for internal usage.

PinpointTownes: It looks far better when completely removing the generic `IOAuthAuthenticationNotifications` parameters: https://github.com/PinpointTownes/Security/commit/511334d4726e1aedd6deb0ed4240d1c85fe586eb

src/Microsoft.AspNet.Authentication.Google/GoogleAuthenticationOptions.cs
```diff
@@ -28,5 +29,12 @@ public GoogleAuthenticationOptions()
         /// access_type. Set to 'offline' to request a refresh token.
         /// </summary>
         public string AccessType { get; set; }
+
+        /// <summary>
+        /// Defines whether access and refresh tokens should be stored in the
+        /// <see cref="ClaimsPrincipal"/> instance after a successful authentication.
+        /// This option is turned off by default to reduce the size of the final authentication cookie.
+        /// </summary>
+        public override bool SaveTokensAsClaims { get; set; } = false;
```
HaoK: I don't know if I like overriding a property just to change the default to false...  Seems better to just keep the property non virtual and set it in the ctor...

src/Microsoft.AspNet.Authentication.OAuth/Notifications/OAuthAuthenticatedContext.cs
HaoK: Couldn't the user just be null in that case?

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs
```diff
@@ -109,5 +101,13 @@ public string Caption
...
         /// </summary>
         public ISecureDataFormat<AuthenticationProperties> StateDataFormat { get; set; }
+
+        /// <summary>
+        /// Defines whether access and refresh tokens should be stored in the
+        /// <see cref="ClaimsPrincipal"/> instance after a successful authentication.
+        /// You can set this property to <see cref="false"/> to reduce
+        /// the size of the final authentication cookie.
+        /// </summary>
+        public virtual bool SaveTokensAsClaims { get; set; } = true;
```
HaoK: Yeah I don't think an auto property really needs to be virtual

src/Microsoft.AspNet.Authentication.Google/GoogleAuthenticationOptions.cs
PinpointTownes: If you look carefully at the description, you'll see it's not exactly the same :smile:

src/Microsoft.AspNet.Authentication.OAuth/Notifications/OAuthAuthenticatedContext.cs
PinpointTownes: Well yeah (or an empty `JObject`), but semantically, it's not totally satisfactory.

src/Microsoft.AspNet.Authentication.Google/GoogleAuthenticationOptions.cs
HaoK: So we are making it virtual to add doc comments now? :)

PinpointTownes: Yeah: https://github.com/aspnet/Security/pull/257/files#r30449626 :smile:

HaoK: So what I don't like about this vs the old pattern is its pretty hard to tell what's changing, before in the ctor, you can easily see which defaults were affected just by looking at the ctor.  Now you have to wade through all of the various inline'd autoproperties, its just not as obvious.  I'm not sure its worth it just for doc comments.  I am sort of leaning towards going back to simple auto properties and derived classes can just change the defaults in the ctor, you can continue to put any doc comments about the defaults changing in the ctor still.  Its also just less code...  @Tratcher thoughts?

Tratcher: I've thought too much about this PR already. :-)

I don't think the ability to change docs is compelling enough, especially when all that changes is the default. I also agree that the defaults are easier to understand in the constructor.

PinpointTownes: Okay, I'll revert this change then, and merge the two comments to indicate that the default value used by the social providers is not the same as the one used by the OAuth2 base middleware.

Any remark concerning the new generic approach?

HaoK: I think its fine, the usage is more or less unchanged now correct?

PinpointTownes: Yup.

Just note that with this new pattern, OAuth2 subclasses cannot define custom notifications anymore, at least as easily as it was before (you'll probably need tons of casts if you try to define your own notifications class :smile:).

src/Microsoft.AspNet.Authentication.OAuth/Notifications/OAuthAuthenticatedContext.cs
blowdart: If we're arguing semantics it may not be a user at all, it could be the result of machine to machine, and the round trip gives you machine permissions ...

I wouldn't want it in the base class though, as that's unrealistic. It indicates there will be a back channel communication to populate it in all cases.

PinpointTownes: I agree.

If we really want to have a `User` property at the `OAuthAuthenticatedContext` level, then we should consider implementing a default user-info client in `CreateTicketAsync`, that would throw if it didn't receive a valid JSON payload, telling the user that he should implement his own notification delegate or his own handler.

Tratcher: Yuck, no. There's a reason the retrieval code wasn't already in the base class. All the providers do it differently, a default client will almost never succeed.

PinpointTownes: Sure, they all have different implementations, but often share three common things: they often use GET, send the access token via the `Authorization` header and return JSON.

src/Microsoft.AspNet.Authentication.Google/GoogleAuthenticationHandler.cs
```diff
@@ -30,45 +30,48 @@ public GoogleAuthenticationHandler(HttpClient httpClient)
...
+            {
+                Properties = properties,
+                Principal = new ClaimsPrincipal(identity)
+            };
+            
             if (!string.IsNullOrEmpty(context.Id))
             {
                 identity.AddClaim(new Claim(ClaimTypes.NameIdentifier, context.Id,
                     ClaimValueTypes.String, Options.ClaimsIssuer));
             }
```
brentschmaltz: A slight tweak here could improve performance.
ClaimIdentity.AddClaim(...) performs an object.ReferenceEquals(claim.Subject, this). If false, then claim.Clone is called.  
If you did identity.AddClaim(..., identity), then claim.Clone would not be called.

PinpointTownes: @brentschmaltz ah, nice tweak, thanks :smile: 

As I'm pretty sure there are tons of places where we could use it, I'd suggest opening a new ticket to track it and implement it everywhere at the same time.

PinpointTownes:
@HaoK @Tratcher @blowdart @brentschmaltz do you approve the last design? Are you okay to remove the ability to create custom/derived `IOAuthAuthenticationNotifications`?

HaoK:
If none of our existing providers today take advantage of this functionality. Seems like we aren't losing much. They can always just cast the interface as you mentioned anyways.

Tratcher:
Disagree, adding notifications is an intentional form of extensibility, disabling it is not ok.  And did you miss the fact that these generics look terrible and should just be removed?

PinpointTownes:
@Tratcher did you take a look at the recent commits? Everything got better when removing the TNotifications generic parameter.

HaoK:
Might be best to not try to make this change as part of save tokens since there's no consensus

PinpointTownes:
Concerning the ability to add notifications, I agree with @HaoK, it doesn't seem to be a big deal (though I admit it was my idea when we discussed the very first OAuth2 base handler design).

The base notifications are generic enough - specially `OnAuthenticated` - to support every possible scenario.

Tratcher:
Ah, that's much less terrible. A little strange, but not terrible.

PinpointTownes:
If you need evidence, here's one: if you take a look at the OWIN providers developed by the community - that I'll port to ASP.NET 5 with @jerriep - (https://github.com/RockstarLabs/OwinOAuthProviders), none uses custom notifications.

PinpointTownes:
Another design: make everything non-generic (except the options), mark `OAuthAuthenticatedContext` as sealed, and add a new `GetUser<TUser>` to `OAuthAuthenticatedContext` that would return a strongly-typed object containing the properties we used to have in the provider-specific `*AuthenticatedContext`.

PinpointTownes:
Basically, something like that:

```
// https://console.developers.google.com/project
app.UseGoogleAuthentication(options => {
    options.ClientId = "560027070069-37ldt4kfuohhu3m495hk2j4pjp92d382.apps.googleusercontent.com";
    options.ClientSecret = "n2Q-GEw9RQjzcRbU3qhfTj8f";

    options.Notifications = new OAuthAuthenticationNotifications {
        OnAuthenticated = async notification => {
            var user = notification.GetUser<GoogleUser>();
            var email = user.Email;
        }
    };
});
```

PinpointTownes:
@Haok @Tratcher ping.

Tratcher:
This PR seems OK as is. @HaoK ?

HaoK:
Yep looks good to me!

PinpointTownes:
@HaoK @Tratcher okay! No regret? :smile: 
https://github.com/PinpointTownes/Security/commit/e8d4c77acd29cf5478cfb1c90eac47d758419c63

src/Microsoft.AspNet.Authentication.Google/GoogleAuthenticationMiddleware.cs
```diff
@@ -37,7 +37,7 @@ public class GoogleAuthenticationMiddleware : OAuthAuthenticationMiddleware<Goog
         {
             if (Options.Notifications == null)
             {
-                Options.Notifications = new GoogleAuthenticationNotifications();
+                Options.Notifications = new OAuthAuthenticationNotifications<GoogleAuthenticatedContext>();
```
PinpointTownes: I'll try to remove that and insert it in the base middleware.

src/Microsoft.AspNet.Authentication.MicrosoftAccount/MicrosoftAccountAuthenticationMiddleware.cs
```diff
@@ -35,7 +35,7 @@ public class MicrosoftAccountAuthenticationMiddleware : OAuthAuthenticationMiddl
         {
             if (Options.Notifications == null)
             {
-                Options.Notifications = new MicrosoftAccountAuthenticationNotifications();
+                Options.Notifications = new OAuthAuthenticationNotifications<MicrosoftAccountAuthenticatedContext>();
```
PinpointTownes: Same remark.

src/Microsoft.AspNet.Authentication.OAuth/Notifications/OAuthAuthenticatedContext.cs
```diff
@@ -22,46 +22,53 @@ public class OAuthAuthenticatedContext : BaseContext<OAuthAuthenticationOptions>
...
         /// <summary>
         /// Gets the access token expiration time.
         /// </summary>
-        public TimeSpan? ExpiresIn { get; protected set; }
+        public TimeSpan? ExpiresIn
+        {
+            get
+            {
+                int value;
+                if (int.TryParse(TokenResponse.ExpiresIn, NumberStyles.Integer, CultureInfo.InvariantCulture, out value)) {
```
PinpointTownes: Ahhhhhh, K&R style FTW! :smile: 
Will fix that.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
```diff
@@ -156,17 +179,30 @@ protected override AuthenticationTicket AuthenticateCore()
...
+            return new OAuthTokenResponse(oauth2Token);
         }
 
-        protected virtual async Task<AuthenticationTicket> GetUserInformationAsync(AuthenticationProperties properties, TokenResponse tokens)
+        protected virtual async Task<AuthenticationTicket> CreateTicketAsync(ClaimsIdentity identity, AuthenticationProperties properties, OAuthTokenResponse tokens)
         {
-            var context = new OAuthGetUserInformationContext(Context, Options, Backchannel, tokens)
+            if (typeof(TAuthenticationContext) != typeof(OAuthAuthenticatedContext))
+            {
+                throw new InvalidOperationException();
```
PinpointTownes: We need a nice message here (well, in the .resx file actually). Any idea? :smile:

PinpointTownes:
@HaoK @Tratcher did you take a look at the other design, that doesn't use generics at all (https://github.com/PinpointTownes/Security/commit/e8d4c77acd29cf5478cfb1c90eac47d758419c63)?

That would be great if we could progress a little bit with this PR so feel free to ping me when the design is definitely adopted :smile:

amattie:
Will this be merged soon? Just wondering b/c it seems to be approved.

PinpointTownes:
@amattie we are finalizing a few things, stay tuned! :smile: 

As a side note, I'd also like to add a `UseTokenLifetime` option to the OAuth2 middleware, similar to the one exposed by the OIDC handler. But of course, it'll be part of another PR.

src/Microsoft.AspNet.Authentication.Google/GoogleAuthenticationMiddleware.cs
```diff
@@ -33,23 +33,7 @@ public class GoogleAuthenticationMiddleware : OAuthAuthenticationMiddleware<Goog
...
             [NotNull] IOptions<GoogleAuthenticationOptions> options,
             ConfigureOptions<GoogleAuthenticationOptions> configureOptions = null)
-            : base(next, dataProtectionProvider, loggerFactory, encoder, externalOptions, options, configureOptions)
-        {
-            if (Options.Notifications == null)
-            {
-                Options.Notifications = new GoogleAuthenticationNotifications();
-            }
-
-            if (Options.Scope.Count == 0)
```
Tratcher: The default scopes where here on purpose, don't move them to the options constructor.

HaoK: Yeah perhaps just whack the TODO (might be my todo that i added)

src/Microsoft.AspNet.Authentication.MicrosoftAccount/MicrosoftAccountAuthenticationMiddleware.cs
```diff
@@ -31,19 +31,7 @@ public class MicrosoftAccountAuthenticationMiddleware : OAuthAuthenticationMiddl
             [NotNull] IOptions<ExternalAuthenticationOptions> externalOptions,
             [NotNull] IOptions<MicrosoftAccountAuthenticationOptions> options,
             ConfigureOptions<MicrosoftAccountAuthenticationOptions> configureOptions = null)
-            : base(next, dataProtectionProvider, loggerFactory, encoder, externalOptions, options, configureOptions)
-        {
-            if (Options.Notifications == null)
-            {
-                Options.Notifications = new MicrosoftAccountAuthenticationNotifications();
-            }
-            if (Options.Scope.Count == 0)
```
Tratcher: Same scope change to revert.

src/Microsoft.AspNet.Authentication.Google/GoogleAuthenticationMiddleware.cs
HaoK: Yeah makes more sense here, as you don't want the options to have defaults, only if the user doesn't specify any do you then want to populate default scopes

PinpointTownes: @Tratcher it was your idea, actually: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Google/GoogleAuthenticationMiddleware.cs#L47

I think it's better to have it directly in the options, so the user can add additional scopes without making the base scopes "automagically" disappear. If you don't want to use the basic scope, you can clear `Options.Scope`.

Tratcher:
Nice. I think we're just about done here.

src/Microsoft.AspNet.Authentication.Google/GoogleAuthenticationHelper.cs
```diff
@@ -1,68 +1,45 @@
...
-        public string Profile { get; private set; }
+        public static string GetProfile(JObject user) => TryGetValue(user, "url");
 
         /// <summary>
         /// Gets the user's email.
         /// </summary>
-        public string Email { get; private set; }
+        public static string GetEmail(JObject user) => TryGetFirstValue(user, "emails", "value");
 
         private static string TryGetValue(JObject user, string propertyName)
```
HaoK: Maybe we should move TryGetValue on JObject to a public extension method in the parent OAuth assembly so all of the helpers can reuse it

PinpointTownes: IMHO, we should simply use `user.Value<string>("id")`, which basically does the same thing (it returns `null` the item is missing from the JSON payload).

HaoK: That's fine too, just cringe seeing 3? copies of the same method...

src/Microsoft.AspNet.Authentication.Google/GoogleAuthenticationMiddleware.cs
Tratcher: There was a long debate over this back in katana. Let's get this PR finished and open a separate issue if you want to pursue this tangent.

HaoK:
Yep much cleaner now!  Only took 150 comments and I don't know how many iterations to get there :)

src/Microsoft.AspNet.Authentication.Google/GoogleAuthenticationMiddleware.cs
PinpointTownes: I'll revert this change and open a new ticket tomorrow.

src/Microsoft.AspNet.Authentication.OAuth/Notifications/OAuthAuthenticatedContext.cs
```diff
@@ -20,55 +22,97 @@ public class OAuthAuthenticatedContext : BaseContext<OAuthAuthenticationOptions>
...
         /// <summary>
-        /// Gets the <see cref="ClaimsIdentity"/> representing the user.
+        /// Gets the <see cref="ClaimsPrincipal"/> representing the user.
         /// </summary>
         public ClaimsPrincipal Principal { get; set; }
 
         /// <summary>
+        /// Gets the main identity exposed by <see cref="Principal"/>.
+        /// </summary>
+        public ClaimsIdentity Identity => (ClaimsIdentity) Principal.Identity;
```
PinpointTownes: @HaoK @Tratcher is it okay to have a new `Identity` property here, that simply wraps `Principal.Identity`?
It makes adding new claims to the principal much easier from the `OnAuthenticated` notification, as you don't have to cast the identity anymore from the user code (see the social sample).

src/Microsoft.AspNet.Authentication.Google/GoogleAuthenticationHelper.cs
PinpointTownes: Fixed, they all have disappeared :smile:

src/Microsoft.AspNet.Authentication.OAuth/Notifications/OAuthAuthenticatedContext.cs
HaoK: I'm not sure we should add an alias like this, but I do agree this is pretty poor behavior in ClaimsIdentity, no chance we can get them to fix that property to be typed correctly in the new world?

PinpointTownes: I don't think so, `ClaimsPrincipal` should also be fixed in full CLR, which is probably far too late.

Tratcher: I think this accessor is fine, especially when we just made the identity.

PinpointTownes: I'll make it safe when `Principal` is `null` though.

Tratcher: `Principal?.Identity`

PinpointTownes: You're not fast enough, sir :smile:

PinpointTownes:
@HaoK @Tratcher don't hesitate to make a final review: I'll fix everything tomorrow.

src/Microsoft.AspNet.Authentication.Google/GoogleAuthenticationMiddleware.cs
PinpointTownes: Reverted :smile:

src/Microsoft.AspNet.Authentication.MicrosoftAccount/MicrosoftAccountAuthenticationMiddleware.cs
PinpointTownes: Reverted too.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs
PinpointTownes: We decided we didn't to change that.

PinpointTownes:
Squashed.

Tratcher:
Looks good. @hoak can you check it and merge it?

PinpointTownes:
@HaoK any chance you could have a look? :smile:

HaoK:
Yeah I need to review it again since Lou had some potential concerns which I think we avoided but I'll need to double check

PinpointTownes:
@HaoK any chance you could share more info concerning these potential concerns? :smile:

HaoK:
So he is fine with us nuking all of the derived generics/options/interfaces.  His main concern is that we must keep the core INotifications interface so you can replace the workers on the options via DI.  

Also when you get a chance, can you rebase this against dev?  I sort of blew all sorts of stuff up :)

PinpointTownes:
> So he is fine with us nuking all of the derived generics/options/interfaces. His main concern is that we must keep the core INotifications interface so you can replace the workers on the options via DI.

Ah yeah, I couldn't agree more. You'll have to convince @Tratcher though: https://github.com/aspnet/Security/issues/47#issuecomment-108294547 :smile: 

> Also when you get a chance, can you rebase this against dev? I sort of blew all sorts of stuff up :)
Sure. Any idea when the CI builds including your changes will be ready? It would be easier than hacking global.json :smile:

HaoK:
CI might take a day or two, you can try using the dev volatile feed which is what we use internally:
https://www.myget.org/F/aspnetvolatile/

I wouldn't recommend doing this in general, unless you like being broken...a lot :)

HaoK:
Well the status quo wins in general, its more like we have to convince @lodejard that this is the right thing to do, he felt pretty strongly against removing the interfaces, and if you actually agree with him, then that's 2:1 in favor of the status quo.  I mostly wanted the generics gone, so I don't have a strong opinion either way on how this goes, which in practice means I'm leaning towards leaving the interface alone as well then...

PinpointTownes:
@HaoK well yeah, but the current situation is not satisfactory, as we have both patterns in the code base, which is terribly inconsistent.

samples/SocialSample/Startup.cs
```diff
@@ -131,56 +133,61 @@ 127.0.0.1 MsSecSample.localhost.this
...
+                                ClaimValueTypes.String, notification.Options.ClaimsIssuer));
                         }
-
-                        context.Principal = new ClaimsPrincipal(identity);
                     },
                 };
             });
 
             // Choose an authentication type
-            app.Map("/login", signoutApp =>
```
PinpointTownes: Definitely not something I wanted to change, but it sounds like @HaoK's recent changes broke something here: 

image

It seems that the `redirect_uri` is now computed using the `PathBase` of the handler/middleware that triggers the `Challenge` call, which of course, produces an incorrect `redirect_uri` (`http://localhost:54540/login/signin-google` vs `http://localhost:54540/signin-google`).

samples/SocialSample/Startup.cs
```diff
@@ -199,13 +206,10 @@ 127.0.0.1 MsSecSample.localhost.this
...
-            app.Map("/logout", signoutApp =>
-            {
-                signoutApp.Run(async context =>
+                    return;
+                }
+                
+                if (context.Request.Path.StartsWithSegments(new PathString("/logout")))
                 {
                     await context.Authentication.SignOutAsync(CookieAuthenticationDefaults.AuthenticationScheme);
                     context.Response.ContentType = "text/html";
```
PinpointTownes: @HaoK now that `Challenge`/`SignIn`/`SignOut` are not deferred anymore, this kind of code seems really dangerous, specially if the authentication middleware writes to the response stream from `SignOut`.

Tratcher:
I primarily want to reduce it from triple extensibility to just one pattern. I'm less concerned which pattern, though I do think the delegate pattern we use in OIDC is the simplest.

samples/SocialSample/Startup.cs
HaoK: Very possible i regressed something until the ci goes green. Music store isn't green yet due to lots of issues so I will probably fix this today

Tratcher: Hmm, the middleware may need to snapshot the base address when the request first arrives.

HaoK: What is the correct url that is registered for these apps?  I'm seeing errors saying: "http://localhost:54540/login/signin-google-token" did not match a registered redirect uri

HaoK: nm, I should scroll up :)

HaoK: I'm going to port this fix manually to a standalone so social samples actually work again

HaoK:
Ok so sounds like we have consensus to add back just the base `IOAuthAuthenticationNotifications` (RIP Generics and derived options), and use that as the pattern everywhere?

samples/SocialSample/Startup.cs
HaoK: Yeah well, as you can tell, the samples code for security is basically a little test app (that I never ever run clearly)  I mostly rely on MVC/Identity/MusicStore for functional verification, but I should use this as a canary as well (until we get real functional tests...in the repo sigh)

PinpointTownes:
> Ok so sounds like we have consensus to add back just the base IOAuthAuthenticationNotifications (RIP Generics and derived notifications), and use that as the pattern everywhere?

Yup! I haven't removed `IOAuthAuthenticationNotifications`, so we're fine for this PR, but we'll definitely need to fix the other providers in a future PR :smile:

samples/SocialSample/Startup.cs
PinpointTownes: I started porting `AspNet.Security.OpenIdConnect.Server` to the new authentication APIs... and I almost had a heart attack: the new `ChallengeAsync/SignInAsync/SignOutAsync` are really fragile.

For instance, returning a simple `new HttpStatusCodeResult(200)` result from MVC after invoking one of these methods may result in either a no-op thing... or in an "headers already sent" exception thrown by the host :scream:

HaoK: Instead of fragile, I prefer the term powerful and immediate :)  But yeah the behavior is very different, I don't think we entirely know all the ramifications yet (i.e. map path issue)

HaoK: I would argue calling these methods which immediately do something is easier to grok, than the old behavior of stuff happens at some undetermined future time (whenever the response is sent).  Perhaps you are just used to the old behavior, and the new will grow on you.  I didn't play with the old behavior much (so I'm missing any fond memories of it)

PinpointTownes: I'm not saying that the new behavior is necessarily bad, it's just **surprising** and... obviously different :smile:

What I'm saying is that we can't ship these new methods without changing the way we use them: altering the headers after invoking `SignInAsync/SignOutAsync/ChallengeAsync` is now very risky, specially if the authentication handler started writing to the response stream.

For instance, this snippet will throw an exception:

```csharp
await Context.Authentication.SignInAsync(OpenIdConnectDefaults.AuthenticationScheme, new ClaimsPrincipal(identity));

return new HttpStatusCodeResult(200);
```
... while this one, that uses `EmptyResult` (which doesn't change the status code), won't:

```csharp
await Context.Authentication.SignInAsync(OpenIdConnectDefaults.AuthenticationScheme, new ClaimsPrincipal(identity));

return new EmptyResult();
```

image

HaoK: Yep totally agree, this is something that definitely needs to go on the Announcements as a heads up/breaking change

samples/SocialSample/Startup.cs
PinpointTownes: Thanks to your fix, I can now revert this change :smile:

samples/SocialSample/Startup.cs
PinpointTownes: I moved `context.Response.ContentType = "text/html";` before the `SignOut` call.

PinpointTownes:
Rebased, squashed. I also removed the volatile feed from NuGet.Config.

HaoK:
merged 9bb8b6114697f14a42e747acd1d1a41e6eed5af1 thanks for grinding through this and all the rebases :)

PinpointTownes:
@HaoK wooo, thank you for merging it! :bow:

PR #258, opened 2015-05-14
Adding code only flow for sign in.

tushargupta51:
Also changes for integrating wilson logging to OIDC middleware.


src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -520,6 +540,14 @@ protected override AuthenticationTicket AuthenticateCore()
             }
         }
 
+        protected virtual IdentityModel.Clients.ActiveDirectory.AuthenticationResult RedeemAuthorizationCode(string authorizationCode)
```
PinpointTownes: Oh please no, no dependency on AAD stuff... specially for such a simple thing.
Why not simply using a good old `HttpClient`?

https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/blob/vNext/samples/Mvc/Mvc.Client/Startup.cs#L76-L103

src/Microsoft.AspNet.Authentication.OpenIdConnect/DefaultIdentityModelLoggingListener.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.Diagnostics.Tracing;
+using Microsoft.Framework.Logging;
+
+namespace Microsoft.AspNet.Authentication.OpenIdConnect
+{
+    /// <summary>
+    /// Default logging event listener for wilson event source. It logs the data to the target where other Asp.Net logs go.
+    /// </summary>
+    public class DefaultIdentityModelLoggingListener : EventListener
```
PinpointTownes: Is another abstraction really needed?

I know `EventListener` is internally used by Wilson, but why not directly using `ILogger`?

PinpointTownes:
I have absolutely no idea what you're doing with the git history, but it's a real mess... you should really use fast forward merge :smile:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -520,6 +540,14 @@ protected override AuthenticationTicket AuthenticateCore()
...
+        {
+            AuthenticationContext authContext = new AuthenticationContext(Options.Authority, false);
+            ClientCredential credential = new ClientCredential(Options.ClientId, Options.ClientSecret);
+            var tokens = authContext.AcquireTokenByAuthorizationCodeAsync(authorizationCode, new Uri(Options.RedirectUri), credential).GetAwaiter().GetResult();
```
PinpointTownes: What? Why using a blocking `.GetAwaiter().GetResult()` when you can make the method fully async?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -238,6 +243,7 @@ protected override AuthenticationTicket AuthenticateCore()
         /// <remarks>Uses log id's OIDCH-0000 - OIDCH-0025</remarks>
         protected override async Task<AuthenticationTicket> AuthenticateCoreAsync()
         {
+            bool isCodeOnlyFlow = (Options.ResponseType == OpenIdConnectResponseTypes.CodeOnly);
```
HaoK: nit: var

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -429,6 +474,7 @@ protected override AuthenticationTicket AuthenticateCore()
                         return null;
                     }
 
+
```
HaoK: nit: extra line

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
HaoK: :+1:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
```diff
@@ -17,6 +17,9 @@
 using Microsoft.Framework.OptionsModel;
 using Microsoft.IdentityModel.Protocols;
 using Microsoft.Framework.Internal;
+using Microsoft.IdentityModel.Logging;
+using System.Diagnostics.Tracing;
```
HaoK: nit: Sort where System namespaces go first

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
```diff
@@ -47,6 +50,16 @@ public class OpenIdConnectAuthenticationMiddleware : AuthenticationMiddleware<Op
             ConfigureOptions<OpenIdConnectAuthenticationOptions> configureOptions = null)
             : base(next, options, loggerFactory, encoder, configureOptions)
         {
+            if (Options.IdentityModelEventSourceListener != null)
+            {
+                Options.IdentityModelEventSourceListener.EnableEvents(IdentityModelEventSource.Logger, EventLevel.Informational);
+            }
+            else
+            {
+                DefaultIdentityModelLoggingListener eventListener = new DefaultIdentityModelLoggingListener(Logger);
```
HaoK: nit: var

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: And BTW, given that `Options.RedirectUri` is not a mandatory setting, `Uri`'s constructor will throw an exception if you don't set it.

src/Microsoft.AspNet.Authentication.OpenIdConnect/DefaultIdentityModelLoggingListener.cs
tushargupta51: Do you mean using ILogger directly for Wilson? This is the listener that I hook to Wilson Event Source. That is needed otherwise how Wilson logs can be logged.

tushargupta51:
@PinpointTownes Regarding the git history, are you referring to the merges because all other commits are relevant to the logging and code only flow changes?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: @PinpointTownes We want to use ADAL for redeeming code for tokens instead of doing a direct httpClient call. That is why we have adal for :smile: We have plans to extend this to authorization servers beyond AAD but that will be the second step.

src/Microsoft.AspNet.Authentication.OpenIdConnect/DefaultIdentityModelLoggingListener.cs
PinpointTownes: Yep, exactly.

The logging stack has .NET 4.5/portable targets that you could use for Wilson: https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Framework.Logging.Abstractions/project.json

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: Then rename this middleware to `ActiveDirectoryAuthenticationMiddleware` and let the community offer a real provider-agnostic OpenID Connect middleware :smile: 

Out of curiosity, what's the advantage of using ADAL here?

PinpointTownes:
@tushargupta51 sure, my remark was only about the merge commits.

I'd like to play with the bits offered by this PR but the static ADAL version you specified doesn't seem to exist. Could you please upload it on the MyGet nightly build repository?

tushargupta51:
@PinpointTownes It is there on MyGet. Here: https://www.myget.org/gallery/azureadwebstacknightly

PinpointTownes:
Hum, I see, it failed because the Azure AD has been removed from NuGet.Config when the JWT security token handler bits have been published to aspnetvnext.

You should add it again if you plan to take a dependency on ADAL, or upload it to the aspnetvnext repository.

(BTW, that's why the build failed on AppVeyor: https://ci.appveyor.com/project/aspnetci/security/build/1.0.107)

src/Microsoft.AspNet.Authentication.OpenIdConnect/DefaultIdentityModelLoggingListener.cs
tushargupta51: I am not sure if I understand this. Wilson uses IdentityModelEventSource to log things. To make it log into the same place as aspnet, I need to hook an eventListener to override the default behavior in Wilson and tell it to redirect logs to aspnet logger. How else can I integrate the IdentityModelEventSource in aspnet? Or are you suggesting not using the IdentityModelEventSource at all?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: Nice catch. Thanks!

src/Microsoft.AspNet.Authentication.OpenIdConnect/DefaultIdentityModelLoggingListener.cs
PinpointTownes: > Or are you suggesting not using the IdentityModelEventSource at all?

That's exactly what I'm suggesting :smile:
In Wilson, why not using the new logging stack from aspnet/Logging?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: Caching. That's one major advantage of using ADAL. I understand we need to be provider agnostic and that is the plan we have for GA. Adding @vibronet if he has anything to add here.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: Same remark for `Options.Authority`. Can `AuthenticationContext` live with it if it's null?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: God no, not this mess. You can't even imagine how many people lost their hair on the #owin channel because of token caching :smile:

And why would you use caching here? New code, new token. Absolutely no need for caching.

@vibronet hey, it's been a long time... :smile:

vibronet:
for the pure sign in flow, the plan is to move to HttpClient. When we'll look after improving the access token acquisition experience we will be using ADAL mostly for its caching benefits.

PinpointTownes:
@vibronet I still don't understand why you would integrate ADAL directly in the OIDC client middleware instead of simply using the notifications hooks to store the access tokens exactly where you want to.

src/Microsoft.AspNet.Authentication.OpenIdConnect/DefaultIdentityModelLoggingListener.cs
tushargupta51: Wilson is not an asp.net component and we do not want it to take a dependency on aspnet logging stack.

Tratcher:
Agreed, OIDC should not take a direct AAD dependency, even temporarily.

src/Microsoft.AspNet.Authentication.OpenIdConnect/DefaultIdentityModelLoggingListener.cs
PinpointTownes: Despite its name, aspnet/Logging is not an ASP.NET component - which explains why it has a neutral namespace: "Microsoft.Framework.Logging".

tushargupta51: Even though the namespace is different, I would still consider aspnet/Logging as part of asp.net.

Tratcher: AspNet/Logging a library developed by the ASP.NET team that we are considering contributing back to the CLR team/repo. That will not be resolved any time soon, so it's fine not to use it.

PinpointTownes: Well, aspnet/Logging is heavily used by Entity Framework 7, which - AFAICT - is not an ASP.NET component, no? :smile:

tushargupta51: @PinpointTownes You mean this: https://github.com/aspnet/EntityFramework? Shouldn't we have it somewhere else then?

Tratcher: Don't confuse org/team structures with technologies.

vibronet:
let me restate - for the sign in flow we won't have a dependency on ADAL. For the access token flow the notifications route is already available today, but requires far too much code and is prone to mistakes. We need a more integrated solution. I am not saying that such solution will show up in the OIDC middleware, though it is a possibility given that the code request and redemption machinery is largely the same

vibronet:
also note, "store the access token" is a very simplistic way of describing what happens. It's not enough to provide a storage, you need a model to come with it.  Token caching is a combination of caching access token, refresh tokens, and context (source, client app, resource, user and many others). When at a later time a new token is requested, there is very sophisticated logic that infers whether there is an access token and/or a refresh token and/or a multi resource refresh token and so on. All those rules can get very complicated and have a strong dependency on the provider. Generic classes that work across the board can't ingest this level of sophistication and surface the details to the developer. Conversely, ADAL's assumptions allow it to take care of all that logic without bothering the developer. Once again, I am not saying that this will end up in the OIDC middleware. But when it comes to clients and token management strategies, there is no such thing as a provider-agnostic approach AND a viable programming model. I am just back from BUILD and Ignite, during which I heard over and over again from customers how useless they find generic purpose libraries for OAuth2 (again, not OIDC).

src/Microsoft.AspNet.Authentication.OpenIdConnect/DefaultIdentityModelLoggingListener.cs
Eilon: `ILogger` is the logging system to use here, not EventListener. As @Tratcher says, `ILogger` is *not* an ASP.NET component. Rather, it is a general purpose logging abstraction used by all DNX-based apps, and even used outside of DNX scenarios (e.g. EF).

There will separately be an EventListener-based implementation of ILogger. We don't want half the system to use ILogger, and half to use EventListener. That's why we have an abstraction: to allow all components to talk to the same abstraction regardless of where the sink is.

Eilon: Also as @Tratcher said, `ILogger` will eventually move out to the `dotnet` org and won't be in the ASP.NET org.

src/Microsoft.AspNet.Authentication.OpenIdConnect/project.json
```diff
@@ -3,8 +3,11 @@
     "description": "ASP.NET 5 middleware that enables an application to support OpenIdConnect authentication workflow.",
     "dependencies": {
         "Microsoft.AspNet.Authentication": "1.0.0-*",
+        "Microsoft.IdentityModel.Protocol.Extensions": "2.0.0-beta4-*",
+        "System.Threading.Tasks": "4.0.10-beta-*",
+        "System.Runtime": "4.0.20-beta-*",
         "Microsoft.Framework.NotNullAttribute.Sources": { "type": "build", "version": "1.0.0-*" },
-        "Microsoft.IdentityModel.Protocol.Extensions": "2.0.0-beta4-*"
+        "Microsoft.AspNet.Authentication.OAuth": "1.0.0-*"
```
PinpointTownes: Why taking a dependency on `Microsoft.AspNet.Authentication.OAuth`?

IMHO, you should use the super-cool `OpenIdConnectMessage` primitive instead of the more limited `TokenResponse` we created for the OAuth2 client middleware.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -520,6 +545,25 @@ protected override AuthenticationTicket AuthenticateCore()
...
+            var tokenRequestParameters = new Dictionary<string, string>()
+            {
+                { "client_id", Options.ClientId },
+                { "redirect_uri", Options.RedirectUri },
+                { "client_secret", Options.ClientSecret },
+                { "code", authorizationCode },
+                { "grant_type", "authorization_code" },
+            };
+            var requestMessage = new HttpRequestMessage(HttpMethod.Post, _configuration.TokenEndpoint);
+            requestMessage.Content = new FormUrlEncodedContent(tokenRequestParameters);
```
PinpointTownes: In the OAuth2 client handler, we only support flowing the client credentials via the request body. It would be probably useful to also support `client_secret_basic`, specially since we can easily know if it's supported using the provider configuration if it's available.

src/Microsoft.AspNet.Authentication.OpenIdConnect/project.json
brentschmaltz: @PinpointTownes @tushargupta51 I think it helps in crafting the message url and avoiding missssspellings, since the properties like 'ClientId' avoid that category of error.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -321,6 +333,45 @@ protected override AuthenticationTicket AuthenticateCore()
                     _configuration = await Options.ConfigurationManager.GetConfigurationAsync(Context.RequestAborted);
                 }
 
+                if (!string.IsNullOrWhiteSpace(message.Code))
```
brentschmaltz: Switching the order here is problematic in the case of id_token + code. We haven't validated the code w.r.t. id_token (chash) or even validated the token. This should be fixed.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
brentschmaltz: You could use OIDCMessage instead of Dictionary..., use first class properties such as ClientId etc, then do FormUrlEncodedContext(oidc.Prameters)

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: I moved the notification so that user can plug in their own code for exchanging "code" for tokens in case of "Code only flow". Now, I think they can do so in messageReceivedNotification of by overriding RedeemAuthorizationCode method (which requires more work!). I moved the authorizationCodeReceivedNotification back to where it was earlier.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: Yeah, it's definitely the right thing to do, for both the token request and the response.

PinpointTownes: As a side note, that would be great to have new notifications to alter the token request before it is sent and after the response is received. 

I already planned to submit a PR to add these hooks to the OAuth2 handler when we are done with the other PR.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -520,6 +557,73 @@ protected override AuthenticationTicket AuthenticateCore()
...
+                AccessToken = jsonTokenResponse.Value<string>("access_token"),
+                IdToken = jsonTokenResponse.Value<string>("id_token"),
+                TokenType = jsonTokenResponse.Value<string>("token_type"),
+                ExpiresIn = jsonTokenResponse.Value<string>("expires_in")
+            };
+        }
+
+        protected virtual async Task<AuthenticationTicket> GetUserInformationAsync(AuthenticationProperties properties, OpenIdConnectMessage message, AuthenticationTicket ticket)
+        {
+            var requestMessage = new HttpRequestMessage(HttpMethod.Get, _configuration.UserInfoEndpoint);
```
PinpointTownes: According to the OIDC discovery specifications, `userinfo_endpoint` is not a "required" parameter but a "recommended" one. Chances are high to crash here if it was missing from the discovery response and if the developer has not provided one in `app.UseOpenIdConnectAuthentication`.

Any chance you could update `GetUserInformationAsync` to work in a degraded mode in this case?

http://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -520,6 +557,73 @@ protected override AuthenticationTicket AuthenticateCore()
...
+
+            var requestMessage = new HttpRequestMessage(HttpMethod.Post, _configuration.TokenEndpoint);
+            requestMessage.Content = new FormUrlEncodedContent(openIdMessage.Parameters);
+            var responseMessage = await Backchannel.SendAsync(requestMessage);
+            responseMessage.EnsureSuccessStatusCode();
+            var tokenResonse = await responseMessage.Content.ReadAsStringAsync();
+            var jsonTokenResponse = JObject.Parse(tokenResonse);
+            return new OpenIdConnectMessage()
+            {
+                AccessToken = jsonTokenResponse.Value<string>("access_token"),
```
PinpointTownes: Any chance you could use `OpenIdConnectParameterNames` to remove some of these magic strings?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -520,6 +557,73 @@ protected override AuthenticationTicket AuthenticateCore()
...
+        {
+            var requestMessage = new HttpRequestMessage(HttpMethod.Get, _configuration.UserInfoEndpoint);
+            requestMessage.Headers.Authorization = new AuthenticationHeaderValue("Bearer", message.AccessToken);
+            var responseMessage = await Backchannel.SendAsync(requestMessage);
+            responseMessage.EnsureSuccessStatusCode();
+            var userInfoResponse = await responseMessage.Content.ReadAsStringAsync();
+            var user = JObject.Parse(userInfoResponse);
+
+            var identity = (ClaimsIdentity)ticket.Principal.Identity;
+            var subject = identity.FindFirst(ClaimTypes.NameIdentifier).ToString();
```
PinpointTownes: Out of curiosity, why using `ToString` (that internally concatenates `Claim.Type` and `Claim.Value`) instead of simply using `Claim.Value`?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -520,6 +557,73 @@ protected override AuthenticationTicket AuthenticateCore()
...
+
+            // check if the sub claim matches
+            var userInfoSubject = user.Value<string>("sub");
+            if (userInfoSubject == null || !string.Equals(userInfoSubject.ToString(), subject, StringComparison.OrdinalIgnoreCase))
```
PinpointTownes: Same remark (and BTW, `userInfoSubject` is already a string).

I don't even understand how this check would succeed, as you compare two different things: `subject` (= `type + value`) with `userInfoSubject` (= `value`)

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -520,6 +557,73 @@ protected override AuthenticationTicket AuthenticateCore()
...
+            {
+                Logger.LogError(Resources.OIDCH_0038_Subject_Claim_Mismatch);
+                throw new ArgumentException(Resources.OIDCH_0038_Subject_Claim_Mismatch); 
+            }
+
+            var userInfoIdentity = new ClaimsIdentity(identity);
+            foreach (KeyValuePair<string, JToken> pair in user)
```
davidfowl: var?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -520,6 +557,73 @@ protected override AuthenticationTicket AuthenticateCore()
...
+            {
+                JToken value;
+                var claimValue = user.TryGetValue(pair.Key, out value) ? value.ToString() : null;
+
+                IDictionary<string, string> inboundClaimTypeMap = new Dictionary<string, string>(JwtSecurityTokenHandler.InboundClaimTypeMap);
```
davidfowl: var?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
```diff
@@ -100,6 +105,11 @@ public class OpenIdConnectAuthenticationMiddleware : AuthenticationMiddleware<Op
                 Options.TokenValidationParameters.ValidAudience = Options.ClientId;
             }
 
+            Backchannel = new HttpClient(ResolveHttpMessageHandler(Options));
+            Backchannel.DefaultRequestHeaders.UserAgent.ParseAdd("Microsoft ASP.NET OAuth middleware");
+            Backchannel.Timeout = Options.BackchannelTimeout;
+            Backchannel.MaxResponseContentBufferSize = 1024 * 1024 * 10; // 10 MB
```
davidfowl: Configurable via options?

src/Microsoft.AspNet.Authentication.OpenIdConnect/project.json
```diff
@@ -3,8 +3,10 @@
     "description": "ASP.NET 5 middleware that enables an application to support OpenIdConnect authentication workflow.",
     "dependencies": {
         "Microsoft.AspNet.Authentication": "1.0.0-*",
-        "Microsoft.Framework.NotNullAttribute.Sources": { "type": "build", "version": "1.0.0-*" },
-        "Microsoft.IdentityModel.Protocol.Extensions": "2.0.0-beta4-*"
+        "Microsoft.IdentityModel.Protocol.Extensions": "2.0.0-beta4-*",
+        "System.Threading.Tasks": "4.0.10-beta-*",
```
davidfowl: Move these into the dnxcore50 specific dependencies. All the System. packages

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
```diff
@@ -100,6 +105,11 @@ public class OpenIdConnectAuthenticationMiddleware : AuthenticationMiddleware<Op
                 Options.TokenValidationParameters.ValidAudience = Options.ClientId;
             }
 
+            Backchannel = new HttpClient(ResolveHttpMessageHandler(Options));
+            Backchannel.DefaultRequestHeaders.UserAgent.ParseAdd("Microsoft ASP.NET OAuth middleware");
```
Tratcher: Change this value to say OpenIdConnect instead of OAuth.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
Tratcher: It isn't configurable on any of our providers. https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Twitter/TwitterAuthenticationMiddleware.cs#L80

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: @PinpointTownes How can I developer provide "userinfo_endpoint" value in app.UseOpenIdConnectAuthentication? I couldn't find any property on Options or TokenValidationParameters.

PinpointTownes: You can provide your own `OpenIdConnectConfiguration` instance instead of relying on metadata discovery: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs#L152

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/K/src/Microsoft.IdentityModel.Protocol.Extensions/Configuration/OpenIdConnectConfiguration.cs#L252

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -321,6 +334,18 @@ protected override AuthenticationTicket AuthenticateCore()
                     _configuration = await Options.ConfigurationManager.GetConfigurationAsync(Context.RequestAborted);
                 }
 
+                // Redeeming authorization code for tokens
+                if (string.IsNullOrWhiteSpace(message.IdToken) && isCodeOnlyFlow)
```
brentschmaltz: just because Options.ResponseType == CodeOnly, that does not mean the message is 'code' only. We need to determine the type of response from the actual message received. This is particularly important when a site is being attacked.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -520,6 +557,74 @@ protected override AuthenticationTicket AuthenticateCore()
...
 
+        protected virtual async Task<OpenIdConnectMessage> RedeemAuthorizationCode(string authorizationCode)
+        {
+            var openIdMessage = new OpenIdConnectMessage()
+            {
+                ClientId = Options.ClientId,
+                ClientSecret = Options.ClientSecret,
+                Code = authorizationCode,
+                GrantType = "authorization_code",
+                RedirectUri = Options.RedirectUri
```
brentschmaltz: redirectUri needs to be obtained from the request (authenticationProperties) sent as part of state if available. Otherwise use this default.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -299,12 +310,16 @@ protected override AuthenticationTicket AuthenticateCore()
...
 
                 var properties = GetPropertiesFromState(message.State);
+
                 if (properties == null)
                 {
                     Logger.LogError(Resources.OIDCH_0005_MessageStateIsInvalid);
                     return null;
                 }
 
+                var isCodeOnlyFlow = (properties.Items.ContainsKey(OpenIdConnectParameterNames.ResponseType) ?
```
PinpointTownes: I'm still not sure it's the most appropriate way to determine whether an OIDC callback really corresponds to a "code only" flow: nothing prevents a server from ignoring your `response_type` and returning an `id_token` or a `token`.

IMHO, making sure there's no `id_token` or `token` in the message would be far more robust: https://github.com/aspnet/Security/pull/279/files#diff-5ba88763cd4e76f8ecfcccff1ca9984aR266

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -1,11 +1,14 @@
-// Copyright (c) .NET Foundation. All rights reserved.
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
PinpointTownes: BOM?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: I do check for a null id_token in the incoming message along with this boolean before the code is used to redeem tokens. This boolean is an indication to take an action when a null id_token is received. If this boolean is not set and the id_token is null, we go directly to processing the code.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: Added the response_type property to the state.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -520,6 +566,90 @@ protected override AuthenticationTicket AuthenticateCore()
             }
         }
 
+        protected virtual async Task<OpenIdConnectMessage> RedeemAuthorizationCode(string authorizationCode, string redirectUri)
+        {
+            var openIdMessage = new OpenIdConnectMessage()
+            {
+                ClientId = Options.ClientId,
+                ClientSecret = Options.ClientSecret,
```
brentschmaltz: We can start with ClientSecret, but we will need an issue for: Certs and SignedJwt (AssertionCredentials).

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -520,6 +566,90 @@ protected override AuthenticationTicket AuthenticateCore()
...
+                Code = authorizationCode,
+                GrantType = "authorization_code",
```
brentschmaltz: Look for a constant in Wilson, if not open an issue in Wilson.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -520,6 +566,90 @@ protected override AuthenticationTicket AuthenticateCore()
...
+
+            var requestMessage = new HttpRequestMessage(HttpMethod.Get, userInfoEndpoint);
+            requestMessage.Headers.Authorization = new AuthenticationHeaderValue("Bearer", message.AccessToken);
+            var responseMessage = await Backchannel.SendAsync(requestMessage);
+            responseMessage.EnsureSuccessStatusCode();
+            var userInfoResponse = await responseMessage.Content.ReadAsStringAsync();
+            var user = JObject.Parse(userInfoResponse);
+
+            var identity = (ClaimsIdentity)ticket.Principal.Identity;
+            var subject = identity.FindFirst(ClaimTypes.NameIdentifier).Value;
```
brentschmaltz: Should we specifically look for the 'sub' claim. Best to check with Mike Jones. Need to be mind-full of how claims mapping will influence this. When you use the ClaimsIdentity, a mapping has been performed by the JwtSecurityTokenHandler.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -520,6 +566,90 @@ protected override AuthenticationTicket AuthenticateCore()
...
+                throw new ArgumentException(Resources.OIDCH_0038_Subject_Claim_Mismatch); 
+            }
+
+            var userInfoIdentity = new ClaimsIdentity(identity);
+            foreach (var pair in user)
+            {
+                JToken value;
+                var claimValue = user.TryGetValue(pair.Key, out value) ? value.ToString() : null;
+
+                var inboundClaimTypeMap = new Dictionary<string, string>(JwtSecurityTokenHandler.InboundClaimTypeMap);
```
brentschmaltz: Each time a claim is mapped by the JwtSecurityTokenHandler the 'short' name is captured in the properties of the claim. See JwtSecurityTokenHandler.ShortNameProperty

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -164,6 +165,11 @@ public string Caption
         public bool DefaultToCurrentUriOnRedirect { get; set; }
 
         /// <summary>
+        /// Boolean to set whether the middleware should go to user info endpoint to retrieve claims or not.
```
brentschmaltz: We want to indicate that this is an additional step after setting the Identity to the Id_token that arrived from the TokenEndpoint.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -183,6 +193,12 @@ protected override void ApplyResponseChallenge()
                 State = OpenIdConnectAuthenticationDefaults.AuthenticationPropertiesKey + "=" + UrlEncoder.UrlEncode(Options.StateDataFormat.Protect(properties))
             };
 
+            // if response_type=code, nonce is not required.
+            if (Options.ResponseType.Equals(OpenIdConnectResponseTypes.Code))
+            {
+                Options.ProtocolValidator.RequireNonce = false;
```
brentschmaltz: This will influence id_token + code that comes on a different message.
We need to make setting this independent of the value on ProtocolValidator.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -183,6 +193,12 @@ protected override void ApplyResponseChallenge()
                 State = OpenIdConnectAuthenticationDefaults.AuthenticationPropertiesKey + "=" + UrlEncoder.UrlEncode(Options.StateDataFormat.Protect(properties))
             };
 
+            // if response_type=code, nonce is not required.
+            if (Options.ResponseType.Equals(OpenIdConnectResponseTypes.Code))
```
brentschmaltz: Need to check the value of OpenIdConnectMessage.ResponseType, since it can be set in RedirectToIdentityProvider later. So any micro changes should be based off of the final request sent to the IDP.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: Issue opened in Wilson https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/186

tushargupta51:
Closing this, will send another PR later.

PR #267, opened 2015-05-19
[Prototype] generic notifications

HaoK:
This seems to work and not look too bad, do you want to try something like this in your PR @PinpointTownes 

Pick something more reasonable for the naming :)

cc @Tratcher


PinpointTownes:
@HaoK I'm trying a similar approach locally, but with `Authenticated` defined at the base notifications class level (we need it for the case where you don't override the handler and only use the delegate approach).

HaoK:
Yeah this looks promising, I think we just need a new common base class that all of the social providers can use which has the common Authenticated/user properties that they all share.  I'll leave the details for you to play with, I just wanted to make sure it worked

Tratcher:
OAuth is the common base class for the social providers...

HaoK:
So you think we can just roll all the changes I made in OAuthSquared into the base classes?

Tratcher:
We should try. We would need a very compelling reason not to.

PinpointTownes:
@HaoK we can't, we MUST :smile:

HaoK:
Yeah, I'm happy as long as all the duplicated code finally goes away :)

PinpointTownes:
https://github.com/PinpointTownes/Security/commit/466b9a59dac810e8223d760bf76a945e3342f988

Issue #265, opened 2015-05-19
[AuthZ] Consider new sugar for adding auth handler


HaoK:
@divega suggested something like this to make it easy to register a handler/requirement all at once without implementing a new class/interface (the sugar only would exist for the sync scenario):

[Need a new name other than Magic...]

```
        policy.AddRequirement(new MagicRequirement(context, req => { context.Fail() }));

        // Add this come up name
        public class MagicRequirement : AuthorizationHandler<MagicRequirement>, IAuthorizationRequirement
        {
            public Action<AuthorizationContext, MagicRequirement> Handler { get;  }

            public MagicRequirement(Action<AuthorizationContext, MagicRequirement> handleMe)
            { Handler = handleMe; }

            protected override void Handle(AuthorizationContext context, MagicRequirement requirement) 
            { Handler.Invoke(context, requirement); }
```

HaoK:
cc @blowdart thoughts?

blowdart:
They'd have to be very simple requirements then if you wanted to specify them like that, but, meh, I have no objections, as long as the existing methods still work for more "grown up" requirements

HaoK:
https://github.com/aspnet/Security/pull/283

Issue #235, opened 2015-05-01
Make ClaimsTransformation Async?


HaoK:
Requested by customers

leastprivilege:
Chances are high you want to interact with a data store and / or cache.

HaoK:
https://github.com/aspnet/Security/pull/283

PR #269, opened 2015-05-19
React to Diego :)

HaoK:
Some suggestions Diego had:

- DelegateRequirement which lets you register a handler directly without creating a IRequirement passthrough handler class
- Fix issue with one of the AuthHandlers where Async path wasn't correct
- Misc cleanup

Fixes https://github.com/aspnet/Security/issues/265
Fixes https://github.com/aspnet/Security/issues/235

cc @divega @lodejard @blowdart

### DIFF
--- a/src/Microsoft.AspNet.Authentication/ClaimsTransformationOptions.cs
+++ b/src/Microsoft.AspNet.Authentication/ClaimsTransformationOptions.cs
@@ -3,11 +3,42 @@
 
 using System;
 using System.Security.Claims;
+using System.Threading.Tasks;
 
 namespace Microsoft.AspNet.Authentication
 {
     public class ClaimsTransformationOptions
     {
-        public Func<ClaimsPrincipal, ClaimsPrincipal> Transformation { get; set; }
+        public IClaimsTransformer Transformer { get; set; }
+    }
+
+    public interface IClaimsTransformer
+    {
+        Task<ClaimsPrincipal> TransformAsync(ClaimsPrincipal principal);
+        ClaimsPrincipal Transform(ClaimsPrincipal principal);
+    }
+
+    public class ClaimsTransformer : IClaimsTransformer
+    {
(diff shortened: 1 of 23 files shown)


HaoK:
Updated PR making ClaimsTransformation async

HaoK:
@divega @Tratcher can you guys review?

Tratcher:
You may have to get someone else, I have no idea what's going on here :-)

src/Microsoft.AspNet.Authentication/ClaimsTransformationOptions.cs
```diff
@@ -3,11 +3,42 @@
...
+
+    public interface IClaimsTransformer
+    {
+        Task<ClaimsPrincipal> TransformAsync(ClaimsPrincipal principal);
+        ClaimsPrincipal Transform(ClaimsPrincipal principal);
+    }
+
+    public class ClaimsTransformer : IClaimsTransformer
+    {
+        public Func<ClaimsPrincipal, Task<ClaimsPrincipal>> TransformAsyncDelegate { get; set; }
```
HaoK: Change this to a ctor that takes only sync, and one that takes both, and no longer make these settable

src/Microsoft.AspNet.Authorization/DelegateRequirement.cs
```diff
@@ -0,0 +1,22 @@
...
+        public DelegateRequirement(Action<AuthorizationContext, DelegateRequirement> handleMe)
+        {
+            Handler = handleMe;
+        }
+
+        protected override void Handle(AuthorizationContext context, DelegateRequirement requirement)
+        {
+            Handler.Invoke(context, requirement);
+        }
+    }
```
HaoK: We'll keep delegate as sync only, because its only supposed to take those two inputs, and is unable to use injection

HaoK:
Combined into https://github.com/aspnet/Security/pull/283

Issue #246, opened 2015-05-06
Failed authorization for an authenticated user redirects to login page instead of access denied page


leastprivilege:
A failed role check with the Authorize attribute redirects an authenticated user back to the login page. I haven't checked the internals, but that seems to indicate you are emitting a 401.

My assumption was that in that situation a 403 should be emitted and the cookie middleware will use the AccessDeniedPath instead.

Tratcher:
See https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs#L319

brockallen:
Yes, that API is very easy to read :)

leastprivilege:
OK - but I am returning a 403 for an authenticated user - and this code path does not seem to get hit..

Tratcher:
You'd think that 403 would be the appropriate thing to return, but what they eventually settled on is returning a 401 and having the auth middleware that produced the user identity (cookies) change it to a 403 OR redirect to AccessDeniedPath.

HaoK:
Yep, need to convince @lodejard if you want to change this behavior...

PinpointTownes:
Related ticket: https://github.com/aspnet/Security/issues/134

leastprivilege:
OK - fair enough. 

But still it does not work.

I authenticate using cookie MW and hit a Authorize attribute with a role check - role check fails and I get redirected to login page. I have an access denied page set on the cookie MW.

Am I doing something wrong?

Tratcher:
@HaoK ?

bojanrajkovic:
So, according to @lodejard, they should already be acting like this: https://twitter.com/loudej/status/601074953836834816

In my tests with 1.0.0-beta5-12413 versions of AuthN/AuthZ packages, this isn't working -- an authenticated, but not authorized, user is still getting a 302 redirect to the login page.

HaoK:
Looks like the issue is 401->403 not working for automatic authentication, which is always the case for the Authorize(Roles = "Foo").   Things seem to work via an explicit policy where ActiveAuthenticationScheme includes the Cookie...  Still investigating

HaoK:
Yup, confirmed this workaround should work @leastprivilege @bojanrajkovic can you try this and see fixes the issue?

```
                // This policy cannot succeed since the claim is never added
                options.AddPolicy("Cookie", policy => {
                    policy.AddAuthenticationSchemes("Cookies");
[...]

Issue #273, opened 2015-05-22
[AuthN] Clean up 401-403 AuthenticateCalled


HaoK:
Currently for 401-403 to work, the AuthenticateCalled flag must be set.  This is a pretty obscure thing for subclasses to be required to do and is very easy to miss.  Cookie was missing this and as a result broke 401-403.  We should revisit this and see if there's anyway to make this more seamless...

@blowdart

HaoK:
# Summary of plan:

## HttpAbstractions:
```
public enum ChallengeBehavior { Automatic, Unauthorized, Forbidden }

Add ChallengeBehavior property to ChallengeContext

new methods on AuthenticationManager:

// calls Challenge with Forbidden behavior
public void Forbidden(string authenticationScheme, AuthenticationProperties properties)

// Existing challenge uses Automatic behavior (where the middleware decides challenge behavior
```

## AuthenticationHandler changes:

- AuthenticationScheme is required for SIgnIn/SignOut, no more overloads that sign in/sign out all automaticAuthentication middlewares (this setting now only affects ApplyResponseChallenge automatically getting called on the way out, and Authenticate getting called on the way in)

- Immediately change status codes/set headers/cookies for SignIn/SignOut/Challenge instead of deferring the changes until ApplyResponse
- ApplyResponseChallenge will still need to check for the AutomaticAuthentication chase which will then call handler's Challenge method explicitly (the non chained version)

- There will be a helper method in AuthenticationHandler which uses new ChallengeBehavior property:
```
        public virtual void Challenge(ChallengeContext context)
        {
            if (ShouldHandleScheme(context.AuthenticationScheme))
            {
                HandleChallenge(context)
                context.Accept();
            }

            if (PriorHandler != null)
            {
                PriorHandler.Challenge(context);
            }
        }

// If this returns true, challenge method should stop and just return 
protected virtual void HandleChallenge(ChallengeContext context) {
   switch (behavior) {
     case Automatic:
        if (Response.StatusCode != 401) return false;
        var ticket = Authenticate(options.AuthenticationScheme);
        var foundPrincipal = ticket.Principal != null;
        if (!foundPrincipal) {
           HandleForbidden();
         }
         HandleUnauthorized();
         return;
     case Forbidden:
         HandleForbidden()
         return;
     case Unauthorized:
         HandleUnauthorized();
         return;
   }
}

// Cookie Middleware would override this to 302 to AccessDenied if options set
public virtual void HandleForbidden() {
[...]

PR #274, opened 2015-05-22
Fix 401-403 for automatic cookies

HaoK:
Cookies was overriding the base impl of AuthenticateAsync so the flag wasn't getting set which made the 401-403 check fail...  

Definitely some smell here, filed https://github.com/aspnet/Security/issues/273 to fix

cc @blowdart @Tratcher @lodejard


HaoK:
Fixes https://github.com/aspnet/Security/issues/246

HaoK:
@blowdart do we want this for beta 5 (aka today?)

blowdart:
Yea, we should get it in.

HaoK:
merged e54d088c462b18c76a8f218b11ea2614e41622b6

PR #276, opened 2015-05-24
Incorporate community feedback from https://github.com/aspnet/Security/pull/112

PinpointTownes:
This pull request fixes a few things that were not solved by @brentschmaltz's PR (https://github.com/aspnet/Security/pull/112).
Please note that it's still a WIP and that I'll probably add more commits.

/cc @HaoK @brentschmaltz

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationMiddleware.cs
+++ b/src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationMiddleware.cs
@@ -2,12 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
-using System.IdentityModel.Tokens;
-using System.IdentityModel.Tokens.Jwt;
 using System.Net.Http;
-using Microsoft.AspNet.Authentication;
 using Microsoft.AspNet.Builder;
 using Microsoft.Framework.Internal;
 using Microsoft.Framework.Logging;
@@ -41,11 +37,6 @@ namespace Microsoft.AspNet.Authentication.OAuthBearer
             if (Options.Notifications == null)
             {
                 Options.Notifications = new OAuthBearerAuthenticationNotifications();
-            }
-
-            if (Options.SecurityTokenValidators == null)
-            {
-                Options.SecurityTokenValidators = new List<ISecurityTokenValidator> { new JwtSecurityTokenHandler() };
             }
 
             if (string.IsNullOrEmpty(Options.TokenValidationParameters.ValidAudience) && !string.IsNullOrEmpty(Options.Audience))
--- a/src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
+++ b/src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
@@ -4,8 +4,8 @@
 using System;
 using System.Collections.Generic;
 using System.IdentityModel.Tokens;
+using System.IdentityModel.Tokens.Jwt;
 using System.Net.Http;
-using Microsoft.Framework.Internal;
 using Microsoft.IdentityModel.Protocols;
 using Microsoft.IdentityModel.Protocols.OpenIdConnect;
 
@@ -16,21 +16,12 @@ namespace Microsoft.AspNet.Authentication.OAuthBearer
     /// </summary>
     public class OAuthBearerAuthenticationOptions : AuthenticationOptions
     {
-        private ICollection<ISecurityTokenValidator> _securityTokenValidators;
-        private TokenValidationParameters _tokenValidationParameters;
-
         /// <summary>
         /// Creates an instance of bearer authentication options with default values.
         /// </summary>
         public OAuthBearerAuthenticationOptions() : base()
         {
             AuthenticationScheme = OAuthBearerAuthenticationDefaults.AuthenticationScheme;
-            BackchannelTimeout = TimeSpan.FromMinutes(1);
-            Challenge = OAuthBearerAuthenticationDefaults.AuthenticationScheme;
-            Notifications = new OAuthBearerAuthenticationNotifications();
-            RefreshOnIssuerKeyNotFound = true;
-            SystemClock = new SystemClock();
-            TokenValidationParameters = new TokenValidationParameters();
         }
 
         /// <summary>
@@ -54,15 +45,14 @@ namespace Microsoft.AspNet.Authentication.OAuthBearer
         /// <summary>
         /// Gets or sets the challenge to put in the "WWW-Authenticate" header.
         /// </summary>
-        /// TODO - brentschmaltz, should not be null.
-        public string Challenge { get; set; }
+        public string Challenge { get; set; } = OAuthBearerAuthenticationDefaults.AuthenticationScheme;
 
         /// <summary>
         /// The object provided by the application to process events raised by the bearer authentication middleware.
         /// The application may implement the interface fully, or it may create an instance of OAuthBearerAuthenticationProvider
         /// and assign delegates only to the events it wants to process.
         /// </summary>
-        public OAuthBearerAuthenticationNotifications Notifications { get; set; }
+        public OAuthBearerAuthenticationNotifications Notifications { get; set; } = new OAuthBearerAuthenticationNotifications();
 
         /// <summary>
         /// The HttpMessageHandler used to retrieve metadata.
@@ -74,7 +64,7 @@ namespace Microsoft.AspNet.Authentication.OAuthBearer
         /// <summary>
         /// Gets or sets the timeout when using the backchannel to make an http call.
         /// </summary>
-        public TimeSpan BackchannelTimeout { get; set; }
+        public TimeSpan BackchannelTimeout { get; set; } = TimeSpan.FromMinutes(1);
 
 #if DNX451
         /// <summary>
@@ -104,48 +94,24 @@ namespace Microsoft.AspNet.Authentication.OAuthBearer
         /// Gets or sets if a metadata refresh should be attempted after a SecurityTokenSignatureKeyNotFoundException. This allows for automatic
         /// recovery in the event of a signature key rollover. This is enabled by default.
         /// </summary>
-        public bool RefreshOnIssuerKeyNotFound { get; set; }
+        public bool RefreshOnIssuerKeyNotFound { get; set; } = true;
 
         /// <summary>
         /// Used to know what the current clock time is when calculating or validating token expiration. When not assigned default is based on
         /// DateTimeOffset.UtcNow. This is typically needed only for unit testing.
         /// </summary>
-        public ISystemClock SystemClock { get; set; }
+        public ISystemClock SystemClock { get; set; } = new SystemClock();
 
         /// <summary>
-        /// Gets or sets the <see cref="SecurityTokenValidators"/> for validating tokens.
+        /// Gets the ordered list of <see cref="ISecurityTokenValidator"/> used to validate access tokens.
         /// </summary>
-        /// <exception cref="ArgumentNullException">if 'value' is null.</exception>
-        public ICollection<ISecurityTokenValidator> SecurityTokenValidators
-        {
-            get
-            {
-                return _securityTokenValidators;
-            }
-
-            [param: NotNull]
-            set
-            {
-                _securityTokenValidators = value;
-            }
-        }
+        public IList<ISecurityTokenValidator> SecurityTokenValidators { get; } = new List<ISecurityTokenValidator> { new JwtSecurityTokenHandler() };
 
         /// <summary>
-        /// Gets or sets the TokenValidationParameters
+        /// Gets or sets the parameters used to validate identity tokens.
         /// </summary>
         /// <remarks>Contains the types and definitions required for validating a token.</remarks>
         /// <exception cref="ArgumentNullException">if 'value' is null.</exception>
-        public TokenValidationParameters TokenValidationParameters
-        {
-            get
-            {
-                return _tokenValidationParameters;
-            }
-            [param: NotNull]
-            set
-            {
-                _tokenValidationParameters = value;
-            }
-        }
+        public TokenValidationParameters TokenValidationParameters { get; set; } = new TokenValidationParameters();
     }
 }
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
@@ -5,11 +5,12 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.IdentityModel.Tokens;
+using System.IdentityModel.Tokens.Jwt;
 using System.Net.Http;
+using System.Security.Claims;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Authentication;
 using Microsoft.Framework.Caching.Distributed;
-using Microsoft.Framework.Internal;
 using Microsoft.IdentityModel.Protocols;
 using Microsoft.IdentityModel.Protocols.OpenIdConnect;
 
@@ -20,13 +21,6 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
     /// </summary>
     public class OpenIdConnectAuthenticationOptions : AuthenticationOptions
     {
-        private TimeSpan _backchannelTimeout;
-        private OpenIdConnectProtocolValidator _protocolValidator;
-        private ICollection<ISecurityTokenValidator> _securityTokenValidators;
-        private ISecureDataFormat<AuthenticationProperties> _stateDataFormat;
-        private ISecureDataFormat<string> _stringDataFormat;
-        private TokenValidationParameters _tokenValidationParameters;
-
         /// <summary>
         /// Initializes a new <see cref="OpenIdConnectAuthenticationOptions"/>
         /// </summary>
@@ -55,16 +49,7 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
         public OpenIdConnectAuthenticationOptions(string authenticationScheme)
         {
             AuthenticationScheme = authenticationScheme;
-            BackchannelTimeout = TimeSpan.FromMinutes(1);
             Caption = OpenIdConnectAuthenticationDefaults.Caption;
-            GetClaimsFromUserInfoEndpoint = false;
-            ProtocolValidator = new OpenIdConnectProtocolValidator() { RequireState = false };
-            RefreshOnIssuerKeyNotFound = true;
-            ResponseMode = OpenIdConnectResponseModes.FormPost;
-            ResponseType = OpenIdConnectResponseTypes.CodeIdToken;
-            Scope = OpenIdConnectScopes.OpenIdProfile;
-            TokenValidationParameters = new TokenValidationParameters();
-            UseTokenLifetime = true;
         }
 
         /// <summary>
@@ -103,23 +88,7 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
         /// Gets or sets the timeout when using the backchannel to make an http call.
         /// </summary>
         [SuppressMessage("Microsoft.Usage", "CA2208:InstantiateArgumentExceptionsCorrectly", Justification = "By design we use the property name in the exception")]
-        public TimeSpan BackchannelTimeout
-        {
-            get
-            {
-                return _backchannelTimeout;
-            }
-
-            set
-            {
-                if (value <= TimeSpan.Zero)
-                {
-                    throw new ArgumentOutOfRangeException(nameof(BackchannelTimeout), value, Resources.OIDCH_0101_BackChallnelLessThanZero);
-                }
-
-                _backchannelTimeout = value;
-            }
-        }
+        public TimeSpan BackchannelTimeout { get; set; } = TimeSpan.FromSeconds(60);
 
         /// <summary>
         /// Get or sets the text that the user can display on a sign in user interface.
@@ -192,25 +161,14 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
         /// <summary>
         /// Gets or sets the <see cref="OpenIdConnectAuthenticationNotifications"/> to notify when processing OpenIdConnect messages.
         /// </summary>
-        public OpenIdConnectAuthenticationNotifications Notifications { get; set; }
+        public OpenIdConnectAuthenticationNotifications Notifications { get; set; } = new OpenIdConnectAuthenticationNotifications();
 
         /// <summary>
         /// Gets or sets the <see cref="OpenIdConnectProtocolValidator"/> that is used to ensure that the 'id_token' received
         /// is valid per: http://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation 
         /// </summary>
         /// <exception cref="ArgumentNullException">if 'value' is null.</exception>
-        public OpenIdConnectProtocolValidator ProtocolValidator
-        {
-            get
-            {
-                return _protocolValidator;
-            }
-            [param: NotNull]
-            set
-            {
-                _protocolValidator = value;
-            }
-        }
+        public OpenIdConnectProtocolValidator ProtocolValidator { get; set; } = new OpenIdConnectProtocolValidator { RequireState = false };
 
         /// <summary>
         /// Gets or sets the 'post_logout_redirect_uri'
@@ -230,7 +188,7 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
         /// Gets or sets if a metadata refresh should be attempted after a SecurityTokenSignatureKeyNotFoundException. This allows for automatic
         /// recovery in the event of a signature key rollover. This is enabled by default.
         /// </summary>
-        public bool RefreshOnIssuerKeyNotFound { get; set; }
+        public bool RefreshOnIssuerKeyNotFound { get; set; } = true;
 
         /// <summary>
         /// Gets or sets the 'resource'.
@@ -240,103 +198,49 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
         /// <summary>
         /// Gets or sets the 'response_mode'.
         /// </summary>
-        public string ResponseMode { get; set; }
+        public string ResponseMode { get; set; } = OpenIdConnectResponseModes.FormPost;
 
         /// <summary>
         /// Gets or sets the 'response_type'.
         /// </summary>
-        public string ResponseType { get; set; }
+        public string ResponseType { get; set; } = OpenIdConnectResponseTypes.CodeIdToken;
 
         /// <summary>
-        /// Gets or sets the 'scope'.
+        /// Gets the list of permissions to request.
         /// </summary>
-        public string Scope { get; set; }
+        public IList<string> Scope { get; } = new List<string> { "openid", "profile" };
 
         /// <summary>
-        /// Gets or sets the SignInScheme which will be used to set the <see cref="System.Security.Claims.ClaimsIdentity.AuthenticationType"/>.
+        /// Gets or sets the SignInScheme which will be used to set the <see cref="ClaimsIdentity.AuthenticationType"/>.
         /// </summary>
         public string SignInScheme { get; set; }
 
         /// <summary>
         /// Gets or sets the type used to secure data handled by the middleware.
         /// </summary>
-        public ISecureDataFormat<AuthenticationProperties> StateDataFormat
-        {
-            get
-            {
-                return _stateDataFormat;
-            }
-            [param: NotNull]
-            set
-            {
-                _stateDataFormat = value;
-            }
-        }
+        public ISecureDataFormat<AuthenticationProperties> StateDataFormat { get; set; }
 
         /// <summary>
         /// Gets or sets the type used to secure strings used by the middleware.
         /// </summary>
-        public ISecureDataFormat<string> StringDataFormat
-        {
-            get
-            {
-                return _stringDataFormat;
-            }
-            [param: NotNull]
-            set
-            {
-                _stringDataFormat = value;
-            }
-        }
+        public ISecureDataFormat<string> StringDataFormat { get; set; }
 
         /// <summary>
-        /// Gets or sets the <see cref="SecurityTokenValidators"/> for validating tokens.
+        /// Gets or sets the <see cref="ISecurityTokenValidator"/> used to validate identity tokens.
         /// </summary>
-        /// <exception cref="ArgumentNullException">if 'value' is null.</exception>
-        public ICollection<ISecurityTokenValidator> SecurityTokenValidators
-        {
-            get
-            {
-                return _securityTokenValidators;
-            }
-
-            set
-            {
-                if (value == null)
-                {
-                    throw new ArgumentNullException("SecurityTokenValidators");
-                }
-
-                _securityTokenValidators = value;
-            }
-        }
+        public ISecurityTokenValidator SecurityTokenValidator { get; set; } = new JwtSecurityTokenHandler();
 
         /// <summary>
-        /// Gets or sets the TokenValidationParameters
+        /// Gets or sets the parameters used to validate identity tokens.
         /// </summary>
         /// <remarks>Contains the types and definitions required for validating a token.</remarks>
-        public TokenValidationParameters TokenValidationParameters
-        {
-            get
-            {
-                return _tokenValidationParameters;
-            }
-            [param: NotNull]
-            set
-            {
-                _tokenValidationParameters = value;
-            }
-        }
+        public TokenValidationParameters TokenValidationParameters { get; set; } = new TokenValidationParameters();
 
         /// <summary>
         /// Indicates that the authentication session lifetime (e.g. cookies) should match that of the authentication token.
         /// If the token does not provide lifetime information then normal session lifetimes will be used.
         /// This is enabled by default.
         /// </summary>
-        public bool UseTokenLifetime
-        {
-            get;
-            set;
-        }
+        public bool UseTokenLifetime { get; set; } = true;
     }
 }
--- a/src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationMiddleware.cs
+++ b/src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationMiddleware.cs
@@ -63,6 +63,11 @@ namespace Microsoft.AspNet.Authentication.OAuth
                 throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(Options.TokenEndpoint)));
             }
 
+            if (Options.Notifications == null)
+            {
+                Options.Notifications = new OAuthAuthenticationNotifications();
+            }
+
             if (Options.StateDataFormat == null)
             {
                 var dataProtector = dataProtectionProvider.CreateProtector(
--- a/src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs
+++ b/src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs
@@ -7,7 +7,6 @@ using System.Net.Http;
 using System.Security.Claims;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Authentication;
-using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Authentication.OAuth
 {
@@ -83,7 +82,7 @@ namespace Microsoft.AspNet.Authentication.OAuth
         /// <summary>
         /// Gets or sets the <see cref="IOAuthAuthenticationNotifications"/> used to handle authentication events.
         /// </summary>
-        public IOAuthAuthenticationNotifications Notifications { get; [param: NotNull] set; } = new OAuthAuthenticationNotifications();
+        public IOAuthAuthenticationNotifications Notifications { get; set; } = new OAuthAuthenticationNotifications();
 
         /// <summary>
         /// A list of permissions to request.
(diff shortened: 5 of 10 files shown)


HaoK:
Looks good so far...

HaoK:
I actually like this suggestion from the PR you refer to (assuming its true)...

Rename OAuthBearerAuthenticationMiddleware => TokenAuthenticationMiddleware

HaoK:
But I see lots of stuff that doesn't look general enough to call this TokenAuthenticationOptions in here though... with a bunch of OpenId stuff on there.

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs

PinpointTownes:
Renaming the OAuth2 bearer middleware was not part of the plan simply because there was no consensus concerning my suggestion :smile: 

IMHO, the best approach (as explained here: https://github.com/aspnet/Security/pull/112#issuecomment-69846740 and here: https://github.com/aspnet/Security/pull/112#discussion_r22930679) would probably consist in splitting the current OAuth2 bearer middleware into 2 middleware:
* A basic `TokenAuthenticationMiddleware`, designed like the OAuth2 interactive middleware (with a public handler whose methods are protected and virtual and with generic notifications if you don't want to subclass it).
* A "new" `OpenIdConnectTokenAuthenticationMiddleware`, subclassing `TokenAuthenticationMiddleware` and using the magic OIDC provider configuration discovery to validate tokens.

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
```diff
@@ -103,53 +93,25 @@ public OAuthBearerAuthenticationOptions() : base()
...
-                return _securityTokenValidators;
-            }
-
-            [param: NotNull]
-            set
-            {
-                _securityTokenValidators = value;
-            }
-        }
+        public IList<ISecurityTokenValidator> SecurityTokenValidators { get; } = new List<ISecurityTokenValidator>();
```
tushargupta51: Why there is no setter on this?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -246,96 +212,30 @@ public OpenIdConnectProtocolValidator ProtocolValidator
...
-            {
-                if (value == null)
-                {
-                    throw new ArgumentNullException("value");
-                }
-
-                _stateDataFormat = value;
-            }
-        }
+        public ISecureDataFormat<AuthenticationProperties> StateDataFormat { get; set; }
```
tushargupta51: Should we have a [param: NotNull] to make sure this property does not get set to null?

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
PinpointTownes: Writable collection properties are generally considered as a bad practice: https://msdn.microsoft.com/fr-fr/library/ms182327(v=vs.140).aspx

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: It cannot be null: the OIDC middleware always creates a default format if you don't provide one.

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs#L55-L64

tushargupta51: That is for setting the default, the checks that you have removed were making sure that user cannot set it to null even explicitly.

PinpointTownes: I'm not sure that adding a NotNull check for a property whose default value is never set by the class itself is really useful. 

If you want to add a NotNull attribute, I think we'll have to add everywhere we use these "format properties" to be consistent.

HaoK: Not null doesn't makes sense when we use null to mean the default. If someone goes and sets it back to null later. That seems like programmer error at least in the pattern we have today. 

If we want to change this pattern it needs to be done consistently across the board all at once.

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
brentschmaltz: I have seen that rule many times, i find it leads to some odd code. How will a user set the validators they want to use?  It is a little more cumbersome. I think that rule was written before the new syntax like: new foo { x = this, y = that ...}

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: hang on, are you saying we should allow the user to set this to null?

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
```diff
@@ -103,53 +93,25 @@ public OAuthBearerAuthenticationOptions() : base()
...
-            {
-                if (value == null)
-                {
-                    throw new ArgumentNullException("TokenValidationParameters");
-                }
-
-                _tokenValidationParameters = value;
-            }
-        }
+        public TokenValidationParameters TokenValidationParameters { get; } = new TokenValidationParameters();
```
brentschmaltz: why remove the setter?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -246,96 +212,30 @@ public OpenIdConnectProtocolValidator ProtocolValidator
...
-            {
-                if (value == null)
-                {
-                    throw new ArgumentNullException("value");
-                }
-
-                _stringDataFormat = value;
-            }
-        }
+        public ISecureDataFormat<string> StringDataFormat { get; set; }
```
brentschmaltz: you can't allow null here

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -246,96 +212,30 @@ public OpenIdConnectProtocolValidator ProtocolValidator
...
-            {
-                if (value == null)
-                {
-                    throw new ArgumentNullException("value");
-                }
-
-                _tokenValidationParameters = value;
-            }
-        }
+        public TokenValidationParameters TokenValidationParameters { get; } = new TokenValidationParameters();
```
brentschmaltz: we need a non-null setter

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
HaoK: If they want sure, its just options, as @PinpointTownes mentioned the handler has default behavior it uses when not specified.  Null means effectively use your default.  Options are configured once at startup typically, and not meant to be twiddled with later, so yes, this should be nullable.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -246,96 +212,30 @@ public OpenIdConnectProtocolValidator ProtocolValidator
...
-            {
-                if (value == null)
-                {
-                    throw new ArgumentNullException("SecurityTokenValidators");
-                }
-
-                _securityTokenValidators = value;
-            }
-        }
+        public IList<ISecurityTokenValidator> SecurityTokenValidators { get; } = new List<ISecurityTokenValidator>();
```
brentschmaltz: I know the .net guidelines call for this pattern. But I don't like it, it makes it harder to code to and was designed for a different pattern.

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
```diff
@@ -103,53 +93,25 @@ public OAuthBearerAuthenticationOptions() : base()
...
 
         /// <summary>
         /// Used to know what the current clock time is when calculating or validating token expiration. When not assigned default is based on
         /// DateTimeOffset.UtcNow. This is typically needed only for unit testing.
         /// </summary>
-        public ISystemClock SystemClock { get; set; }
+        public ISystemClock SystemClock { get; set; } = new SystemClock();
 
         /// <summary>
         /// Gets or sets the <see cref="SecurityTokenValidators"/> for validating tokens.
```
brentschmaltz: This comment doesn't match the change, which I hope we don't make.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -100,23 +93,7 @@ public OpenIdConnectAuthenticationOptions(string authenticationScheme)
...
-            {
-                if (value <= TimeSpan.Zero)
-                {
-                    throw new ArgumentOutOfRangeException(nameof(BackchannelTimeout), value, Resources.OIDCH_0101_BackChallnelLessThanZero);
-                }
-
-                _backchannelTimeout = value;
-            }
-        }
+        public TimeSpan BackchannelTimeout { get; set; } = TimeSpan.FromSeconds(60);
```
brentschmaltz: Where is the check for negative values?  You can't do this.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: I disagree, you can't allow this to be set to null.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
HaoK: I think its reasonable choice for Options to be plain POCOs (with defaults).  And for consumers of the options (the handler/middlewares) to add whatever error handling/default behaviors they want when consuming the options.

brentschmaltz: Are you saying it makes sense to have a negative value for a timeout?

tushargupta51:
I do not understand the reason for this pull request. Are there any issues opened that this PR is addressing?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: __You CAN'T have null reference exceptions__: `OpenIdConnectAuthenticationMiddleware` always creates a default value if you don't set it: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs#L63

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
PinpointTownes: Being able to both replace the `TokenValidationParameters` and modify its properties led to much friction on JabbR (and even on GitHub: https://github.com/IdentityServer/IdentityServer3/issues/1176#issuecomment-95268305 :))

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: Sure we can, see @HaoK's comment.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: See my remark concerning `OAuthBearerAuthenticationOptions`.

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
PinpointTownes: I'll update it, thanks :smile:

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
PinpointTownes: Very easily, with `Add`: https://github.com/aspnet/Security/pull/276/files#diff-b15c3246438a45f3bed149e5f60b1e22R144

And BTW, the object initializer syntax cannot be used anymore, as we're now using delegates to set the options:

```
app.UseCookieAuthentication(options => {
    options.AutomaticAuthentication = true;
});
```

PinpointTownes:
@tushargupta51 see @brentschmaltz's previous PR: https://github.com/aspnet/Security/pull/112

Tons of things have been omitted and this PR only aims at incorporating them :smile:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: @brentschmaltz I modified it to make it consistent with the OAuth2 client handler: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs#L80

BTW, `HttpClient` will throw a similar exception if the timeout is negative but not infinite, so I'm not sure it's worth keep the check in the options: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs#L123

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: Harder to code? Not sure to understand why.

BTW, I replaced the collection by a list, to make sure people realize that the order is important.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: BTW, you've added ugly tabs there, I'll fix that too :smile:

brentschmaltz:
@PinpointTownes specifically what issues are you addressing?

I see some formatting and style changes, some that break extensiblity, make the product more fragile.

For example, if you can set Options.StateDataFormat to null, then any consumer of that properties must expect that it can be null.

Another example, setting timeout negative, what would be the result of that?

Siting the .Net guidelines as a reason for a change seems to imply that following those guidelines is required for code changes. I don't think that is true.

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
tushargupta51: @PinpointTownes The issue on Github is not related to the ability to set the TVP. It is about the ordering problem. That issue can occur even without the setter if user sets TVP.AuthenticationType after setting the SetDefaultSignInAsAuthenticationType.Also, that issue does not exist in vNext (as far as I can tell) as we do not set the SignInScheme the same way we were setting the SetDefaultSignInAsAuthenticationType

PinpointTownes: It doesn't exist in vNext because I fixed it in a previous PR :smile:

Gimme a strong reason to make it settable (with a code sample, heh :smile:), and I'll happily revert this change (given that I already fixed the other issue).

PinpointTownes:
> @PinpointTownes specifically what issues are you addressing?

No particular issue, just formatting and style things you didn't include when you pushed your previous PR :hushed: 

> I see some formatting and style changes, some that break extensiblity, make the product more fragile.
For example, if you can set Options.StateDataFormat to null, then any consumer of that properties must expect that it can be null.

AGAIN, IT __CANNOT__ BE NULL: a default format relying on the data protection block is automatically instantiated if you leave it to null.

> Another example, setting timeout negative, what would be the result of that?
`TimeSpan.FromMilliseconds(-1.0);` is a valid value for `HttpClient.Timeout`. But I only removed the backing field to make it consistent with the OAuth2 client middleware, where @Tratcher only used an automatic property.

> Siting the .Net guidelines as a reason for a change seems to imply that following those guidelines is required for code changes. I don't think that is true.

But rules are there for a reason, and you need a reason to break them :smile:

PinpointTownes:
@brentschmaltz here's something I'll like to fix: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationHandler.cs#L178

The bearer middleware shouldn't throw by default, as it returns a 500 response instead of a correct challenge. And it's bad, bad, bad :smile:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -226,116 +185,49 @@ public OpenIdConnectProtocolValidator ProtocolValidator
...
         /// Gets or sets the 'response_type'.
         /// </summary>
-        public string ResponseType { get; set; }
+        public string ResponseType { get; set; } = OpenIdConnectResponseTypes.CodeIdToken;
 
         /// <summary>
-        /// Gets or sets the 'scope'.
+        /// Gets the list of permissions to request.
         /// </summary>
-        public string Scope { get; set; }
```
PinpointTownes: Replaced by a list for consistency: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs#L92

PinpointTownes:
Other things I'd like to fix:

* https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs#L175 : if `IssuerAddress` is empty, where will the user agent be redirect to? We should fix that and forbid empty addresses and null configurations.

* https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs#L167 : this kind of null check is just awful and is the direct result of having both a `Configuration` property in the options and a `ConfigurationManager`. We should fix that and make it consistent.

* https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs#L150 : this option is basically pointless and exists nowhere else. I suggest removing it and always using the current URL when there's no explicit `RedirectUri` in the authentication properties (just like the other providers).

* https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs#L164 : storing the callback endpoint in the state parameter is pointless as it always refers to `Options.RedirectUri`, that never changes. It just makes the state - and the whole OIDC payload - unnecessary bigger. It also complicates the code flow (see @tushargupta51's PR and your recent remark).

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: "You CAN'T have null reference exceptions: OpenIdConnectAuthenticationMiddleware always creates a default value if you don't set it" this is only true if you use the default middleware.

Options is passed to all the notifications and anyone can set it too null. Allowing null tells developers that null is OK. 

I will not agree to this.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: I will not agree to this.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: why is order important? There is no reason to make this an IList.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: I don't care what OAuth2 client does. I am not interested in writing code that is broken.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: Modifying the options using the notifications has never been supported and would be thread safe anyway.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: Meh. Sure there's a reason: when the first validator returns true in CanReadToken, the other ones are always excluded.
Surprising you don't even know what your own code does :smile:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: But you DO write broken code, like everybody else :smile: 

@Tratcher thoughts?

Tratcher: Taking offline.... we'll get back to you.

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
tushargupta51: I dont think it works that way. If you are suggesting a change that we take out the setter then you should back it up with a strong reason instead of me giving a reason for having it. "Friction on jabbr" is not a strong enough reason.

PinpointTownes: TokenValidationParameters is fully mutable after being instantiated, so there's no need to replace the whole instance. 

It's your turn now :smile:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: Respectfully, I am not convinced. We are developing a 24/7 framework, we need to minimize the places where a developer can get into trouble. By allowing a property to be set to null, we are telling developers it is OK to set it to null.

PinpointTownes: But IT IS okay to set it to null, as the middleware always creates a default format. 

Making the various options classes immutable to make sure we can't update them from notifications would be far more useful, at least IMHO :smile:

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
brentschmaltz: The Jabbr conversation, has nothing to do with this change. The discussion relates to the need for ClaimsIdentity.AuthenticationType to match the MW. This IMHO is a weakness in leakage between layers.

I get that there are many ways to do things. But there is no strong reason to make this change and force users to take a code change.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: Commented above.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: Well, what I am saying order shouldn't matter. In previous versions we checked that only one handler per token type was in the list. We relaxed that for this release as it seemed a forced restriction.

ICollection supports IList, so I am not following.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: Respectfully, I don't agree. As above we are writing a 7/24 framework. If we allow a negative value, then we are saying the runtime works fine with negative timespans, but in fact it doesn't.

brentschmaltz:
Your points above:
1. IssuerAddress: A user can hook RedirectToIdentityProvider to set the IssuerAddress so the thought is we would block those users if we fault too early.
2. I agree it is not ideal, we need ConfigurationManager to obtain metadata automatically AND for users to be able to set metadata independently. We haven't revisited this yet. What do you propose?
3 / 4. The current way we capture the 'redirect_uri' is broken and needs to be changed. We  need the exact 'redirect_uri' sent to the IdentityProvider for the ConfidentialClient 'code' -> 'access_token' flow. We are capturing too early, after RedirectToIdentityProvider seems better, as the user could have set the value. I don't think the flag will then be needed. My thoughts are the logic is similar to 'state' changes in https://github.com/aspnet/Security/pull/233 where we capture after the notifications.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: > Well, what I am saying order shouldn't matter.

What's sure is that order matters *in the current bits* :smile: 

> ICollection supports IList, so I am not following.

`IList<>` is definitely better suited for enumerations where order is important because it offers an `Insert` method that allows you to choose where your item will be inserted in the list.

That's exactly why MVC 6 uses lists almost everywhere in the options: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/MvcOptions.cs#L83

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
PinpointTownes: Let's make a deal:

I'll revert this change if you promise me you'll __NEVER__ add new properties to the OIDC options wrapping the `TokenValidationParameters` property :smile:

Something like:

```csharp
public string Issuer {
    get { return TokenValidationParameters.ValidIssuer; }
    set { TokenValidationParameters.ValidIssuer = value; }
}
```

PinpointTownes:
> 1. IssuerAddress: A user can hook RedirectToIdentityProvider to set the IssuerAddress so the thought is we would block those users if we fault too early.
2. I agree it is not ideal, we need ConfigurationManager to obtain metadata automatically AND for users to be able to set metadata independently. We haven't revisited this yet. What do you propose?

I'm not saying that we shouldn't have these two properties (it's probably unavoidable BTW), but the way we're currently using them is just weird:
* If you don't provide a metadata address (explicitly or implicitly using the authority) and don't set `Options.Configuration`, your app won't have any information concerning the issuer and will miserably fail during the challenge processing. Even enabling logging won't be helpful in this case, as it will simply log the generic (and cryptic) `OIDCH_0036: Uri.IsWellFormedUriString(redirectUri, UriKind.Absolute) returned 'false', redirectUri is: {0}", (redirectUri ?? "null"))` message.

* Allowing developers to change the issuer address at runtime using notifications is not a bad thing, but it's definitely a "niche market". I'm totally convinced we should make `AuthorizationEndpoint` mandatory and fail very early if we can't find it in the static or dynamic configuration. You can always provide a dummy value and replace it at runtime using the `RedirectToIdentityProvider` notification. The same remark applies to every other OIDC parameter like `redirect_uri`, which is currently not required by the OIDC middleware, though it's declared as mandatory by the specifications.

> 3 / 4. The current way we capture the 'redirect_uri' is broken and needs to be changed. We need the exact 'redirect_uri' sent to the IdentityProvider for the ConfidentialClient 'code' -> 'access_token' flow. 

Being able to "dynamically" change the `redirect_uri` has always been a weird thing for me. I still don't understand why we can do that and why `Options.CallbackPath` is not a mandatory property (BTW, the path check is wrong: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs#L244). What would be the advantage of that?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: I suppose, making them immutable from notifications is one option, but the have to get set somewhere.  Now we are starting to talk about 'friends' and 'internals'. These sometimes make testing harder.

What I am saying is, not to trick developers into believing setting them to null is ok, cause it is not.

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
brentschmaltz: Fair enough. I have no plan to do that.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: What I am trying to get across about order is that devs should not count on it, tends to be fragile. In fact it could be the case that two handlers are in the collection that can process the same token.
Not sure how to think about that matrix yet, but it is on my mind.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: Done.

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
PinpointTownes: Done :smile:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: So, what's the final decision? :smile:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: > Taking offline.... we'll get back to you.

So? :octopus:

brentschmaltz:
Any changes we take here should wait till 315 is merged.

PinpointTownes:
@brentschmaltz don't worry, I won't break your changes :smile: 

This PR needs to be rebased anyway.

PinpointTownes:
Rebased.

I limited this PR to the code style issues we discussed in @brentschmaltz's PR (https://github.com/aspnet/Security/pull/112) and removed the controversial changes. This PR also replaces `Scope` by a list of items (we could consider using a set instead) and removes the validators list by a single instance, as suggested by @brentschmaltz.

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
```diff
@@ -54,15 +46,14 @@ public OAuthBearerAuthenticationOptions() : base()
...
-        public string Challenge { get; set; }
+        public string Challenge { get; [param: NotNull] set; } = OAuthBearerAuthenticationDefaults.AuthenticationScheme;
 
         /// <summary>
         /// The object provided by the application to process events raised by the bearer authentication middleware.
         /// The application may implement the interface fully, or it may create an instance of OAuthBearerAuthenticationProvider
         /// and assign delegates only to the events it wants to process.
         /// </summary>
-        public OAuthBearerAuthenticationNotifications Notifications { get; set; }
+        public OAuthBearerAuthenticationNotifications Notifications { get; [param: NotNull] set; } = new OAuthBearerAuthenticationNotifications();
```
Tratcher: FYI: we've been asked not to use NotNull in this context (more details coming soon). Let's revert back to setting these in the middleware constructor for now.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -240,103 +199,49 @@ public OpenIdConnectProtocolValidator ProtocolValidator
...
-            {
-                return _stringDataFormat;
-            }
-            [param: NotNull]
-            set
-            {
-                _stringDataFormat = value;
-            }
-        }
+        public ISecureDataFormat<string> StringDataFormat { get; set; }
```
Tratcher: Should this be NonceSecureFormat?

Tratcher:
:watch: Almost ready.

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
PinpointTownes: Do you want me to revert all the inline properties? This syntax makes the code much clearer IMHO.

Tratcher: You can keep those. We just don't want to put the non-null enforcement code here.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: Ideally, this thing shouldn't exist. We don't need a full data protector (that offers both encryption and hashing) for nonces:

> The nonce parameter value needs to include per-session state and be unguessable to attackers. One method to achieve this for Web Server Clients is to store a cryptographically random value as an HttpOnly session cookie and use a cryptographic hash of the value as the nonce parameter. In that case, the nonce in the returned ID Token is compared to the hash of the session cookie to detect ID Token replay by third parties.

http://openid.net/specs/openid-connect-core-1_0.html#NonceNotes

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationOptions.cs
PinpointTownes: Okay. `param: NotNull` is currently used in a few places. Do you want me to remove them?

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationOptions.cs#L41
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication/Notifications/RedirectToIdentityProviderNotification.cs#L26

Tratcher: Don't worry about the existing ones yet, just don't add more.

PinpointTownes: Reverted.

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationMiddleware.cs
```diff
@@ -38,16 +34,6 @@ public class OAuthBearerAuthenticationMiddleware : AuthenticationMiddleware<OAut
             ConfigureOptions<OAuthBearerAuthenticationOptions> configureOptions)
             : base(next, options, loggerFactory, encoder, configureOptions)
         {
-            if (Options.Notifications == null)
-            {
-                Options.Notifications = new OAuthBearerAuthenticationNotifications();
```
Tratcher: Can you revert this half too please?

PinpointTownes: Done. I also reverted the same change for OIDC and added a check for the OAuth2 generic middleware.

Issue #217, opened 2015-04-16
OpenID connect: support response_mode=query


PinpointTownes:
`response_mode=query` support has been a recurrent demand on JabbR, given that the optional `response_mode=form_post` is not implemented everywhere.

Even if authorization servers should make sure they don't return `token` or `id_token` when using `response_mode=query`, we should reject such requests in the middleware itself.

https://github.com/aspnet/Security/pull/202#discussion_r28379573

/cc @HaoK @brentschmaltz @Praburaj

PinpointTownes:
@brentschmaltz if you agree with the general idea, I'll submit a PR when yours is merged.

brentschmaltz:
It's not entirely my call, other folks involved have indicated that 'query_mode' should not be supported by default.

PinpointTownes:
You should maybe ping them so they can participate to this conversation :construction_worker: 

/cc @selfissued

brentschmaltz:
@PinpointTownes I don't think there is much support for query mode.

brentschmaltz:
@PinpointTownes did you ping MikeJones (selfissued)?

PinpointTownes:
@brentschmaltz I mentioned him in my last post, so he probably received a notification from GitHub.

I'll ping him via Twitter :smile:

PinpointTownes:
Tweet sent: https://twitter.com/PinpointTownes/status/601111713853890560

selfissued:
First, response_mode should only be used when a non-default response mode is being requested for the response_type, per http://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#ResponseModes and http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest.  Given that it's never safe to use the query response_mode when fragment encoding is the default, this means that it is only ever safe to use response_mode=query when response_type=code is used.  But in that case, it's unnecessary and NOT RECOMMENDED, because query is already the default for the code response_mode.  So I short, response_mode=query should never be used.

The query response mode is, of course, used by default for the most common response_type - code.

I would close this one as invalid, because it reflects a misunderstanding of the specifications.

PinpointTownes:
@selfissued by `response_mode=query`, I mean GET callbacks (the last being the result of the explicit response_mode or implied from the `response_type` used), that are currently NOT supported by the OIDC client middleware.

PinpointTownes:
> it's never safe to use the query response_mode when fragment encoding is the default

[...]

PR #279, opened 2015-05-28
Add response_mode=query support for OpenID Connect

PinpointTownes:
Fixes https://github.com/aspnet/Security/issues/217

/cc @brentschmaltz @HaoK @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
@@ -139,10 +139,18 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
                 // [brentschmaltz] - #215 this should be a property on RedirectToIdentityProviderNotification not on the OIDCMessage.
                 RequestType = OpenIdConnectRequestType.AuthenticationRequest,
                 Resource = Options.Resource,
-                ResponseMode = Options.ResponseMode,
                 ResponseType = Options.ResponseType,
                 Scope = Options.Scope
             };
+
+            // Omitting the response_mode parameter when it already corresponds to the default
+            // response_mode used for the specified response_type is recommended by the specifications.
+            // See http://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#ResponseModes
+            if (!string.Equals(Options.ResponseType, OpenIdConnectResponseTypes.Code, StringComparison.Ordinal) ||
+                !string.Equals(Options.ResponseMode, OpenIdConnectResponseModes.Query, StringComparison.Ordinal))
+            {
+                message.ResponseMode = Options.ResponseMode;
+            }
 
             if (Options.ProtocolValidator.RequireNonce)
             {
@@ -236,8 +244,22 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
 
             OpenIdConnectMessage message = null;
 
+            if (string.Equals(Request.Method, "GET", StringComparison.OrdinalIgnoreCase))
+            {
+                message = new OpenIdConnectMessage(Request.Query);
+
+                // response_mode=query (explicit or not) and a response_type containing id_token
+                // or token are not considered as a safe combination and MUST be rejected.
+                // See http://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#Security
+                if (!string.IsNullOrWhiteSpace(message.IdToken) || !string.IsNullOrWhiteSpace(message.Token))
+                {
+                    Logger.LogError("An OpenID Connect response cannot contain an identity token " +
+                                    "or an access token when using response_mode=query");
+                    return null;
+                }
+            }
             // assumption: if the ContentType is "application/x-www-form-urlencoded" it should be safe to read as it is small.
-            if (string.Equals(Request.Method, "POST", StringComparison.OrdinalIgnoreCase)
+            else if (string.Equals(Request.Method, "POST", StringComparison.OrdinalIgnoreCase)
               && !string.IsNullOrWhiteSpace(Request.ContentType)
               // May have media/type; charset=utf-8, allow partial match.
               && Request.ContentType.StartsWith("application/x-www-form-urlencoded", StringComparison.OrdinalIgnoreCase)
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -248,8 +256,24 @@ protected override AuthenticationTicket AuthenticateCore()
...
+            {
+                message = new OpenIdConnectMessage(Request.Query);
+
+                // response_mode=query (explicit or not) and a response_type containing id_token
+                // or token are not considered as a safe combination and MUST be rejected.
+                // See http://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#Security
+                if (!string.IsNullOrWhiteSpace(message.IdToken) ||
+                    !string.IsNullOrWhiteSpace(message.Token))
+                {
+                    Logger.LogError("An OpenID Connect response cannot contain an identity token " +
```
PinpointTownes: I'll remove the hardcoded message to the .resx file when we're done with the other changes.

brentschmaltz:
I good with this PR, ack.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -248,8 +256,24 @@ protected override AuthenticationTicket AuthenticateCore()
...
+                                    "or an access token when using response_mode=query");
+                    return null;
+                }
+            }
+
```
Tratcher: nit: no empty line before an else statement.

Tratcher:
Looks good. Could use some tests.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: Fixed, thanks.

PinpointTownes:
@Tratcher yup, I agree, but the new OpenID Connect tests are a bit cryptic for me, I'm not sure where I'm supposed to add these tests.

@brentschmaltz a suggestion? :smile:

Eilon:
@Tratcher please merge when ready.

PinpointTownes:
@Eilon right now, merging this PR is pointless as it depends on https://github.com/aspnet/Security/pull/258, which doesn't seem to be ready yet.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: Is moving logging messages to .resx still the recommended approach?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -177,12 +177,20 @@ protected override void ApplyResponseChallenge()
...
-                ResponseMode = Options.ResponseMode,
                 ResponseType = Options.ResponseType,
                 Scope = Options.Scope,
                 State = OpenIdConnectAuthenticationDefaults.AuthenticationPropertiesKey + "=" + UrlEncoder.UrlEncode(Options.StateDataFormat.Protect(properties))
             };
 
+            // Omitting the response_mode parameter when it already corresponds to the default
+            // response_mode used for the specified response_type is recommended by the specifications.
+            // See http://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#ResponseModes
+            if (!string.Equals(Options.ResponseType, OpenIdConnectResponseTypes.Code, StringComparison.Ordinal) ||
```
PinpointTownes: `OpenIdConnectResponseTypes.CodeOnly` has been renamed to `OpenIdConnectResponseTypes.Code`.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/156

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -248,8 +256,22 @@ protected override AuthenticationTicket AuthenticateCore()
...
+            if (string.Equals(Request.Method, "GET", StringComparison.OrdinalIgnoreCase))
+            {
+                message = new OpenIdConnectMessage(Request.Query);
+
+                // response_mode=query (explicit or not) and a response_type containing id_token
+                // or token are not considered as a safe combination and MUST be rejected.
+                // See http://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#Security
+                if (!string.IsNullOrWhiteSpace(message.IdToken) || !string.IsNullOrWhiteSpace(message.Token))
+                {
+                    Logger.LogError("An OpenID Connect response cannot contain an identity token " +
```
brentschmaltz: This needs to be in a resource file.

PinpointTownes: I asked @Eilon if adding logging messages to the .resx file was still the recommended and got no answer :smile: 

I'll fix that, thanks.

Eilon: I'm pretty sure we don't localize log messages. They are currently intended to be English only because there's no way to do localized logs. The problem with having localized log messages is which culture do you use for them? Surely you don't want to use the current thread's culture, because that is usually set to whatever the *end user's* culture is. So if a French user visits, I get French logs. And the next user is German, so the next log is German. Clearly that is not going to work. So in the meantime we decided to not localize them.

Tratcher:
@PinpointTownes I think all the PRs you need went in. Are you waiting on anything else? If not then please rebase.

PinpointTownes:
@Tratcher I had planned to move the warning message to the .resx file, but according to @Eilon, this doesn't seem to be the appropriate approach anymore. 

I'll rebase it when back home (you can cherry pick the unique commit if it's urgent)

PinpointTownes:
@Tratcher rebased. As expected, there was no conflict :smile:

PR #283, opened 2015-06-05
AuthenticationHandler now fully Async, SignOut/SignIn/Challenge API changes 

HaoK:
Depends on https://github.com/aspnet/HttpAbstractions/pull/323

## Overview of AuthenticationHandler changes:
- Response is mutated directly as part of SignIn/SignOut Challenge invocation (instead of on Teardown)
- Consume ChallengeBehavior: 
   Automatic means (403 if Authenticate produces an identity, 401 otherwise)
   Unauthorized/Forbidden explicitly call the appropriate code path
- new ChallengeCalled flag so ApplyResponseChallenge is only called once
- Cookie now overrides TeardownAsync to continue the old renew/refresh behavior

## New protected virtual methods
- HandleChallenge: default behavior for implementing the ChallengeBehavior functionality
- HandleUnauthorized: (turns around and just calls the existing ApplyResponseChallenge)
- HandleForbidden: sets 403 by default, Cookie overrides to add support for AccessDenied redirect page
- HandleSignIn/Out: lifted out of ApplyResponseGrant, now done immediately

## TODO:
- Its still a bit messy in AuthenticationHandler, need to iterate on that more
- Lou wants the AuthenticationManager.Challenge setting of 401 to move into the handler (that change currently doesn't completely work with the passive challenge flow intercepting 401s)

cc @lodejard @Tratcher @blowdart @divega @davidfowl @muratg

### DIFF
--- a/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -8,6 +8,7 @@ using System.Security.Claims;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Authentication;
+using Microsoft.AspNet.Http.Features.Authentication;
 using Microsoft.Framework.Logging;
 
 namespace Microsoft.AspNet.Authentication.Cookies
@@ -99,7 +100,6 @@ namespace Microsoft.AspNet.Authentication.Cookies
 
                 await Options.Notifications.ValidatePrincipal(context);
 
-                AuthenticateCalled = true;
                 return new AuthenticationTicket(context.Principal, context.Properties, Options.AuthenticationScheme);
             }
             catch (Exception exception)
@@ -315,49 +315,50 @@ namespace Microsoft.AspNet.Authentication.Cookies
             return path[0] == '/' && path[1] != '/' && path[1] != '\\';
         }
 
+        protected override void HandleForbidden(ChallengeContext context)
+        {
+            // HandleForbidden by redirecting to AccessDeniedPath if set
+            if (Options.AccessDeniedPath.HasValue)
+            {
+                try
+                {
+                    var accessDeniedUri =
+                        Request.Scheme +
+                        "://" +
+                        Request.Host +
+                        Request.PathBase +
+                        Options.AccessDeniedPath;
+
+                    var redirectContext = new CookieApplyRedirectContext(Context, Options, accessDeniedUri);
+                    Options.Notifications.ApplyRedirect(redirectContext);
+                }
+                catch (Exception exception)
+                {
+                    var exceptionContext = new CookieExceptionContext(Context, Options,
+                        CookieExceptionContext.ExceptionLocation.ApplyResponseChallenge, exception, ticket: null);
+                    Options.Notifications.Exception(exceptionContext);
+                    if (exceptionContext.Rethrow)
+                    {
+                        throw;
+                    }
+                }
+            }
+            else
+            {
+                base.HandleForbidden(context);
+            }
+        }
+
         protected override void ApplyResponseChallenge()
         {
-            if (ShouldConvertChallengeToForbidden())
-            {
-                // Handle 403 by redirecting to AccessDeniedPath if set
-                if (Options.AccessDeniedPath.HasValue)
-                {
-                    try
-                    {
-                        var accessDeniedUri =
-                            Request.Scheme +
-                            "://" +
-                            Request.Host +
-                            Request.PathBase +
-                            Options.AccessDeniedPath;
+            // REVIEW: should we check if status code is 401?
 
-                        var redirectContext = new CookieApplyRedirectContext(Context, Options, accessDeniedUri);
-                        Options.Notifications.ApplyRedirect(redirectContext);
-                    }
-                    catch (Exception exception)
-                    {
-                        var exceptionContext = new CookieExceptionContext(Context, Options,
-                            CookieExceptionContext.ExceptionLocation.ApplyResponseChallenge, exception, ticket: null);
-                        Options.Notifications.Exception(exceptionContext);
-                        if (exceptionContext.Rethrow)
-                        {
-                            throw;
-                        }
-                    }
-                }
-                else
-                {
-                    Response.StatusCode = 403;
-                }
-                return;
-            }
-
-            if (Response.StatusCode != 401 || !Options.LoginPath.HasValue )
+            if (!Options.LoginPath.HasValue)
             {
                 return;
             }
 
-            // Automatic middleware should redirect on 401 even if there wasn't an explicit challenge.
+            // Should redirect even if there wasn't an explicit challenge when Automatic.
             if (ChallengeContext == null && !Options.AutomaticAuthentication)
             {
                 return;
--- a/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -27,12 +27,7 @@ namespace Microsoft.AspNet.Authentication.Cookies
         private DateTimeOffset _renewExpiresUtc;
         private string _sessionKey;
 
-        protected override AuthenticationTicket AuthenticateCore()
-        {
-            return AuthenticateCoreAsync().GetAwaiter().GetResult();
-        }
-
-        protected override async Task<AuthenticationTicket> AuthenticateCoreAsync()
+        public override async Task<AuthenticationTicket> AuthenticateAsync()
         {
             AuthenticationTicket ticket = null;
             try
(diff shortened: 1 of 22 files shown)


HaoK:
Updated cookie tests to have the correct set of Unauthorized/401/403/accessdenied/login tests now

PinpointTownes:
FWIW, I'm not sure you should add a new `ApplyResponseStartingAsync` method called from an OSH callback (specially if we don't make it async): it can be really, really dangerous and create awful deadlocks when trying to write to the response stream.

That's the unique reason why we had to use `TeardownCoreAsync` in ASOS: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/blob/vNext/src/AspNet.Security.OpenIdConnect.Server/OpenIdConnectServerHandler.cs#L365-L377

HaoK:
AuthHandlers are fully async now and more of a sandbox with only a few protected virtual methods.

HandleUnauthorizedAsync
HandleSignInAsync
HandleSignOutAsync
FinishResponseAsync (cookie needs this to renew the cookie if none of the other operations were done)

AuthenticateAsync remains pure abstract

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -90,48 +77,47 @@ internal AuthenticationOptions BaseOptions
             }
         }
 
-        private static void OnSendingHeaderCallback(object state)
+        private static async Task OnResponseStartingCallback(object state)
```
HaoK: @Tratcher Is there a reason this is a static passing this in, instead of just making this an instance method with null state (soon to be using the stateless sugar method)?

PinpointTownes: I guess that's to avoid a closure when calling `Response.OnResponseStarting`.

Tratcher: Exactly.

HaoK: Okay so I'll add the stateless extension but we won't use it in the AuthHandler

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -59,9 +51,6 @@ internal AuthenticationOptions BaseOptions
             get { return _baseOptions; }
         }
 
-        // REVIEW: Overriding Authenticate and not calling base requires manually calling this for 401-403 to work
-        protected bool AuthenticateCalled { get; set; }
-
         public IAuthenticationHandler PriorHandler { get; set; }
 
         public bool Faulted { get; set; }
```
Tratcher: Do you need to remove `Faulted`? It looks like you've removed it's usage.

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -223,171 +172,62 @@ public AuthenticationTicket Authenticate()
...
         }
 
-        public virtual void SignIn(SignInContext context)
+        public async Task SignInAsync(SignInContext context)
         {
             if (ShouldHandleScheme(context.AuthenticationScheme))
             {
-                SignInContext = context;
-                SignOutContext = null;
+                SignInCalled = true;
```
Tratcher: Should this be SignInAccepted?

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -395,11 +235,48 @@ public virtual bool ShouldConvertChallengeToForbidden()
...
-            ApplyResponseChallenge();
-            return Task.FromResult(0);
+            Response.StatusCode = 401;
+            return Task.FromResult(false);
+        }
+
+        public async Task ChallengeAsync(ChallengeContext context)
+        {
+            bool handled = false;
+            ChallengeCalled = true;
```
Tratcher: Should ChallengeCalled be inside the ShouldHandleScheme block like SignIn and SignOut?

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -395,11 +235,48 @@ public virtual bool ShouldConvertChallengeToForbidden()
...
+                        }
+                        break;
+                    case ChallengeBehavior.Unauthorized:
+                        handled = await HandleUnauthorizedAsync(context);
+                        break;
+                    case ChallengeBehavior.Forbidden:
+                        handled = await HandleForbiddenAsync(context);
+                        break;
+                }
+                context.Accept();
```
Tratcher: Does it make sense to Accept if !handled?

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
blowdart: Good, faulted smacks horribly of WCF and no-one ever checked it there :)

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -395,11 +235,48 @@ public virtual bool ShouldConvertChallengeToForbidden()
...
         /// </summary>
-        /// <returns></returns>
-        protected virtual Task ApplyResponseChallengeAsync()
+        /// <param name="context"></param>
+        /// <returns>True if no other handlers should be called</returns>
+        protected virtual Task<bool> HandleUnauthorizedAsync(ChallengeContext context)
         {
-            ApplyResponseChallenge();
-            return Task.FromResult(0);
+            Response.StatusCode = 401;
```
Tratcher: Don't set the status code if you're not going to handle it. Same for Forbidden.

HaoK: Should we just flip these to true to block the chaining?  We would need to set 401 explicitly in all of the children if we don't do it here.  This is how we get the default 401/403 behavior if you don't override these methods

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
HaoK: This one is slightly different, since we are using it as a guard around HandleAutomaticChallengeIfNeeded, I will rename the other two to Accepted and leave this one called

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
HaoK: I think so, they are two different things.  Handled is really don't call anyone else.  If you don't accept, we blow up :/

HaoK: The naming is bad I see though... ShouldHandle perhaps should be ShouldAccept?  Or the flag handled should be renamed...ugh naming is hard

HaoK: Accept and Handle sound wayyyyy too similar...

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -315,60 +382,49 @@ private static bool IsHostRelative(string path)
...
+            else
             {
-                return;
-            }
-
-            // Automatic middleware should redirect on 401 even if there wasn't an explicit challenge.
-            if (ChallengeContext == null && !Options.AutomaticAuthentication)
-            {
-                return;
+                return base.HandleForbiddenAsync(context);
```
HaoK: I think its correct that the default base implementation doesn't block chaining, since we want to just continue in this case.

HaoK: And if base.handleForbidden just set 403 and returned true, that wouldn't let any other handlers execute afterwards, which isn't really what we are trying to enforce here is it?

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
Tratcher: What does Forbid(Facebook) even mean? I'd be tempted to throw NotSupportedExcetpion since this is now applied directly from user code.

HaoK: Well today it would be you get a 403 as the status code?  I'm ok with overidding the default behavior to just throw NotSupported/InvalidOperation instead on the OAuth providers if we think that's better?

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -90,48 +75,47 @@ internal AuthenticationOptions BaseOptions
...
 
-        protected virtual Task InitializeCoreAsync()
+        private async Task FinishResponseOnce()
         {
-            return Task.FromResult(0);
+            if (!_finishCalled)
+            {
+                _finishCalled = true;
+                await FinishResponseAsync();
+                await HandleAutomaticChallengeIfNeeded();
```
Tratcher: Ordering: I would expect Challenge to be applied before Finish.

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -90,48 +75,47 @@ internal AuthenticationOptions BaseOptions
             }
         }
 
-        private static void OnSendingHeaderCallback(object state)
+        private async Task OnResponseStartingCallback()
```
Tratcher: This was supposed to stay static so we only allocate one delegate ever, not one per request.

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -76,9 +63,7 @@ internal AuthenticationOptions BaseOptions
 
             RegisterAuthenticationHandler();
 
-            Response.OnResponseStarting(OnSendingHeaderCallback, this);
-
-            await InitializeCoreAsync();
+            Response.OnResponseStarting(OnResponseStartingCallback);
```
Tratcher: Needs to be updated to OnStarting.

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -90,48 +75,47 @@ internal AuthenticationOptions BaseOptions
...
-        protected virtual Task TeardownCoreAsync()
+        /// <summary>
+        /// Called once per request after Initialize and Invoke.
+        /// </summary>
+        /// <returns>async completion</returns>
+        internal async Task TeardownAsync()
         {
-            return Task.FromResult(0);
+            await FinishResponseOnce();
+            UnregisterAuthenticationHandler();
```
Tratcher: try Finish finally Unregister

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -90,48 +75,47 @@ internal AuthenticationOptions BaseOptions
...
+                await HandleUnauthorizedAsync(new ChallengeContext(BaseOptions.AuthenticationScheme));
             }
-
-            await TeardownCoreAsync();
-            UnregisterAuthenticationHandler();
         }
 
-        protected virtual Task TeardownCoreAsync()
+        /// <summary>
+        /// Called once per request after Initialize and Invoke.
```
Tratcher: Is this comment current? It sounds like it got copied from somewhere else. This doesn't get called until unwind.

src/Microsoft.AspNet.Authentication.OAuthBearer/OAuthBearerAuthenticationHandler.cs
```diff
@@ -179,30 +175,11 @@ protected override AuthenticationTicket AuthenticateCore()
...
-            }
-
+            Response.StatusCode = 401;
             await Options.Notifications.ApplyChallenge(new AuthenticationChallengeNotification<OAuthBearerAuthenticationOptions>(Context, Options));
-        }
-
-        protected override void ApplyResponseGrant()
-        {
-            // N/A
+            return true;
```
Tratcher: false, Bearer should chain.

HaoK:
Updated with last PR fixes

1ae4c24a5fd43108b691272ebc6b0e9878ee49a7
3a8ea672eae590785a78d74777abb76bb9664130

PR #286, opened 2015-06-09
set correct expires of cookie when IsPersistent is true.

RickyLin:
The `Microsoft.AspNet.Http.CookieOptions.Expires` is of type `DateTime?`,  the `Microsoft.Net.Http.Headers.SetCookieHeaderValue.Expires` is of type `DateTimeOffset?`. Looking into the following source code:
```c#
// CookieAuthenticationHandler.cs, line 184
signInContext.CookieOptions.Expires = expiresUtc.ToUniversalTime().DateTime;

// ChunkingCookieManager.cs, line 129
var template = new SetCookieHeaderValue(escapedKey)
{
    Domain = options.Domain,

    // The result of conversion is unexpected because options.Expires has a UTC value 
    // but its Kind is DateTimeKind.Unspecified
    Expires = options.Expires, 

    HttpOnly = options.HttpOnly,
    Path = options.Path,
    Secure = options.Secure,
};
```

To fix it, using `expiresUtc.UtcDateTime` instead of `expiresUtc.ToUniversalTime().DateTime`.

I also added a test for this.

### DIFF
--- a/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -161,7 +161,7 @@ namespace Microsoft.AspNet.Authentication.Cookies
             var cookieOptions = BuildCookieOptions();
             if (ticket.Properties.IsPersistent && _renewExpiresUtc.HasValue)
             {
-                cookieOptions.Expires = _renewExpiresUtc.Value.ToUniversalTime().DateTime;
+                cookieOptions.Expires = _renewExpiresUtc.Value.UtcDateTime;
             }
 
             Options.CookieManager.AppendResponseCookie(
@@ -236,7 +236,7 @@ namespace Microsoft.AspNet.Authentication.Cookies
                 if (signInContext.Properties.IsPersistent)
                 {
                     var expiresUtc = signInContext.Properties.ExpiresUtc ?? issuedUtc.Add(Options.ExpireTimeSpan);
-                    signInContext.CookieOptions.Expires = expiresUtc.ToUniversalTime().DateTime;
+                    signInContext.CookieOptions.Expires = expiresUtc.UtcDateTime;
                 }
 
                 model = new AuthenticationTicket(signInContext.Principal, signInContext.Properties, signInContext.AuthenticationScheme);
--- a/test/Microsoft.AspNet.Authentication.Test/Cookies/CookieMiddlewareTests.cs
+++ b/test/Microsoft.AspNet.Authentication.Test/Cookies/CookieMiddlewareTests.cs
@@ -539,6 +539,27 @@ namespace Microsoft.AspNet.Authentication.Cookies
         }
 
         [Fact]
+        public async Task CookieIsPersistent()
+        {
+            var expected = " expires=Tue, 11 Jun 2013 12:40:56 GMT";
+            var clock = new TestClock();
+            var server = CreateServer(options =>
+            {
+                options.SystemClock = clock;
+            },
+            context =>
+                context.Authentication.SignInAsync("Cookies",
+                    new ClaimsPrincipal(new ClaimsIdentity(new GenericIdentity("Alice", "Cookies"))),
+                    new AuthenticationProperties() { ExpiresUtc = clock.UtcNow.AddMinutes(6), IsPersistent = true }));
+
+            var transaction1 = await SendAsync(server, "http://example.com/testpath");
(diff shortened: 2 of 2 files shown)


RickyLin:
I have already signed a CLA when I sent PR to MVC repo, Is it expired?

Tratcher:
We recently switched from MS OpenTech CLAs to .NET Foundation CLAs, so you may need to re-sign. However, it should be easier this time.

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -181,7 +181,7 @@ protected override void ApplyResponseGrant()
                     if (signInContext.Properties.IsPersistent)
                     {
                         var expiresUtc = signInContext.Properties.ExpiresUtc ?? issuedUtc.Add(Options.ExpireTimeSpan);
-                        signInContext.CookieOptions.Expires = expiresUtc.ToUniversalTime().DateTime;
+                        signInContext.CookieOptions.Expires = expiresUtc.LocalDateTime;
```
brentschmaltz: I am confused by this. Why the switch to LocalTime?

RickyLin: Because the `Kind` property of `expiresUtc.ToUniversalTime().DateTime` is `DateTimeKind.Unspecified` rather than `DateTimeKind.Local`. So there is a problem in this code:
```c#
// ChunkingCookieManager.cs, line 129
var template = new SetCookieHeaderValue(escapedKey)
{
    Domain = options.Domain,

    // convert DateTime to DateTimeOffset implicitly here
    // because options.Expires.Kind == DateTimeKind.Unspecified,
    // options.Expires is treated as a local DateTime during the conversion while it's already a UTC DateTime
    Expires = options.Expires,

    HttpOnly = options.HttpOnly,
    Path = options.Path,
    Secure = options.Secure,
};
```

RickyLin:
@brentschmaltz, If

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -236,7 +236,7 @@ private CookieOptions BuildCookieOptions()
                 if (signInContext.Properties.IsPersistent)
                 {
                     var expiresUtc = signInContext.Properties.ExpiresUtc ?? issuedUtc.Add(Options.ExpireTimeSpan);
-                    signInContext.CookieOptions.Expires = expiresUtc.ToUniversalTime().DateTime;
+                    signInContext.CookieOptions.Expires = expiresUtc.LocalDateTime;
```
RickyLin: @brentschmaltz regarding your confusion,
The `Kind` value of `expiresUtc.ToUniversalTime().DateTime` is `DateTimeKind.Unspecified` rather than `DateTimeKind.Utc` (Surprise!), this results in a bug in the following code:

```c#
// ChunkingCookieManager.cs, line 129
var template = new SetCookieHeaderValue(escapedKey)
{
    Domain = options.Domain,

    // there is an implicit conversion (from DateTime to DateTimeOffset) here.
    // options.Expires.Kind == DateTimeKind.Unspecified
    // hence options.Expires is treated as a local DateTime during the conversion while it's a UTC DateTime.
    Expires = options.Expires,

    HttpOnly = options.HttpOnly,
    Path = options.Path,
    Secure = options.Secure,
};
```

brentschmaltz: @RickyLin I am trying to understand why we the need to use LocalTime.

RickyLin: @brentschmaltz  I think using LocalDateTime is the simplest workaround to make the conversion work as expected. Maybe DateTimeOffset class should have a `UniversalDateTime` property which set its `Kind` value to `DateTimeKind.Utc`.

muratg:
@HaoK @Tratcher Could you have a look at this please?

HaoK:
Looks good to me

HaoK:
Actually I think we should change Expires to DateTimeOffset instead of DateTime? like https://github.com/aspnet/HttpAbstractions/issues/386 suggests

muratg:
@HaoK would you like to take it in? 
cc @Tratcher

HaoK:
@Tratcher and I both think we should take the fix in HttpAbstractions instead, we are going to go with a lower level fix instead of taking this PR, thanks for the submission!

PR #289, opened 2015-06-15
Fix Forbidden Challenge for AutomaticAuthentication

andresmoschini:
`AuthenticateCalled` is not marked as true when `AutomaticAuthentication` is used, then `ShouldConvertChallengeToForbidden` was returning `false` and the resulting StatusCode was 401 in place of 403.


HaoK:
Yep good catch, we're actually reworking (removing all of this, so we likely won't take this fix, but its totally correct)

HaoK:
PR https://github.com/aspnet/Security/pull/283 is the wrecking ball that's approaching

andresmoschini:
Thanks for the information. I think that it is a good idea to rework on security middlewares.

Issue #212, opened 2015-04-08
Replace INonceCache by IDistributedCache


PinpointTownes:
https://github.com/aspnet/Security/pull/202#discussion_r27964163

Eilon:
@PinpointTownes do you want to take a stab at this?

PinpointTownes:
@Eilon yup, feel free to assign me this issue :smile:

Eilon:
Can't actually assign to you because you have to be in our GitHub org, but we know it's you :smile:

PinpointTownes:
> Can't actually assign to you because you have to be in our GitHub org, but we know it's you :smile:

Actually, that's a strong sign I should be in this organization :smile:

PinpointTownes:
I got two questions concerning the general design we should adopt for this PR:
* Should the nonce cache be used by default? Currently, there's absolutely no `INonceCache` implementation offered by the OIDC package, which means that it's turned off by default. Should we change that?

* Where should the distributed cache be retrieved from?
  * Directly from the DI container: in this case, it will necessarily use the distributed cache defined at the app-level.
  * From the OIDC options: with this approach, it won't be retrieved from the DI container and will be isolated from the rest of the application.
  * From the OIDC options or from the DI container: that's the hybrid scenario, where the distributed cache is first retrieved from the options when available, and from the DI container when the user didn't provide its own cache.

FWIW, I opted for the second approach in `AspNet.Security.OpenIdConnect.Server`, but I'll probably replace it by the hybrid solution: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/blob/vNext/src/AspNet.Security.OpenIdConnect.Server/OpenIdConnectServerOptions.cs#L195

/cc @Eilon @HaoK @Tratcher

Tratcher:
This needs to be under developer control. I don't think the presence or absence of an IDistributedCache in DI is sufficient to turn this on or off since the cache may be used for other components as well. Maybe just a bool that toggles nonce cache vs nonce cookies? I would still default to cookies, as cache requires dependencies and setup.

Check how we use IDistributedCache in Session. It may not be quite the same since technically it's optional here. The hybrid approach sounds the most accessible.

PinpointTownes:
> Maybe just a bool that toggles nonce cache vs nonce cookies?

Yup, that's what I had in mind.

> I would still default to cookies, as cache requires dependencies and setup.

[...]

PR #293, opened 2015-06-22
Replace INonceCache by IDistributedCache

PinpointTownes:
Fixes https://github.com/aspnet/Security/issues/212.

For the first attempt, I opted for the "hybrid" approach (https://github.com/aspnet/Security/issues/212#issuecomment-112533342): the distributed cache is retrieved from the options when available, and from the DI container when the user didn't provide his own cache.

Storing nonces in the distributed cache is not enabled by default but can easily be using `StoreNoncesInCache`. I'm not super satisfied with the name, so don't hesitate to suggest another one :smile: 

FYI, I haven't been able to use the `OpenIdConnectSample` app because it delegates authentication to AAD, that doesn't allow me to log in using my live.com account. I worked around this issue using `AspNet.Security.OpenIdConnect.Server` (https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server).

Of course, I'll revert these changes before merging (though depending on AAD is not really a good idea IMHO).

image

/cc @Eilon @HaoK @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
@@ -12,6 +12,7 @@ using Microsoft.AspNet.Authentication.Notifications;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Authentication;
 using Microsoft.AspNet.Http.Features.Authentication;
+using Microsoft.Framework.Caching.Distributed;
 using Microsoft.Framework.Internal;
 using Microsoft.Framework.Logging;
 using Microsoft.IdentityModel.Protocols;
@@ -149,13 +150,18 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
             if (Options.ProtocolValidator.RequireNonce)
             {
                 message.Nonce = Options.ProtocolValidator.GenerateNonce();
-                if (Options.NonceCache != null)
+                if (Options.CacheNonces)
                 {
-                    if (!Options.NonceCache.TryAddNonce(message.Nonce))
+                    if (await Options.NonceCache.GetAsync(message.Nonce) != null)
                     {
-                        Logger.LogError(Resources.OIDCH_0033_TryAddNonceFailed, message.Nonce);
-                        throw new OpenIdConnectProtocolException(string.Format(CultureInfo.InvariantCulture, Resources.OIDCH_0033_TryAddNonceFailed, message.Nonce));
+                        Logger.LogError(Resources.OIDCH_0033_NonceAlreadyExists, message.Nonce);
+                        throw new OpenIdConnectProtocolException(string.Format(CultureInfo.InvariantCulture, Resources.OIDCH_0033_NonceAlreadyExists, message.Nonce));
                     }
+
+                    await Options.NonceCache.SetAsync(message.Nonce, new byte[0], new DistributedCacheEntryOptions
+                    {
+                        AbsoluteExpirationRelativeToNow = Options.ProtocolValidator.NonceLifetime
+                    });
                 }
                 else
                 {
@@ -389,11 +395,16 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
                     }
 
                     string nonce = jwt.Payload.Nonce;
-                    if (Options.NonceCache != null)
+                    if (Options.CacheNonces)
                     {
-                        // if the nonce cannot be removed, it was used
-                        if (!Options.NonceCache.TryRemoveNonce(nonce))
+                        if (await Options.NonceCache.GetAsync(nonce) != null)
                         {
+                            await Options.NonceCache.RemoveAsync(nonce);
+                        }
+                        else
+                        {
+                            // If the nonce cannot be removed, it was
+                            // already used and MUST be rejected.
                             nonce = null;
                         }
                     }
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
@@ -13,11 +13,13 @@ using Microsoft.AspNet.Authentication.DataHandler.Serializer;
 using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.DataProtection;
 using Microsoft.AspNet.Http;
+using Microsoft.Framework.Caching.Distributed;
+using Microsoft.Framework.DependencyInjection;
+using Microsoft.Framework.Internal;
 using Microsoft.Framework.Logging;
 using Microsoft.Framework.OptionsModel;
-using Microsoft.IdentityModel.Protocols;
-using Microsoft.Framework.Internal;
 using Microsoft.Framework.WebEncoders;
+using Microsoft.IdentityModel.Protocols;
 
 namespace Microsoft.AspNet.Authentication.OpenIdConnect
 {
@@ -42,6 +44,7 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
             [NotNull] IDataProtectionProvider dataProtectionProvider,
             [NotNull] ILoggerFactory loggerFactory,
             [NotNull] IUrlEncoder encoder,
+            [NotNull] IServiceProvider services,
             [NotNull] IOptions<ExternalAuthenticationOptions> externalOptions,
             [NotNull] IOptions<OpenIdConnectAuthenticationOptions> options,
             ConfigureOptions<OpenIdConnectAuthenticationOptions> configureOptions = null)
@@ -124,6 +127,13 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
                     httpClient.MaxResponseContentBufferSize = 1024 * 1024 * 10; // 10 MB
                     Options.ConfigurationManager = new ConfigurationManager<OpenIdConnectConfiguration>(Options.MetadataAddress, httpClient);
                 }
+            }
+
+            if (Options.CacheNonces && Options.NonceCache == null)
+            {
+                // Use the global distributed cache if the user has not provided his own instance.
+                // Note: GetRequiredService will throw an exception if caching services have not been registered.
+                Options.NonceCache = services.GetRequiredService<IDistributedCache>();
             }
         }
 
--- a/src/Microsoft.AspNet.Authentication/AuthenticationServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNet.Authentication/AuthenticationServiceCollectionExtensions.cs
@@ -32,6 +32,5 @@ namespace Microsoft.Framework.DependencyInjection
         {
             return services.Configure<ClaimsTransformationOptions>(o => o.Transformer = new ClaimsTransformer { TransformAsyncDelegate = asyncTransform });
         }
-
     }
 }
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
@@ -8,6 +8,7 @@ using System.IdentityModel.Tokens;
 using System.Net.Http;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Authentication;
+using Microsoft.Framework.Caching.Distributed;
 using Microsoft.Framework.Internal;
 using Microsoft.IdentityModel.Protocols;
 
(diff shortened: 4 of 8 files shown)


src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -186,13 +187,16 @@ protected override void ApplyResponseChallenge()
...
                 {
-                    if (!Options.NonceCache.TryAddNonce(message.Nonce))
+                    if (await Options.NonceCache.GetAsync(message.Nonce) != null)
                     {
                         Logger.LogError(Resources.OIDCH_0033_TryAddNonceFailed, message.Nonce);
                         throw new OpenIdConnectProtocolException(string.Format(CultureInfo.InvariantCulture, Resources.OIDCH_0033_TryAddNonceFailed, message.Nonce));
                     }
+
+                    // Review: do we need to define an expiration date?
+                    await Options.NonceCache.SetAsync(message.Nonce, new byte[0], new DistributedCacheEntryOptions());
```
Tratcher: Yes, specify an expiration. The Nonce's are only good for an hour(?).

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
```diff
@@ -42,6 +43,7 @@ public class OpenIdConnectAuthenticationMiddleware : AuthenticationMiddleware<Op
             [NotNull] IDataProtectionProvider dataProtectionProvider,
             [NotNull] ILoggerFactory loggerFactory,
             [NotNull] IUrlEncoder encoder,
+            [NotNull] IDistributedCache cache,
```
Tratcher: The trouble with adding IDistributedCache as a constructor parameter is that it makes it required. What if you took in an IServiceProvider instead and then tried to do an optional lookup of IDistributedCache?

src/Microsoft.AspNet.Authentication/AuthenticationServiceCollectionExtensions.cs
```diff
@@ -12,6 +12,7 @@ public static class AuthenticationServiceCollectionExtensions
     {
         public static IServiceCollection AddAuthentication([NotNull] this IServiceCollection services)
         {
+            services.AddCaching();
```
Tratcher: No, we don't want to add caching services and dependencies for all auth providers, just OIDC.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
HaoK: I'm not positive but I think you can just do make it optional by adding cache = null.  But I'm not sure this works on all DI implementations consistently...

HaoK:
In regards to name, would CacheNonces be correct?

src/Microsoft.AspNet.Authentication/AuthenticationServiceCollectionExtensions.cs
HaoK: Yeah this is tricky then, if it requires services, as it's going to require its own AddOIDC extension then?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
PinpointTownes: @Tratcher I used this approach for `HtmlEncoder` in `OpenIdConnectServerMiddleware`, but TBH, I was not totally satisfied... it rather sounded like an anti-pattern: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/blob/vNext/src/AspNet.Security.OpenIdConnect.Server/OpenIdConnectServerMiddleware.cs#L65

src/Microsoft.AspNet.Authentication/AuthenticationServiceCollectionExtensions.cs
PinpointTownes: Okay, I'll revert this change but...

Where are we supposed to register caching services?
* In a new `services.AddOpenIdConnectAuthentication()` extension that you would have to call from your Startup.cs (would be rather weird IMO).
* Directly in Startup.cs: the developer would just have to manually call `services.AddCaching()`. :smile:

Tratcher: If it's optional then you don't need an extension, caching services would be added when the dev configured caching.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: Is it worth making it configurable?
If yes, I guess we would have to modify cookies' lifetime to match this new parameter.

src/Microsoft.AspNet.Authentication/AuthenticationServiceCollectionExtensions.cs
PinpointTownes: This means no `ÌDistributedCache` imported from the constructor :smile: 

If we can't resolve it from the service provider, can we create our own `LocalCache` and use it for `NonceCache`?

Tratcher: No, never create your own. They need to set the service or the option.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
Tratcher: Use the existing config.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/src/Microsoft.IdentityModel.Protocol.Extensions/OpenIdConnectProtocolValidator.cs#L108
Options.ProtocolValidator.NonceLifetime

Tratcher: The cookie doesn't set a timeout because we want it to expire as soon as you close the tab. Setting a timeout may extend it beyond the tab lifetime.

PinpointTownes: Well, IIRC, most browsers actually clear sessions cookies when their process is shut down, not just when the tab is closed. And it's not rare to see people who never close their browser: in this case, session cookies can have a "virtually unlimited lifetime" :smile: 

Anyway, that's now fixed.

src/Microsoft.AspNet.Authentication/AuthenticationServiceCollectionExtensions.cs
PinpointTownes: It really sounds like a PITA for almost no benefits, as the memory caching package that offers the `services.AddCaching()` extension doesn't depend on packages that are not referenced by the other security packages.

But meh, fixed anyway.

PinpointTownes:
@HaoK name updated, thanks :smile:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -186,13 +187,18 @@ protected override void ApplyResponseChallenge()
             if (Options.ProtocolValidator.RequireNonce)
             {
                 message.Nonce = Options.ProtocolValidator.GenerateNonce();
-                if (Options.NonceCache != null)
+                if (Options.CacheNonces)
                 {
-                    if (!Options.NonceCache.TryAddNonce(message.Nonce))
+                    if (await Options.NonceCache.GetAsync(message.Nonce) != null)
                     {
                         Logger.LogError(Resources.OIDCH_0033_TryAddNonceFailed, message.Nonce);
```
Tratcher: This resource name is out of date. It was probably too specific.

PinpointTownes: Updating it is on my todo list, but it's a bit painful as I have to edit the .resx file and the designer class manually. I'll do that when the rest is done :smile:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
```diff
@@ -125,6 +128,20 @@ public class OpenIdConnectAuthenticationMiddleware : AuthenticationMiddleware<Op
                     Options.ConfigurationManager = new ConfigurationManager<OpenIdConnectConfiguration>(Options.MetadataAddress, httpClient);
                 }
             }
+
```
Tratcher: Silent failure/fallback for misconfiguration is hard to debug. I'd expect something more like:
```
if  (Options.CacheNonces && Options.NonceCache == null)
{
  Options.NonceCache = services.GetRequiredService<IDistributedCache>(); // throws
}
```
Or maybe GetServices and check for null to throw your own error.

src/Microsoft.AspNet.Authentication/project.json
```diff
@@ -5,6 +5,7 @@
         "Microsoft.AspNet.DataProtection": "1.0.0-*",
         "Microsoft.AspNet.Http": "1.0.0-*",
         "Microsoft.AspNet.Http.Extensions": "1.0.0-*",
+        "Microsoft.Framework.Caching.Abstractions": "1.0.0-*",
```
Tratcher: Revert this

PinpointTownes: I will. But honestly, depending on the caching packages was not really a big deal :smile:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
Tratcher: You shouldn't have to update the designer manually. `build resx` should update the designer.

PinpointTownes: Except that there are currently many designers that are not located in the `Properties` directory: `build resx` doesn't detect them and create new designers under `Properties`, which creates duplicate files. I could remove the existing designers, but that's not something I want to do as part of this PR.

Tratcher: Ah, file a separate bug for that please.

HaoK: Yeah build resx is busted for security repo

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
PinpointTownes: Fixed using the first option :smile:

Tratcher:
Looks good. Remove aspnet-contrib and squash.

PinpointTownes:
@Tratcher is there a public identity provider supporting OIDC that we could use to replace AAD?
Because not being able to run the samples is a bit annoying, specially since I'd like to start working on the POST authorization requests thing, that will require much testing.

PinpointTownes:
.resx item renamed. And squashed :smile:

Tratcher:
None that I know of. That has been one of our big problems verifying the middleware at all.

NuGet.Config
```diff
@@ -4,5 +4,6 @@
     <add key="AspNetVNext" value="https://www.myget.org/F/aspnetvnext/api/v2" />
     <add key="NuGet" value="https://nuget.org/api/v2/" />
     <add key="AzureAD" value="http://www.myget.org/F/azureadwebstacknightly"/>
+    <add key="aspnet-contrib" value="https://www.myget.org/F/aspnet-contrib"/>
```
Tratcher: I thought you were removing this?

PinpointTownes: I'll remove everything related to the embedded OIDC server once we're done with all the changes... testing an OIDC middleware without a server is a bit hard :tada:

Tratcher: Push it into a separate branch, I don't want to merge this into dev.

PinpointTownes: Meh, removing it from this PR will be easier. Just tell me we're done with this PR, and I'll revert these changes.

Tratcher: That's what I meant. I think we're done and ready to merge.

PinpointTownes: Ah, that was not clear :smile: 

Changes reverted.

Tratcher: @haok can you merge this? I'm going offline.
________________________________

HaoK: Monday. I don't do Friday afternoon merges without good reason :)

PR #301, opened 2015-06-24
State and tests

brentschmaltz:
Replacement https://github.com/aspnet/Security/pull/233
Changed logic for user-state
Added additional testing


src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -291,36 +303,39 @@ protected override AuthenticationTicket AuthenticateCore()
...
+                    properties = SetUserStateOnMessage(message);
+                    if (properties == null)
+                    {
+                        Logger.LogError(Resources.OIDCH_0005_MessageStateIsInvalid);
+                        return null;
+                    }
                 }
 
-                // devs will need to hook AuthenticationFailedNotification to avoid having 'raw' runtime errors displayed to users.
+                // if any of the error fields are set, return null
```
brentschmaltz: Fix this comment.

Tratcher:
Close your old PR if you're done with it.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -23,6 +23,9 @@ public class OpenIdConnectAuthenticationHandler : AuthenticationHandler<OpenIdCo
     {
         private const string NonceProperty = "N";
         private const string UriSchemeDelimiter = "://";
+        // should we consider adding a first class property on AuthenticationProperties similar to RedirectUri.
+        // if we don't, then we may want to think about where to make this value public.
+        private const string UserState = "userstate";
```
Tratcher: You could put it in OpenIdConnectAuthenticationDefaults.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -69,21 +72,20 @@ protected override void ApplyResponseGrant()
...
+                    message.PostLogoutRedirectUri = properties.RedirectUri;
                 }
                 else if (!string.IsNullOrWhiteSpace(Options.PostLogoutRedirectUri))
                 {
-                    openIdConnectMessage.PostLogoutRedirectUri = Options.PostLogoutRedirectUri;
+                    message.PostLogoutRedirectUri = Options.PostLogoutRedirectUri;
                 }
 
-                var notification = new RedirectToIdentityProviderNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions>(Context, Options)
+                var notification = new RedirectToIdentityProviderNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions>(Context, Options, message);
```
Tratcher: Will the compiler let you omit the generics now that you're passing everything in as a direct argument?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -69,21 +72,20 @@ protected override void ApplyResponseGrant()
...
+                if (Options.Notifications.RedirectToIdentityProvider != null)
```
Tratcher: Inconsistent null checking. You check it here but not in ApplyResponseChallenge. Didn't we decide to just use NotNull for these on the Notifications class so we wouldn't have to check it everywhere?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -174,18 +176,36 @@ protected override void ApplyResponseChallenge()
...
                 message.Nonce = Options.ProtocolValidator.GenerateNonce();
+            }
+
+            var redirectToIdentityProviderNotification =
+                new RedirectToIdentityProviderNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions>(Context, Options, message);
+
+            await Options.Notifications.RedirectToIdentityProvider(redirectToIdentityProviderNotification);
+            if (redirectToIdentityProviderNotification.HandledResponse)
+            {
+                Logger.LogInformation(Resources.OIDCH_0034_RedirectToIdentityProviderNotificationHandledResponse);
```
Tratcher: LogVerbose?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -174,18 +176,36 @@ protected override void ApplyResponseChallenge()
...
+                return;
+            }
+            else if (redirectToIdentityProviderNotification.Skipped)
+            {
+                Logger.LogInformation(Resources.OIDCH_0035_RedirectToIdentityProviderNotificationSkipped);
+                return;
+            }
+
+            if (!string.IsNullOrWhiteSpace(message.Nonce))
```
Tratcher: Use IsNullOrEmpty everywhere, not IsNullOrWhiteSpace. @Eilon

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -586,34 +605,19 @@ private string ReadNonceCookie(string nonce)
             return null;
         }
 
-        private AuthenticationProperties GetPropertiesFromState(string state)
+        private AuthenticationProperties SetUserStateOnMessage(OpenIdConnectMessage message)
```
Tratcher: DetaingleUserStateAndProperties? :-)

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationNotifications.cs
```diff
@@ -44,7 +46,19 @@ public OpenIdConnectAuthenticationNotifications()
         /// <summary>
         /// Invoked to manipulate redirects to the identity provider for SignIn, SignOut, or Challenge.
         /// </summary>
-        public Func<RedirectToIdentityProviderNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions>, Task> RedirectToIdentityProvider { get; set; }
+        public Func<RedirectToIdentityProviderNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions>, Task> RedirectToIdentityProvider
+        {
+            get { return _redirectToIdentityProvider; }
+            set
```
Tratcher: NotNull

src/Microsoft.AspNet.Authentication/Notifications/RedirectToIdentityProviderNotification.cs
```diff
@@ -1,16 +1,43 @@
...
 
-        public TMessage ProtocolMessage { get; set; }
+        /// <summary>
+        /// Gets or sets the <see cref="{TMessage}"/>.
+        /// </summary>
+        /// <exception cref="ArgumentNullException">if 'value' is null.</exception>
+        public TMessage ProtocolMessage
+        {
+            get { return _message; }
+            set
```
Tratcher: NotNull

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
```diff
@@ -102,214 +58,207 @@ public void LoggingLevel()
...
+            {
+                var formater = new AuthenticationPropertiesFormaterKeyValue();
+                var properties = new AuthenticationProperties();
+                var dataset = new TheoryData<Action<OpenIdConnectAuthenticationOptions>, OpenIdConnectMessage>();
+
+                var message = new OpenIdConnectMessage();
+                message.State = UrlEncoder.Default.UrlEncode(formater.Protect(properties));
+                message.Code = Guid.NewGuid().ToString();
+                message.Parameters.Add(ExpectedStateParameter, null);
+                dataset.Add(SetStateOptions, message);
```
Tratcher: Add comments describing each test scenario.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
Eilon: Yeah the rule we have with logging level Info and higher (e.g. also Warning and Error) is that there should be at most a *few* of them per request. If this is really just for a dev to diagnose some odd behavior, then Verbose or Debug are more appropriate.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
Eilon: Yeah I don't think a whitespace nonce would ever happen here, even in an error case. The check for whitespace is merely more expensive for no added value. The intention of the code presumably is about a *missing* nonce, not a nonce that is provided but is all whitespace.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
brentschmaltz: I think that would work, will make that change.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
brentschmaltz: I will try and see what happens.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
brentschmaltz: Thanks, I will review https://github.com/aspnet/Logging/wiki/Guidelines
and make adjustments accordingly.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
brentschmaltz: I am fine with making that adjustment.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationNotifications.cs
brentschmaltz: Will do.

src/Microsoft.AspNet.Authentication/Notifications/RedirectToIdentityProviderNotification.cs
brentschmaltz: Will do.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
brentschmaltz: Sure.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
brentschmaltz: I removed some code for the null checks from the previous PR.
I would like to have a comprehensive POR before any other modifications. Then perform a consistency pass.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
brentschmaltz: I think that is a good idea.

brentschmaltz:
@Tratcher @Eilon can you guys have a look at the latest.
I think we should make the change for leaving the notifications null and only engaging if !null.

PinpointTownes:
@brentschmaltz I'm not sure it's really worth it. Anyway, we'll probably switch to good old methods in a future PR to match the pattern used by the OAuth2 middleware, so you won't have to care about null values in the handler itself: https://github.com/aspnet/Security/pull/257#issuecomment-115818763

src/Microsoft.AspNet.Authentication/Notifications/RedirectToIdentityProviderNotification.cs
PinpointTownes:
```
public TMessage ProtocolMessage { get; [param: NotNull] set; }
```

I'm not sure we should have a setter though:
* `OpenIdConnectMessage` is a fully mutable type, so replacing it adds no real value.
* You need cumbersome `message = notification.ProtocolMessage` everywhere to make sure you flow the notification replaced by the user.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -586,34 +621,13 @@ private string ReadNonceCookie(string nonce)
             return null;
         }
 
-        private AuthenticationProperties GetPropertiesFromState(string state)
+        private AuthenticationProperties SetUserStateOnMessage(OpenIdConnectMessage message, AuthenticationProperties properties)
```
PinpointTownes: This method seems rather trivial and is used only once in the handler. Maybe you can remove it?

brentschmaltz: @PinpointTownes  can you help me understand about leaving the notification null.  I thinks there is a gain here, if the user hasn't set a notification, then no need to new up the Notification and call a no-op.

+1 on removing the mini-method: SetUserStateOnMessage

src/Microsoft.AspNet.Authentication/Notifications/RedirectToIdentityProviderNotification.cs
brentschmaltz: @PinpointTownes Thanks for the syntax lesson :-)

If we don't have the setter, we can pass [NoNull] as a param.
How do you feel about that?

brentschmaltz:
@Eilon this PR has been marked cla-required.  What do I need to do?

src/Microsoft.AspNet.Authentication/Notifications/RedirectToIdentityProviderNotification.cs
PinpointTownes: Having an `HttpContext` in the `ProtocolMessage` property is a weird thing. Maybe we should just allow null values in the constructor and remove the setter?

Eilon:
@brentschmaltz I sent you mail about the CLA.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
```diff
@@ -102,214 +57,223 @@ public void LoggingLevel()
...
+            var errors = new List<Tuple<LogEntry, LogEntry>>();
+            var expectedLogs = LoggingUtilities.PopulateLogEntries(expectedLogIndexes);
+            var handler = new OpenIdConnectAuthenticationHandlerForTestingAuthenticate(EmptyTask, EmptyTask);
+            var loggerFactory = new ReturnsLoggerLoggerFactory(logLevel);
+            var server = CreateServer(new ConfigureOptions<OpenIdConnectAuthenticationOptions>(action), UrlEncoder.Default, loggerFactory, handler);
+
+            await server.CreateClient().PostAsync("http://localhost", new FormUrlEncodedContent(message.Parameters));
+            LoggingUtilities.CheckLogs(loggerFactory.Logger.Logs, expectedLogs, errors);
+            Debug.WriteLine(LoggingUtilities.LoggingErrors(errors));
+            Assert.True(errors.Count == 0, LoggingUtilities.LoggingErrors(errors));
```
Eilon: Can use `Assert.Equal`.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
```diff
@@ -394,9 +356,42 @@ private static void SecurityTokenReceivedSkippedOptions(OpenIdConnectAuthenticat
                 };
         }
 
-        private static void SecurityTokenValidatedHandledOptions(OpenIdConnectAuthenticationOptions options)
+        private static void SecurityTokenValidatorCannotReadToken(OpenIdConnectAuthenticationOptions options)
+        {
+            AuthenticationErrorHandledOptions(options);
+            Mock<ISecurityTokenValidator> mockValidator = new Mock<ISecurityTokenValidator>();
```
Eilon: Can `var` a lot of these.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
```diff
@@ -422,14 +417,31 @@ private static void SecurityTokenValidatedSkippedOptions(OpenIdConnectAuthentica
                 };
         }
 
-        #endregion
+        private static void StateNullOptions(OpenIdConnectAuthenticationOptions options)
```
Eilon: All three of the methods here have the same implementation - is that deliberate?

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectMiddlewareTests.cs
```diff
@@ -33,25 +29,31 @@ public class OpenIdConnectMiddlewareTests
     {
         static string noncePrefix = "OpenIdConnect." + "Nonce.";
         static string nonceDelimiter = ".";
+        const string Challenge = "/challenge";
+        const string ChallengeWithOutContext = "/challengeWithOutContext";
+        const string ChallengeWithProperties = "/challengeWithProperties";
+        const string DefaultHost = @"https://example.com";
+        const string DefaultAuthority = @"https://login.windows.net/common";
+        const string ExpectedAuthorizeRequest = @"https://expectedauthorize.com/signin";
```
Eilon: Only use `.example.com/org/net` for fake URLs.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/ScenarioHelperTypes.cs
```diff
@@ -0,0 +1,225 @@
...
+using Microsoft.Framework.Logging;
+using Microsoft.Framework.OptionsModel;
+using Microsoft.Framework.WebEncoders;
+
+namespace Microsoft.AspNet.Authentication.Tests.OpenIdConnect
+{
+    /// <summary>
+    /// This formatter creates an easy to read string of the format: "'key1' 'value1' ..."
+    /// </summary>
+    public class AuthenticationPropertiesFormaterKeyValue : ISecureDataFormat<AuthenticationProperties>
```
Eilon: One top-level type per file.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/ScenarioHelperTypes.cs
```diff
@@ -0,0 +1,225 @@
...
+
+                _logEntries.Add(logEntry);
+                Debug.WriteLine(logEntry.ToString());
+            }
+        }
+
+        public List<LogEntry> Logs { get { return _logEntries; } }
+    }
+
+    public class ReturnsLoggerLoggerFactory : ILoggerFactory
```
Eilon: Use a mock for these?

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/TestUtilities.cs
```diff
@@ -10,6 +16,8 @@ namespace Microsoft.AspNet.Authentication.Tests.OpenIdConnect
     /// </summary>
     public class TestUtilities
     {
+        public const string DefaultHost = @"http://localhost";
+
         public static bool AreEqual<T>(object obj1, object obj2, Func<object, object, bool> comparer = null) where T : class
```
Eilon: How is this method different from the existing xUnit.net method? https://github.com/xunit/xunit/blob/master/src/xunit.assert/Asserts/EqualityAsserts.cs#L30

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/TestUtilities.cs
```diff
@@ -105,4 +108,194 @@ private static bool AreEqual(Exception exception1, Exception exception2)
             return AreEqual<Exception>(exception1.InnerException, exception2.InnerException);
         }
     }
+
+    /// <summary>
+    /// Used to set up different configurations of metadata for different tests
+    /// </summary>
+    public class ConfigurationManager
```
Eilon: One top-level type per file

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
brentschmaltz: fixed

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
brentschmaltz: Yes, it helps in debugging. To see the action in the debugger.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectMiddlewareTests.cs
brentschmaltz: updated.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/ScenarioHelperTypes.cs
brentschmaltz: updated

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/ScenarioHelperTypes.cs
brentschmaltz: I need an actual instance of InMemoryLogger created.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/TestUtilities.cs
brentschmaltz: handles multiple types.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/TestUtilities.cs
brentschmaltz: updated.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
brentschmaltz: yes but there is no overload that has a message.

Tratcher:
Making progress. You'll need to rebase after all the changes that went into Security last week.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
Eilon: But these methods all do the same thing... we don't generally write code that is just for the debugger. What does this do that a debugger watch doesn't do?

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/TestUtilities.cs
Eilon: The xUnit one is generic - it handles any two types.

brentschmaltz:
@Tratcher there are too many diff's to rebase.I will have to craft a new PR.

PinpointTownes:
@brentschmaltz why are you so afraid of rebasing? :smile: 
It's actually far easier to rebase (even when there are merge conflicts) than creating a whole new PR.

brentschmaltz:
updated

PR #304, opened 2015-06-26
Fix social samples

HaoK:
Port fix for social samples from https://github.com/aspnet/Security/pull/257

cc @Tratcher @PinpointTownes


Tratcher:
This doesn't fix the regression, it just works around it. We need to handle the case where Map has modified the paths.

HaoK:
replacing with a real fix

PR #306, opened 2015-06-26
Fix OAuthHandler to use OriginalPathBase

HaoK:
Fixes Map path issue, and consolidates the logic into common helpers now

(Unfortunately can't write a test for twitter like fb since it doesn't surface the redirectUrl easily)

cc @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
@@ -117,7 +117,7 @@ namespace Microsoft.AspNet.Authentication.OAuth
                 }
 
                 var requestPrefix = Request.Scheme + "://" + Request.Host;
-                var redirectUri = requestPrefix + RequestPathBase + Options.CallbackPath;
+                var redirectUri = requestPrefix + OriginalPathBase + Options.CallbackPath;
 
                 var tokens = await ExchangeCodeAsync(code, redirectUri);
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
```diff
@@ -117,7 +117,7 @@ public OAuthAuthenticationHandler(HttpClient backchannel)
                 }
 
                 var requestPrefix = Request.Scheme + "://" + Request.Host;
-                var redirectUri = requestPrefix + RequestPathBase + Options.CallbackPath;
+                var redirectUri = requestPrefix + OriginalPathBase + Options.CallbackPath;
```
Tratcher: Can you use your new BuildRedirectUri helper here?

Tratcher:
:shipit:

test/Microsoft.AspNet.Authentication.Test/Facebook/FacebookMiddlewareTests.cs
```diff
@@ -62,6 +64,37 @@ public class FacebookMiddlewareTests
         }
 
         [Fact]
+        public async Task MapWillNotAffectRedirect()
```
PinpointTownes: Is it worth adding another test where `app.UseFacebookAuthentication` and `app.Map` are both in a global `app.Map`?.

```csharp
var server = CreateServer(app => {
    app.Map("/base", map => {
        map.UseFacebookAuthentication();
        map.Map("/login", signoutApp => signoutApp.Run(context => context.Authentication.ChallengeAsync("Facebook", new AuthenticationProperties() { RedirectUri = "/" })));
     });
});
```

In this case, the `redirect_uri` should be `http://example.com/base/signin-facebook`.

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationHandler.cs
HaoK: yep will fix

test/Microsoft.AspNet.Authentication.Test/Facebook/FacebookMiddlewareTests.cs
HaoK: Sure fair enough

HaoK:
78cf7f99ffd0b18f5beac89ef37345e4aef66dc1

Issue #309, opened 2015-06-29
[Cookies/Identity] ValidateIdentity => SignIn => ValidateIdentity infinite loop


HaoK:
Looks like the new design of immediate sign in breaks the old way that identity checking the security stamp.  It was registering a Identity validator, and then calling sign in with the new identity.  SignIn is applied immediately now so this triggers an infinite loop...weee

@Tratcher

HaoK:
b9f152ebb1c37845a89889ab7ae80a1395a40455

PR #310, opened 2015-06-29
Fix cookie refresh and authenticate caching

HaoK:
Fixes https://github.com/aspnet/Security/issues/309

Also addresses Authenticate being called repeatedly which was a regression (ticket should be generated once and cached)

Added a ShouldRenew for the CookieValidateContext which explicitly tells the cookie to renew the cookie.  Identity used to call SignInDirectly but that would trigger infinite validation loops.  Not sure this is the right fix but it appears to work.  Thoughts? @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -24,8 +24,8 @@ namespace Microsoft.AspNet.Authentication.Cookies
         private const string SessionIdClaim = "Microsoft.AspNet.Authentication.Cookies-SessionId";
 
         private bool _shouldRenew;
-        private DateTimeOffset _renewIssuedUtc;
-        private DateTimeOffset _renewExpiresUtc;
+        private DateTimeOffset? _renewIssuedUtc;
+        private DateTimeOffset? _renewExpiresUtc;
         private string _sessionKey;
 
         protected override async Task<AuthenticationTicket> AuthenticateAsync()
@@ -77,23 +77,28 @@ namespace Microsoft.AspNet.Authentication.Cookies
                     return null;
                 }
 
-                var context = new CookieValidatePrincipalContext(Context, ticket, Options);
-
-                await Options.Notifications.ValidatePrincipal(context);
-
                 var allowRefresh = ticket.Properties.AllowRefresh ?? true;
                 if (issuedUtc != null && expiresUtc != null && Options.SlidingExpiration && allowRefresh)
                 {
                     var timeElapsed = currentUtc.Subtract(issuedUtc.Value);
                     var timeRemaining = expiresUtc.Value.Subtract(currentUtc);
 
-                    if (context.ShouldRenew || timeRemaining < timeElapsed)
+                    if (timeRemaining < timeElapsed)
                     {
                         _shouldRenew = true;
                         _renewIssuedUtc = currentUtc;
                         var timeSpan = expiresUtc.Value.Subtract(issuedUtc.Value);
                         _renewExpiresUtc = currentUtc.Add(timeSpan);
                     }
+                }
+
+                var context = new CookieValidatePrincipalContext(Context, ticket, Options);
+
+                await Options.Notifications.ValidatePrincipal(context);
+
+                if (context.ShouldRenew)
+                {
+                    _shouldRenew = true;
                 }
 
                 return new AuthenticationTicket(context.Principal, context.Properties, Options.AuthenticationScheme);
@@ -134,9 +139,12 @@ namespace Microsoft.AspNet.Authentication.Cookies
         {
             var cookieOptions = BuildCookieOptions();
 
-            if (_shouldRenew)
+            if (_renewIssuedUtc.HasValue)
             {
                 model.Properties.IssuedUtc = _renewIssuedUtc;
+            }
+            if (_renewExpiresUtc.HasValue)
+            {
                 model.Properties.ExpiresUtc = _renewExpiresUtc;
             }
 
@@ -152,9 +160,9 @@ namespace Microsoft.AspNet.Authentication.Cookies
 
             var cookieValue = Options.TicketDataFormat.Protect(model);
 
-            if (model.Properties.IsPersistent)
+            if (model.Properties.IsPersistent && _renewExpiresUtc.HasValue)
             {
-                cookieOptions.Expires = _renewExpiresUtc.ToUniversalTime().DateTime;
+                cookieOptions.Expires = _renewExpiresUtc.Value.ToUniversalTime().DateTime;
             }
 
             Options.CookieManager.AppendResponseCookie(
--- a/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -28,7 +28,7 @@ namespace Microsoft.AspNet.Authentication.Cookies
         private DateTimeOffset _renewExpiresUtc;
         private string _sessionKey;
 
-        public override async Task<AuthenticationTicket> AuthenticateAsync()
+        protected override async Task<AuthenticationTicket> AuthenticateAsync()
         {
             AuthenticationTicket ticket = null;
             try
@@ -77,13 +77,17 @@ namespace Microsoft.AspNet.Authentication.Cookies
                     return null;
                 }
 
+                var context = new CookieValidatePrincipalContext(Context, ticket, Options);
+
+                await Options.Notifications.ValidatePrincipal(context);
+
                 var allowRefresh = ticket.Properties.AllowRefresh ?? true;
                 if (issuedUtc != null && expiresUtc != null && Options.SlidingExpiration && allowRefresh)
                 {
                     var timeElapsed = currentUtc.Subtract(issuedUtc.Value);
                     var timeRemaining = expiresUtc.Value.Subtract(currentUtc);
 
-                    if (timeRemaining < timeElapsed)
+                    if (context.ShouldRenew || timeRemaining < timeElapsed)
                     {
                         _shouldRenew = true;
                         _renewIssuedUtc = currentUtc;
@@ -91,10 +95,6 @@ namespace Microsoft.AspNet.Authentication.Cookies
                         _renewExpiresUtc = currentUtc.Add(timeSpan);
                     }
                 }
-
-                var context = new CookieValidatePrincipalContext(Context, ticket, Options);
-
-                await Options.Notifications.ValidatePrincipal(context);
 
                 return new AuthenticationTicket(context.Principal, context.Properties, Options.AuthenticationScheme);
             }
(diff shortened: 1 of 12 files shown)


src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -77,13 +77,17 @@ internal class CookieAuthenticationHandler : AuthenticationHandler<CookieAuthent
...
+                await Options.Notifications.ValidatePrincipal(context);
+
                 var allowRefresh = ticket.Properties.AllowRefresh ?? true;
                 if (issuedUtc != null && expiresUtc != null && Options.SlidingExpiration && allowRefresh)
                 {
                     var timeElapsed = currentUtc.Subtract(issuedUtc.Value);
                     var timeRemaining = expiresUtc.Value.Subtract(currentUtc);
 
-                    if (timeRemaining < timeElapsed)
+                    if (context.ShouldRenew || timeRemaining < timeElapsed)
```
Tratcher: Should notification renewal be dependent on sliding refresh and all those conditions in the outer if?

HaoK:
Updated splitting notification refresh from sliding refresh as discussed

Tratcher:
:shipit:

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -134,8 +139,14 @@ private CookieOptions BuildCookieOptions()
         {
```
brentschmaltz: suggest replacing 'model' with 'ticket' it would read better.

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -149,9 +160,9 @@ private CookieOptions BuildCookieOptions()
 
             var cookieValue = Options.TicketDataFormat.Protect(model);
 
-            if (model.Properties.IsPersistent)
+            if (model.Properties.IsPersistent && _renewExpiresUtc.HasValue)
```
brentschmaltz: The use of .HasValue is inconsistent in this code. In other areas == null is checked.
Should be consistent.

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -149,9 +160,9 @@ private CookieOptions BuildCookieOptions()
...
             {
-                cookieOptions.Expires = _renewExpiresUtc.ToUniversalTime().DateTime;
+                cookieOptions.Expires = _renewExpiresUtc.Value.ToUniversalTime().DateTime;
```
brentschmaltz: Why call ToUniversalTime? the name of the variable implies UTC, if it is not, that is an error.

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -181,7 +192,7 @@ private CookieOptions BuildCookieOptions()
                 return;
             }
 
-            var model = await AuthenticateAsync();
+            var model = await AuthenticateOnceAsync();
```
brentschmaltz: suggest renaming model => ticket

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -155,14 +157,24 @@ public void GetDescriptions(DescribeSchemesContext describeContext)
             }
         }
 
+        protected Task<AuthenticationTicket> AuthenticateOnceAsync()
+        {
+            if (_authenticate == null)
+            {
+                _authenticate = AuthenticateAsync();
```
brentschmaltz: suggest renaming _authenticate => _ticket or _authenticationTicket

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -155,14 +157,24 @@ public void GetDescriptions(DescribeSchemesContext describeContext)
             }
         }
 
+        protected Task<AuthenticationTicket> AuthenticateOnceAsync()
```
brentschmaltz: What is the need for this logic of remembering if _authenticate was set?
What happens if the AuthenticateAsync() returns null?

src/Microsoft.AspNet.Authentication/DataHandler/Serializer/TicketSerializer.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.AspNet.Authentication.DataHandler.Serializer
 {
     public class TicketSerializer : IDataSerializer<AuthenticationTicket>
     {
-        private const int FormatVersion = 2;
+        private const int FormatVersion = 3;
```
brentschmaltz: Are you relying on this to determine the correct deserialization routine to use?
If so, this will not work in a heterogeneous where some servers are upgraded and some are not.

src/Microsoft.AspNet.Authentication/DataHandler/Serializer/TicketSerializer.cs
```diff
@@ -41,21 +41,29 @@ public static void Write([NotNull] BinaryWriter writer, [NotNull] Authentication
...
+            {
+                writer.Write(principal.Identities.Count());
+                foreach (var identity in principal.Identities)
                 {
-                    WriteWithDefault(writer, claim.Type, identity.NameClaimType);
-                    writer.Write(claim.Value);
-                    WriteWithDefault(writer, claim.ValueType, DefaultValues.StringValueType);
-                    WriteWithDefault(writer, claim.Issuer, DefaultValues.LocalAuthority);
-                    WriteWithDefault(writer, claim.OriginalIssuer, claim.Issuer);
+                    var authenticationType = string.IsNullOrWhiteSpace(identity.AuthenticationType) ? string.Empty : identity.AuthenticationType;
```
brentschmaltz: Why the translation: whitespace => string.Empty?

src/Microsoft.AspNet.Authentication/DataHandler/Serializer/TicketSerializer.cs
Tratcher: Deserialization would fail in a heterogeneous environment even without this check, the expected formats are different. This check is only to make the failure deterministic and debuggable rather than having random IndexOutOfRangeExceptions or NullReferenceExceptions.

src/Microsoft.AspNet.Authentication/DataHandler/Serializer/TicketSerializer.cs
Tratcher: FYI any usage of IsNullOrWhiteSpace should be replaced with IsNullOrEmpty.

test/Microsoft.AspNet.Authentication.Test/Cookies/CookieMiddlewareTests.cs
```diff
@@ -342,6 +342,132 @@ private Task SignOutAsWrong(HttpContext context)
...
+
+            clock.Add(TimeSpan.FromMinutes(6));
+
+            var transaction4 = await SendAsync(server, "http://example.com/me/Cookies", transaction3.CookieNameValue);
+
+            clock.Add(TimeSpan.FromMinutes(11));
+
+            var transaction5 = await SendAsync(server, "http://example.com/me/Cookies", transaction4.CookieNameValue);
+
+            transaction2.SetCookie.ShouldNotBe(null);
```
brentschmaltz: Do you have to call each SendAsync before the FindClaimValue(...) ?
If not, suggest grouping them together.

test/Microsoft.AspNet.Authentication.Test/DataHandler/Encoder/TicketSerializerTests.cs
```diff
@@ -0,0 +1,56 @@
...
+                TicketSerializer.Write(writer, ticket);
+                stream.Position = 0;
+                var readTicket = TicketSerializer.Read(reader);
+                readTicket.Principal.Identities.Count().ShouldBe(0);
+                readTicket.Properties.RedirectUri.ShouldBe("bye");
+                readTicket.AuthenticationScheme.ShouldBe("Hello");
+            }
+        }
+    }
+}
```
brentschmaltz: Are there comprehensive tests on Serializing complex ClaimsPrincipals with fully populated and complex ClaimsIdentities?

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
HaoK: Sure

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
HaoK: There's a community PR out for this https://github.com/aspnet/Security/pull/286

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
HaoK: You aren't supposed to return null tasks, the idea is only to call authenticate once, and subsequent calls should return the same result/ticket/null

test/Microsoft.AspNet.Authentication.Test/Cookies/CookieMiddlewareTests.cs
HaoK: Yeah that's what I usually do but that's how these tests were written I guess I can clean them up

test/Microsoft.AspNet.Authentication.Test/DataHandler/Encoder/TicketSerializerTests.cs
HaoK: Nope

HaoK: I believe the plan is to switch to ClaimsPrincipal.WriteTo once its available

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
Tratcher: Ah, the naming could use work. _authenticateTask?

HaoK: Yeah sure, I ressurected the old name from before, but I agree its unclear, _authenticateTask works for me

HaoK:
b9f152ebb1c37845a89889ab7ae80a1395a40455

src/Microsoft.AspNet.Authentication/DataHandler/Serializer/TicketSerializer.cs
brentschmaltz: @Tratcher  @HaoK  @Eilon @tushargupta51 
You can design for this, I have done it multiple times.
When I designed .Net 4.5 to shim ClaimsPrincipal behind WindowsPrincipal, RolePrincipal and GenericPrincipal I had to design for .Net 4.0 and .Net 4.5 servers running in the same environment. A .Net 4.0 server would handle the new format and 4.0 handle the new.

The new .WriteTo OM in 4.6 is designed to flow through multiple versions without a version identifier.

PR #311, opened 2015-06-30
Fix exception parameter name

hishamco:
This PK made according to the issue that raised in the following commit https://github.com/aspnet/WebSockets/commit/208fa4af366e23b51f2b943f709e00b6d48eec22
/cc @Eilon

### DIFF
--- a/src/Microsoft.AspNet.Authentication.Facebook/FacebookAuthenticationMiddleware.cs
+++ b/src/Microsoft.AspNet.Authentication.Facebook/FacebookAuthenticationMiddleware.cs
@@ -37,11 +37,11 @@ namespace Microsoft.AspNet.Authentication.Facebook
         {
             if (string.IsNullOrWhiteSpace(Options.AppId))
             {
-                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(Options.AppId)));
+                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(Options)));
             }
             if (string.IsNullOrWhiteSpace(Options.AppSecret))
             {
-                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(Options.AppSecret)));
+                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(Options)));
             }
 
             if (Options.Notifications == null)
--- a/src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationMiddleware.cs
+++ b/src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationMiddleware.cs
@@ -43,27 +43,27 @@ namespace Microsoft.AspNet.Authentication.OAuth
             // todo: review error handling
             if (string.IsNullOrWhiteSpace(Options.AuthenticationScheme))
             {
-                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(Options.AuthenticationScheme)));
+                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(Options)));
             }
 
             if (string.IsNullOrWhiteSpace(Options.ClientId))
             {
-                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(Options.ClientId)));
+                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(Options)));
             }
 
             if (string.IsNullOrWhiteSpace(Options.ClientSecret))
             {
-                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(Options.ClientSecret)));
+                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(Options)));
             }
 
             if (string.IsNullOrWhiteSpace(Options.AuthorizationEndpoint))
             {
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNet.Authentication.Facebook/FacebookAuthenticationMiddleware.cs
```diff
@@ -37,11 +37,11 @@ public class FacebookAuthenticationMiddleware : OAuthAuthenticationMiddleware<Fa
         {
             if (string.IsNullOrWhiteSpace(Options.AppId))
             {
-                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(Options.AppId)));
+                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(Options)));
```
Eilon: This doesn't look right to me. There's no parameter named `Options` either. Plus, supposing there was, the exception message won't make much sense to the user: what they really need to know is that the `AppId` property wasn't set on the options.

hishamco: @Eilon the ```Options``` parameter is there!!, I'm agree that the message won't make sense, but I think we need to modify the message, not the exception message while ```Options``` is the original source for ```AppId```

Eilon: There's an `options` parameter, not an `Options` parameter.

hishamco: Look to this line of code https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Facebook/FacebookAuthenticationMiddleware.cs#L38

Eilon: That code is using the `Options` property, not the `options` parameter.

hishamco: What I did is I use the original source ```Options``` as exception parameter instead of ```Options.AppId```

Eilon: The parameter to an ArgumentException is the name of the argument. Is there a better way to do this exception?

hishamco: I think ```ArgumentException``` is suitable but i'm still asking why we are using ```Options``` instead of ```options```?!!

PinpointTownes:
I understand the reason behind this change, but sadly, it will make all these null checks totally useless, as the user will have no way to know which specific option was invalid.

Eilon:
@PinpointTownes exactly.

PinpointTownes:
@Eilon I need to buy glasses, I had not realized you had posted a similar comment :smile: :eyeglasses:

src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationMiddleware.cs
```diff
@@ -43,27 +43,27 @@ public class OAuthAuthenticationMiddleware<TOptions, TNotifications> : Authentic
             // todo: review error handling
             if (string.IsNullOrWhiteSpace(Options.AuthenticationScheme))
             {
-                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(Options.AuthenticationScheme)));
+                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(Options)));
```
brentschmaltz: Will this change result in the user being informed that 'Options.AuthenticationScheme' is missing or "Options".
What does the exception message look like?

hishamco: You are right @brentschmaltz, but we faced an issue with ```nameof``` while it returns the name of the last identifier in the property path. I know the latest change doesn't make sense from meaning point of you. I'm looking for a decent solution. Hope i hear from you guys :smile:

hishamco: FYI the ```nameof``` issue has been suggested here https://github.com/dotnet/roslyn/issues/3759

PinpointTownes:
Let's make a compromise then:
* Mention the whole property path in the exception message.
* Use `ArgumentException`'s constructor taking a parameter name and set it to `nameof(options)`.

hishamco:
@PinpointTownes we don't write the whole property as string, if we did what's the benefit of introducing ```nameof``` here

PinpointTownes:
@hishamco if you really want to use nameof everywhere, just mention the option name:

```csharp
throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(options.AuthenticationScheme)), nameof(options));
```

hishamco:
:+1: @PinpointTownes 
While the exception may caused if the ```AuthenticationScheme``` is whitespace, so it will be nice if we change it to
```csharp
throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(options.AuthenticationScheme)), $"{nameof(options)}.{nameof(options.AuthenticationScheme)}");
```

PinpointTownes:
But in this case, `paramName` won't be equal to `options`, which is the name of the constructor parameter... @Eilon won't be happy :smile:

Eilon:
@PinpointTownes correct.

@hishamco the parameter to an `ArgumentException` is *the name of the parameter*. It's not for "other information that might be relevant". That's what the exception *message* is for.

hishamco:
But we should use ```$"{nameof(buffer)}.{nameof(buffer.Offset)}``` here https://github.com/aspnet/WebSockets/commit/208fa4af366e23b51f2b943f709e00b6d48eec22#diff-2f20d3473af8bc60eee727dd88bbb804R595 Ami I right?

PinpointTownes:
@hishamco you can use `$"{nameof(buffer)}.{nameof(buffer.Offset)}` in the message, but not for the parameter name.

hishamco:
Sorry I think I got confused little bit :smile:

Eilon:
Looks like this PR still has unaddressed issues, so closing this out for now. If you have time to correct the issues and send a new PR please let us know. Thanks!

Issue #270, opened 2015-05-20
Cookie MW null reference exception


dolly22:
I'm getting null reference exception from cookie authentication middleware when protecting authentication ticket. This happens when the following conditions occur on request:

1. Sliding expiration is enabled and enough time passed so middleware tries to refresh authentication ticket (setting `_shouldRenew = true`)
2. Principal is rejected in `Options.Notifications.ValidatePrincipal` notification (setting `Context.Principal = null`)
3. `_shouldRenew` path is hit inside `ApplyResponseGrantAsync()` and null reference is thrown from `Options.TicketDataFormat.Protect(model);` because `Principal` was set to null.

Also `Security/src/Microsoft.AspNet.Authentication.Cookies/Notifications/CookieValidateIdentityContext.cs` should be probably renamed to `CookieValidatePrincipalContext.cs`

muratg:
@HaoK Could you have a look?  We can move it out of Beta 6 based on your workload.

HaoK:
d7ce42dacc226c74056a6c0883f9b3e2be7828d7

PR #312, opened 2015-06-30
Handle Null Principal in TicketSerializer

HaoK:
Rebased vs dev

cc @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Authentication/DataHandler/Serializer/TicketSerializer.cs
+++ b/src/Microsoft.AspNet.Authentication/DataHandler/Serializer/TicketSerializer.cs
@@ -41,21 +41,29 @@ namespace Microsoft.AspNet.Authentication.DataHandler.Serializer
             writer.Write(FormatVersion);
             writer.Write(model.AuthenticationScheme);
             var principal = model.Principal;
-            writer.Write(principal.Identities.Count());
-            foreach (var identity in principal.Identities)
+            if (principal == null)
             {
-                var authenticationType = string.IsNullOrWhiteSpace(identity.AuthenticationType) ? string.Empty : identity.AuthenticationType;
-                writer.Write(authenticationType);
-                WriteWithDefault(writer, identity.NameClaimType, DefaultValues.NameClaimType);
-                WriteWithDefault(writer, identity.RoleClaimType, DefaultValues.RoleClaimType);
-                writer.Write(identity.Claims.Count());
-                foreach (var claim in identity.Claims)
+                // Use -1 to signal null
+                writer.Write(-1);
+            }
+            else
+            {
+                writer.Write(principal.Identities.Count());
+                foreach (var identity in principal.Identities)
                 {
-                    WriteWithDefault(writer, claim.Type, identity.NameClaimType);
-                    writer.Write(claim.Value);
-                    WriteWithDefault(writer, claim.ValueType, DefaultValues.StringValueType);
-                    WriteWithDefault(writer, claim.Issuer, DefaultValues.LocalAuthority);
-                    WriteWithDefault(writer, claim.OriginalIssuer, claim.Issuer);
+                    var authenticationType = string.IsNullOrWhiteSpace(identity.AuthenticationType) ? string.Empty : identity.AuthenticationType;
+                    writer.Write(authenticationType);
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Authentication/DataHandler/Serializer/TicketSerializer.cs
```diff
@@ -41,21 +41,29 @@ public static void Write([NotNull] BinaryWriter writer, [NotNull] Authentication
             writer.Write(FormatVersion);
```
Tratcher: A break like this may require changing the FormatVersion.

Tratcher:
Why would you serialize a null ClaimsPrincipal?

src/Microsoft.AspNet.Authentication/DataHandler/Serializer/TicketSerializer.cs
```diff
@@ -69,27 +77,35 @@ public static AuthenticationTicket Read([NotNull] BinaryReader reader)
             }
             string authenticationScheme = reader.ReadString();
             int identityCount = reader.ReadInt32();
-            var identities = new ClaimsIdentity[identityCount];
-            for (int i = 0; i != identityCount; ++i)
+            ClaimsPrincipal principal = null;
+
+            // Negative values are used to signify null
+            if (identityCount > 0)
```
Tratcher: I think this should be >= 0. Right now if you serialize a non-null ClaimsPrincipal with 0 identities, you will get back a null principal.

HaoK:
We support null principal in the auth ticket generally today, and you can serialize the properties still.  We don't have guard anywhere today, so this seemed reasonable.  But yes, I agree this is some more code smell...

test/Microsoft.AspNet.Authentication.Test/Facebook/FacebookMiddlewareTests.cs
```diff
@@ -125,6 +125,67 @@ public class FacebookMiddlewareTests
         }
 
         [Fact]
+        public async Task NestedMapWillNotAffectRedirect()
```
PinpointTownes: Bad rebase?

src/Microsoft.AspNet.Authentication/DataHandler/Serializer/TicketSerializer.cs
PinpointTownes: I'm not sure: the serialization format is not really changed (this PR only gives `-1` a special meaning).

Tratcher: Old code will fail to deserialize a new ticket that contains -1.

PinpointTownes: Ah yeah, it will likely crash in the for loop, due to the way it's coded :+1: 

`for (int i = 0; i != identityCount; ++i)`

test/Microsoft.AspNet.Authentication.Test/Facebook/FacebookMiddlewareTests.cs
HaoK: No clue, I wouldn't worry about it, I always squash/merge

HaoK: Ah yeah probably because  this was based of the map fix branch and I rebased to dev so its confused

HaoK: Oh its a dupe, yeah i'll remove :)

src/Microsoft.AspNet.Authentication/DataHandler/Serializer/TicketSerializer.cs
HaoK: Updated Version and added test for empty principal

Tratcher:
:shipit:
Is there an issue tracking this bug?

PinpointTownes:
@Tratcher https://github.com/aspnet/Security/issues/270

HaoK:
d7ce42dacc226c74056a6c0883f9b3e2be7828d7

PR #315, opened 2015-07-01
State changes

brentschmaltz:
Additional Tests
Check for null RedirectToIdentityProviderNotification != null before preparing the call
Additional state can be easily set by user in RedirectToIdentityProviderNotification

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
@@ -62,7 +62,11 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
 
                 if (Options.Notifications.RedirectToIdentityProvider != null)
                 {
-                    var redirectToIdentityProviderNotification = new RedirectToIdentityProviderNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions>(Context, Options, message);
+                    var redirectToIdentityProviderNotification = new RedirectToIdentityProviderNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions>(Context, Options)
+                    {
+                        ProtocolMessage = message
+                    };
+
                     await Options.Notifications.RedirectToIdentityProvider(redirectToIdentityProviderNotification);
                     if (redirectToIdentityProviderNotification.HandledResponse)
                     {
@@ -166,7 +170,10 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
             if (Options.Notifications.RedirectToIdentityProvider != null)
             {
                 var redirectToIdentityProviderNotification =
-                    new RedirectToIdentityProviderNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions>(Context, Options, message);
+                    new RedirectToIdentityProviderNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions>(Context, Options)
+                    {
+                        ProtocolMessage = message
+                    };
 
                 await Options.Notifications.RedirectToIdentityProvider(redirectToIdentityProviderNotification);
                 if (redirectToIdentityProviderNotification.HandledResponse)
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
@@ -28,19 +28,6 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
         private const string UriSchemeDelimiter = "://";
         private OpenIdConnectConfiguration _configuration;
 
-        private string CurrentUri
-        {
-            get
-            {
-                return Request.Scheme +
-                       UriSchemeDelimiter +
-                       Request.Host +
-                       Request.PathBase +
-                       Request.Path +
-                       Request.QueryString;
-            }
-        }
-
         /// <summary>
         /// Handles Signout
         /// </summary>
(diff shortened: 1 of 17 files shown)


src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -125,7 +124,7 @@ private string CurrentUri
             // When redeeming a 'code' for an AccessToken, this value is needed
             if (!string.IsNullOrWhiteSpace(Options.RedirectUri))
             {
-                properties.Items.Add(OpenIdConnectAuthenticationDefaults.RedirectUriUsedForCodeKey, Options.RedirectUri);
+                properties.Items.Add(OpenIdConnectAuthenticationDefaults.RedirectUriForCodePropertiesKey, Options.RedirectUri);
```
Tratcher: This if block can be merged with the prior one.

Replace IsNullOrWhiteSpace with IsNullOrEmpty.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -66,30 +53,42 @@ private string CurrentUri
...
                 }
-                else if (!string.IsNullOrWhiteSpace(Options.PostLogoutRedirectUri))
+                else if (!string.IsNullOrEmpty(Options.PostLogoutRedirectUri))
                 {
-                    openIdConnectMessage.PostLogoutRedirectUri = Options.PostLogoutRedirectUri;
+                    message.PostLogoutRedirectUri = Options.PostLogoutRedirectUri;
                 }
 
-                var notification = new RedirectToIdentityProviderNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions>(Context, Options)
+                if (Options.Notifications.RedirectToIdentityProvider != null)
```
PinpointTownes: For consistency, you should avoid adding this kind of changes in this PR.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -169,24 +167,36 @@ private string CurrentUri
...
+                    return false;
+                }
+
+                if (!string.IsNullOrEmpty(redirectToIdentityProviderNotification.ProtocolMessage.State))
+                {
+                    properties.Items[OpenIdConnectAuthenticationDefaults.UserstatePropertiesKey] = redirectToIdentityProviderNotification.ProtocolMessage.State;
+                }
+
+                message = redirectToIdentityProviderNotification.ProtocolMessage;
+                message.State = Options.StateDataFormat.Protect(properties);
```
Tratcher: message.State needs to be set outside of this notification block. There is still state even if there was no notification event.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationNotifications.cs
```diff
@@ -23,7 +23,6 @@ public OpenIdConnectAuthenticationNotifications()
             MessageReceived = notification => Task.FromResult(0);
             SecurityTokenReceived = notification => Task.FromResult(0);
             SecurityTokenValidated = notification => Task.FromResult(0);
-            RedirectToIdentityProvider = notification => Task.FromResult(0);
```
Tratcher: We should sit down and review the whole notification design before you apply this to any additional events. We want to standardize the notification pattern across all the auth handlers.

brentschmaltz: This is the pattern I am pushing for.
Let's decide now.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
brentschmaltz: Sure enough, that was a copy paste issue.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
brentschmaltz: Thanks, cut and paste error.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationNotifications.cs
HaoK: This PR is big enough, there's no need to couple it to orthogonal consistency changes, don't change the notification pattern as part of this PR

HaoK: Once we decide on the pattern to use everywhere, one single PR should apply the changes everywhere.

brentschmaltz:
@Tratcher @PinpointTownes to move things forward, I put back the default RedirectToIdentityProvider = notification => Task.FromResult(0);
I still check for null, since the user can (with existing code) set the handler to null.

PinpointTownes:
If you really want to prevent a user from setting notifications to null, why not simply adding a `[param: NotNull]` attribute on their setter?

brentschmaltz:
Well, I don't actually. For a very busy server, handling a lot of request. Setting up for notifications that just no-op seems a waste. So I would like to start with them null and take action if the user actually hooked one.

PinpointTownes:
Awaiting a completed task (like the one returned by `Task.FromResult`) is not that expensive and never causes a thread switching. We're really talking about micro-optimization here (btw, in a code path that shouldn't be hit thousands times a second :smile:)

Do you have a benchmark confirming your feeling?

brentschmaltz:
It is not just the Task.FromResult, but the build up to make the call. We have to new up objects etc.
We took changes like using: string.IsNullOrEmpty rather than string.IsNullOrWhiteSpace just because it seemed like the right thing.
This feels the same way, and is the way traditional asp.net / WIF works.  Sometimes you have to go, ya that seems right.

Eilon:
Checks for whitespace are *functionally incorrect* in 99% of cases. That is different from doing an optimization where the functionality is identical.

brentschmaltz:
I am not sure exactly sure what you mean by functionally incorrect or how you came to 99%, but I removed the default task handler and put in a check for null since the existing API allows for the users to set the notification to null.

Is there anything else blocking this PR?

Eilon:
@brentschmaltz we'll finish this up next week when @Tratcher is back from vacation.

Tratcher:
Rebased, squashed, and merged.

PR #316, opened 2015-07-01
Tweak SecurityHelper for MVC usage

HaoK:
Pass/Return HttpContext.user instead of manipulating the context to make this reusable for MVC

cc @Tratcher


Tratcher:
:shipit: There's no behavior change, correct?

HaoK:
I certainly hope not!

HaoK:
While I'm touching this, should we consider renaming the helper?  We could even make it an extension method on ClaimsPrincipal instead?

Tratcher:
PrincipalHelper?
I don't recommend changing it to an extension method.

HaoK:
If we are going to keep helper I guess security isn't bad

HaoK:
5e92de8009901aa73ccdbd1d9ebe4fc89dbc7add

Issue #341, opened 2015-07-07
Need to rename the protected AuthenticationTicket AuthenticateAsync to not clash


HaoK:
public async Task AuthenticateAsync(AuthenticateContext context)
protected abstract Task<AuthenticationTicket> AuthenticateAsync();

Just a little confusing...

HaoK:
Leaning towards CreateTicketAsync (which happens to be the same name @PinpointTownes used for the OAuth stuff, which is good)

PinpointTownes:
> which happens to be the same name @PinpointTownes used for the OAuth stuff, which is good

And I'll also use it for the future OpenID 2.0 generic middleware: https://github.com/aspnet-contrib/AspNet.Security.OpenId.Providers/pull/2/files#diff-5cf36d7a3098ef19627e2a76097ca8b6R255 :smile: 

That said, I'm not sure that renaming `AuthenticateAsync` to `CreateTicketAsync` is actually a good idea:

1. It would clash with the `CreateTicketAsync` method we use in the OAuth2 base handler (the two `CreateTicketAsync` methods would only differ by their arguments) :laughing: 

2. `CreateTicketAsync` doesn't seem 'generic' enough to describe what could happen during the authentication process. For instance, `AuthenticateAsync` is also the place where the cookie middleware determines whether the cookie should be renewed. `CreateTicketAsync` would finally be much more than a simple method creating a ticket.

Why not simply renaming the protected abstract `AuthenticateAsync` to `AuthenticateCoreAsync`? (well, like in Katana 3, actually) :smile:

HaoK:
Well Authenticate and AuthenticateCore might as well both be Authenticate in my book.  I was trying to tease apart what the base method does (which is deal with the Http IAuthenticationHandler stuff), and name the abstract "Core" something else, I guess to follow my own naming scheme, this could just be:

`public Task<AuthenticationTicket> HandleAuthenticateAsync()'

HaoK:
Fixed in https://github.com/aspnet/Security/pull/343

PinpointTownes:
:+1: `HandleAuthenticateAsync` seems a good option (though `HandleAuthenticationAsync` seems more natural).

HaoK:
Yeah the naming pattern I used was simply adding a Handle prefix in front of the non virtual public method name (as opposed to the old Core suffix which I strongly disliked :)

HaoK:
ab4ba794e546074a573fbf76d9150ffcf9752c89

Issue #338, opened 2015-07-07
AuthenticationHandler.BaseInitializeAsync calls ShouldHandleScheme with string.Empty


brentschmaltz:
The comment says: // Automatic authentication is the empty scheme
This does not match the logic in ShouldHandleScheme

```c#
 return string.Equals(BaseOptions.AuthenticationScheme, authenticationScheme, StringComparison.Ordinal) ||
                (BaseOptions.AutomaticAuthentication && string.IsNullOrWhiteSpace(authenticationScheme));
```

1. What is the purpose of BaseOptions.AutomaticAuthentication if the comment is true?
2. The comment says string.IsEmpty, but null, and any whitespace will be true.

HaoK:
Fixed in https://github.com/aspnet/Security/pull/343

Issue #325, opened 2015-07-01
[Cookies] Reject/throw null principal authentication tickets


HaoK:
Also make serializer throw if principal is null instead

brentschmaltz:
Can you be more specific on what you are proposing?
Thanks.

HaoK:
Rather than the serializer handling nulls as it does today, it will throw instead, and cookie middleware will reject/not support storing tickets with no principal.

HaoK:
ab4ba794e546074a573fbf76d9150ffcf9752c89

PR #343, opened 2015-07-07
Rename abstract Authenticate => HandleAuthenticate

HaoK:
Also:
- fix ShouldHandleScheme to only check for string.Empty
- use the Once version of Authenticate more consistently

Fixes https://github.com/aspnet/Security/issues/341, https://github.com/aspnet/Security/issues/338

cc @Tratcher


HaoK:
Add a few tests

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -161,7 +161,7 @@ protected Task<AuthenticationTicket> AuthenticateOnceAsync()
         {
             if (_authenticateTask == null)
             {
-                _authenticateTask = AuthenticateAsync();
+                _authenticateTask = HandleAuthenticateAsync();
```
Tratcher: It's odd that you renamed AuthenticateAsync -> HandleAuthenticateAsync, but left AuthenticateOnceAsync as is. HandleAuthenticateOnceAsync?

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -189,12 +189,12 @@ protected Task<AuthenticationTicket> AuthenticateOnceAsync()
...
         }
 
-        protected abstract Task<AuthenticationTicket> AuthenticateAsync();
+        protected abstract Task<AuthenticationTicket> HandleAuthenticateAsync();
 
         public bool ShouldHandleScheme(string authenticationScheme)
         {
             return string.Equals(BaseOptions.AuthenticationScheme, authenticationScheme, StringComparison.Ordinal) ||
-                (BaseOptions.AutomaticAuthentication && string.IsNullOrWhiteSpace(authenticationScheme));
+                (BaseOptions.AutomaticAuthentication && authenticationScheme == string.Empty);
```
Tratcher: Use IsNullOrEmpty. Never use == with strings.

Tratcher:
:shipit:

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
HaoK: Brent brought up an issue that we use automatic with an empty string never null. I am on the fence. 

But I am fine with treating null and empty the same as we were before if you feel strongly about it.

Tratcher: I could have sworn there was a dot net design guideline for treating string.Empty and null as the same, but I can't find a reference for it.

HaoK: Well I don't like treating them differently in general, but since we will be using [NotNull] for authenticationScheme everywhere, I don't care too strongly about this.  You have the tiebreaking vote :)

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
HaoK: Fair enough I'll rename that too

HaoK:
Updated PR to fix cookie issues since this was touching the HandleAuthenticate code

@Tratcher can you review again, updated Cookie and TicketSerializer (to prevent nulls)

HaoK:
Fixes https://github.com/aspnet/Security/issues/355 and https://github.com/aspnet/Security/issues/325 now as well

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -28,74 +28,88 @@ internal class CookieAuthenticationHandler : AuthenticationHandler<CookieAuthent
         private DateTimeOffset? _renewExpiresUtc;
         private string _sessionKey;
 
-        protected override async Task<AuthenticationTicket> AuthenticateAsync()
+        private async Task<AuthenticationTicket> ConfigureTicket()
```
Tratcher: RetrieveTicketAsync? Also, is there any point in having this run more than once? RetrieveTicketOnceAsync?

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -299,7 +314,7 @@ private CookieOptions BuildCookieOptions()
             catch (Exception exception)
             {
                 var exceptionContext = new CookieExceptionContext(Context, Options,
-                    CookieExceptionContext.ExceptionLocation.ApplyResponseGrant, exception, model);
+                    CookieExceptionContext.ExceptionLocation.ApplyResponseGrant, exception, ticket);
```
Tratcher: These ExceptionLocations are out of date.

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -183,7 +197,7 @@ private CookieOptions BuildCookieOptions()
                 return;
             }
 
-            var ticket = await AuthenticateOnceAsync();
+            var ticket = await HandleAuthenticateOnceAsync();
             try
             {
                 await ApplyCookie(ticket);
```
Tratcher: This is the only place that ApplyCookie is called. Consider inlining it or refactoring it so that it can also be called from HandleSignIn

Tratcher:
:watch:

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
HaoK: Went with EnsureCookieTicket read and stored in a field

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
HaoK: Added enums for each virtual (SignIn, SignOut, Unauthorized, Forbidden, FinishResponse) this seems a bit much to me...

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
HaoK: I'll see if I can combine it with the sign in one

HaoK: Moved to a ApplyHeaders(bool shouldRedirectToReturnUrl) helper

HaoK:
PR updated

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -27,75 +27,103 @@ internal class CookieAuthenticationHandler : AuthenticationHandler<CookieAuthent
...
+                            _renewIssuedUtc = currentUtc;
+                            var timeSpan = expiresUtc.Value.Subtract(issuedUtc.Value);
+                            _renewExpiresUtc = currentUtc.Add(timeSpan);
+                        }
+                    }
+
+                    // Finally we have a valid ticket
+                    _cookieTicket = ticket;
+                }
+                finally
```
Tratcher: drop the try/finally and just set _cookieRead at the top.

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -27,75 +27,103 @@ internal class CookieAuthenticationHandler : AuthenticationHandler<CookieAuthent
...
+        }
 
-                var context = new CookieValidatePrincipalContext(Context, ticket, Options);
+        protected override async Task<AuthenticationTicket> HandleAuthenticateAsync()
+        {
+            AuthenticationTicket ticket = null;
+            try
+            {
+                await EnsureCookieTicket();
+                ticket = _cookieTicket;
```
Tratcher: Did you notice that every usage of `await EnsureCookieTicket();` is immediately followed by `ticket = _cookieTicket;`? Why not use the same pattern as HandleAuthenticateOnceAsync?

HaoK: You mean the pattern in the base class?  Its for the exception handler which sometimes changes the ticket... for some of these methods...  Yes I noticed this code looks bad :)  Let me take a look again

HaoK: Alright cleaned things up to use the pattern I used in AuthHandler

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -27,75 +27,102 @@ internal class CookieAuthenticationHandler : AuthenticationHandler<CookieAuthent
         private DateTimeOffset? _renewIssuedUtc;
         private DateTimeOffset? _renewExpiresUtc;
         private string _sessionKey;
+        private Task<AuthenticationTicket> _cookieTicketTask;
+        private bool _cookieRead;
```
Tratcher: _cookieRead is no longer used.

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -27,75 +27,102 @@ internal class CookieAuthenticationHandler : AuthenticationHandler<CookieAuthent
...
-            AuthenticationTicket ticket = null;
-            try
+            // We only need to read the ticket once
+            if (_cookieTicketTask == null)
+            {
+                _cookieTicketTask = ReadCookieTicket();
+            }
+            return _cookieTicketTask;
+        }
+
```
Tratcher: nit: extra line

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -270,36 +289,13 @@ private CookieOptions BuildCookieOptions()
...
-
-                Response.Headers.Set(
-                    HeaderNamePragma,
-                    HeaderValueNoCache);
-
-                Response.Headers.Set(
-                    HeaderNameExpires,
-                    HeaderValueMinusOne);
-
                 var shouldLoginRedirect = Options.LoginPath.HasValue && Request.Path == Options.LoginPath;
```
Tratcher: Request.Path may be wrong (altered by Map) when SignIn is applied immediately. Do we need to add base.OriginalPath like we have for OriginalPathBase?

HaoK: Yeah looks like it would be the same issue, I'll add a test and fix if its busted

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -387,36 +377,34 @@ private static bool IsHostRelative(string path)
...
-                    }
-                }
-                return Task.FromResult(true);
+            try
+            {
+                var accessDeniedUri =
+                    Request.Scheme +
+                    "://" +
+                    Request.Host +
+                    Request.PathBase +
```
Tratcher: Same base.OriginalPathBase issue here.

Tratcher: And in HandleUnauthorizedAsync

HaoK:
Updated with new tests and fixed signin/signout to reflect map correctly.  I'll push this so I can start on the API minor changes, and address any further PR cleanup in the next PR

Tratcher:
:shipit:

HaoK:
ab4ba794e546074a573fbf76d9150ffcf9752c89

Issue #339, opened 2015-07-07
Create an Authorization.Abstractions package for frameworks to reference


rynowak:
API review feedback. MVC is taking a reference on Authorization+Authentication just to use the attribute to call into the authorization service.

Basically the hope is that the code in the MVC `AuthorizeFilter` would only need this dependency.

We want to create a slim package with minimal surface area.
- AuthorizeAttribute
- IAuthorizationService

HaoK:
Splitting securityhelper into shared sources so MVC can remove dependency on Auth

https://github.com/aspnet/Mvc/pull/2792

HaoK:
@rynowak @davidfowl AuthZ doesn't bring in much so with AuthN dependency broken is this sufficient?

rynowak:
The extra reference that we're getting here is `Microsoft.AspNet.Http.Features`

Tratcher:
@rynowak It's expected that middleware reference Microsoft.AspNet.Http.Features. Microsoft.AspNet.Http.Abstractions depends on Features anyways.

PR #344, opened 2015-07-08
Switch to using shared SecurityHelper

HaoK:
Depends on https://github.com/aspnet/Common/pull/27

cc @Tratcher


src/Microsoft.AspNet.Authentication/SecurityHelper.cs
```diff
@@ -21,16 +19,7 @@ public static class SecurityHelper
...
-            // New principal identities go first
-            newPrincipal.AddIdentities(additionalPrincipal.Identities);
-
-            // Then add any existing non empty or authenticated identities
-            if (existingPrincipal != null)
-            {
-                newPrincipal.AddIdentities(existingPrincipal.Identities.Where(i => i.IsAuthenticated || i.Claims.Count() > 0));
-            }
-            return newPrincipal;
+            return Microsoft.Framework.Internal.SecurityHelper.MergeUserPrincipal(existingPrincipal, additionalPrincipal);
```
Tratcher: You should be able to just delete this file and use the other one directly.

HaoK: Yeah but this was a public helper for other middleware authors to use before, I mean I can certainly nuke it if we want to just rip that utility out...

Tratcher: MVC should do whatever we ask other middleware to do. We don't want both the shared sources and this weird public version.  Should we consider moving this into a lower package instead? Http.Extensions?

HaoK: That's what I suggested to @davidfowl but he wanted shared sources so we could remove the dependency from MVC to AuthN, see also https://github.com/aspnet/Security/issues/339

Tratcher: Then remove SecurityHelper from this project. It's only called in one place in the base class anyways.

HaoK: Ok, so the scenario we are cutting is making it easy for other auth middlewares to merge principals 'correctly' via this helper.  @blowdart ok if we cut this?  Its definitely cleaner not having a public shim around our internal static :)

blowdart: What's the risk in removing it? Middleware identities get merged incorrectly?

Tratcher: We have yet to see a middleware that needed to do the merge itself, the base class does it.

HaoK: Pretty much fun stuff like your identity no longer being Authenticated since we do the whole promoting/reordering thing here, and if we do muck with the principal selector it would be in this method as well.

blowdart: If the base class does it, kill it.

blowdart: Wait, so which is it? Is it done elsewhere in the base class, or must middleware specifically call this? (In which case shouldn't we be wrapping?)

HaoK: AutomaticAuthentication = true does this in the base Authenticationhandler today.

blowdart: And that's not confusing to anyone :) You know if we could detect if there was more than one automatic it'd go a long way to stopping pit of failures.

HaoK: Well there's nothing wrong with having more than one automatic middleware on in principal

blowdart: In practice though things go wonky. But sure, if there's another way to call this, without having to rely on it being called manually I lean towards killing.

HaoK:
Nuked everything

efc35302e84f8e564564249d45916ae5c1538d07

Tratcher:
:shipit:

PR #346, opened 2015-07-08
Make AddAuthorization() idempotent

rynowak:
Found this issue which looking into making AddMvc() idempotent. You'll end
up with multiple pass-through handlers registered if two components call
AddAuthorization(). This is very possible to happen if used two frameworks
in the same app.


HaoK:
Ah yes, good catch thanks! :shipit:

Eilon:
@rynowak is this merged? Or waiting for anything?

rynowak:
Looks like not merged, I'll take care of it.

rynowak:
831785fe9fc4e3fd00889ce59b38c68a6c021a9a

Issue #27, opened 2014-07-22
Revisit Cookie's custom AJAX redirect


Tratcher:
The cookies middleware has custom redirect logic for AJAX requests, where it sends a 200 and embeds the real status code and location header inside a custom response header.  Everybody I've asked says this was done to support the first version of templates, but I can't find any reference to it there. The templates have since been modified. Do we still need this strange custom behavior?
RE: https://katanaproject.codeplex.com/workitem/264

Eilon:
Remove this behavior in Cookies.DefaultBehavior.

HaoK:
cd237445ec49f7def7978db5a6e1a44e691450da

PR #352, opened 2015-07-09
Remove special cookie ajax redirect behavior

HaoK:
We did have a test for this!

cc @Tratcher


HaoK:
Fixes https://github.com/aspnet/Security/issues/27

Tratcher:
:shipit:

Issue #278, opened 2015-05-27
OpenIdConnect needs additional validation of the response.


brentschmaltz:
Currently OIDC handler needs to consider the entire message when validating if a message is protocol compliant. Several areas need addressing. 

Some are easy: id_token + access_token arrives, the at_hash needs to be evaluated, similar to c_hash. The nonce claim is required in the 'code' and 'token' hybrid flows.

Some are more subtle (from spec): 
If the auth_time Claim was requested, either through a specific request for this Claim or by using the max_age parameter, the Client SHOULD check the auth_time Claim value and request re-authentication if it determines too much time has elapsed since the last End-User authentication. 

If the JWT alg Header Parameter uses a MAC based algorithm such as HS256, HS384, or HS512, the octets of the UTF-8 representation of the client_secret corresponding to the client_id contained in the aud (audience) Claim are used as the key to validate the signature. For MAC based algorithms, the behavior is unspecified if the aud is multi-valued or if an azp value is present that is different than the aud value. 

etc. some we should support, others we may want to leave the validation to the Client since the spec calls out that many times the behavior is Client specific.

At a minimum I suggest:
1. change the API to OpenIdConnectProtocolValidator to take the OpenIdConnectMessage and TBD info.
2. we introduce a delegate, possibly on OIDCPValidator, that can be called for additional validation so that users don't have to derive for custom processing.

Eilon:
@blowdart by when should we get this done?

blowdart:
Before the next beta

Eilon:
@blowdart got it.

@Tratcher putting this on your beta7 plate.

PinpointTownes:
> If the JWT alg Header Parameter uses a MAC based algorithm such as HS256, HS384, or HS512, the octets of the UTF-8 representation of the client_secret corresponding to the client_id contained in the aud (audience) Claim are used as the key to validate the signature. For MAC based algorithms, the behavior is unspecified if the aud is multi-valued or if an azp value is present that is different than the aud value.

I guess this part depends on https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/53?

The rest sounds like a job for `OpenIdConnectProtocolValidator` (specially `at_hash` validation), which is part of Wilson. I'd be curious to know what needs to be added/changed in the OIDC middleware itself.

Tratcher:
[...]

PR #353, opened 2015-07-10
Code flow sign in

tushargupta51:


tushargupta51:
I will rebase the PR before merging to remove those "merge" commits.

tushargupta51:
@Tratcher @HaoK @brentschmaltz

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -514,6 +556,111 @@ public class OpenIdConnectAuthenticationHandler : AuthenticationHandler<OpenIdCo
...
+                GrantType = "authorization_code",
+                RedirectUri = redirectUri
+            };
+
+            var requestMessage = new HttpRequestMessage(HttpMethod.Post, _configuration.TokenEndpoint);
+            requestMessage.Content = new FormUrlEncodedContent(openIdMessage.Parameters);
+            var responseMessage = await Backchannel.SendAsync(requestMessage);
+            responseMessage.EnsureSuccessStatusCode();
+            var tokenResonse = await responseMessage.Content.ReadAsStringAsync();
+            var jsonTokenResponse = JObject.Parse(tokenResonse);
```
Tratcher: In OAuth we found that people wanted direct access to the jsonTokenResponse because sometimes there are additional parameters returned. Will this ever be true for OIDC? If so, should there be a notification at this point?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
```diff
@@ -125,7 +130,7 @@ public class OpenIdConnectAuthenticationMiddleware : AuthenticationMiddleware<Op
                     var httpClient = new HttpClient(ResolveHttpMessageHandler(Options));
                     httpClient.Timeout = Options.BackchannelTimeout;
                     httpClient.MaxResponseContentBufferSize = 1024 * 1024 * 10; // 10 MB
-                    Options.ConfigurationManager = new ConfigurationManager<OpenIdConnectConfiguration>(Options.MetadataAddress, httpClient);
+                    Options.ConfigurationManager = new ConfigurationManager<OpenIdConnectConfiguration>(Options.MetadataAddress, Backchannel);
```
Tratcher: You changed this to use the primary Backchannel, but you didn't remove the other httpClient instance.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -514,6 +556,111 @@ public class OpenIdConnectAuthenticationHandler : AuthenticationHandler<OpenIdCo
         }
 
         /// <summary>
+        /// Redeems the authorization code for tokens at the token endpoint
+        /// </summary>
+        /// <param name="authorizationCode">The authorization code to redeem.</param>
+        /// <param name="redirectUri">Uri that was passed in the request sent for the authorization code.</param>
+        /// <returns>OpenIdConnect message that has tokens inside it.</returns>
+        protected virtual async Task<OpenIdConnectMessage> RedeemAuthorizationCode(string authorizationCode, string redirectUri)
```
Tratcher: RedeemAuthorizationCodeAsync

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -302,6 +311,21 @@ public class OpenIdConnectAuthenticationHandler : AuthenticationHandler<OpenIdCo
...
+                // Redeeming authorization code for tokens
+                if (string.IsNullOrEmpty(message.IdToken) && !string.IsNullOrEmpty(message.Code))
+                {
+                    Logger.LogDebug(Resources.OIDCH_0038_Redeeming_Auth_Code, message.Code);
+
+                    var redirectUri = properties.Items.ContainsKey(OpenIdConnectAuthenticationDefaults.RedirectUriForCodePropertiesKey) ?
+                        properties.Items[OpenIdConnectAuthenticationDefaults.RedirectUriForCodePropertiesKey] : Options.RedirectUri;
+
+                    tokensRetrievedUsingCode = await RedeemAuthorizationCode(message.Code, redirectUri);
+                    if (tokensRetrievedUsingCode != null)
```
Tratcher: RedeemAuthorizationCode always returns non-null.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -514,6 +556,111 @@ public class OpenIdConnectAuthenticationHandler : AuthenticationHandler<OpenIdCo
...
+            if (subjectClaimType == null)
+            {
+                Logger.LogError(string.Format(CultureInfo.InvariantCulture, Resources.OIDCH_0041_Subject_Claim_Not_Found, identity.ToString()));
+                return ticket;
+            }
+
+            var userInfoSubjectClaimValue = user.Value<string>("sub");
+
+            // check if the sub claim matches
+            if (subjectClaimType.Properties[JwtSecurityTokenHandler.ShortClaimTypeProperty] != "sub" || userInfoSubjectClaimValue == null || !string.Equals(userInfoSubjectClaimValue, subjectClaimType.Value, StringComparison.Ordinal))
```
Tratcher: Wrap long lines.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -514,6 +556,111 @@ public class OpenIdConnectAuthenticationHandler : AuthenticationHandler<OpenIdCo
...
+                    claim.Properties[JwtSecurityTokenHandler.ShortClaimTypeProperty] : string.Empty;
+
+                // checking if claim in the identity (generated from id_token) has the same type as a claim retrieved from userinfo endpoint
+                JToken value;
+                var isClaimIncluded = user.TryGetValue(claim.Type, out value) ? true : user.TryGetValue(shortClaimTypeName, out value);
+
+                // if a same claim exists (matching both type and value) both in id_token identity and userinfo response, remove the json entry from the userinfo response
+                if (isClaimIncluded && claim.Value.Equals(value.ToString(), StringComparison.Ordinal))
+                {
+                    var isClaimRemoved = user.Remove(claim.Type) ? true : user.Remove(shortClaimTypeName);
```
Tratcher: Why are you setting a variable here?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: To be consistent with the OAuth2 handler, you could also use `ExchangeCodeAsync`.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -197,6 +196,14 @@ public class OpenIdConnectAuthenticationHandler : AuthenticationHandler<OpenIdCo
 
             message.State = Options.StateDataFormat.Protect(properties);
 
+            if (!string.IsNullOrEmpty(Options.RedirectUri))
```
PinpointTownes: Why not using `message.RedirectUri` instead of `Options.RedirectUri`?
One should be able to replace the `RedirectUri` in the `RedirectToIdentityProvider` notification.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -302,6 +311,21 @@ public class OpenIdConnectAuthenticationHandler : AuthenticationHandler<OpenIdCo
                     _configuration = await Options.ConfigurationManager.GetConfigurationAsync(Context.RequestAborted);
                 }
 
+                // Redeeming authorization code for tokens
+                if (string.IsNullOrEmpty(message.IdToken) && !string.IsNullOrEmpty(message.Code))
```
PinpointTownes: Hum, this check is odd. Why not retrieving an access/identity token from the token endpoint when an id_token has been received from the authorization endpoint? (`string.IsNullOrEmpty(message.IdToken)`)

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -514,6 +556,111 @@ public class OpenIdConnectAuthenticationHandler : AuthenticationHandler<OpenIdCo
...
+        {
+            string userInfoEndpoint = null;
+            if (_configuration != null)
+            {
+                userInfoEndpoint = _configuration.UserInfoEndpoint;
+            }
+
+            if (string.IsNullOrEmpty(userInfoEndpoint))
+            {
+                Logger.LogError("UserInfo endpoint is not set. Request to retrieve claims from userinfo endpoint cannot be completed.");
```
PinpointTownes: Is it really an error? The OIDC specs aren't really clear on this point, but AFAIK, the userinfo endpoint is not mandatory. Logging an error seems a bit abusive. A warning to indicate that the user set `GetClaimsFromUserInfoEndpoint` to true while the provider didn't support it may be more appropriate.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -514,6 +556,111 @@ public class OpenIdConnectAuthenticationHandler : AuthenticationHandler<OpenIdCo
...
+
+            var identity = (ClaimsIdentity)ticket.Principal.Identity;
+            var subjectClaimType = identity.FindFirst(ClaimTypes.NameIdentifier);
+            if (subjectClaimType == null)
+            {
+                Logger.LogError(string.Format(CultureInfo.InvariantCulture, Resources.OIDCH_0041_Subject_Claim_Not_Found, identity.ToString()));
+                return ticket;
+            }
+
+            var userInfoSubjectClaimValue = user.Value<string>("sub");
```
PinpointTownes: No constant for `sub` in Wilson?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -514,6 +556,111 @@ public class OpenIdConnectAuthenticationHandler : AuthenticationHandler<OpenIdCo
...
+
+            foreach (var claim in identity.Claims)
+            {
+                // If this claimType is mapped by the JwtSeurityTokenHandler, then this property will be set
+                var shortClaimTypeName = claim.Properties.ContainsKey(JwtSecurityTokenHandler.ShortClaimTypeProperty) ?
+                    claim.Properties[JwtSecurityTokenHandler.ShortClaimTypeProperty] : string.Empty;
+
+                // checking if claim in the identity (generated from id_token) has the same type as a claim retrieved from userinfo endpoint
+                JToken value;
+                var isClaimIncluded = user.TryGetValue(claim.Type, out value) ? true : user.TryGetValue(shortClaimTypeName, out value);
```
PinpointTownes: `user.TryGetValue(claim.Type, out value) || user.TryGetValue(shortClaimTypeName, out value)`?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -514,6 +556,111 @@ public class OpenIdConnectAuthenticationHandler : AuthenticationHandler<OpenIdCo
...
+                {
+                    var isClaimRemoved = user.Remove(claim.Type) ? true : user.Remove(shortClaimTypeName);
+                }
+            }
+
+            // adding remaining unique claims from userinfo endpoint to the identity
+            foreach (var pair in user)
+            {
+                JToken value;
+                var claimValue = user.TryGetValue(pair.Key, out value) ? value.ToString() : null;
```
PinpointTownes: I'm not a JSON.NET expert, but you should be able to use `user.Value<string>(pair.Key)`, that returns null when the property is missing.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
tushargupta51: Fixed.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: Good Catch. Removed that check.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: using constant instead of "sub" removed the need for subjectClaimType.Properties[JwtSecurityTokenHandler.ShortClaimTypeProperty] != "sub" condition, so the check is shorter now.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: I wanted to remove the shortClaimType if claim.Type did not succeed. Removed the variable and added if (!user.Remove(claim.Type)) { user.Remove(shortClaimTypeName); }

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: Good point. I will update it to look into message.RedirectUri. If that is not set then fall back to Options.RedirectUri. I wanted to set it after the notificaiton is called to make sure it is send in the message and user has not removed it, like we do with state.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: That is in our POR to use code in id_token + code response type too. Currenlty, we are redeeming it only for the sigin flow ("Code" response type)

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: Agreed. Changed it to Warning message.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: There is one. Updated.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: Wilson maps all short names to long names and preserve the shortname in ShortClaimTypeProperty. This is to check if a claim with either short or long name exists in the list of claims returned from the UI endpoint.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: I have used user.Value<string>(pair.Key) syntax in RedeemAUthorizationCode method. In this case, the value is of type JToken and I want to finally convert it to string. user.Value<string>(pair.Key) makes it easier.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: And is supporting token exchange with the hybrid flows planned?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: Actually, I was not suggesting a better syntax. It should do exactly the same thing.

tushargupta51: Ah, didn't notice the difference in the syntax in your comment. user.TryGetValue(claim.Type, out value) || user.TryGetValue(shortClaimTypeName, out value) (y)

tushargupta51:
I have added AuthorizationCodeRedeemedNotification and had to duplicate the AuthorizationCodeReceivedNotification. Because in case of "code" flow, calling the code received notification is of no use to user, it should be called before in the flow. We will be revisiting this flow as part of this issue: https://github.com/aspnet/Security/issues/278

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
brentschmaltz: Seems like a good idea.
Before or after it is parsed?

Tratcher: After, you want to give access to both the JObject and OpenIdConnectMessage.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -302,174 +305,99 @@ public class OpenIdConnectAuthenticationHandler : AuthenticationHandler<OpenIdCo
                     _configuration = await Options.ConfigurationManager.GetConfigurationAsync(Context.RequestAborted);
                 }
 
-                AuthenticationTicket ticket = null;
-                JwtSecurityToken jwt = null;
-
-                // OpenIdConnect protocol allows a Code to be received without the id_token
-                if (!string.IsNullOrEmpty(message.IdToken))
+                if (string.IsNullOrEmpty(message.IdToken) && !string.IsNullOrEmpty(message.Code))
```
Tratcher: This is getting complicated enough that I recommend extracting it to its own method. Same goes for the else case.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -609,6 +635,251 @@ private AuthenticationProperties GetPropertiesFromState(string state)
...
+                if (expires != DateTime.MinValue)
+                {
+                    ticket.Properties.ExpiresUtc = expires;
+                }
+            }
+
+            return ticket;
+        }
+
+        private async void ValidateOpenIdConnectProtocolAsync(JwtSecurityToken jwt, OpenIdConnectMessage message, bool ValidateCHash = true)
```
Tratcher: Never use `async void`. The callers need to `await` this call.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -302,174 +305,99 @@ public class OpenIdConnectAuthenticationHandler : AuthenticationHandler<OpenIdCo
...
+                        if (authorizationCodeReceivedNotification.HandledResponse)
+                        {
+                            return authorizationCodeReceivedNotification.AuthenticationTicket;
+                        }
+                        else if (authorizationCodeReceivedNotification.Skipped)
+                        {
+                            return null;
+                        }
                     }
                 }
```
Tratcher: is there an else scenario for when both token and code are missing? It seems like you'd at least have a debug log for that.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -514,6 +432,114 @@ public class OpenIdConnectAuthenticationHandler : AuthenticationHandler<OpenIdCo
...
+                Logger.LogWarning("UserInfo endpoint is not set. Request to retrieve claims from userinfo endpoint cannot be completed.");
+                return ticket;
+            }
+
+            var requestMessage = new HttpRequestMessage(HttpMethod.Get, userInfoEndpoint);
+            requestMessage.Headers.Authorization = new AuthenticationHeaderValue("Bearer", message.AccessToken);
+            var responseMessage = await Backchannel.SendAsync(requestMessage);
+            responseMessage.EnsureSuccessStatusCode();
+            var userInfoResponse = await responseMessage.Content.ReadAsStringAsync();
+            var user = JObject.Parse(userInfoResponse);
```
Tratcher: This is another potential extension point where developers want direct access to the JObject.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: Updated to return Task instead of void.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: It shouldn't really happen but having a debug log makes sense. I will add the else scenario.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: Does it make sense to include the JObject inside OIDCMessage parameters dictionary?

Tratcher: Well, if you move RunAuthorizationCodeRedeemedNotificationAsync inside of RedeemAuthorizationCode then you could just expose the JObject and OIDCMessage on that notification directly.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: Refactored.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: I think the notification should be outside of the redemption method. I have added a wrapper class OpenIdConnectTokenEndpointResponse for this.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: Let's handle this one separately as an issue since this require some modifications to existing SecurityTokenValidated notification. Please open an issue or I can open one if you want.

tushargupta51:
@Tratcher do the changes look good?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
Tratcher: Go ahead and open an issue to track this.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: > I think the notification should be outside of the redemption method.

Curious: why?

Tratcher:
Looks good. Rebase, squash, and force push back to your branch to update the PR for a final check.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: https://github.com/aspnet/Security/issues/358

tushargupta51:
@Tratcher done

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: They are 2 different extensibility points. RedeemAuthorizationCode would let users control how a code should be redeemed. RunAuthorizationCodeRedeemedNotificationAsync let users control what to do with the tokens, cahing access or refresh tokens if they want. Putting the notification inside the redemption method would make a user overriding RedeemAuthorizationCode ensuring that code redeemed notification is called too. I don't want users to think about notification if all they want to do is override redemption method.

Tratcher:
:shipit:

Issue #259, opened 2015-05-14
Have an UseCookieAuthentication overload that takes in instance of DataProtectionProvider


suhasj:
Based on the discussions to make the cookie compatibility between v4.5 and v5.0 apps, we need to have `UseCookieAuthentication` method take an instance of DataProtectionProvider which is shared among the apps.

Tratcher:
Or put it on the options.

HaoK:
+1 to options, I can do this together with the identity changes

PinpointTownes:
IMHO, the best option would be to provide an extension method around `CookieAuthenticationOptions`, that would internally set the ticket data format using a given data protector.

```
app.UseCookieAuthentication(options => {
    options.UseDataProtector(protector);
});
```

HaoK:
Well, today looks like there's the TicketDataFormat which takes the DataProtector in its ctor.  

So if we leave cookies alone as is, identity could simply configure things via

`options.TicketDataFormat = new TicketDataFormat(dataprotector);`

This doesn't look unreasonable, maybe we should just leave things be?

Tratcher:
We have a token binding feature coming that may require us to work directly with the DataProtection classes on a per request basis. In that case I think we'd want it on the options.

Eilon:
Related to https://github.com/aspnet/DataProtection/issues/85

HaoK:
As discussed in https://github.com/aspnet/Security/issues/363, what we have today with overriding the TicketDataFormat is enough for cookie sharing

PR #363, opened 2015-07-16
Support specifying DataProtector in CookieOptions

HaoK:
Fixes https://github.com/aspnet/Security/issues/259

cc @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationMiddleware.cs
+++ b/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationMiddleware.cs
@@ -32,8 +32,8 @@ namespace Microsoft.AspNet.Authentication.Cookies
             }
             if (Options.TicketDataFormat == null)
             {
-                var dataProtector = dataProtectionProvider.CreateProtector(
-                    typeof(CookieAuthenticationMiddleware).FullName, Options.AuthenticationScheme, "v2");
+                var dataProtector = Options.DataProtector ??
+                    dataProtectionProvider.CreateProtector(typeof(CookieAuthenticationMiddleware).FullName, Options.AuthenticationScheme, "v2");
                 Options.TicketDataFormat = new TicketDataFormat(dataProtector);
             }
             if (Options.CookieManager == null)
--- a/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationMiddleware.cs
+++ b/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationMiddleware.cs
@@ -32,8 +32,7 @@ namespace Microsoft.AspNet.Authentication.Cookies
             }
             if (Options.TicketDataFormat == null)
             {
-                var dataProtector = Options.DataProtector ??
-                    dataProtectionProvider.CreateProtector(typeof(CookieAuthenticationMiddleware).FullName, Options.AuthenticationScheme, "v2");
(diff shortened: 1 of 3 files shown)


Tratcher:
:shipit:

PinpointTownes:
Curious: why not providing a `options.UseDataProtector(protector)` extension to avoid having two disconnected properties (`TicketDataFormat` vs `DataProtector`)?

```csharp
public static CookieAuthenticationOptions UseDataProtector(this CookieAuthenticationOptions options, IDataProtector protector) {
    options.TicketDataFormat = new TicketDataFormat(protector);

    return options;
}

**Comment by Tratcher** [comment]
@PinpointTownes I asked him to hold onto it because I'm going to need it later for https://github.com/aspnet/Security/issues/263

**Comment by PinpointTownes** [comment]
@Tratcher I'm a total noob concerning this new TLS extension, but isn't this new token a per-client thing? I'm not sure to understand how being able to set a global data protector (i.e shared by the different clients) may be related to this feature.

**Comment by Tratcher** [comment]
@PinpointTownes IDataProtector lets you create a derived protector. If the client is providing key data then your encryption needs to take into account both the server keys and the client keys.

**Comment by PinpointTownes** [comment]
@Tratcher, so the cookies middleware will directly use `Options.DataProtector` instead of `Options.TicketDataFormat` but only when `Options.DataProtector` is explicitly set?
IMHO, it doesn't seem very intuitive.

Instead, have you considered refactoring `ISecureDataFormat.Protect`/`ISecureDataFormat.Unprotect` to take an additional parameter that could be used to flow this new TLS binding token?

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication/DataHandler/SecureDataFormat.cs

**Comment by PinpointTownes** [comment]
Something like that:

```csharp
public string Protect(TData data, string purpose)
{
    var protector = _protector;
    if (!string.IsNullOrEmpty(purpose))
    {
        protector = protector.CreateProtector(purpose);
    }

    var userData = _serializer.Serialize(data);
    var protectedData = protector.Protect(userData);
    var protectedText = _encoder.Encode(protectedData);
    return protectedText;
}
```

```csharp
public TData Unprotect(string protectedText, string purpose)
{
    try
    {
        var protector = _protector;
        if (!string.IsNullOrEmpty(purpose))
        {
            protector = protector.CreateProtector(purpose);
        }

        if (protectedText == null)
        {
            return default(TData);
        }

        var protectedData = _encoder.Decode(protectedText);
        if (protectedData == null)
        {
            return default(TData);
        }

        var userData = protector.Unprotect(protectedData);
        if (userData == null)
        {
            return default(TData);
        }

        var model = _serializer.Deserialize(userData);
        return model;
    }
    catch
    {
        // TODO trace exception, but do not leak other information
        return default(TData);
    }
}
```

Tratcher:
Hmm, it may come to that. We haven't worked through the details yet.

PinpointTownes:
In this case, I guess that this new `DataProtector` property won't be necessary :smile:

HaoK:
@Tratcher Where did we end up finally on this?  Can we just replace the data protector now?

Tratcher:
@PinpointTownes is probably right, we don't need to save the DP on the options, at least not yet.

HaoK:
Ok so now this PR basically only consists of adding a test to verify that we can share data protectors in two different TicketDataFormats.  We are happy with the existing TicketDataFormat extensibility correct?

Tratcher:
Yes, for now.

HaoK:
a3f0ee3330e936cad13b4391a1f7a13ff26764fd

PR #366, opened 2015-07-20
Makes FacebookAuthenticationHandler respect Options.UserInformationEndpoint

### LINKED ISSUE #365: Facebook.UserInformationEndpoint + access_token make Invalid Backchannel URL
When the back-channel URL graph call is constructed [here](https://github.com/aspnet/Security/blob/9bb8b6114697f14a42e747acd1d1a41e6eed5af1/src/Microsoft.AspNet.Authentication.Facebook/FacebookAuthenticationHandler.cs#L54) it is possible for `Options.UserInformationEndpoint` to already contain `?`. However, as-is currently `?access_token` is appended to the endpoint without checking if `?` already exists.

This can lead to an malformed URL API call to Facebook.

For example,
```
var fops = new FacebookAuthenticationOptions()
    {
        AppId = "appID",
        AppSecret = "Secret",
        Scope = {"public_profile", "email"},
        Provider = new FacebookAuthenticationProvider
            {
                OnAuthenticated = async (ctx) =>
                    {
                        ctx.Identity.AddClaim(new Claim("timezone", ctx.User['timezone'].ToString()));
                    }
            },
        UserInformationEndpoint = "https://graph.facebook.com/me?fields=email,verified,currency,timezone,picture"
    };
```
Would result in a malformed URL when making a request to Facebook because the URL would contain two `?`.

bchavez:
Fixes #365

### DIFF
--- a/src/Microsoft.AspNet.Authentication.Facebook/FacebookAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.Facebook/FacebookAuthenticationHandler.cs
@@ -51,10 +51,10 @@ namespace Microsoft.AspNet.Authentication.Facebook
 
         protected override async Task<AuthenticationTicket> CreateTicketAsync(ClaimsIdentity identity, AuthenticationProperties properties, OAuthTokenResponse tokens)
         {
-            var endpoint = Options.UserInformationEndpoint + "?access_token=" + UrlEncoder.UrlEncode(tokens.AccessToken);
+            var endpoint = QueryHelpers.AddQueryString(Options.UserInformationEndpoint, "access_token", tokens.AccessToken);
             if (Options.SendAppSecretProof)
             {
-                endpoint += "&appsecret_proof=" + GenerateAppSecretProof(tokens.AccessToken);
+                endpoint = QueryHelpers.AddQueryString(endpoint, "appsecret_proof", GenerateAppSecretProof(tokens.AccessToken));
             }
 
             var response = await Backchannel.GetAsync(endpoint, Context.RequestAborted);
--- a/test/Microsoft.AspNet.Authentication.Test/Facebook/FacebookMiddlewareTests.cs
+++ b/test/Microsoft.AspNet.Authentication.Test/Facebook/FacebookMiddlewareTests.cs
@@ -3,15 +3,21 @@
 
 using System;
 using System.Net;
+using System.Net.Http;
+using System.Collections.Generic;
+using System.Text;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Authentication.OAuth;
 using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.DataProtection;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Authentication;
 using Microsoft.AspNet.TestHost;
 using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.WebEncoders;
 using Shouldly;
+using Newtonsoft.Json;
 using Xunit;
 
 namespace Microsoft.AspNet.Authentication.Facebook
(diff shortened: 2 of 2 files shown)


HaoK:
Do you think you could add a test verifying your fix to this class?

https://github.com/aspnet/Security/blob/dev/test/Microsoft.AspNet.Authentication.Test/Facebook/FacebookMiddlewareTests.cs

bchavez:
Sure, no problem, be back in a few. :+1:

src/Microsoft.AspNet.Authentication.Facebook/FacebookAuthenticationHandler.cs
```diff
@@ -51,7 +51,12 @@ public FacebookAuthenticationHandler(HttpClient httpClient)
 
         protected override async Task<AuthenticationTicket> CreateTicketAsync(ClaimsIdentity identity, AuthenticationProperties properties, OAuthTokenResponse tokens)
         {
-            var endpoint = Options.UserInformationEndpoint + "?access_token=" + UrlEncoder.UrlEncode(tokens.AccessToken);
+            var endpoint = Options.UserInformationEndpoint;
+            var accessToken = "access_token=" + UrlEncoder.UrlEncode(tokens.AccessToken);            
+                        
+            endpoint += endpoint.Contains("?") ? "&" : "?";
```
PinpointTownes: Note: you can also use `QueryHelpers.AddQueryString` :family:

bchavez: @PinpointTownes Yeah, I was thinking about using a helper to build the query string, but wasn't sure about @HaoK 's thoughts on the subject since the other Auth providers don't do anything like that. I was exercising some caution not to move so much code. @HaoK , would you prefer that I use `QueryHelpers`?

HaoK: Yeah please use QueryHelpers, the manual query string manipulation stood out when I saw the change as well.

bchavez:
@HaoK @PinpointTownes Should be ok now. My new Unit Test is passing.

src/Microsoft.AspNet.Authentication.Facebook/FacebookAuthenticationHandler.cs
```diff
@@ -51,10 +51,10 @@ public FacebookAuthenticationHandler(HttpClient httpClient)
 
         protected override async Task<AuthenticationTicket> CreateTicketAsync(ClaimsIdentity identity, AuthenticationProperties properties, OAuthTokenResponse tokens)
         {
-            var endpoint = Options.UserInformationEndpoint + "?access_token=" + UrlEncoder.UrlEncode(tokens.AccessToken);
+            var endpoint = QueryHelpers.AddQueryString(Options.UserInformationEndpoint, "access_token", UrlEncoder.UrlEncode(tokens.AccessToken));
```
PinpointTownes: When using `QueryHelpers.AddQueryString`, `key` and `value` are automatically url-encoded for you.
Pre-encoding them will result in a terrible double encoding. It's probably not a big deal since access tokens issued by Facebook only use alphanumerical chars, but it's not a guarantee and it may change in the future.

https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.WebUtilities/QueryHelpers.cs#L59-L62

bchavez: @PinpointTownes Ah, thank you, I totally missed that. I've updated my branch accordingly. Thanks so much.

bchavez:
1f8368e2efeada70c00d24bbb104b1dde4173c44 now voids double URL encoding with `QueryHelpers`. Thanks @PinpointTownes .

PinpointTownes:
You're welcome, @bchavez! :smile: 

On a side note, it seems that Facebook now supports the `Authorization` header with the `Bearer` scheme, just like the other OAuth2 providers.I just tested it and it works with the latest graph API.

Sadly, I can't find any reference :cry: 
That's really unfortunate, since flowing the access token in the headers is safer (intermediate proxies won't cache it)... but without an official reference, it's hard to know if it will be supported in the long term.

bchavez:
@PinpointTownes Sounds great, hoping they publish a reference soon rather than passing `access_token` around the URL string.

Slightly off topic, but I tried to run the unit test again just to make sure everything was okay. But ran into a problem. Everything was working prior to opening the project today. After opening this project in VS2015 RTM, waited for everything to load, then saw the "package manager" console go crazy like it was doing some updates to `dnx` automatically. I'd prefer VS not do this without permission because I suspect the recent auto-update actions by VS broke my unit testing environment with this project.

```
------ Discover test started ------
------ Test started: Project: Microsoft.AspNet.Authentication.Test ------
Starting  Microsoft.Framework.TestHost [C:\Users\Profile\.dnx\runtimes\dnx-clr-win-x86.1.0.0-beta5\bin\dnx.exe --appbase "C:\Code\Projects\Public\Security\test\Microsoft.AspNet.Authentication.Test" Microsoft.Framework.ApplicationHost --port 47363 Microsoft.Framework.TestHost --port 47424]
System.MissingMethodException: Method not found: 'Boolean Microsoft.Framework.Runtime.Project.TryGetProject(System.String, Microsoft.Framework.Runtime.Project ByRef, System.Collections.Generic.ICollection`1<Microsoft.Framework.Runtime.DiagnosticMessage>)'.
   at Microsoft.Framework.TestHost.Program.<>c__DisplayClass2_0.<<Main>b__0>d.MoveNext()
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.Start[TStateMachine](TStateMachine& stateMachine)
   at Microsoft.Framework.TestHost.Program.<>c__DisplayClass2_0.<Main>b__0()
   at Microsoft.Framework.Runtime.Common.CommandLine.CommandLineApplication.<>c__DisplayClass68_0.<OnExecute>b__0()
   at Microsoft.Framework.Runtime.Common.CommandLine.CommandLineApplication.Execute(String[] args)
   at Microsoft.Framework.TestHost.Program.Main(String[] args)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.Framework.Runtime.Common.EntryPointExecutor.Execute(Assembly assembly, String[] args, IServiceProvider serviceProvider)
   at Microsoft.Framework.ApplicationHost.Program.ExecuteMain(DefaultHost host, String applicationName, String[] args)
   at Microsoft.Framework.ApplicationHost.Program.Main(String[] args)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.Framework.Runtime.Common.EntryPointExecutor.Execute(Assembly assembly, String[] args, IServiceProvider serviceProvider)
   at dnx.host.Bootstrapper.RunAsync(List`1 args, IRuntimeEnvironment env)
   at dnx.host.RuntimeBootstrapper.ExecuteAsync(String[] args)
   at dnx.host.RuntimeBootstrapper.Execute(String[] args)

Unable to start Microsoft.Framework.TestHost
========== Discover test finished: 0 found (0:00:03.4269352) ==========
```
:(

PinpointTownes:
I guess that's due to the fact you're using a beta5 DNX runtime while contributing on beta7 packages :smile: 
You can upgrade to a recent DNX version but sadly, there's currently a bug that prevents the latest one from working correctly. You can either wait for the next nightly build or use the volatile packages:

```
SET DNX_FEED=https://www.myget.org/F/aspnetvolatile/api/v2
dnvm upgrade -r clr
```

bchavez:
I see. Thanks @PinpointTownes. My pull-request for Facebook here will eventually propagate to ASP.NET 4 too right? Ultimately, v4 is what I'm looking to fix at the moment. Thanks for all your help. :+1:

PinpointTownes:
If by "ASP.NET 4" you mean **Katana** (aka `Microsoft.Owin`), nope, changes introduced in ASP.NET 5 are never backported. Actually, Katana is not under development anymore :smile:

bchavez:
Oh noes, so is no way to fix this Nuget package? 

https://www.nuget.org/packages/Microsoft.Owin.Security.Facebook/

:( :( :(

PinpointTownes:
Absolutely no way, I'm afraid. I guess the only reason the ASP.NET team would update a Katana package would be to fix a major security flaw.

Tratcher:
:shipit:

Tratcher:
Merged.

PinpointTownes:
Looks like the `Authorization` header support is not new (https://jira.spring.io/browse/SOCIAL-96), but the official page announcing it has since disappeared: http://developers.facebook.com/docs/authentication/oauth2_updates

:facepunch:

PR #370, opened 2015-07-21
Remove usage of IOptions for auth 

HaoK:
- SharedAuthenticationOptions are still used
- Also add sugar to configure them in AddAuthentication
- Nuked all of the ConfigureXyzAuthentication methods
- Also switched all the IsNullOrWhitespaces => IsNullOrEmpty

cc @Tratcher @divega


davidfowl:
@HaoK What's the deal with this?

HaoK:
Old iteration nuking

PR #377, opened 2015-07-27
Add default policy support / React to options changes

HaoK:
New default policy added to AuthorizationOptions

```
            options.DefaultPolicy = new AuthorizationPolicyBuilder("default").RequireClaim("default").Build();
```

The default policy is used only when combining default AuthorizeAttributes (no policy / role specified)

cc @divega @blowdart


HaoK:
cc @Tratcher I accidentaly applied my Options/cleanup changes to this branch too, Can you review as well?

I nuked the AuthenticationHandler base that didn't take TOptions since it wasn't used.

I changed CookieAuthenticationOptions to implement IOptions and added a UseCookie overload that can be used via an instance which will ignore the shared cookie options in DI.

We could do something similar for the social auth middleware if we wanted.

global.json
```diff
@@ -1,3 +1,3 @@
 {
-    "projects": ["src"]
+    "projects": ["src", "C:\\Github\\Options\\src"]
```
Tratcher: Revert

src/Microsoft.AspNet.Authentication.Cookies/CookieAppBuilderExtensions.cs
```diff
@@ -20,13 +20,10 @@ public static class CookieAppBuilderExtensions
         /// <param name="configureOptions">Used to configure the options for the middleware</param>
         /// <param name="optionsName">The name of the options class that controls the middleware behavior, null will use the default options</param>
         /// <returns>The original app parameter</returns>
-        public static IApplicationBuilder UseCookieAuthentication([NotNull] this IApplicationBuilder app, Action<CookieAuthenticationOptions> configureOptions = null, string optionsName = "")
+        public static IApplicationBuilder UseCookieAuthentication([NotNull] this IApplicationBuilder app, Action<CookieAuthenticationOptions> configureOptions = null)
```
Tratcher: Didn't you say you added an overload that took a CookieAuthenticationOptions directly?

src/Microsoft.AspNet.Authentication.Cookies/CookieAppBuilderExtensions.cs
```diff
@@ -20,13 +20,10 @@ public static class CookieAppBuilderExtensions
...
         {
             return app.UseMiddleware<CookieAuthenticationMiddleware>(
-                new ConfigureOptions<CookieAuthenticationOptions>(configureOptions ?? (o => { }))
-                {
-                    Name = optionsName
-                });
+                new ConfigureOptions<CookieAuthenticationOptions>(configureOptions ?? (o => { })));
```
PinpointTownes: Too bad your newest design doesn't allow removing these ugly `ConfigureOptions`... :cry:

HaoK: I know :(  Need it to preserve the lamda override...  Its kind of gross I agree

src/Microsoft.AspNet.Authentication.Cookies/CookieAppBuilderExtensions.cs
HaoK: Yeah I need to merge those changes back in from the other dead PR... will update soon

HaoK: Updated PR, added overload taking IOptions directly, and also added 2 tests verifying that the options are not shared

HaoK:
Merged changes from https://github.com/aspnet/Security/pull/370

Also pulled over replacing IsStringOrWhiteSpace => IsStringOrEmpty
and sugar for configuring sharedAuthenticationOptions inline with AddAuthentication

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationOptions.cs
```diff
@@ -152,5 +153,13 @@ public string CookieName
         /// to the client. This can be used to mitigate potential problems with very large identities.
         /// </summary>
         public IAuthenticationSessionStore SessionStore { get; set; }
+
+        CookieAuthenticationOptions IOptions<CookieAuthenticationOptions>.Value
```
PinpointTownes: This pattern is so weird :gun:

HaoK: Yeah its a bit of a gimmick to get UseMiddleware to bypass the arguments.

PinpointTownes: But why not providing a wrapper implementing `IOptions<T>`? 

    public class Options<T> : IOptions<T> where T : class, new() {
        public Options(T value) {
            Value = value;
        }

        public T Value { get; }
    }

HaoK: I tried, no good, abstract base class messes with all the inheritance chains.  It was really messy and ended up adding `AuthenticationOptions<TOptions>` : `Options<TOptions>` and it exploded into all the base options...  My conclusion NOT worth it

HaoK: Cookies is the only one that NEEDS this, the social auths mostly will only be 1 per app, so they don't need the extra overload or gimmick for the options

PinpointTownes: Heh, I'm not suggesting creating a base class (note that it's not abstract BTW) but a wrapper that you could use to convert a `TOptions` instance to `IOptions<TOptions>`:

    public static IApplicationBuilder UseCookieAuthentication([NotNull] this IApplicationBuilder app, CookieAuthenticationOptions options)
    {
        return app.UseMiddleware<CookieAuthenticationMiddleware>(
            new Options<CookieAuthenticationOptions>(options),
            new ConfigureOptions<CookieAuthenticationOptions>(o => { }));
    }

HaoK: Oh, well we already have precedence for this weird options implementing IOptions of itself already, so that's why I just went with that... I'm not sure its worth another wrapper, ConfigureOptions is already a plain wrapper now :/

PinpointTownes: > Oh, well we already have precedence for this weird options implementing IOptions of itself already, so that's why I just went with that...

Having to implement `IOptions<>` in every options class is a real PITA and not really DRY IMHO, since you end up duplicating the same code again and again (even in the Caching repo!).

> I'm not sure its worth another wrapper, ConfigureOptions is already a plain wrapper now :/

You could see it as `IOptions<>`'s default implementation :smile:

PinpointTownes: > Cookies is the only one that NEEDS this, the social auths mostly will only be 1 per app, so they don't need the extra overload or gimmick for the options

And OAuth2? And OIDC?

HaoK: So I don't disagree, its just right now I'm only adding the overload to UseCookie.  When/if the time comes that we add this overload to more places, we can revisit this again.  For now its opt in and we can make this better as part of the larger sweep.  I don't want to add the Options class just for Cookies, I had it in my local changes, but reverted it as it wasn't used anywhere except for cookies yet...

HaoK: This discussion actually belongs as part of the Options repo not security anyways... @divega we should discuss what we want to do for the consistency sweep

src/Microsoft.AspNet.Authentication.Cookies/CookieAppBuilderExtensions.cs
```diff
@@ -20,13 +20,24 @@ public static class CookieAppBuilderExtensions
...
         }
+
+        /// <summary>
+        /// Adds a cookie-based authentication middleware to your web application pipeline.
+        /// </summary>
+        /// <param name="app">The IApplicationBuilder passed to your configuration method</param>
+        /// <param name="configureOptions">Used to configure the options for the middleware</param>
+        /// <param name="optionsName">The name of the options class that controls the middleware behavior, null will use the default options</param>
+        /// <returns>The original app parameter</returns>
+        public static IApplicationBuilder UseCookieAuthentication([NotNull] this IApplicationBuilder app, IOptions<CookieAuthenticationOptions> options)
```
Tratcher: Why `IOptions<CookieAuthenticationOptions>` and not just `CookieAuthenticationOptions`?

src/Microsoft.AspNet.Authentication.Cookies/CookieAppBuilderExtensions.cs
```diff
@@ -20,13 +20,24 @@ public static class CookieAppBuilderExtensions
...
-                });
+                new ConfigureOptions<CookieAuthenticationOptions>(configureOptions ?? (o => { })));
         }
+
+        /// <summary>
+        /// Adds a cookie-based authentication middleware to your web application pipeline.
+        /// </summary>
+        /// <param name="app">The IApplicationBuilder passed to your configuration method</param>
+        /// <param name="configureOptions">Used to configure the options for the middleware</param>
+        /// <param name="optionsName">The name of the options class that controls the middleware behavior, null will use the default options</param>
```
Tratcher: param comment cleanup

src/Microsoft.AspNet.Authentication.Cookies/CookieAppBuilderExtensions.cs
HaoK: The budding consensus was that we should use IOptions everywhere, and just make it easier to opt out for the special cases where you want instances (like cookies).  see https://github.com/aspnet/Options/issues/78#issuecomment-127441235

Tratcher: I understand using ` IOptions<CookieAuthenticationOptions>` in the CookieAuthenticationMiddleware constructor, but why use it here in this extension method? What use do you get from IOptions here?

HaoK: Oh, yeah the previous incarnation of this used the concrete options.  I think the IOptions gives us slightly more flexibility.  But I'm not hardset on this, this pattern was used in a lot of other repos (for testing I think mostly)

Tratcher: Hmm, could you have two extensions? One for  `IOptions<CookieAuthenticationOptions>` and one for  `CookieAuthenticationOptions`? It would make the intellisense clearer.

HaoK: Meh so many overloads, I think we should just pick one or the other, I'm not opposed to just taking the concrete cookie options.  @divega thoughts on the overall pattern?

HaoK: One thing to note, this has some bearing on whether we wrap the CookieOptions inside or not
Like what @PinpointTownes suggested:
```
public static IApplicationBuilder UseCookieAuthentication([NotNull] this IApplicationBuilder app, CookieAuthenticationOptions options)
{
    return app.UseMiddleware<CookieAuthenticationMiddleware>(
        new Options<CookieAuthenticationOptions>(options),
        new ConfigureOptions<CookieAuthenticationOptions>(o => { }));
}
```

Tratcher: But you wouldn't need to wrap it if it already implements `IOptions<T>`.

HaoK: Yeah but the discussion was whether we want to set the pattern where our options classes implement it like this PR has, or if we wrap instead.

HaoK: Pros for the wrapper are it keeps the options POCO clean without the super hacky IOptions return this which just stands out as a bit of wtf

Tratcher: Yeah, it was a hack I did in MemoryCache because you have to call the constructor directly, and we don't support multiple constructors in DI. Here we can add better sugar.

PinpointTownes:
> Small change in plans, instead of removing usage of IOptions, the new pattern for options is simply to make each options instance implement IOptions of itself, so they can be directly passed in, which will bypass whatever shared options were in DI. Still working through the details of exactly how the new UseXyzAuth methods will be most likely:

```csharp
UseXyzAuth(options => options.Foo = "whatever") // same as today, action tweaks options from DI
UseXyzAuth(new XyzOptions() { options.Foo = "whatever" }) // will use only this instance, ignores DI options
```

Naive question: if we configure - as suggested - some global options in `ConfigureServices` and add a middleware in `Configure` that overrides the global options, won't it also override the options used by the next middleware in the pipeline, since the options instance is a singleton?

```csharp
public void ConfigureServices(IServiceCollection services) {
    services.ConfigureCookieAuthentication(options => {
        options.CookieName = "Global_cookie_name";
    });
}

public void Configure(IApplicationBuilder app) {
    // This middleware will use Local_cookie_name as the cookie name.
    app.UseCookieAuthentication(options => {
        options.CookieName = "Local_cookie_name";
    });

    // This one should use Global_cookie_name.
    app.UseCookieAuthentication();
}
```

HaoK:
Doesn't override so much as each updates the shared options, so yes the next middleware will 'see it'.  Really want to use the instanced options for situations like that (i.e. cookies)

HaoK:
Basically singleton shared options are exactly that, its a bit weird but everyone who is using that shared instance will see any updates to it.  Not the best thing to use when placing multiples of the same middleware in the pipeline (should use the non shared overloads for that via newing up a stand alone options for each middleware)  I still need to finish updating identity to see what that looks like...  its not pretty either :/

PinpointTownes:
Named options had at least one merit then :laughing: 

I guess it would be better to avoid offering a way to "override" the options from the `app.Use*Authentication` if it has such unpredictable side-effects.

The pattern I had suggested in my prototype sounds more robust (https://github.com/PinpointTownes/Security/commit/26b511f3d294f4b9545e1fbc57e4ef5226672118):

> Basically, `app.Use*Authentication` extensions have 2 versions: one that relies on the options defined by `ConfigureServices` *and is defined by `authenticationScheme`* (the current vNext solution), and another one that directly takes the options instance and wraps it in a `Options` wrapper instead of using the options defined by `ConfigureServices` (Katana's approach).

HaoK:
yeah we'll need to figure out something that works consistently everywhere

HaoK:
Replaced by https://github.com/aspnet/Security/pull/425

PR #385, opened 2015-08-06
Microsoft.AspNet.Authentication.Test -> Microsoft.AspNet.Authentication.Tests & Microsoft.AspNet.Authorization.Test -> Microsoft.AspNet.Authorization.Tests

hishamco:
According to the guidelines https://github.com/aspnet/Home/wiki/Engineering-guidelines#solution-and-project-folder-structure-and-naming, we should rename the test project to ```Microsoft.AspNet.Authentication.Tests``` instead of ```Microsoft.AspNet.Authentication.Test``` and ```Microsoft.AspNet.Authorization.Tests``` instead of ```Microsoft.AspNet.Authorization.Test```


Eilon:
See https://github.com/aspnet/EventNotification/pull/12#issuecomment-128427434.

Issue #98, opened 2014-11-19
OAuthAuthentication* naming concerns


Tratcher:
The `UseOAuthAuthentication` naming may be misleading because by default all it does is ask the user to _authorize_ your app to access their remote service (e.g. Github). Derived classes or extenders may choose to take the resulting access token and use that to _authenticate_ by retrieving user information.

The `GetUserInformationAsync` notification may be more generally named as `AccessTokenReceived`

PR #391, opened 2015-08-07
Rename OAuthAuthentication* to OAuth* and Authenticated to AccessTokenReceived.

Tratcher:
#98
@HaoK @PinpointTownes @Eilon @brockallen

Note I did not rename the Handler, Middleware, or Options classes because A) the type names are not referenced directly in user code, and B) I didn't want to confuse the inheritance heirarchy for derived classes: FacebookAuthenticationOptions -> OAuthAuthenticationOptions -> AuthenticationOptions

I also cleaned out some obsolete Context classes.


PinpointTownes:
Semantically, this naming change is totally appropriate: OAuth2 is an authorization protocol (though we use it for pseudo-authentication).

Technically, I'm not sure it's really pertinent: `OAuthAuthenticationMiddleware` is an authentication middleware, no matter what we think. It's derived from `AuthenticationMiddleware`, it calls `SignInAsync` during the callback process and the user is considered as authenticated after a complete "authorization flow", even if you don't provide an explicit `AccessTokenReceived` notification. 

IMHO, removing `Authentication` from the extension name will just make things less clear.

test/Microsoft.AspNet.Authentication.Test/Facebook/FacebookMiddlewareTests.cs
```diff
@@ -40,7 +40,7 @@ public class FacebookMiddlewareTests
                     {
                         options.AppId = "Test App Id";
                         options.AppSecret = "Test App Secret";
-                        options.Notifications = new OAuthAuthenticationNotifications
+                        options.Notifications = new OAuth.OAuthNotifications
```
PinpointTownes: Hum, why do you need `OAuth.` here?

HaoK:
Or Maybe we should shorten everything to auth?

UseGoogleAuth
UseFacebookAuth

PinpointTownes:
Or **app.UseOAuthAuthenrization** :trollface:

HaoK:
:shipit:

Tratcher:
So just close this and leave everything as is?

Tratcher:
Aborted

Issue #295, opened 2015-06-22
Support POST OpenID Connect authentication requests


PinpointTownes:
Supporting POST authentication/authorization requests is not mandatory for clients, but would be a nice addition to support heavy OIDC payloads. 

> Authorization Servers MUST support the use of the HTTP GET and POST methods defined in RFC 2616 [RFC2616] at the Authorization Endpoint. Clients MAY use the HTTP GET or POST methods to send the Authorization Request to the Authorization Server. If using the HTTP GET method, the request parameters are serialized using URI Query String Serialization, per Section 13.1. If using the HTTP POST method, the request parameters are serialized using Form Serialization, per Section 13.2.

http://openid.net/specs/openid-connect-core-1_0.html

Implementing this thing will require much carefulness to make sure we don't create deadlocks on critical async/sync paths (@Tratcher already knows everything about this issue, but you can take a look at this comment for more information: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/blob/vNext/src/AspNet.Security.OpenIdConnect.Server/OpenIdConnectServerHandler.cs#L365-L377)

/cc @brentschmaltz @Eilon @Tratcher

Tratcher:
Right now we use redirects to send the client over to the auth server (as a GET). What would we send to the client that would result in them doing a POST?

PinpointTownes:
An auto-post HTML form pointing to the authorization endpoint, similar to the one an OIDC server returns for a **response_mode=form_post** response.

Tratcher:
Ah, yes that's a challenge.  This might be easier with @HaoK's new prototypes where Challenge takes effect immediately rather than during OnResponseStarted, it could write the body directly. This would definitely have to be opt-in though.

Eilon:
Marking as up-for-grabs in case anyone wants to do it.

PinpointTownes:
@Eilon I'm rather busy these days, but I'm your man if nobody wants to do it :smile: 
That said, I'll probably need @brentschmaltz's help to add the appropriate tests (the logging testing is a bit cryptic for me).

@brentschmaltz out of curiosity, does AAD support POST authorization requests?

Eilon:
@PinpointTownes I'm out of town all next week but maybe we could sync up the week after that to see how this could be arranged.

PinpointTownes:
@Eilon I guess you've found nobody to implement it? :cry:

Eilon:
[...]

PR #392, opened 2015-08-07
Support POST OpenID Connect authentication/logout requests

PinpointTownes:
Fixes https://github.com/aspnet/Security/issues/295.

@Eilon as promised in my last mail, here are the first bits of the new POST authentication/authorization requests support for OIDC.

/cc @Eilon @HaoK @Tratcher

### DIFF
--- a/samples/OpenIdConnectSample/Startup.cs
+++ b/samples/OpenIdConnectSample/Startup.cs
@@ -31,11 +31,11 @@ namespace OpenIdConnectSample
             });
 
             app.UseOpenIdConnectAuthentication(options =>
-                {
-                    options.ClientId = "fe78e0b4-6fe7-47e6-812c-fb75cee266a4";
-                    options.Authority = "https://login.windows.net/cyrano.onmicrosoft.com";
-                    options.RedirectUri = "http://localhost:42023";
-                });
+            {
+                options.ClientId = "fe78e0b4-6fe7-47e6-812c-fb75cee266a4";
+                options.Authority = "https://login.windows.net/cyrano.onmicrosoft.com";
+                options.RedirectUri = "http://localhost:42023";
+            });
 
             app.Run(async context =>
             {
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
@@ -11,6 +11,7 @@ using System.Linq;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Security.Claims;
+using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Authentication;
@@ -19,6 +20,7 @@ using Microsoft.Framework.Caching.Distributed;
 using Microsoft.Framework.Internal;
 using Microsoft.Framework.Logging;
 using Microsoft.IdentityModel.Protocols.OpenIdConnect;
+using Microsoft.Net.Http.Headers;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.AspNet.Authentication.OpenIdConnect
@@ -30,6 +32,22 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
     {
         private const string NonceProperty = "N";
         private const string UriSchemeDelimiter = "://";
+
+        private const string InputTagFormat = @"<input type=""hidden"" name=""{0}"" value=""{1}"" />";
+        private const string HtmlFormFormat = @"<!doctype html>
+<html>
+<head>
+    <title>Please wait while you're being redirected to the identity provider</title>
+</head>
+<body>
+    <form name=""form"" method=""post"" action=""{0}"">
+        {1}
+        <noscript>Click here to finish the process: <input type=""submit"" /></noscript>
+    </form>
+    <script>document.form.submit();</script>
+</body>
+</html>";
+
         private OpenIdConnectConfiguration _configuration;
 
         protected HttpClient Backchannel { get; private set; }
@@ -93,13 +111,43 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
                     message = redirectToIdentityProviderNotification.ProtocolMessage;
                 }
 
-                var redirectUri = message.CreateLogoutRequestUrl();
-                if (!Uri.IsWellFormedUriString(redirectUri, UriKind.Absolute))
+                if (Options.AuthenticationMethod == OpenIdConnectAuthenticationMethod.RedirectGet)
                 {
-                    Logger.LogWarning(Resources.OIDCH_0051_RedirectUriLogoutIsNotWellFormed, redirectUri);
-                }
+                    var redirectUri = message.CreateLogoutRequestUrl();
+                    if (!Uri.IsWellFormedUriString(redirectUri, UriKind.Absolute))
+                    {
+                        Logger.LogWarning(Resources.OIDCH_0051_RedirectUriLogoutIsNotWellFormed, redirectUri);
+                    }
 
-                Response.Redirect(redirectUri);
+                    Response.Redirect(redirectUri);
+                }
+                else if (Options.AuthenticationMethod == OpenIdConnectAuthenticationMethod.FormPost)
+                {
+                    var inputs = new StringBuilder();
+                    foreach (var parameter in message.Parameters)
+                    {
+                        var name = Options.HtmlEncoder.HtmlEncode(parameter.Key);
+                        var value = Options.HtmlEncoder.HtmlEncode(parameter.Value);
+
+                        var input = string.Format(CultureInfo.InvariantCulture, InputTagFormat, name, value);
+                        inputs.AppendLine(input);
+                    }
+
+                    var issuer = Options.HtmlEncoder.HtmlEncode(message.IssuerAddress);
+
+                    var content = string.Format(CultureInfo.InvariantCulture, HtmlFormFormat, issuer, inputs);
+                    var buffer = Encoding.UTF8.GetBytes(content);
+
+                    Response.ContentLength = buffer.Length;
+                    Response.ContentType = "text/html;charset=UTF-8";
+
+                    // Emit Cache-Control=no-cache to prevent client caching.
+                    Response.Headers.Set(HeaderNames.CacheControl, "no-cache");
+                    Response.Headers.Set(HeaderNames.Pragma, "no-cache");
+                    Response.Headers.Set(HeaderNames.Expires, "-1");
+
+                    await Response.Body.WriteAsync(buffer, 0, buffer.Length);
+                }
             }
         }
 
@@ -218,14 +266,50 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
 
             message.State = Options.StateDataFormat.Protect(properties);
 
-            var redirectUri = message.CreateAuthenticationRequestUrl();
-            if (!Uri.IsWellFormedUriString(redirectUri, UriKind.Absolute))
+            if (Options.AuthenticationMethod == OpenIdConnectAuthenticationMethod.RedirectGet)
             {
-                Logger.LogWarning(Resources.OIDCH_0036_UriIsNotWellFormed, redirectUri);
+                var redirectUri = message.CreateAuthenticationRequestUrl();
+                if (!Uri.IsWellFormedUriString(redirectUri, UriKind.Absolute))
+                {
+                    Logger.LogWarning(Resources.OIDCH_0036_UriIsNotWellFormed, redirectUri);
+                }
+
+                Response.Redirect(redirectUri);
+
+                return true;
+            }
+            else if (Options.AuthenticationMethod == OpenIdConnectAuthenticationMethod.FormPost)
+            {
+                var inputs = new StringBuilder();
+                foreach (var parameter in message.Parameters)
+                {
+                    var name = Options.HtmlEncoder.HtmlEncode(parameter.Key);
+                    var value = Options.HtmlEncoder.HtmlEncode(parameter.Value);
+
+                    var input = string.Format(CultureInfo.InvariantCulture, InputTagFormat, name, value);
+                    inputs.AppendLine(input);
+                }
+
+                var issuer = Options.HtmlEncoder.HtmlEncode(message.IssuerAddress);
+
+                var content = string.Format(CultureInfo.InvariantCulture, HtmlFormFormat, issuer, inputs);
+                var buffer = Encoding.UTF8.GetBytes(content);
+
+                Response.ContentLength = buffer.Length;
+                Response.ContentType = "text/html;charset=UTF-8";
+
+                // Emit Cache-Control=no-cache to prevent client caching.
+                Response.Headers.Set(HeaderNames.CacheControl, "no-cache");
+                Response.Headers.Set(HeaderNames.Pragma, "no-cache");
+                Response.Headers.Set(HeaderNames.Expires, "-1");
+
+                await Response.Body.WriteAsync(buffer, 0, buffer.Length);
+
+                return true;
             }
 
-            Response.Redirect(redirectUri);
-            return true;
+            Logger.LogError("An unsupported authentication method has been configured: {0}", Options.AuthenticationMethod);
+            return false;
         }
 
         /// <summary>
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
@@ -11,6 +11,7 @@ using System.Security.Claims;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Authentication;
 using Microsoft.Framework.Caching.Distributed;
+using Microsoft.Framework.WebEncoders;
 using Microsoft.IdentityModel.Protocols;
 using Microsoft.IdentityModel.Protocols.OpenIdConnect;
 
@@ -191,6 +192,11 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
         public bool RefreshOnIssuerKeyNotFound { get; set; } = true;
 
         /// <summary>
+        /// Gets or sets the method used to redirect the user agent to the identity provider.
+        /// </summary>
+        public OpenIdConnectAuthenticationMethod AuthenticationMethod { get; set; }
+
+        /// <summary>
         /// Gets or sets the 'resource'.
         /// </summary>
         public string Resource { get; set; }
@@ -242,5 +248,10 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
         /// This is enabled by default.
         /// </summary>
         public bool UseTokenLifetime { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets the <see cref="IHtmlEncoder"/> used to sanitize HTML outputs.
+        /// </summary>
+        public IHtmlEncoder HtmlEncoder { get; set; }
     }
 }
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMethod.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMethod.cs
@@ -1,0 +1,21 @@
+﻿namespace Microsoft.AspNet.Authentication.OpenIdConnect
+{
+    /// <summary>
+    /// Lists the different authentication methods used to
+    /// redirect the user agent to the identity provider.
+    /// </summary>
+    public enum OpenIdConnectAuthenticationMethod
+    {
+        /// <summary>
+        /// Emits a 302 response to redirect the user agent to
+        /// the OpenID Connect provider using a GET request.
+        /// </summary>
+        RedirectGet = 0,
+
+        /// <summary>
+        /// Emits an HTML form to redirect the user agent to
+        /// the OpenID Connect provider using a POST request.
+        /// </summary>
+        FormPost = 1
+    }
+}
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationMiddleware.cs
@@ -54,6 +54,11 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
                 Options.SignInScheme = sharedOptions.Options.SignInScheme;
             }
 
+            if (Options.HtmlEncoder == null)
+            {
+                Options.HtmlEncoder = services.GetHtmlEncoder();
+            }
+
             if (Options.StateDataFormat == null)
             {
                 var dataProtector = dataProtectionProvider.CreateProtector(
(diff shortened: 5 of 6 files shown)


samples/OpenIdConnectSample/NuGet.Config
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="aspnet-contrib" value="https://www.myget.org/F/aspnet-contrib/" />
```
PinpointTownes: I'll undo these changes when we're ready to merge this PR.

samples/OpenIdConnectSample/Startup.cs
```diff
@@ -31,28 +38,63 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
...
+                };
+            });
 
             app.Run(async context =>
             {
                 if (!context.User.Identities.Any(identity => identity.IsAuthenticated))
                 {
                     await context.Authentication.ChallengeAsync(OpenIdConnectAuthenticationDefaults.AuthenticationScheme, new AuthenticationProperties { RedirectUri = "/" });
-
-                    context.Response.ContentType = "text/plain";
```
PinpointTownes: Updating the headers collection at this stage is no longer safe with POST requests since we start writing the auto-post form to the response stream in the middleware (invoked via `context.Authentication.ChallengeAsync`).

These two lines were pointless anyway, since the user agent was not supposed to see the printed text.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -217,6 +218,47 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
...
+            if (Options.UsePostAuthorizationRequests)
+            {
+                using (var buffer = new MemoryStream())
+                using (var writer = new StreamWriter(buffer))
+                {
+                    writer.WriteLine("<!doctype html>");
+                    writer.WriteLine("<html>");
+                    writer.WriteLine("<body>");
+
+                    writer.WriteLine("<form name='form' method='post' action='" + Options.HtmlEncoder.HtmlEncode(message.IssuerAddress) + "'>");
```
PinpointTownes: The last time I tried `IHtmlEncoder` (well, its default implementation), using it with single quotes was safe.
@blowdart 'zat still true?

PinpointTownes: Note: `OpenIdConnectMessage`, which derives from `AuthenticationProtocolMessage`, has a `BuildFormPost` method that is supposed to do exactly the same thing. Sadly it's totally busted as it internally uses `Uri.EscapeDataString` which messes up the action attribute. Instead, I used the "manual" approach I already used for `AspNet.Security.OpenIdConnect.Server`: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/blob/vNext/src/AspNet.Security.OpenIdConnect.Server/OpenIdConnectServerHandler.cs#L709

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -217,6 +218,47 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
 
             message.State = Options.StateDataFormat.Protect(properties);
 
+            if (Options.UsePostAuthorizationRequests)
+            {
+                using (var buffer = new MemoryStream())
+                using (var writer = new StreamWriter(buffer))
+                {
+                    writer.WriteLine("<!doctype html>");
```
HaoK: Nit: This handcrafted html, can we put this into a string constant at least and just format string the action and input block in a which you generate as well?  

Something like this so you can just put all the html in one literal string constant block formatted how you like without all the writelines

```
var inputs = new StringBuilder();
foreach (var parameter in message.Parameters) {
   inputs.Add(string.Format(InputTagFormat, name, value);
}
writer.Write(string.Format(HtmlFormat, messageAction, inputs));
```

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -217,6 +218,47 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
...
+                    writer.WriteLine("<script>document.form.submit();</script>");
+                    writer.WriteLine("</body>");
+                    writer.WriteLine("</html>");
+                    writer.Flush();
+
+                    Response.ContentLength = buffer.Length;
+                    Response.ContentType = "text/html;charset=UTF-8";
+
+                    // Emit Cache-Control=no-cache to prevent client caching.
+                    Response.Headers.Set("Cache-Control", "no-cache");
```
PinpointTownes: We could generalize this practice and also emit these headers for GET requests, but AFAIK, browsers are not supposed to cache 302 responses so I only applied it for POST requests.

/cc @Tratcher

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -233,6 +234,11 @@ public OpenIdConnectProtocolValidator ProtocolValidator
         public bool RefreshOnIssuerKeyNotFound { get; set; }
 
         /// <summary>
+        /// Determines whether authorization requests should be made using POST.
+        /// </summary>
+        public bool UsePostAuthorizationRequests { get; set; }
```
PinpointTownes: If you have a better name in mind, don't hesitate :smile:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: Another note: this code was a good candidate for string interpolation. Sadly, since `CultureInfo.CurrentCulture` is the default culture used by string interpolation, I preferred using the good old method to avoid future issues with the localization stuff.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: Yep, good idea.

PinpointTownes: It's indeed much clearer, thanks! :smile:

HaoK: Yep looks good, I don't have enough context about the actual feature so that's about all I can contribute to this PR :)

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -217,6 +235,35 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
...
+                    var inputs = new StringBuilder();
+                    foreach (var parameter in message.Parameters)
+                    {
+                        inputs.AppendFormat(InputTagFormat, parameter.Key, parameter.Value);
+                    }
+
+                    writer.Write(string.Format(CultureInfo.InvariantCulture, HtmlFormat, message.IssuerAddress, inputs));
+                    writer.Flush();
+
+                    Response.ContentLength = buffer.Length;
```
PinpointTownes: @Tratcher I used a local buffer to avoid chunking. I'm fine removing it if you don't like it.

Tratcher: No, that's fine.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
Tratcher: Enum? AuthorizationMode = FormPost or RedirectGet

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -29,6 +30,22 @@ public class OpenIdConnectAuthenticationHandler : AuthenticationHandler<OpenIdCo
     {
         private const string NonceProperty = "N";
         private const string UriSchemeDelimiter = "://";
+
+        private const string InputTagFormat = @"<input type=""hidden"" name=""{0}"" value=""{1}"" />";
+        private const string HtmlFormat = @"<!doctype html>
```
Tratcher: AuthorizationFormFormat?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -217,6 +234,40 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
...
+                    writer.Write(string.Format(CultureInfo.InvariantCulture, HtmlFormat, issuer, inputs));
+                    writer.Flush();
+
+                    Response.ContentLength = buffer.Length;
+                    Response.ContentType = "text/html;charset=UTF-8";
+
+                    // Emit Cache-Control=no-cache to prevent client caching.
+                    Response.Headers.Set("Cache-Control", "no-cache");
+                    Response.Headers.Set("Pragma", "no-cache");
+                    Response.Headers.Set("Expires", "-1");
```
Tratcher: Recommend using constants: https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.Net.Http.Headers/HeaderNames.cs

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: Or `AuthenticationMethod`/`AuthorizationMethod`?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: Updated.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: Ah nice, I had not realized `Microsoft.Net.Http.Headers` was indirectly referenced. Thanks for the tip! :smile:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
Tratcher: ` AuthorizationMethod `

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -217,6 +235,40 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
 
             message.State = Options.StateDataFormat.Protect(properties);
 
+            if (Options.UsePostAuthorizationRequests)
+            {
+                using (var buffer = new MemoryStream())
+                using (var writer = new StreamWriter(buffer))
```
Tratcher: Is the default encoding UTF-8? Does it emit BOM characters?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: The OIDC middleware currently only uses `Authentication`, though both options are totally valid (OIDC being both an authorization and an authentication protocol). To make this property consistent, I guess it would be better to use `Authentication`.

> An Authentication Request is an OAuth 2.0 Authorization Request that requests that the End-User be authenticated by the Authorization Server.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -217,6 +235,40 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
...
+                    {
+                        var name = Options.HtmlEncoder.HtmlEncode(parameter.Key);
+                        var value = Options.HtmlEncoder.HtmlEncode(parameter.Value);
+
+                        inputs.AppendFormat(CultureInfo.InvariantCulture, InputTagFormat, name, value);
+                    }
+
+                    var issuer = Options.HtmlEncoder.HtmlEncode(message.IssuerAddress);
+
+                    writer.Write(string.Format(CultureInfo.InvariantCulture, AuthorizationFormFormat, issuer, inputs));
```
Tratcher: writer.Write is only called once. Why not just call Encoding.UTF8.GetBytes directly? Then you don't need the StreamWriter, the MemoryStream, etc..

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: Nope, it doesn't. This constructor internally uses `StreamWriter.UTF8NoBOM` as the default encoding. You'd prefer UTF-8 with BOM?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
Tratcher: *brain explodes*

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
Tratcher: No, we don't want BOMs, some clients don't read them correctly.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: We could, but note that `Encoding.UTF8` always emits a BOM. If we don't want BOMs, we'll have to instantiate a custom `UTF8Encoding` with `encoderShouldEmitUTF8Identifier` set to `false`.

Tratcher: GetBytes doesn't include BOMs, StreamWriter adds them using Encoding.GetPreamble.

PinpointTownes: Ah nice, let's use `GetBytes` directly then :smile:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
PinpointTownes: Done :smile: 

Allowing the developer to change the authentication method directly from the `RedirectToIdentityProvider` notification would be nice, but sadly, `RedirectToIdentityProviderNotification` is not OIDC-specific and is part of `Microsoft.AspNet.Authentication`.

To work around that, we could create a derived class or add a "generic" property to `RedirectToIdentityProvider` containing the HTTP method.

Tratcher: Lets not go there yet. It wouldn't even be an OIDC specific option, it would be an OIDC challenge specific option.  RedirectToIdentityProvider is also used for SignOut.

That said, does the protocol support POSTs for SignOut? It does not appear to: http://openid.net/specs/openid-connect-session-1_0.html#RPLogout

PinpointTownes: Well, the `OpenID Connect Session Management` specs are very unclear about the methods an authorization server should or must support, since they only use the `redirect` keyword:

> In this case, the RP, after having logged the End-User out of the RP, redirects the End-User's User Agent to the OP's logout endpoint URL.

Personally, I decided to support both methods in `AspNet.Security.OpenIdConnect.Server`, as I couldn't think of a valid reason to reject POST requests.

Tratcher:
Forking the comment thread so GitHub will stop hiding it: https://github.com/aspnet/Security/pull/392#discussion_r36679805
@PinpointTownes, I've forwarded your question to our OIDC spec guy. We'll see what they have to say.

PinpointTownes:
@Tratcher thanks!

[sarcastic mode]
I hope your OIDC spec guy will come with a more useful answer than the one he had for `response_mode=query` because I had the weird impression I was treated like a complete noob last time :smile:
[/sarcastic mode]

NuGet.Config
```diff
@@ -4,5 +4,8 @@
     <add key="AspNetVNext" value="https://www.myget.org/F/aspnetvnext/api/v2" />
     <add key="NuGet" value="https://nuget.org/api/v2/" />
     <add key="AzureAD" value="http://www.myget.org/F/azureadwebstacknightly"/>
+    <add key="aspnet-contrib" value="https://www.myget.org/F/aspnet-contrib/" />
+    <add key="Roslyn" value="https://www.myget.org/F/roslyn-nightly/" />
```
PinpointTownes: These changes have nothing to do in this PR but are required to make CI happy.
I opened a ticket to track this omission (https://github.com/aspnet/Security/issues/399) and I'll remove them before squashing.

PinpointTownes:
@Tratcher hey. Did you hear back from your OIDC specs specialist?

Tratcher:
There is still some debate, but they recommended supporting it.

PinpointTownes:
No words about the methods the logout endpoint should support?

https://github.com/aspnet/Security/pull/392#discussion_r36679805

Tratcher:
That's what I meant, they said both endpoint should support POSTs.

PinpointTownes:
Ah, interesting. Should we add another property to control the logout method, or should we use AuthenticationMethod for that?

Tratcher:
I'd control them both from the same option. I don't know why you'd want to separate them.

PinpointTownes:
Well, I guess an authentication server could support POST for the authorization endpoint (since it's mandatory :smile:) but not for the end session endpoint.

samples/OpenIdConnectSample/NuGet.Config
PinpointTownes: Reverted.

PinpointTownes:
PR rebased and updated to support POST logout requests.

FYI, I'll be absent this week, but I guess this PR is not critical and can wait a bit. I'll react to your feedback next week :+1:

PinpointTownes:
I'm back! :smile: 

@blowdart no security concern with this PR?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -92,13 +110,43 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
...
+                        var value = Options.HtmlEncoder.HtmlEncode(parameter.Value);
 
-                Response.Redirect(redirectUri);
+                        var input = string.Format(CultureInfo.InvariantCulture, InputTagFormat, name, value);
+                        inputs.AppendLine(input);
+                    }
+
+                    var issuer = Options.HtmlEncoder.HtmlEncode(message.IssuerAddress);
+
+                    var content = string.Format(CultureInfo.InvariantCulture, HtmlFormFormat, issuer, inputs);
```
blowdart: Use string interpolation rather than {0}, {1}

PinpointTownes: Then we'll have to de-constant-ize (I love that verb :smile:) `HtmlFormFormat`.

See this comment for why I prefer avoiding using string interpolation on .NET 4.5.1 (though it's a great feature): https://github.com/aspnet/Security/pull/392#discussion_r36569188

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -217,14 +265,50 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
...
+                foreach (var parameter in message.Parameters)
+                {
+                    var name = Options.HtmlEncoder.HtmlEncode(parameter.Key);
+                    var value = Options.HtmlEncoder.HtmlEncode(parameter.Value);
+
+                    var input = string.Format(CultureInfo.InvariantCulture, InputTagFormat, name, value);
+                    inputs.AppendLine(input);
+                }
+
+                var issuer = Options.HtmlEncoder.HtmlEncode(message.IssuerAddress);
```
blowdart: Are we assuming IssuerAddress is correctly Url encoded first?

PinpointTownes: Wouldn't URL-encoding it make the resulting URL unusable (since it's used as the form action)?

https://github.com/aspnet/Security/pull/392#discussion_r36568911

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
blowdart: I believe that's still true. There's no separate attribute encode any more.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
blowdart: Which is why I'm asking :) This is configured by the dev right? So they should get it tight. Hopefully.

PinpointTownes: The issuer address (i.e the authorization or the logout endpoint) is either extracted from the OIDC provider metadata or manually configured by the dev. In both cases, it's not supposed to be URL-encoded.

It can also be replaced from the `RedirectToIdentityProvider` notification but I'm not sure there's something we could do to avoid manual encoding.

Tratcher:
:shipit: pending any remaining comments from @blowdart?

blowdart:
Nothing from me.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -30,6 +32,22 @@ public class OpenIdConnectAuthenticationHandler : AuthenticationHandler<OpenIdCo
     {
         private const string NonceProperty = "N";
         private const string UriSchemeDelimiter = "://";
+
+        private const string InputTagFormat = @"<input type=""hidden"" name=""{0}"" value=""{1}"" />";
```
Eilon: @blowdart please make sure you review these codez for HTML injection / CSRF / etc.

blowdart: I did, it's used further down at line 132, with encoding happening in the previous 2 lines.

PR #393, opened 2015-08-07
Cookies now always redirects to Login/AccessDenied Paths

HaoK:
If Paths for these are not set, middleware sets them to the defaults

cc @Tratcher


HaoK:
ugh forgot there's another PR closing this

PR #394, opened 2015-08-07
Cookies now always redirects to Login/AccessDenied Paths

HaoK:
We default values for the paths if they are not set in options

cc @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationMiddleware.cs
+++ b/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationMiddleware.cs
@@ -40,6 +40,18 @@ namespace Microsoft.AspNet.Authentication.Cookies
             {
                 Options.CookieManager = new ChunkingCookieManager(urlEncoder);
             }
+            if (!Options.LoginPath.HasValue)
+            {
+                Options.LoginPath = CookieAuthenticationDefaults.LoginPath;
+            }
+            if (!Options.LogoutPath.HasValue)
+            {
+                Options.LogoutPath = CookieAuthenticationDefaults.LogoutPath;
+            }
+            if (!Options.AccessDeniedPath.HasValue)
+            {
+                Options.AccessDeniedPath = CookieAuthenticationDefaults.AccessDeniedPath;
+            }
         }
 
         protected override AuthenticationHandler<CookieAuthenticationOptions> CreateHandler()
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationMiddleware.cs
```diff
@@ -40,6 +40,18 @@ public class CookieAuthenticationMiddleware : AuthenticationMiddleware<CookieAut
             {
                 Options.CookieManager = new ChunkingCookieManager(urlEncoder);
             }
+            if (!Options.LoginPath.HasValue)
+            {
+                Options.LoginPath = CookieAuthenticationDefaults.LoginPath;
+            }
+            if (!Options.LogoutPath.HasValue)
+            {
+                Options.LogoutPath = CookieAuthenticationDefaults.LogoutPath;
```
Tratcher: Should a logout redirect still be optional? SignOut("Cookies") primary job is to delete the cookie.

HaoK: I don't think this one is a redirect. This one just specified what uri will enable the return uri to trigger
A redirect so I think it's ok to have a default.

They still have to send a query string with return url on it to the actual logout path and call sign out there for this to have any effect

PinpointTownes:
Curious: won't this change conflict with social middleware configured to use the automatic mode?

Having both an OIDC middleware and a cookie middleware using active authentication is very common: in this case, you DON'T want your users to be redirected to a local path (they should be directly redirected to the OIDC provider).

HaoK:
Cookies is supposed to be first. Which means last in the chain. If you put one of the earlier ones I in automatic and it handles the request it should stop cookies from getting hit.

test/Microsoft.AspNet.Authentication.Test/Cookies/CookieMiddlewareTests.cs
```diff
@@ -721,6 +715,22 @@ private Task SignOutAsWrong(HttpContext context)
...
 
         [Fact]
+        public async Task ChallengeDoesNotSet401OnUnauthorized()
+        {
+            var server = TestServer.Create(app =>
+            {
+                app.UseCookieAuthentication();
+                app.Run(async context => {
+                    await Assert.ThrowsAsync<InvalidOperationException>(() => context.Authentication.ChallengeAsync());
+                    context.Response.StatusCode = 200;
```
Tratcher: You should be able to remove this line, it may create false positives.

Tratcher:
:shipit:

HaoK:
b883920befce4afaac89ec031d30d868a48df666

PR #397, opened 2015-08-08
Will we have a Authentication.LinkedIn

PrimeObjects:
Sorry I don't know whether this is a right place to ask a feature like this.


Tratcher:
Questions are better discussed here: https://github.com/aspnet/Security/issues

As for LinkedIn, no, we don't plan to add a middleware for it. However, you can use the generic OAuth middleware as shown here: https://github.com/aspnet/Security/blob/dev/samples/SocialSample/Startup.cs#L127-L185

PinpointTownes:
FYI, the LinkedIn provider will be ported to ASP.NET 5 by @jerriep: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/11

In the meantime, using the generic OAuth2 middleware - as suggested by @Tratcher - is your best option.

PinpointTownes:
@PrimeObjects thanks to @jerriep, the LinkedIn provider is now on MyGet: https://www.myget.org/F/aspnet-contrib/ :tada:

PR #403, opened 2015-08-21
Additional OIDC message validation.

Tratcher:
#278 @brentschmaltz @muratg @tushargupta51

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
@@ -409,6 +409,8 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
                 ticket = await GetUserInformationAsync(properties, tokenEndpointResponse.Message, ticket);
             }
 
+            await ValidateOpenIdConnectProtocolAsync(jwt, null);
+
             var securityTokenValidatedNotification = await RunSecurityTokenValidatedNotificationAsync(message, ticket);
             if (securityTokenValidatedNotification.HandledResponse)
             {
@@ -418,9 +420,6 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
             {
                 return null;
             }
-
-            // If id_token is received using code only flow, no need to validate chash.
-            await ValidateOpenIdConnectProtocolAsync(jwt, message, false);
 
             return ticket;
         }
@@ -443,6 +442,8 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
             var validationParameters = Options.TokenValidationParameters.Clone();
             ticket = ValidateToken(message.IdToken, message, properties, validationParameters, out jwt);
 
+            await ValidateOpenIdConnectProtocolAsync(jwt, message);
+
             var securityTokenValidatedNotification = await RunSecurityTokenValidatedNotificationAsync(message, ticket);
             if (securityTokenValidatedNotification.HandledResponse)
             {
@@ -452,8 +453,6 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
             {
                 return null;
             }
-
-            await ValidateOpenIdConnectProtocolAsync(jwt, message);
 
             if (message.Code != null)
             {
@@ -879,7 +878,7 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
             return ticket;
         }
 
-        private async Task ValidateOpenIdConnectProtocolAsync(JwtSecurityToken jwt, OpenIdConnectMessage message, bool ValidateCHash = true)
+        private async Task ValidateOpenIdConnectProtocolAsync(JwtSecurityToken jwt, OpenIdConnectMessage message)
         {
             string nonce = jwt.Payload.Nonce;
             if (Options.CacheNonces)
@@ -902,16 +901,13 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
 
             var protocolValidationContext = new OpenIdConnectProtocolValidationContext
             {
+                ProtocolMessage = message,
+                IdToken = jwt,
+                ClientId = Options.ClientId,
                 Nonce = nonce
             };
 
-            // If authorization code is null, protocol validator does not validate the chash
-            if (ValidateCHash)
-            {
-                protocolValidationContext.AuthorizationCode = message.Code;
-            }
-
-            Options.ProtocolValidator.Validate(jwt, protocolValidationContext);
+            Options.ProtocolValidator.Validate(protocolValidationContext);
         }
 
         /// <summary>
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
@@ -58,7 +58,7 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
             BackchannelTimeout = TimeSpan.FromMinutes(1);
             Caption = OpenIdConnectAuthenticationDefaults.Caption;
             GetClaimsFromUserInfoEndpoint = false;
-            ProtocolValidator = new OpenIdConnectProtocolValidator();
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
```diff
@@ -58,7 +58,7 @@ public OpenIdConnectAuthenticationOptions(string authenticationScheme)
             BackchannelTimeout = TimeSpan.FromMinutes(1);
             Caption = OpenIdConnectAuthenticationDefaults.Caption;
             GetClaimsFromUserInfoEndpoint = false;
-            ProtocolValidator = new OpenIdConnectProtocolValidator();
+            ProtocolValidator = new OpenIdConnectProtocolValidator() { RequireState = false };
```
brentschmaltz: We should open an issue that setting 'RequireState = false' this is temporary as proper validation this is required

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -911,7 +914,7 @@ private AuthenticationTicket ValidateToken(string idToken, OpenIdConnectMessage
                 protocolValidationContext.AuthorizationCode = message.Code;
```
brentschmaltz: This if clause, is not needed as the ProtocolValidationContext has the entire message. Also, the Property 'AuthorizationCode' is marked for to be removed in beta8.

Tratcher: which means that ValidateCHash can be removed entirely?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
Tratcher: How can we flow the original state independent of the message?

Tratcher:
Updated

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: We need to use the same strategy as the 'nonce', we include it as part of the nonce. Validation requires that the state returned in the Response message is exactly the same as sent in the Request message.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -879,7 +879,7 @@ private AuthenticationTicket ValidateToken(string idToken, OpenIdConnectMessage
             return ticket;
         }
 
-        private async Task ValidateOpenIdConnectProtocolAsync(JwtSecurityToken jwt, OpenIdConnectMessage message, bool ValidateCHash = true)
+        private async Task ValidateOpenIdConnectProtocolAsync(JwtSecurityToken jwt, OpenIdConnectMessage message)
```
brentschmaltz: We should be calling this validation before firing the SecurityTokenValidated notification, so that users will not perform any work on an invalid message.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: I set the Authorization Code to null to skip the chash validation. If a code is present as ValidationContext.AuthorizationCode in old model or ValidationContext.Message.Code in new model, protocolValidator will try to validate the chash and throw if it cannot find the chash claim which is the case in 'code' only flow.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs#L378

Ideal way to fix this is to have the protocolValidator look at the state and if response_type is 'code' in the request sent out, do not validate the chash.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -419,9 +421,6 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
                 return null;
             }
 
-            // If id_token is received using code only flow, no need to validate chash.
-            await ValidateOpenIdConnectProtocolAsync(jwt, message, false);
```
tushargupta51: I set the Authorization Code to null to skip the chash validation. If a code is present as ValidationContext.AuthorizationCode in old model or ValidationContext.Message.Code in new model, protocolValidator will try to validate the chash and throw if it cannot find the chash claim which is the case in 'code' only flow.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectProtocolValidator.cs#L378

Ideal way to fix this is to have the protocolValidator look at the state and if response_type is 'code' in the request sent out, do not validate the chash.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
Tratcher: https://github.com/aspnet/Security/issues/404

Tratcher:
Updated.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
brentschmaltz: The validator should only look for a 'c_hash' claim if an 'id_token ' is in the message.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
brentschmaltz: I thought that the 'state' issue would block us, hence the temporary flag. Once you have the OIDC sample app working, then I vote ship it, we can deal with #404 later.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: You are right. I missed the outer check for non-null id token. thanks.

tushargupta51:
:shipit:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
tushargupta51: @brentschmaltz In the case, where we get a code using 'code' only flow and then redeem that for an id_token, validating that id_token will throw because the id_token does not have the chash claim.

In this line: https://github.com/aspnet/Security/blob/tratcher/oidc/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs#L412

jwt does not have a chash claim and message.code is not null, it will throw IDX10308. The test here fails: https://github.com/aspnet/Security/blob/tratcher/oidc/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs#L412

brentschmaltz: The original Response message has only a code. That code is redeemed for a token + id_token inside the OIDCMiddleware. I don't know why we would pass the original message + id_token from the tokenendpoint to the validator.

Tratcher:
Updated.

Tratcher:
Updated. Are developers using code flow always going to have to set that, or is the test doing something unusual? It seems odd.

tushargupta51:
@tratcher this is just a test thing

Issue #407, opened 2015-08-24
[OIDC] GetUserInformationAsync needs to fail if the subjects do not match


Tratcher:
OpenIdConnetHander…
protected virtual async Task<AuthenticationTicket> GetUserInformationAsync(AuthenticationProperties properties, OpenIdConnectMessage message, AuthenticationTicket ticket)

The protocol requires that the subject from the UI endpoint must equal the subject of the id_token received from redeeming the code for an ‘access_token + id_token’.
Our code logs and returns original the ticket parameter on failure (there are two failures: 1. The sub claim is not found, 2. The sub claims do not match).. We should fault as something is wrong.
Also the code would be a whole lot simpler if we just fished the ‘sub’ claim out of the ‘id_token’ that arrived with the access token, than run through the ClaimsIdentity.

---
Yes, we should throw instead of just logging. What does the absence of subject claim in the id_token received from token endpoint indicates? Is that an error we should throw or just something that we log?
[brent]OIDC requires the ‘sub’ claim in the id_token, if it’s not there that is not a protocol compliant message from the token endpoint.

Also the code would be a whole lot simpler if we just fished the ‘sub’ claim out of the ‘id_token’ that arrived with the access token, than run through the ClaimsIdentity.
[tushar] not sure if I understand this. Do you mean not looking into the ‘sub’ claim in the ‘id_token’ and just using the one we receive from UI endpoint?
[brent] the flow is: code -> id_token + access token -> UI Endpoint -> JSON claims.

There must be a ‘sub’ claim in the id_token and in the JSON claims. The ‘sub’ must match. IIRC, conversations with Mike Jones indicated if this is not true, we must fault.

PR #409, opened 2015-08-24
OIDC - Fail if the user-info subject does not match

Tratcher:
#407 
@tushargupta51 @brentschmaltz @muratg

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
@@ -403,13 +403,13 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
 
             ticket = ValidateToken(idToken, message, properties, validationParameters, out jwt);
 
+            await ValidateOpenIdConnectProtocolAsync(jwt, null);
+
             if (Options.GetClaimsFromUserInfoEndpoint)
             {
                 Logger.LogDebug(Resources.OIDCH_0040_Sending_Request_UIEndpoint);
                 ticket = await GetUserInformationAsync(properties, tokenEndpointResponse.Message, ticket);
             }
-
-            await ValidateOpenIdConnectProtocolAsync(jwt, null);
 
             var securityTokenValidatedNotification = await RunSecurityTokenValidatedNotificationAsync(message, ticket);
             if (securityTokenValidatedNotification.HandledResponse)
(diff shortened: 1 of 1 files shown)


tushargupta51:
looks good to me

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -529,7 +534,7 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
             if (subjectClaimType == null)
             {
                 Logger.LogError(string.Format(CultureInfo.InvariantCulture, Resources.OIDCH_0041_Subject_Claim_Not_Found, identity.ToString()));
-                return ticket;
+                return null;
```
brentschmaltz: Consider faulting here as this is a protocol error. Returning 'null' will give the next middleware a crack at the message.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -538,7 +543,7 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
             if (userInfoSubjectClaimValue == null || !string.Equals(userInfoSubjectClaimValue, subjectClaimType.Value, StringComparison.Ordinal))
             {
                 Logger.LogError(Resources.OIDCH_0039_Subject_Claim_Mismatch);
-                return ticket;
+                return null;
```
brentschmaltz: Consider faulting here as this is a protocol error. Returning 'null' will give the next middleware a crack at the message.

Tratcher:
Updated

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: No, please, stop throwing exceptions everywhere... or catch them eventually to avoid returning a 500 response (I'll open a ticket for the OAuth2 bearer middleware as it's really an annoying default behavior).

I'm not sure to understand why logging the error and returning `null` wouldn't be enough.

PinpointTownes: https://github.com/aspnet/Security/issues/411

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -528,8 +528,7 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
             var subjectClaimType = identity.FindFirst(ClaimTypes.NameIdentifier);
```
brentschmaltz: this may cause issues if the 'ClaimTypes.NameIdentifier' does not map to sub. It would be less fragile to get the 'sub' claim from the original 'id_token'.

brentschmaltz:
:shipit:

Tratcher:
Merged

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -403,14 +403,14 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
 
             ticket = ValidateToken(idToken, message, properties, validationParameters, out jwt);
 
+            await ValidateOpenIdConnectProtocolAsync(null, message);
```
tushargupta51: @Tratcher Why did you make this change? from (jwt, null) to (null, message).

Tratcher: because @brentschmaltz told me to.

Issue #413, opened 2015-08-26
Consider renaming OAuthBearer


Tratcher:
The OAuthBearer middleware is not a generic bearer token handler, it only supports JWT's. Consider renaming this one to something more specific (JWTBearer? OAughJWT?) so that we can create a more generic one later.

Eilon:
Let's go with `JwtBearer`.

PinpointTownes:
> The OAuthBearer middleware is not a generic bearer token handler, it only supports JWT's.

The OAuth2 bearer middleware is not tied to JWT since it only requires a `ISecurityTokenValidator` (well, a list of `ISecurityTokenValidator` to be exact), which is also implemented by `SamlSecurityTokenHandler` or `Saml2SecurityTokenHandler` (though I'm pretty sure they are still non-finalized things ATM).

Not to mention that you can also use your own token format using the `MessageReceived` or `SecurityTokenReceived` notifications. Renaming this middleware to something JWT-specific would be a bit misleading IMHO.

PinpointTownes:
Related (old) comments about `OAuthBearerAuthenticationMiddleware`'s name:

-----

I agree with @Tratcher: we really need a basic "token validation" middleware that behaves exactly like the OAuth2 bearer token middleware built in Katana, taking no dependency on the OIDC extensions. 

Making `SecurityTokenHandler` the first class citizen in the base bearer middleware is not necessarily a bad thing, but if you replace `ISecureDataFormat<AuthenticationTicket> AccessTokenFormat` by a collection of security token handlers, supporting both serialization mechanisms is a must (think about the scenario where the access token created by the OIDC server is opaque and uses a data protector).

Personally, I'd rename the existing OAuth2 bearer middleware to `TokenAuthenticationMiddleware` and extract all the new OAuth2/OIDC greatness you plan to add to a separate package. `OAuthBearerAuthenticationMiddleware` has been a very confusing name for most people (visit #owin on JabbR if you're not convinced), so I think it's really time to fix that.

https://github.com/aspnet/Security/pull/112#issuecomment-69846740

-------

@brentschmaltz @Tratcher all these changes are great and we definitely need them (the OIDC configuration discovery is just amazing). But we also need a basic token validation middleware that behaves exactly like the existing OAuth2 bearer middleware and allows full pluggability without relying on complicated and OIDC-specific things. Here's my suggestion:

[...]

PR #422, opened 2015-09-01
Rename OAuthBearer to JwtBearer.

Tratcher:
#413
@muratg @Eilon @PinpointTownes


PinpointTownes:
If you really want to rename this middleware to something JWT-specific, you should consider replacing `SecurityTokenValidators` by a single `JwtSecurityTokenHandler`.

(I'm still convinced `TokenAuthenticationMiddleware` would be a clearer name)

Tratcher:
@HaoK

HaoK:
:shipit:

Issue #58, opened 2014-09-09
Port serialization for BootstrapContext. Make sure to increment TicketSerializer.FormatVersion.


PinpointTownes:
Recent changes made in Katana to support `BootstrapContext` have not been ported yet to vNext. When porting them, make sure to increment `TicketSerializer.FormatVersion` to avoid this kind of issue: https://katanaproject.codeplex.com/workitem/347

> Associated topic: https://katanaproject.codeplex.com/discussions/566815

> __Symptoms__: tickets (and by extension, cookies) created by __Katana 2.1__ are not correctly deserialized under __Katana 3__ and lack the `AuthenticationProperties` element. This causes critical properties like `IssuedUtc` or `ExpiresUtc` to be absent from the final ticket. As a result, non-persistent cookies never expire.

> __Root cause__: when __Brent Schmaltz__ added `BootstrapContext` support in `TicketSerializer` (https://katanaproject.codeplex.com/SourceControl/changeset/87f2be297e5b24cbbc2627cc5cac2ded41ddf822), he forgot to increment the version number to make sure tickets serialized before this change were ignored. 

> __Explanation__: when `TicketSerializer` tries to read old tickets, the new `reader.ReadInt32()` call inserted in `TicketSerializer.Read` makes the binary serializer going further than expected by trying to read an element that was not present in the serialized payload. Next line, when `PropertiesSerializer.Read` is called to deserialize the `AuthenticationProperties`, `BinaryReader` fails to read the correct version number associated with `PropertiesSerializer`, which causes the ticket to be returned without its `AuthenticationProperties` companion.

> __Solution__: increment `TicketSerializer.FormatVersion` to make sure `TicketSerializer` won't try to deserialize incompatible tickets. This change will make every existing ticket/cookie unreadable. As a consequence, users would have to re-authenticate. In the meantime, end devs can implement their own `TicketSerializer` and fix this issue by incrementing the version number.

Eilon:
We think the type is now available on Core CLR, so we should be able to do this now.

Eilon:
@Tratcher can you check if the APIs we need are there? If not, we'll have to ping some folks.

Tratcher:
@brentschmaltz What package is the BootstrapContext type available in for CoreCLR?

PinpointTownes:
`BootstrapContext` no longer exists.

`JwtSecurityTokenHandler` sets `ClaimsIdentity.BootstrapContext` with the serialized token (as a string).

Issue #420, opened 2015-09-01
Remove usage of shouldly


davidfowl:
We should just be using xunit

Eilon:
And Shoudly doesn't run on DNXCore, either.

jmloeffler:
If it's ok with everyone, I think that I might be able to tackle this.  I'm looking for ways to contribute and this one seems low-risk.

Eilon:
@jmloeffler go for it. Removing Shoudly is mostly a matter of writing clever regex's in VS search & replace to convert to a regular xUnit assert.

Here's an example where I've previously done it: https://github.com/aspnet/StaticFiles/commit/9c5e78b9dc05ea384db4a59ea0e9e0d26da7ef59

That should show the general pattern of conversions.

jmloeffler:
Yep, that was my plan.  Thanks for the cheat sheet @Eilon!

Eilon:
And BTW most important is changing the project.json to of course remove Shouldly, but more importantly, to add `dnxcore50`.

jmloeffler:
@Eilon It appears that Moq is also a blocker for targeting `dnxcore50`.  Do we have a replacement for Moq (different version or mocking framework) or is that as far as I can take this?

jmloeffler:
FYI - in case we have to do this again somewhere, the following regex replacements in VS2015 converted most (but not all) of the statements for me.  I'll submit a PR with the changes after I spend some time looking into the Moq thing.

`(\s*)(.*)\.ShouldBe\(null\) => $1Assert.Null($2)`
`(\s*)(.*)\.ShouldNotBe\(null\) => $1Assert.NotNull($2)`
`(\s*)(.*)\.ShouldBe\((.*)\) => $1Assert.Equal($3, $2)`
`(\s*)(.*)\.ShouldContain\((.*)\) => $1Assert.Contains($3, $2)`
`(\s*)(.*)\.ShouldNotContain\((.*)\) => $1Assert.DoesNotContain($3, $2)`

Eilon:
No Moq on CoreCLR yet. If the usage of Moq is trivial, you could just remove it. If it's more involved, just leave it and don't enabled CoreCLR.

We're working on Moq for CoreCLR so once that's available we can enable `dnxcore50` for this project.

Issue #324, opened 2015-07-01
CookieMiddleware serialization of the ClaimsPrincipal is lossy.


brentschmaltz:
The AuthenticationTicket serialization is lossy. Not all of the ClaimsIdentities and Claims properties are serialized.

PinpointTownes:
> The AuthenticationTicket serialization is lossy. Not all of the ClaimsIdentities and Claims properties are serialized.

This issue is not specific to cookies and affects everything using `TicketSerializer`.
+1 for `Claim.Properties`, I had to create my own serializer to fix that.

Anyway, these issues will likely disappear when migrating to your new `WriteTo` gem: https://github.com/aspnet/Security/issues/158
That said, I guess this change will have to wait until .NET 4.6 is supported by dnx.

brentschmaltz:
A possible solution is for the runtime could flow the AuthenticationTicket in it's entirety,  rather than chop it up and flow parts then rehydrate them. The claimstransformation point could take action on all of the tickets that are available.

leastprivilege:
Why would you want to serialize the properties? The claims classes are already way to complicated and should be really onyl name/value pairs

PinpointTownes:
I could find 10 examples to legitimate serializing `Claim.Properties` but I can't find a reason not to do that.

leastprivilege:
That makes it clearer.

PinpointTownes:
You want a concrete example? Okay, here's one: In `AspNet.Security.OpenIdConnect.Server`, I use `Claim.Properties` to store a marker indicating whether a claim should be included in the identity token, in the access token or in both.

PinpointTownes:
And BTW, the same remarks applies to `ClaimsIdentity.Actor`, which is currently not serialized.

leastprivilege:
Fair enough - If someone (sensible) would re-create the Claims / CI / CP classes today most of these things would probably not exist. They were included back then to accommodate some of the WS*/SAML features they needed. 

I am not saying the none of them are useful - but the classes are bloated and I am not sure that full fidelity in round-tripping them makes sense. 

Same goes for custom principals - but I guess that is a separate discussion.

PinpointTownes:
> Fair enough - If someone (sensible) would re-create the Claims / CI / CP classes today most of these things would probably not exist.

Are you saying that the guys who introduced claims support in .NET were/are not sensible? :smile: 
[...]

PR #424, opened 2015-09-02
Serialize the ClaimsIdentity.BootstrapContext.

Tratcher:
#58 
@Eilon @PinpointTownes @brentschmaltz

### DIFF
--- a/src/Microsoft.AspNet.Authentication/DataHandler/TicketSerializer.cs
+++ b/src/Microsoft.AspNet.Authentication/DataHandler/TicketSerializer.cs
@@ -11,7 +11,7 @@ namespace Microsoft.AspNet.Authentication
 {
     public class TicketSerializer : IDataSerializer<AuthenticationTicket>
     {
-        private const int FormatVersion = 3;
+        private const int FormatVersion = 4;
 
         public virtual byte[] Serialize(AuthenticationTicket model)
         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Authentication/DataHandler/TicketSerializer.cs
```diff
@@ -63,6 +63,17 @@ public static void Write([NotNull] BinaryWriter writer, [NotNull] Authentication
                         WriteWithDefault(writer, claim.Issuer, DefaultValues.LocalAuthority);
                         WriteWithDefault(writer, claim.OriginalIssuer, claim.Issuer);
                     }
+
+                    var bootstrap = identity.BootstrapContext?.ToString();
```
PinpointTownes: Hum, if `BootstrapContext` is not a string, this may return the type name (which is rather useless). I think I'd probably ensure `BootstrapContext` is a string instead of using `ToString`.

PinpointTownes:
Related issue: https://github.com/aspnet/Security/issues/324

test/Microsoft.AspNet.Authentication.Test/DataHandler/TicketSerializerTests.cs
```diff
@@ -6,7 +6,6 @@
 using System.Linq;
 using System.Security.Claims;
 using Microsoft.AspNet.Http.Authentication;
-using Shouldly;
```
HaoK: Unrelated to this PR, do we want to nuke Shouldly from Authentication tests entirely at some point?

PinpointTownes: @HaoK https://github.com/aspnet/Security/issues/420 ?

Tratcher:
Updated

PinpointTownes:
Looks good :ship: :it:

Issue #307, opened 2015-06-29
Remove default tasks for OIDC notifications, perform null check before any work.


brentschmaltz:
The constructor sets no-op notifications.
```
public OpenIdConnectAuthenticationNotifications()
        {
            AuthenticationFailed = notification => Task.FromResult(0);
            AuthorizationCodeReceived = notification => Task.FromResult(0);
            MessageReceived = notification => Task.FromResult(0);
            SecurityTokenReceived = notification => Task.FromResult(0);
            SecurityTokenValidated = notification => Task.FromResult(0);
            RedirectToIdentityProvider = notification => Task.FromResult(0);
        }
```
This results in the handler doing a bunch of work for nothing. Wrap the whole thing in: 
`if (... != null ) { do work }`
instead of:
```
 var messageReceivedNotification =
     new MessageReceivedNotification<OpenIdConnectMessage, OpenIdConnectAuthenticationOptions>(Context, Options)
     {
         ProtocolMessage = message
     };

 await Options.Notifications.MessageReceived(messageReceivedNotification);
 if (messageReceivedNotification.HandledResponse)
 {
     Logger.LogVerbose(Resources.OIDCH_0002_MessageReceivedNotificationHandledResponse);
     return messageReceivedNotification.AuthenticationTicket;
 }

 if (messageReceivedNotification.Skipped)
 {
     Logger.LogVerbose(Resources.OIDCH_0003_MessageReceivedNotificationSkipped);
     return null;
 }
```

PinpointTownes:
Hum, I'm not sure about this change: is the perf gain worth it, compared to the noise it will add to the code?

If we decide to bring back an `IOpenIdConnectAuthenticationNotifications` interface with good old methods, you won't have to care about null delegates in the handler.

brentschmaltz:
What does IOpenIdConnectAuthenticationNotifications change look like?

PinpointTownes:
Exactly like the OAuth2 client middleware: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OAuth/Notifications/IOAuthAuthenticationNotifications.cs

brentschmaltz:
What is the reason for it?
How does it address this performance issue?

brentschmaltz:
@Tratcher you added a 'need-design' tag. Does that indicate some type of deliverable? If so what does it look like?

PinpointTownes:
I never said it was supposed to address a performance issue :smile: 

[...]

Issue #340, opened 2015-07-07
Create new notification for SignOut


brentschmaltz:
RedirectToIdentityProvider has a dual purpose: Redirect to IDP for signin or signout.
How about we have another notification for signout.  We may want to rename RedirectToIdentityProvider.

Eilon:
Could be done as part of the notifications cleanup work.

Also need to fix this in the cookies middleware.

Issue #327, opened 2015-07-02
Add additional Notification: AuthenticationComplete


brentschmaltz:
This would be fired just before returning from AuthenticateAsync and would be symmetric to OnMessageReceived. It would take the AuthenticationTicket that was produced.

Tratcher:
Something like this?
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Twitter/TwitterAuthenticationHandler.cs#L172

brentschmaltz:
@Tratcher It is not clear what you are suggesting.
Line 172 is: var model = await AuthenticateAsync();

PR #427, opened 2015-09-03
Notification standardization & cleanup

Tratcher:
#47 Standardize structure, rename Notifications to Events that take Contexts.
#307 Remove null checks for notifications
Clean up context generics.
Also regenerated the resource files using the standard tool.
@HaoK @muratg @Eilon @PinpointTownes 

/cc: @brentschmaltz @tushargupta51


samples/SocialSample/Startup.cs
```diff
@@ -135,16 +135,16 @@ 127.0.0.1 MsSecSample.localhost.this
                 options.ClaimsIssuer = "OAuth2-Github";
                 options.SaveTokensAsClaims = false;
                 // Retrieving user information is unique to each provider.
-                options.Notifications = new OAuthAuthenticationNotifications
+                options.Events = new OAuthAuthenticationEvents
```
HaoK: Consider OAuthAuthenticationEvents => OAuthEvents?

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationMiddleware.cs
```diff
@@ -22,9 +22,9 @@ public class CookieAuthenticationMiddleware : AuthenticationMiddleware<CookieAut
             ConfigureOptions<CookieAuthenticationOptions> configureOptions)
             : base(next, options, loggerFactory, urlEncoder, configureOptions)
         {
-            if (Options.Notifications == null)
+            if (Options.Events == null)
             {
-                Options.Notifications = new CookieAuthenticationNotifications();
+                Options.Events = new CookieAuthenticationEvents();
```
HaoK: Same thing, can we shorten to just CookieEvents?

samples/SocialSample/Startup.cs
Tratcher: No, all the things include Authentication. See OAuthAuthenticationOptions, etc...
Also https://github.com/aspnet/Security/pull/391

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
```diff
@@ -7,108 +7,77 @@
...
 
         /// <summary>
         /// Implements the interface method by invoking the related delegate method
         /// </summary>
         /// <param name="context"></param>
-        public virtual void ResponseSignIn(CookieResponseSignInContext context)
-        {
-            OnResponseSignIn.Invoke(context);
-        }
+        public virtual void ResponseSignIn(CookieResponseSignInContext context) => OnResponseSignIn(context);
```
HaoK: Can you remind me why we decided on method + delegate again?  What does the method allow that the public delegate property doesn't?

Tratcher: derivation, selective override, non-breaking addition of new events.

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
```diff
@@ -7,108 +7,77 @@
...
         /// A delegate assigned to this property will be invoked when the related method is called
         /// </summary>
-        public Func<CookieValidatePrincipalContext, Task> OnValidatePrincipal { get; set; }
+        public Func<CookieValidatePrincipalContext, Task> OnValidatePrincipal { get; set; } = context => Task.FromResult(0);
 
         /// <summary>
         /// A delegate assigned to this property will be invoked when the related method is called
         /// </summary>
-        public Action<CookieResponseSignInContext> OnResponseSignIn { get; set; }
+        public Action<CookieResponseSignInContext> OnResponseSignIn { get; set; } = context => { };
```
PinpointTownes: Have we considered making these events async?

Tratcher: Not in this PR :-). File a separate issue.

samples/SocialSample/Startup.cs
HaoK: Well it still feels clunky to me, as we have Authentication in the namespace, it seems redundant to put it in every class name as well.  Its consistently too long :)

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
```diff
@@ -7,108 +7,77 @@
...
         /// A delegate assigned to this property will be invoked when the related method is called
         /// </summary>
-        public Action<CookieResponseSignOutContext> OnResponseSignOut { get; set; }
+        public Action<CookieResponseSignOutContext> OnResponseSignOut { get; set; } = context => { };
 
         /// <summary>
         /// A delegate assigned to this property will be invoked when the related method is called
         /// </summary>
-        public Action<CookieApplyRedirectContext> OnApplyRedirect { get; set; }
+        public Action<CookieApplyRedirectContext> OnApplyRedirect { get; set; } = context => context.Response.Redirect(context.RedirectUri);
```
PinpointTownes: Have you considered removing these "default behaviors" from the events class and moving them directly to the handler?

samples/SocialSample/Startup.cs
HaoK: But I'll file another issue, as its not related to this PR really :)

Tratcher: Then why did we decline https://github.com/aspnet/Security/pull/391?

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
Tratcher: The handler doesn't know if the delegate is set or not, it doesn't call them directly, it calls the methods bellow. We could move the defaults into those methods, but I don't think there's much point to that.

samples/SocialSample/Startup.cs
HaoK: Well in that one we were changing the UseOAuthAuthentication method (and leaving the classes alone).  I'm actually suggesting the opposite now, where we leave the Use/Adds alone, and clean up all of the types instead.

Cookie/Facebook/Google/MicrosoftAccount/OIDC/JwtBearer [Authentication] is already in the namespace that they included to pickup the class no?  For example Mvc doesn't prefix all of their classes with Mvc...

HaoK:
:shipit: I'll randomize in a separate issue :)

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
HaoK: I don't remember if we have any tests exercising this, but we should probably add a common test that actually verifies this works then if we are going through the trouble of adding these everywhere...

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
HaoK: Yeah a benefit of having the methods is to hide the details from the middleware, it just can blindly call the methods.

PinpointTownes: Sure, but you could make `CookieApplyRedirectContext` derive from `BaseControlContext` and check `context.HandledResponse` and `context.SkippedToNextMiddleware` before applying the default behavior directly from the handler. This would make it much more consistent with how we control the other events.

samples/SocialSample/Startup.cs
PinpointTownes: Note: if you remove `Authentication` from the class names, we'll have at least one collision: `CookieOptions`.

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
Tratcher: Even where BaseControlContext is used, it still provides the default behavior in the event delegates. BaseControlContext is a pattern that still needs to be rationalized with the other middleware.

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
PinpointTownes: I will :smile: 
But since this PR is all about unifying the different handlers, I thought it was a good idea to align these events with the OIDC async events.

Tratcher: sure, but that will require deeper functional changes. Everything here is cosmetic.

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
PinpointTownes: Yep, since OIDC's `RedirectToIdentityProvider` or `SecurityTokenValidated` events have their default behavior directly in the handler (I'm sure they are not the only ones) :smile:

src/Microsoft.AspNet.Authentication.OpenIdConnect/Events/RedirectFromIdentityProviderContext.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Http;
+using Microsoft.IdentityModel.Protocols.OpenIdConnect;
+
+namespace Microsoft.AspNet.Authentication.OpenIdConnect
+{
+    public class RedirectFromIdentityProviderContext : BaseControlContext<OpenIdConnectAuthenticationOptions>
```
PinpointTownes: Curious: where is this context used?

PinpointTownes:
It looks fine :+1:

src/Microsoft.AspNet.Authentication.OpenIdConnect/Events/RedirectFromIdentityProviderContext.cs
Tratcher: It's not at the moment, but I kept it to use for https://github.com/aspnet/Security/issues/340

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
Tratcher: SecurityTokenValidated doesn't have any default behavior in the handler. RedirectToIdentityProvider does have default behavior in the handler, but it's a significant amount of code so I can see why it would deviate from the pattern.

src/Microsoft.AspNet.Authentication.OpenIdConnect/Events/RedirectFromIdentityProviderContext.cs
PinpointTownes: Hum, I'm not sure it's a really pertinent name then, since the ASP.NET 5 is always the initiating party when calling `Context.Authentication.SignOutAsync`: you're indeed redirected "from" the identity provider if you provided a `post_logout_redirect_uri` and if the identity provider agreed to redirect the user agent, but this doesn't happen at the same stage.

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
PinpointTownes: Erf, I meant `SecurityTokenReceived` (the rest of the method being the default behavior, actually :smile:)

I'm not opposed to having default behaviors in the events classes themselves, but the logic is clearly not consistent: for instance, if you want to add custom parameters to the final redirect URL, you can't provide a custom `OnApplyRedirect` handler to simply manipulate the URL. You also have to re-write the redirect logic the default notification had. `RedirectToIdentityProvider` on the other hand, allows you to replace the URL without needing to replicate the default logic.

src/Microsoft.AspNet.Authentication.OpenIdConnect/Events/RedirectFromIdentityProviderContext.cs
Tratcher: You're right. Actually I think it was intended as a parallel to OAuthReturnEndpointContext for use in InvokeReplyPathAsync. It would fulfill https://github.com/aspnet/Security/issues/327

I'll just remove it for now.

Issue #390, opened 2015-08-07
[OIDC] Redesign or remove nonce cookie support


Tratcher:
The nonce used during the OIDC login can be stored in a cookie or in a server side cache. The cookie support is buggy and needs to be redesigned or removed.

A simple cookie name cannot be used because it will overwrite the existing cookie if a login flow is in progress in another browser tab. To address this a unique cookie name was used per login.

Now the problem is knowing when to clean up stale cookies since we can't tell if another browser tab is still open and will eventually complete it's login flow.

The most often reported problem is in scenarios with configuration issues where the client gets into a redirect loop and creates additional nonce cookies until they get a 400 from the server saying there are too many headers. In this case the surplus of nonce cookies are a just symptom of a configuration problem and no amount of cleanup effort is likely to make the scenario succeed.

RE: https://github.com/aspnet/Security/issues/179

nenadvicentic:
Regarding one of the issues with Nonce cookie:

> IDX10311: RequireNonce is 'true' (default) but validationContext.Nonce is null. A nonce cannot be validated. If you don't need to check the nonce, set OpenIdConnectProtocolValidator.RequireNonce to 'false'.

There is a case where cookie does not get written in the HTTP response at all. And it's easy to reproduce.

In MVC controller's action (ex `/account/login`), I have 2 lines of code:

    
    httpContext.Session.Abandon(); // Cause of the problem!
    
    context.Authentication.Challenge(
    	new AuthenticationProperties(),
    	OpenIdConnectAuthenticationDefaults.AuthenticationType);
    
I first abandon current `Session` and then call `Authentication.Challenge()` to redirect user to authorization server.

As a consequence of **abandoning the session**, ASP.NET pipeline does not output `Set-Cookie: OpenIdConnect.nonce...` in the redirect response, even though it was previously set by `CookieManager`.

Browser never receives the cookie. After user gets redirected back from authorization server to the relying party, cookie is not there and `nonce` validation fails.

If I replace `Session.Abandon()` with `Session.Clear()` everything works fine.

Core of the issue seems to be related with handling of `Response.Headers` in the Microsoft ASP.NET pipeline, which has some known issues.

*IMPORTANT - To reproduce issue delete ASP.NET session cookie before hitting the page! When cookie is present, example works.*

Tratcher:
[...]

Issue #404, opened 2015-08-21
Flow 'state' independently of the OIDC message for validation


Tratcher:
To fully validate an incoming OIDC message we need to make sure the 'state' parameter matches the one we sent. Can this be flown similar to the nonce (which already has issues)?

RE:
#278 
#390

brentschmaltz:
Additionally, http://tools.ietf.org/html/rfc6819#section-5.3.5 specifies that the 'state' should be tied to the user-agent session.

Tratcher:
https://github.com/aspnet/Security/issues/278#issuecomment-132656395

Tratcher:
Design: Use the same flow as the OAuth middleware's correlation ID, plus OIDC's unique cookie name.

PR #432, opened 2015-09-04
Make the nonce cookie expire.

Tratcher:
#390 
@brentschmaltz @PinpointTownes @HaoK

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
@@ -666,7 +666,8 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
                 new CookieOptions
                 {
                     HttpOnly = true,
-                    Secure = Request.IsHttps
+                    Secure = Request.IsHttps,
+                    Expires = DateTime.UtcNow + Options.ProtocolValidator.NonceLifetime
                 });
         }
 
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationOptions.cs
@@ -169,7 +169,11 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
         /// is valid per: http://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation 
         /// </summary>
         /// <exception cref="ArgumentNullException">if 'value' is null.</exception>
-        public OpenIdConnectProtocolValidator ProtocolValidator { get; set; } = new OpenIdConnectProtocolValidator { RequireState = false };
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -667,7 +667,8 @@ private void WriteNonceCookie(string nonce)
                 new CookieOptions
                 {
                     HttpOnly = true,
-                    Secure = Request.IsHttps
+                    Secure = Request.IsHttps,
+                    Expires = DateTime.UtcNow + Options.ProtocolValidator.NonceLifetime
```
PinpointTownes: Curious (not directly related to this PR): is there a reason why `CookieOptions.Expires` is not a `DateTimeOffset`?

I guess it would be also ideal to use `ISystemClock` here.

PinpointTownes:
Nice, it should help reducing the frequency of these annoying nonce overflows :smile: 

What's the long-term plan for nonce cookies? I'm pretty sure I remember you said once they could be eventually removed.

Tratcher:
Actually, https://github.com/aspnet/Security/issues/404 requires another cookie that can't use the cache, so it's not clear if caching nonces still has much value.

PinpointTownes:
Yup, but using cookies for nonces offers no protection against replay attacks (since you can re-use the same nonce cookie as-is), which defeats the whole purpose of creating a nonce. Using a server-side store is the only way to make sure we can't replay a nonce.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
Tratcher: CookieOptions.Expires might just be an oversight. I don't see a reason. File a bug :-)

We could use ISystemClock, but I wouldn't bother for just one test. Interestingly JwtBearerAuthenticationOptions does have an ISystemClock option but does not actually use it.

Tratcher:
Isn't TVP.TokenReplayCache for replay detection?

PinpointTownes:
> Isn't TVP.TokenReplayCache for replay detection?

It is, but since there's no default implementation, it's no-op :smile: 
If you want to rely on this thing, why not removing nonce validation from the OIDC middleware?

Tratcher:
The middleware doesn't do nonce validation, it just flows the data to the underlying validator. Hence my question if we can just remove the cache.

PinpointTownes:
> The middleware doesn't do nonce validation, it just flows the data to the underlying validator. Hence my question if we can just remove the cache.

Well, it's a bit more subtle: of course it doesn't have any "replay" validation logic, but it does apply "authenticity" validation since the nonce handling relies on the data protection block: `ReadNonceCookie` will consider the nonce as missing if it was not generated by the ASP.NET app.

Tratcher:
https://github.com/aspnet/Security/issues/434

Tratcher:
*Ping* @brentschmaltz

Tratcher:
*ping* @HaoK @blowdart

blowdart:
I'm fine with this.

HaoK:
:shipit:

PR #438, opened 2015-09-08
Verify state via independent cookie.

Tratcher:
#404 
Verify the state parameter with a correlation cookie.
Removed some brittle logging tests.

@HaoK @blowdart @Eilon @brentschmaltz @tushargupta51 @muratg


PinpointTownes:
Nice, having no correlation cookie was a major security concern :+1: 

Have you considered backporting the new correlation process (that now supports concurrent requests) to the OAuth2 middleware?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -394,6 +399,11 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
                     throw new SecurityTokenException(string.Format(CultureInfo.InvariantCulture, Resources.OIDCH_0006_MessageContainsError, message.Error, message.ErrorDescription ?? "ErrorDecription null", message.ErrorUri ?? "ErrorUri null"));
                 }
 
+                if (!ValidateCorrelationId(properties))
+                {
+                    return null;
```
PinpointTownes: Is it worth logging an error before returning `null`?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -714,6 +724,66 @@ private string ReadNonceCookie(string nonce)
...
+
+            var nonceBytes = new byte[32];
+            CryptoRandom.GetBytes(nonceBytes);
+            var correlationId = TextEncodings.Base64Url.Encode(nonceBytes);
+
+            var cookieOptions = new CookieOptions
+            {
+                HttpOnly = true,
+                Secure = Request.IsHttps,
+                Expires = DateTime.UtcNow + Options.ProtocolValidator.NonceLifetime
```
PinpointTownes: State and nonce are two different things, but I guess using the same lifetime is not a bad thing.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -714,6 +724,66 @@ private string ReadNonceCookie(string nonce)
...
+
+            Response.Cookies.Append(correlationKey + correlationId, NonceProperty, cookieOptions);
+        }
+
+        private bool ValidateCorrelationId([NotNull] AuthenticationProperties properties)
+        {
+            var correlationKey = OpenIdConnectAuthenticationDefaults.CookieStatePrefix;
+
+            string correlationId;
+            if (!properties.Items.TryGetValue(
```
PinpointTownes: Were these line wraps really necessary? :trollface:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: Ah yeah, it's already logged elsewhere :+1:

Tratcher:
@PinpointTownes Do you mean the unique cookie name per sign-in? Have you ever seen anyone have issues with cookie conflicts in OAuth2?

PinpointTownes:
@Tratcher yep, the "unique cookie name per sign-in" thing.

> Have you ever seen anyone have issues with cookie conflicts in OAuth2?

Well, if this issue impacts the OIDC middleware, I don't see why it wouldn't impact OAuth2, since the flow is absolutely identical. The logs I've analyzed show the problem exist but is not that frequent, as least when using the passive mode, since there's no risk to trigger an unwanted challenge.

When using passive authentication, the bug is always due to the user clicking multiple times on the login button, which is not the most common case, but can be a problem for mobile users whose connection is not rock-solid.

Tratcher:
@PinpointTownes OAuth does have the same issue but we have never had complaints about it. I think OIDC has been more vulnerable to the issue so far because people are more likely to open multiple tabs referencing the same OIDC login server (e.g. AAD). Since the mitigation has side effects that people have also complained about, I want to see how well the changes work here before applying them to OAuth.

PinpointTownes:
> Since the mitigation has side effects that people have also complained about, I want to see how well the changes work here before applying them to OAuth.

To mitigate the side effects, setting an expiration date is already a great thing. You could also consider adopting a cookies number limit and remove "old" states/nonces when the number becomes too high. Of course, the user won't be able to complete the oldest challenges, but at least, he won't see an ugly "headers too large" error :smile:

Tratcher:
*ping* @HaoK @brentschmaltz

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectMiddlewareTests.cs
```diff
@@ -82,7 +82,8 @@ public class OpenIdConnectMiddlewareTests
                 options.Configuration = TestUtilities.DefaultOpenIdConnectConfiguration;
             });
             var transaction = await SendAsync(server, DefaultHost + Challenge);
-            transaction.SetCookie.Single().ShouldContain(OpenIdConnectAuthenticationDefaults.CookieNoncePrefix);
+            transaction.SetCookie.First().ShouldContain(OpenIdConnectAuthenticationDefaults.CookieNoncePrefix);
+            transaction.SetCookie.Skip(1).First().ShouldContain(OpenIdConnectAuthenticationDefaults.CookieStatePrefix);
         }
```
HaoK: [Nit] Can we switch to using Asserts for any new tests to make getting rid of shoudly easier?

HaoK:
:shipit:

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectMiddlewareTests.cs
Tratcher: I thought about it, but I didn't want to start doing it just for this one test. We'll do a clean sweep later.

Issue #431, opened 2015-09-03
Authentication middleware collision when registering multiple instances


PinpointTownes:
Sometimes, I wonder if you really take my remarks into consideration: https://github.com/aspnet/Security/pull/377#issuecomment-129647531 :smile: 

As expected, removing the named options without introducing an appropriate workaround resulted in a total mess, that you can easily reproduce with the `SocialSample`:

![image](https://cloud.githubusercontent.com/assets/6998306/9672994/fb687eda-529f-11e5-9ded-204518d8285d.png)

HaoK:
Oh I hear you, it was just buried in the avalanche of things that were broken.  The fix for this will probably be similar to Cookies and requires a new overload to be used when registering them multiinstance that takes an an instance of the options.  Thanks for bringing this up though :)  sounds like we have a sample for the generic OAuth but no real test coverage or usage!

PinpointTownes:
> The fix for this will probably be similar to Cookies and requires a new overload to be used when registering them multiinstance that takes an an instance of the options.

This fix is too fragile, IMHO. Having an overload taking no parameter (and retrieving the options exclusively via DI) and a second one taking a delegate (that wouldn't extract the options from DI) would be far less surprising.

HaoK:
I went with a single UseOAuth that requires an options instance to make things simpler.  Do you feel strongly that we need the DI default overload?

HaoK:
We are going to fix this once and for all by standardizing on UseXyz(XyzOptions options) with middlewares not taking IOptions for now.

brockallen:
So does this mean that middleware config in general will be moving away from ConfigureServices? I think this is a good idea (and hopefully MVC does the same).

davidfowl:
@brockallen definitely not. First, MVC isn't even a middleware :smile: . Secondly it has a ton of services that need to be added to the container.

brockallen:
Why is MVC not middleware?

HaoK:
Routing is the middleware that UseMvc actually is registering in the pipeline

HaoK:
This is similar to how UseIdentity actually is registering several instances of Cookies middleware, but there's no identity middleware either

HaoK:
081577e4f4996082a4fc28ef77c84f54c7d04183

PR #440, opened 2015-09-09
Fix OAuth Social sample

HaoK:
Like cookies OAuthAuthenticationOptions now uses instances of options.

@Tratcher @PinpointTownes this fixes things, but things still feel a bit clunky.  I'd love to nuke ConfigureOptions entirely potentially, so then we could take over the lambda to be configuring a new instance of the options instead of tweaking the one that came from DI.  Thoughts?

### DIFF
--- a/samples/SocialSample/Startup.cs
+++ b/samples/SocialSample/Startup.cs
@@ -52,17 +52,21 @@ namespace CookieSample
                 options.AppSecret = "a124463c4719c94b4228d9a240e5dc1a";
             });
 
-            app.UseOAuthAuthentication("Google-AccessToken", options =>
+            var googleOptions = new OAuthAuthenticationOptions
             {
-                options.ClientId = "560027070069-37ldt4kfuohhu3m495hk2j4pjp92d382.apps.googleusercontent.com";
-                options.ClientSecret = "n2Q-GEw9RQjzcRbU3qhfTj8f";
-                options.CallbackPath = new PathString("/signin-google-token");
-                options.AuthorizationEndpoint = GoogleAuthenticationDefaults.AuthorizationEndpoint;
-                options.TokenEndpoint = GoogleAuthenticationDefaults.TokenEndpoint;
-                options.Scope.Add("openid");
-                options.Scope.Add("profile");
-                options.Scope.Add("email");
-            });
+                AuthenticationScheme = "Google-AccessToken",
+                Caption = "Google-AccessToken",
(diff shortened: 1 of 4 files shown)


HaoK:
Fixes https://github.com/aspnet/Security/issues/431

Tratcher:
Agreed, it never made sense to me to pull these from DI

Tratcher:
:shipit: and fix the other extensions separately.

HaoK:
5bc13cbd6b3d3fcc1dc430dcb8841983d3731896

PinpointTownes:
My 2 cents:
* Removing ConfigureOptions is something I've been advocating for a long time, so yeah, nuke it in a future commit (specially since it's now unused, so there's no reason to keep it). 
* Being able to configure the options via DI was a pretty cool thing and was useful to configure per-environment settings. I'm not sure to understand why you don't want to provide an extension that uses options from DI. If you really want to remove it, I think it should be part of a larger work removing it from everywhere it is used (in MVC for instance). 
* Removing the delegate-style extension is clearly a regression from a code style perspective and was not necessary for this fix. Please consider bringing it back.

HaoK:
There's a fix larger issues mixed into these:

- ConfigureOptions existed before mostly just to package both the option name and Action, but now that name is gone, it has no purpose, so that almost certainly will get nuked soonish.

- Well its certainly not going to be removed everywhere at this point, so its just a matter of what makes sense for multi instance middlewares.  Cookies does currently support both DI/instance via 2 overloads.  They should probably be consistent (I could go either way though)

- The main reason for nuking the delegate, was because it conflicts with the DI style overload (which didn't make the cut).  But if it comes back,I'd have lean in favor of the instance one taking the delegate parameters, and requiring UseXyz to take no parameters and just use what's in DI

HaoK:
Discussed with @divega today and we decided that for now, we will go with only UseXyz(XyzOption), nuking all DI support for now for middlewares.  This lets us simplify and also remove bunch of overloads and complexity.  We can always add back sugar to make getting options from DI easier later

PinpointTownes:
Understood. But please, bring back the delegate style extension.

HaoK:
@divega can you summarize again why we decided against adding back the delegate style overload?  I know we agreed not to do it, but I can't recall exactly what our reasoning was for that part.

PinpointTownes:
@divega ping! :package:

divega:
@PinpointTownes sorry I didn't answer on this thread on time. We really wanted to reduce the number of patterns we used to pass options to middleware to the bare minimum that we actually needed. Anyway, as you already saw in https://github.com/aspnet/Security/pull/464, the lambda style overloads survived :smile:

Issue #372, opened 2015-07-22
[OIDC ]Transforming AuthenticationTicket in SecurityTokenValidated notification - and nonce checking bypass bug?!


leastprivilege:
Looking at beta5 but seems to be also the case for dev branch - 

In Katana I was able to transform the contents of the authN ticket in the SecurityTokenValidated event - especially the claims. Typically I want to strip things like iss, nonce, exp, nbf etc - because I don't want these claims in my cookie.

In the current bits - I can only do that if I call HandledResponse() in the notification - otherwise the original authN ticket will be returned. 

But if HandledResponse = true - the nonce checking is not being performed. 

Is that intentional? Is there another way I should transform the claims before they are handed off to cookie middleware.

Letting the user (unintentionally) bypass nonce checking seems like a bug to me.

brentschmaltz:
@leastprivilege we defined setting HandledResponse as telling the runtime to step aside and the ticket has been created, just use it.

Here are some options
JwtSecurityTokenHandler has a static (yes we have a work item to have instance and a static default), InboundClaimFilter. It is currently empty, any values in that set will be filtered if they match claim.Type.

OIDCAOptions.SecurityTokenValidators allows you to plug a derived JwtSecurityTokenHandler which could override CreateClaimsIdentity (sorry no delegate here).

leastprivilege:
Yea - I don't like any of these options. 

I want to be able to replace the authN ticket in SecurityTokenValidated - but after the token is actually *fully* validated (and the nonce cached is cleaned up).

Simply move the nonce handling before firing the notification and everything is fine.

brentschmaltz:
Ok, well then we might have a bug.  If you set the ticket in the SecurityTokenValidated notification we should continue on using that ticket unless you call Handled or Skipped.

leastprivilege:
Yes - that's one part to it.

The other part is that i don't get why there is an event called SecurityTokenValidated - when the token is not validated yet. And exiting there leaves protocol artifacts that cannot be easily cleaned up manually.

brentschmaltz:
The design behind exiting after notifications such as SecurityTokenReceived / Validated was to give users the ability to take over validation. It gives one a place to take complete control.
Having just reviewed HandleUnauthorizedAsync, it looks like we write the cookie nonce before calling RedirectToIdentityProvider, that needs to be adjusted for this to be symmetric and address artifacts lying around.

[...]

Issue #358, opened 2015-07-14
Extension point to give users access to user info endpoint json response


tushargupta51:
OpenIdConnectAuthenticationHandler.GetUserInformationAsync goes to UI endpoint to retrieve user claims. response is parsed into a json object. We should provide users access to this json response in case they want to retrieve some additional information.

PR #442, opened 2015-09-09
More OIDC events, better data flow

Tratcher:
Add new OIDC events:
#340 RedirectToIdentityProvider -> RedirectForAuthentication & RedirectForSignOut
#358 UserInformationReceived
#327 AuthenticationCompleted

Also, #372 make sure any mutable event state flows after the event.

@HaoK @brentschmaltz @tushargupta51 @PinpointTownes @Eilon @muratg


src/Microsoft.AspNet.Authentication.OpenIdConnect/Events/RedirectForSignOutContext.cs
```diff
@@ -0,0 +1,27 @@
...
+        public RedirectForSignOutContext([NotNull] HttpContext context, [NotNull] OpenIdConnectAuthenticationOptions options)
+            : base(context, options)
+        {
+        }
+
+        /// <summary>
+        /// Gets or sets the <see cref="OpenIdConnectMessage"/>.
+        /// </summary>
+        /// <exception cref="ArgumentNullException">if 'value' is null.</exception>
+        public OpenIdConnectMessage ProtocolMessage { get; [param: NotNull] set; }
```
PinpointTownes: Isn't NotNull discouraged for properties?

I still don't know why replacing the message was important for @brentschmaltz since it's already fully mutable.

PinpointTownes: Since these contexts are OIDC-specific, consider using more appropriate names like LogoutRequest.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -525,9 +530,11 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
             {
                 return null;
             }
+            message = securityTokenReceivedContext.ProtocolMessage;
+            var idToken = securityTokenReceivedContext.SecurityToken;
```
PinpointTownes: SecurityToken is far too generic. And actually, I wonder if it's not the access token.

src/Microsoft.AspNet.Authentication.OpenIdConnect/Events/RedirectForSignOutContext.cs
Tratcher: NotNull removed.

I actually like the standard property name across all the OIDC events.

PinpointTownes:
For the new events, I wonder if RedirectToAuthorizationEndpoint and RedirectToLogoutEndpoint wouldn't be a bit more expressive.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -467,14 +465,16 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
...
             // Redeeming authorization code for tokens
-            Logger.LogDebug(Resources.OIDCH_0038_Redeeming_Auth_Code, message.Code);
+            Logger.LogDebug(Resources.OIDCH_0038_Redeeming_Auth_Code, code);
 
-            tokenEndpointResponse = await RedeemAuthorizationCodeAsync(message.Code, authorizationCodeReceivedContext.RedirectUri);
+            tokenEndpointResponse = await RedeemAuthorizationCodeAsync(code, authorizationCodeReceivedContext.RedirectUri);
             idToken = tokenEndpointResponse.Message.IdToken;
 
-            var authorizationCodeRedeemedContext = await RunAuthorizationCodeRedeemedEventAsync(message, tokenEndpointResponse);
+            var authorizationCodeRedeemedContext = await RunAuthorizationCodeRedeemedEventAsync(message, code, tokenEndpointResponse);
```
PinpointTownes: You should consider exposing both authorization response and the token response when you can.

src/Microsoft.AspNet.Authentication.OpenIdConnect/Events/RedirectForSignOutContext.cs
HaoK: If this is an exact copy of the other Redirect context, can we not just have a single RedirectContext we use for both?  I'm really not a huge fan of all of these context objects in general.  Even if we have to add to the context in the future, I don't think sharing in this case is really going to haunt us will it?

HaoK: Or if we want unique context classes per event, we should at least consider pulling the ProtocolMessage property that appears in all of them I think to a OpenIdBaseContext?  That would at least make them empty child classes

Tratcher:
@PinpointTownes It's actually called the EndSessionEndpoint. RedirectToEndSessionEndpoint? Seems ok.

PinpointTownes:
Yeah, but this name sounds weird (that's why I never use it). That said, if you want to use the official name, I'm fine.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
Tratcher: I don't follow. We're including the token response here. What authorization response are you looking for?

Tratcher:
Updated.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
PinpointTownes: Wooops, wrong line. This remark was for the `SecurityTokenReceived` /`SecurityTokenValidated` notifications, where the token response is not added to the context. 

This is a very annoying thing that currently prevents adding the access token retrieved from the token endpoint to the final `ClaimsIdentity`, since only the authorization response is exposed. (that's why I haven't updated this sample: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/blob/vNext/samples/Mvc/Mvc.Client/Startup.cs#L101-L109)

Tratcher: SecurityTokenReceived only runs in the IdToken flow, where the token response is only available in the code flow. That said, yes, I can add the token response to the  SecurityTokenValidated event.

Tratcher:
Updated. @HaoK @PinpointTownes @brentschmaltz

PinpointTownes:
I'll test it in a real world app. Some renamed notifications look weird.

PinpointTownes:
There's a terrible bug with the authorization code flow. I'll investigate.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -484,21 +480,27 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
...
             }
 
+            message = authorizationCodeRedeemedContext.ProtocolMessage;
+
             // no need to validate signature when token is received using "code flow" as per spec [http://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation].
             var validationParameters = Options.TokenValidationParameters.Clone();
             validationParameters.ValidateSignature = false;
 
-            ticket = ValidateToken(idToken, message, properties, validationParameters, out jwt);
+            ticket = ValidateToken(message.IdToken, message, properties, validationParameters, out jwt);
```
PinpointTownes: This line is bugged: `message` actually represents the authorization response, not the token response. And since we're in a path that only supports the code flow (and not the hybrid flow), `message.IdToken` is always null. You must instead extract the `id_token` from the token response.

> error   : [Microsoft.AspNet.Authentication.OpenIdConnect.OpenIdConnectAuthentica
tionMiddleware] OIDCH_0017: Exception occurred while processing message.
System.ArgumentNullException: La valeur ne peut pas être null.
Nom du paramètre : IDX10000: The parameter 'tokenstring' cannot be a 'null' or a
n empty string.
   à Microsoft.IdentityModel.Logging.LogHelper.Throw(String message, Type except
ionType, EventLevel logLevel, Exception innerException)
   à System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler.CanReadToken(String
 tokenString)
   à Microsoft.AspNet.Authentication.OpenIdConnect.OpenIdConnectAuthenticationHa
ndler.ValidateToken(String idToken, OpenIdConnectMessage message, Authentication
Properties properties, TokenValidationParameters validationParameters, JwtSecuri
tyToken& jwt)

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
```diff
@@ -484,21 +480,27 @@ public OpenIdConnectAuthenticationHandler(HttpClient backchannel)
...
-                ticket = await GetUserInformationAsync(properties, tokenEndpointResponse.Message, ticket);
+                ticket = await GetUserInformationAsync(tokenEndpointResponse.ProtocolMessage, ticket);
+                if (ticket == null)
+                {
+                    return null;
+                }
             }
 
-            var securityTokenValidatedContext = await RunSecurityTokenValidatedEventAsync(message, ticket);
+            var securityTokenValidatedContext = await RunIdTokenValidatedEventAsync(message, ticket, tokenEndpointResponse);
```
PinpointTownes: Hum, `IdTokenValidatedContext.ProtocolMessage` will be the authorization response (basically, only containing the code and the state in this scenario). It's definitely not obvious and not really useful. Consider renaming this property to `AuthorizationResponse`.

PinpointTownes:
On a related note, you should consider cleaning this code path: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs#L374-L392

Now that the OIDC middleware has appropriate anti-XSRF countermeasures (`ValidateCorrelationId`), the `state` parameter is mandatory and having a null value should be treated as an error before calling `ValidateCorrelationId`.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectAuthenticationHandler.cs
Tratcher: Good catch, fixed.

Tratcher:
Updated.

PinpointTownes:
I briefly chatted with a friend about these notifications and we agreed they were terribly confusing. I suggest a much more aggressive renaming to match the standard names:

* `MessageReceived` => `AuthorizationResponseReceived`.

* `IdTokenReceived`/`AuthorizationCodeRedeemed`: replace them by `TokenResponseReceived`. When invoked after an authorization response (for what you name "id_token flows"), `IdTokenReceived` is basically useless as you already have  `MessageReceived`.

* `IdTokenValidated` => `AuthorizationResponseValidated` and `TokenResponseValidated`.

Tratcher:
Hmm, there's something to that... lets see...

There's a catch with MessageReceived, you don't actually know if it's an OIDC message / Authorization Response yet, it could be any form post or get query. OIDC does not require a restricted callback path.

Your IdTokenReceived argument seems to apply to AuthorizationCodeReceived as well. I think these have some value over MessageReceived because you're now in a validated OIDC flow (error, state, correlation, etc.). 

AuthorizationCodeRedeemed -> TokenResponseReceived makes sense.

IdTokenValidated -> AuthorizationResponseValidated vs  TokenResponseValidated: There is an authorization response in both flows, but you only validate it in the implicit flow. Really the event is the big "Everything" validated at the end of the flow, almost like AuthenticationCompleted but it still has the details. How about just AuthenticationValidated?

On a related note comparing the two flows, it looks a little odd that code flow runs GetUserInformationAsync before Validated, but the implicit flow runs AuthorizationCodeReceived after Validated. Of the two, I think I'd move GetUserInformationAsync later.

Tratcher:
Or maybe you could merge IdTokenReceived and AuthorizationCodeReceived into a single  AuthorizationResponseReceived.

PinpointTownes:
> There's a catch with MessageReceived, you don't actually know if it's an OIDC message / Authorization Response yet, it could be any form post or get query. OIDC does not require a restricted callback path.

Well, in its current form, `MessageReceived` is never called if the handler is not able to extract the OIDC message from the query string or the request form, which obviously limits the ability to retrieve the message from elsewhere: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L337-L340. I'm not sure that `MessageReceived` is really useful as-is. One evidence is that you were not able to support `response_mode=query` even with this hook.

At this point, yeah, we don't know yet if it's a "standard-compliant" authorization response, but on the other hand, you have no way to *guarantee* that. Not to mention that an OIDC message that contains an error or has no state parameter is still an authorization response, at least semantically speaking.

> Your IdTokenReceived argument seems to apply to AuthorizationCodeReceived as well. I think these have some value over MessageReceived because you're now in a validated OIDC flow (error, state, correlation, etc.).

Do you have a concrete use case for these events?

IMHO, `AuthorizationCodeReceived` would be more useful if it was renamed to `TokenRequestSending` and invoked just before sending the token request (to allow modifying the token request before sending it on the wire).

> Or maybe you could merge IdTokenReceived and AuthorizationCodeReceived into a single AuthorizationResponseReceived.

Yup :+1: 

> IdTokenValidated -> AuthorizationResponseValidated vs TokenResponseValidated: There is an authorization response in both flows, but you only validate it in the implicit flow. Really the event is the big "Everything" validated at the end of the flow, almost like AuthenticationCompleted but it still has the details. How about just AuthenticationValidated?

`AuthenticationValidated` doesn't look too bad, but when implementing the hybrid, we'll need to validate both the authorization response and the token response (http://openid.net/specs/openid-connect-core-1_0.html#HybridAuthResponseValidation and http://openid.net/specs/openid-connect-core-1_0.html#HybridTokenResponseValidation), so I'm not sure it's worth adding a too generic event if we need to distinguish both steps.

> On a related note comparing the two flows, it looks a little odd that code flow runs GetUserInformationAsync before Validated, but the implicit flow runs AuthorizationCodeReceived after Validated. Of the two, I think I'd move GetUserInformationAsync later.

Yeah, good idea :+1:

Tratcher:
The ODICMessage constructor doesn't do any message validation, it would work with almost any query or form post, so MessageReceived will fire for most requests. Listening to it is the only place developers can pre-process or filter incoming messages.

I realized that AuthorizationCodeReceived is called at different places in both flows. In code flow it's just before redeeming the code, but in implicit/hybrid it is your chance to redeem the code yourself. I'll have to think about how to rationalize that. They may need to be two different events.

Tratcher:
Updated PR.

PinpointTownes:
> The ODICMessage constructor doesn't do any message validation, it would work with almost any query or form post, so MessageReceived will fire for most requests. Listening to it is the only place developers can pre-process or filter incoming messages.

I never said it did (luckily, or I wouldn't use it massively :smile:). I just said that an OIDC message posted at client's `redirect_uri` endpoint is semantically an authorization response, even if it doesn't have the parameters you expect.

> OIDC does not require a restricted callback path.

Talking about that, I wonder if this feature is still useful. IIRC, your main argument to keep it was to support "unsolicited assertions". With `state` now being mandatory, it's not possible anymore, since responses lacking a valid `state` - that can only be produced by the client app - will be rejected.

Tratcher:
https://github.com/aspnet/Security/issues/455

Tratcher:
I think the hybrid flow needs more work, but I'll call this good for now. @HaoK @brentschmaltz Can you take a quick pass?

https://github.com/aspnet/Security/issues/456

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -391,6 +387,26 @@ public OpenIdConnectHandler(HttpClient backchannel)
                     _configuration = await Options.ConfigurationManager.GetConfigurationAsync(Context.RequestAborted);
                 }
 
+                Logger.LogDebug("Authorization response received.");
+                var authorizationResponseReceivedContext = new AuthorizationResponseReceivedContext(Context, Options)
```
HaoK: nit: this is one really long local var name...just saying :)

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -846,68 +868,71 @@ private AuthenticationProperties GetPropertiesFromState(string state)
             return authorizationCodeReceivedContext;
         }
 
-        private async Task<AuthorizationCodeRedeemedContext> RunAuthorizationCodeRedeemedEventAsync(OpenIdConnectMessage message, OpenIdConnectTokenEndpointResponse tokenEndpointResponse)
+        private async Task<TokenResponseReceivedContext> RunTokenResponseReceivedEventAsync(OpenIdConnectMessage message, OpenIdConnectTokenEndpointResponse tokenEndpointResponse)
         {
-            Logger.LogDebug(Resources.OIDCH_0042_AuthorizationCodeRedeemed, message.Code);
-            var authorizationCodeRedeemedContext = new AuthorizationCodeRedeemedContext(Context, Options)
+            Logger.LogDebug("Token response received.");
```
HaoK: Are you meaning to nuke all of the log messages or just not add any new ones?

HaoK:
:shipit:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -350,36 +347,35 @@ public OpenIdConnectHandler(HttpClient backchannel)
...
                     return null;
                 }
+                message = messageReceivedContext.ProtocolMessage;
 
-                var properties = new AuthenticationProperties();
-
-                // if state is missing, just log it
-                if (string.IsNullOrEmpty(message.State))
+                // if any of the error fields are set, throw error null
+                if (!string.IsNullOrEmpty(message.Error))
```
PinpointTownes: Consider moving this block after the `state` check to reject requests missing a valid `state` since it must be returned by the server even on errored responses:

http://openid.net/specs/openid-connect-core-1_0.html#AuthError
https://tools.ietf.org/html/rfc6749#section-4.1.2.1

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -350,36 +347,35 @@ public OpenIdConnectHandler(HttpClient backchannel)
...
+                if (string.IsNullOrEmpty(message.State))
+                {
+                    Logger.LogError(Resources.OIDCH_0004_MessageStateIsNullOrEmpty);
+                    return null;
                 }
 
-                // if any of the error fields are set, throw error null
-                if (!string.IsNullOrEmpty(message.Error))
+                // if state exists and we failed to 'unprotect' this is not a message we should process.
+                var properties = Options.StateDataFormat.Unprotect(Uri.UnescapeDataString(message.State));
```
PinpointTownes: Is `Uri.UnescapeDataString` needed here? The OAuth2 middleware doesn't use it: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs#L100

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: We will eventually de-localize all the log statements. So far I'm just de-localizing the ones that need updating.

Issue #443, opened 2015-09-09
Certificate verification tests do not work on Mono


Tratcher:
All of the custom certificate verification code fails on Mono.
https://travis-ci.org/aspnet/Security/builds/79531362
CertificateSubjectKeyIdentifierValidator
CertificateSubjectPublicKeyInfoValidator - uses PInvokes
CertificateThumbprintValidator

The public key validator won't ever work because it PInvokes. The other two might work if we had appropriate certs to test with.

Consider removing these. Only Twitter uses them and removing them will allow us to remove the associated BackchannelCertificateValidator and related complexity on all the middleware. https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Twitter/TwitterAuthenticationOptions.cs#L66
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Twitter/TwitterAuthenticationMiddleware.cs#L101-L110

@blowdart @muratg

blowdart:
This makes me sad, but I won't stand in its way.

davidfowl:
@blowdart wrote this code right?

blowdart:
Yes I did. It's horribly windows centric, and I doubt there's a way to make it cross platform as neither mono or core will expose wrappers for the cert structures I need without us producing X509Certificate3.

davidfowl:
Let's nuke it then.

PR #444, opened 2015-09-10
Remove custom certificate validators.

Tratcher:
#443 This enables testing on Travis.
@HaoK @muratg @davidfowl


HaoK:
Much cleaner too :)  :shipit:

Issue #434, opened 2015-09-04
Remove the nonce distributed cache


Tratcher:
OIDC has two different ways of tracking out of band nonces, via cookies or IDistributedCache. Recent changes have reduced the value of the cache approach (e.g. we need a cookie anyways for state validation, so now it can just be used to reduce response sizes). Rather than trying to toggle between cookies and cache in the middleware, consider factoring out an INonceManager and provide separate implementations for cookie vs cache. Something like:
```
interface INonceManager
{
     Task ApplyNonce(HttpContext, string);
     Task<string> RetrieveNonce(HttpContext);
}
```

Tratcher:
Recommendation: Remove the distributed cache and go back to the cookie approach. We can add the extensibility later if required.

PR #446, opened 2015-09-11
Remove the nonce cache.

Tratcher:
#434 
@HaoK @PinpointTownes @brentschmaltz 
cc: @Eilon @muratg

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
@@ -16,7 +16,6 @@ using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Authentication;
 using Microsoft.AspNet.Http.Features.Authentication;
-using Microsoft.Framework.Caching.Distributed;
 using Microsoft.Framework.Internal;
 using Microsoft.Framework.Logging;
 using Microsoft.IdentityModel.Protocols.OpenIdConnect;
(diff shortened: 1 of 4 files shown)


Eilon:
:shipit:

PinpointTownes:
If you really want to remove the server-side cache, I won't fight for it. But you have to realize that this nonce feature offers absolutely no protection against replay attacks... it's basically pointless here.

Tratcher:
@PinpointTownes http://openid.net/specs/openid-connect-core-1_0.html#NonceNotes
The use of the cookie is specifically recommended to keep a 3rd party from resubmitting the same token. There's also the TVP.TokenReplayCache.

PinpointTownes:
> The use of the cookie is specifically recommended to keep a 3rd party from resubmitting the same token. 

Yes, but if you don't check the nonce against already-used nonces, you can't detect replays. Again, preventing replay attacks is not mandatory (for both the code and the implicit flows), but clearly recommended by the specs.

> There's also the TVP.TokenReplayCache.

The problem with `TokenReplayCache` is that there's no default implemention (unlike WIF, that had a default in-memory implementation). Do you really think someone will ever implement it? Setting a bool to `true` was much easier :smile:

Tratcher:
You forget the part about setting up a distributed cache.

PinpointTownes:
Ah yeah, `services.AddCaching()` or (`services.AddRedisCache()` for a real distributed implementation) :smile:

PinpointTownes:
The more I think about it, the more I realize that keeping `IDistributedCache` is not harmful. But I think it needs a better name to clearly indicates what it can be used for. What about renaming `CacheNonces` to `DetectReplayedTokens` (the name used by WIF) or `DetectReplayAttacks`?

Tratcher:
@brentschmaltz? This still seems redundant with TokenReplayCache, no?

Tratcher:
I think we're still going to remove it for now.

PR #452, opened 2015-09-16
Add new CookiePolicy middleware

HaoK:
Fixes https://github.com/aspnet/HttpAbstractions/issues/42

cc @Tratcher @divega @lodejard @davidfowl @blowdart 

Note: one quirk of the new instance based options Use methods .  `Configure<CookiePolicyOptions>` in ConfigureServices will ALWAYS be ignored silently.

### DIFF
--- a/src/Microsoft.AspNet.CookiePolicy/CookiePolicyMiddleware.cs
+++ b/src/Microsoft.AspNet.CookiePolicy/CookiePolicyMiddleware.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Http;
@@ -35,7 +36,7 @@ namespace Microsoft.AspNet.CookiePolicy
         {
             public CookiesWrapperFeature(HttpContext context, CookiePolicyOptions options)
             {
-                Wrapper = new CookiesWrapper(context, options, context.Response.Cookies);
+                Wrapper = new CookiesWrapper(context, options);
             }
 
             public IResponseCookies Wrapper { get; }
@@ -51,16 +52,23 @@ namespace Microsoft.AspNet.CookiePolicy
 
         private class CookiesWrapper : IResponseCookies
         {
-            public CookiesWrapper(HttpContext context, CookiePolicyOptions options, IResponseCookies cookies)
+            public CookiesWrapper(HttpContext context, CookiePolicyOptions options)
             {
                 Context = context;
-                Cookies = cookies;
+                Feature = context.Features.Get<IResponseCookiesFeature>() ?? new ResponseCookiesFeature(context.Features);
                 Policy = options;
             }
 
             public HttpContext Context { get; }
 
-            public IResponseCookies Cookies { get; }
+            public IResponseCookiesFeature Feature { get; }
+
+            public IResponseCookies Cookies {
+                get
+                {
+                    return Feature.Cookies;
+                }
+            }
 
             public CookiePolicyOptions Policy { get; }
 
@@ -74,13 +82,21 @@ namespace Microsoft.AspNet.CookiePolicy
                 if (PolicyRequiresCookieOptions() || Policy.OnAppendCookie != null)
                 {
                     Append(key, value, new CookieOptions());
-                    return;
                 }
-                Cookies.Append(key, value);
+                else
+                {
+                    Cookies.Append(key, value);
+
+                }
             }
 
             public void Append(string key, string value, CookieOptions options)
             {
+                if (options == null)
+                {
+                    throw new ArgumentNullException(nameof(options));
+                }
+
                 ApplyPolicy(options);
                 if (Policy.OnAppendCookie != null)
                 {
@@ -97,20 +113,20 @@ namespace Microsoft.AspNet.CookiePolicy
                 if (PolicyRequiresCookieOptions() || Policy.OnDeleteCookie != null)
                 {
                     Delete(key, new CookieOptions());
-                    return;
                 }
-
-                if (Policy.OnDeleteCookie != null)
+                else
                 {
-                    var context = new DeleteCookieContext(Context, options: null, name: key);
-                    Policy.OnDeleteCookie(context);
-                    key = context.CookieName;
+                    Cookies.Delete(key);
                 }
-                Cookies.Delete(key);
             }
 
             public void Delete(string key, CookieOptions options)
             {
+                if (options == null)
+                {
+                    throw new ArgumentNullException(nameof(options));
+                }
+
                 ApplyPolicy(options);
                 if (Policy.OnDeleteCookie != null)
                 {
@@ -133,6 +149,8 @@ namespace Microsoft.AspNet.CookiePolicy
                         break;
                     case SecurePolicy.None:
                         break;
+                    default:
+                        throw new InvalidOperationException();
                 }
                 switch (Policy.HttpOnly)
                 {
@@ -141,6 +159,8 @@ namespace Microsoft.AspNet.CookiePolicy
                         break;
                     case HttpOnlyPolicy.None:
                         break;
+                    default:
+                        throw new InvalidOperationException();
                 }
             }
         }
--- a/src/Microsoft.AspNet.CookiePolicy/CookiePolicyMiddleware.cs
+++ b/src/Microsoft.AspNet.CookiePolicy/CookiePolicyMiddleware.cs
@@ -133,7 +133,6 @@ namespace Microsoft.AspNet.CookiePolicy
                 Cookies.Delete(key, options);
             }
 
-            // Returns true if cookie options need to be applied
             private void ApplyPolicy(CookieOptions options)
             {
                 switch (Policy.Secure)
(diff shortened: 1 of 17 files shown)


src/Microsoft.AspNet.CookiePolicy/CookiePolicyMiddleware.cs
```diff
@@ -0,0 +1,157 @@
...
+        {
+            Options = options;
+            _next = next;
+        }
+
+        public CookiePolicyOptions Options { get; set; }
+
+        public Task Invoke(HttpContext context)
+        {
+            // REVIEW: Do we need to check if there is a Cookie feature already present like SendFile??
```
HaoK: This piece needs review, is it safe to blindly set the feature?

PinpointTownes:
Naive question: why submitting this thing here and not in aspnet/HttpAbstractions or aspnet/BasicMiddleware?

Tratcher:
@PinpointTownes HttpAbstractions can't have any real middleware due to repo layering. Hosting depends on HttpAbstractions, so HttpAbstractions can't have any web app samples or TestHost tests.  BasicMiddleware was considered, it was a toss up between it and Security. I think Security only won because this feature may help make your site more secure.

src/Microsoft.AspNet.CookiePolicy/CookiePolicyMiddleware.cs
Tratcher: Consider wrapping the existing feature (if any) rather than just wrapping context.Response.Cookies. If none, default to a new ResponseCookiesFeature.

src/Microsoft.AspNet.CookiePolicy/CookiePolicyMiddleware.cs
```diff
@@ -0,0 +1,157 @@
...
+            {
+                if (PolicyRequiresCookieOptions())
+                {
+                    Append(key, value, new CookieOptions());
+                    return;
+                }
+
+                if (Policy.OnAppendCookie != null)
+                {
+                    var context = new AppendCookieContext(Context, options: null, name: key, value: value);
```
Tratcher: What if they want to add options in the callback?

src/Microsoft.AspNet.CookiePolicy/CookiePolicyMiddleware.cs
```diff
@@ -0,0 +1,157 @@
...
+                    Policy.OnAppendCookie(context);
+                    key = context.CookieName;
+                    value = context.CookieValue;
+                }
+                Cookies.Append(key, value);
+            }
+
+            public void Append(string key, string value, CookieOptions options)
```
Tratcher: argument validation?

src/Microsoft.AspNet.CookiePolicy/CookiePolicyMiddleware.cs
```diff
@@ -0,0 +1,157 @@
...
+
+            public void Delete(string key)
+            {
+                if (PolicyRequiresCookieOptions())
+                {
+                    Delete(key, new CookieOptions());
+                    return;
+                }
+
+
```
Tratcher: nit: empty line

src/Microsoft.AspNet.CookiePolicy/CookiePolicyMiddleware.cs
Tratcher: I think you'd always call the other overload if the callback is set.

src/Microsoft.AspNet.CookiePolicy/CookiePolicyMiddleware.cs
```diff
@@ -0,0 +1,157 @@
...
+                        break;
+                }
+                switch (Policy.HttpOnly)
+                {
+                    case HttpOnlyPolicy.Always:
+                        options.HttpOnly = true;
+                        break;
+                    case HttpOnlyPolicy.None:
+                        break;
+                }
```
Tratcher: default throw?

src/Microsoft.AspNet.CookiePolicy/project.json
```diff
@@ -0,0 +1,18 @@
+{
+    "version": "1.0.0-*",
+    "description": "ASP.NET 5 cookie policy classes.",
+    "repository": {
+        "type": "git",
+        "url": "git://github.com/aspnet/security"
+    },
+    "dependencies": {
+        "Microsoft.AspNet.Http.Abstractions": "1.0.0-*",
+        "Microsoft.AspNet.Http.Features": "1.0.0-*",
```
Tratcher: Remove transitive dependencies on Abstractions, Features.

test/Microsoft.AspNet.CookiePolicy.Test/CookiePolicyTests.cs
```diff
@@ -0,0 +1,182 @@
...
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.TestHost;
+using Xunit;
+
+namespace Microsoft.AspNet.CookiePolicy.Test
+{
+    public class CookiePolicyTests
+    {
+        [Fact]
+        public async Task UseMapToTestSecurePolicyAndHttpOnlyModes()
```
Tratcher: This is five distinct tests, split it.

test/Microsoft.AspNet.CookiePolicy.Test/project.json
```diff
@@ -0,0 +1,21 @@
...
+        "Microsoft.AspNet.TestHost": "1.0.0-*",
+        "Microsoft.Framework.DependencyInjection": "1.0.0-*",
+        "Moq": "4.2.1312.1622",
+        "xunit.runner.aspnet": "2.0.0-aspnet-*"
+    },
+    "commands": {
+        "test": "xunit.runner.aspnet"
+    },
+    "frameworks": {
+        "dnx451": {
```
Tratcher: Add dnxcore50. no Moq.

src/Microsoft.AspNet.CookiePolicy/CookiePolicyMiddleware.cs
HaoK: There didn't appear to be any argument validation in the base HttpAbstractions ResponseCookies so I just mimiced that behavior since we are just wrapping no?

Tratcher: There's at least a NotNull check on options.

HaoK:
Updated PR, wrapping Cookies feature now and fixed other feedback

src/Microsoft.AspNet.CookiePolicy/CookiePolicyMiddleware.cs
```diff
@@ -0,0 +1,166 @@
...
+            }
+        }
+
+        private class CookiesWrapper : IResponseCookies
+        {
+            public CookiesWrapper(HttpContext context, CookiePolicyOptions options)
+            {
+                Context = context;
+                Feature = context.Features.Get<IResponseCookiesFeature>() ?? new ResponseCookiesFeature(context.Features);
```
Tratcher: It would be clearer that this was nesting if the Get was in Invoke and passed through.

src/Microsoft.AspNet.CookiePolicy/CookiePolicyMiddleware.cs
```diff
@@ -0,0 +1,166 @@
...
+                Policy = options;
+            }
+
+            public HttpContext Context { get; }
+
+            public IResponseCookiesFeature Feature { get; }
+
+            public IResponseCookies Cookies {
```
Tratcher: net: bracket formatting

src/Microsoft.AspNet.CookiePolicy/CookiePolicyMiddleware.cs
```diff
@@ -0,0 +1,166 @@
...
+            public void Append(string key, string value)
+            {
+                if (PolicyRequiresCookieOptions() || Policy.OnAppendCookie != null)
+                {
+                    Append(key, value, new CookieOptions());
+                }
+                else
+                {
+                    Cookies.Append(key, value);
+
```
Tratcher: nit: empty line.

test/Microsoft.AspNet.CookiePolicy.Test/CookiePolicyTests.cs
```diff
@@ -0,0 +1,262 @@
...
+                    throw new NotImplementedException();
+                }
+            }
+        }
+
+        private TestServer CreateTestServer()
+        {
+            return TestServer.Create(app =>
+            {
+                app.Map("/secureAlways", map =>
```
Tratcher: Each of these maps covers just one test, no? So why not just it up in the individual tests? There's no sharing.

Tratcher:
:shipit:

test/Microsoft.AspNet.CookiePolicy.Test/CookiePolicyTests.cs
HaoK: Well my preference would be to do these all via one helper eventually so all of this boilerplate code goes away, and each test would just call something like a one liner feeding in the block that goes in the map, and a second lambda for the assert delegate.  But I don't have time for cleaning that up right now :)

HaoK: But that's why the factoring is the way it is right now, I want every test to look something like RunTest(configurePolicyOptions, configureMiddleware, configureAsserts)

HaoK:
Updated with the test refactoring, I'm thinking this will be how the security tests end up looking too, as its similar already

Tratcher:
Nice, :shipit:

HaoK:
5cc1fea40045041f77a6bb2bd23e231e24f82da7

PR #453, opened 2015-09-16
Removed references to Shouldly

jmloeffler:
* removed Shouldy as a dependency from all tests
* could not enable coreclr50 due to remaining incompatibility with moq, Uri.GetLeftPart(...), UriPartial, et al

#420

### DIFF
--- a/test/Microsoft.AspNet.Authentication.Test/Cookies/CookieMiddlewareTests.cs
+++ b/test/Microsoft.AspNet.Authentication.Test/Cookies/CookieMiddlewareTests.cs
@@ -19,7 +19,6 @@ using Microsoft.AspNet.Http.Authentication;
 using Microsoft.AspNet.Http.Features.Authentication;
 using Microsoft.AspNet.TestHost;
 using Microsoft.Framework.DependencyInjection;
-using Shouldly;
 using Xunit;
 
 namespace Microsoft.AspNet.Authentication.Cookies
@@ -33,7 +32,7 @@ namespace Microsoft.AspNet.Authentication.Cookies
             {
             });
             var response = await server.CreateClient().GetAsync("http://example.com/normal");
-            response.StatusCode.ShouldBe(HttpStatusCode.OK);
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
         }
 
         [Theory]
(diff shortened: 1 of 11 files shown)


davidfowl:
@jmloeffler how many tests had that issue? We could skip those and file the appropriate bugs.

jmloeffler:
It's not clear why the appveyor build failed.  Running `dnx test` from test\Microsoft.AspNet.Authentication.Test works for me.  Any guidance?

jmloeffler:
I've got a branch with my spike trying to enable dnxcore50.  I got pretty close.  I'll put together a list of things that seemed to be in the way and post it here.

davidfowl:
@jmloeffler would you be willing to send that in another PR? I want to get this change in. I'll try it locally to see if appveyor was a random failure.

davidfowl:
Works for me as well. Might be a random failure.

jmloeffler:
I just found an issue with my PR anyway where so I'm going to add a commit anyhow.  That should trigger a new build and we'll see what happens then.

jmloeffler:
@davidfowl Do you want me to submit a separate PR for enabling `dnxcore50`?  I think that is what you were suggesting but I want to be clear.  It is a broken build but if you see value in it then it does show what remains to be worked if we want all of the test projects to be compatible.  I believe Microsoft.AspNet.Authentication.Tests is the only project without compatibility.  This commit message shows what remains as well if that is easier: 6248e21a3482e571820efabd3fc888954c7a4066

Tratcher:
@jmloeffler I've already removed many of those in a related PR: https://github.com/aspnet/Security/pull/442/files#diff-63e5a4ec33f20ffdc309e9cd73b1c6f4

jmloeffler:
@Tratcher Yeah that was a nice big block of dead code hiding a dependency on Moq.  I suppose after your PR goes through the commit I referred to will be moot (so there's no reason for me to PR it) and just the items I pointed out in the commit message will remain.  I'm not sure what to do about those.

Tratcher:
GetLeftPart is an easy one, it just says give me the Uri up to and including the specified part. So GetLeftPart(UriPartial.Path) just means leave off the query and fragment. You can do something similar using GetComponents or UriBuilder.

jmloeffler:
@davidfowl Do you want to create a separate issue for enabling `dnxcore50`?  It seems like that was a secondary goal for #420, which can now be closed.

Tratcher:
@jmloeffler Go for it.

davidfowl:
That would be great

jmloeffler:
Issue #454 created

Issue #433, opened 2015-09-04
Port SaveTokensAsClaims to the OpenID Connect middleware


PinpointTownes:
In one of my recent PRs, I added a new `SaveTokensAsClaims` option for the OAuth2 generic middleware: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OAuth/OAuthAuthenticationOptions.cs#L111-L117.

Having a similar option for the OIDC client middleware would be great! :smile:

Eilon:
@PinpointTownes do you want to send a PR for this?

PinpointTownes:
Yep, you can count on me. I'll start working on it when @Tratcher is done with his PR.

Eilon:
Awesome, thanks!

Eilon:
@PinpointTownes let us know if you are getting close to submitting this - we have 1 week until the end of this release.

PinpointTownes:
@Eilon nope, I haven't started working on it, but in my defense, @Tratcher merged his PR yesterday :smile: 

Unless you want to assign it to someone else, I'll try to submit a first prototype tomorrow.

Eilon:
Oh no hurry at all. Just wanted to let you know our schedule!

PR #463, opened 2015-09-17
Port SaveTokensAsClaims to the OpenID Connect middleware

PinpointTownes:
Fixes https://github.com/aspnet/Security/issues/433.

/cc @Eilon @HaoK @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
@@ -89,6 +89,13 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
                     message.PostLogoutRedirectUri = Options.PostLogoutRedirectUri;
                 }
 
+                if (!string.IsNullOrEmpty(Options.SignInScheme))
+                {
+                    var principal = await Context.Authentication.AuthenticateAsync(Options.SignInScheme);
+
+                    message.IdTokenHint = principal?.FindFirst(OpenIdConnectParameterNames.IdToken)?.Value;
+                }
+
                 var redirectContext = new RedirectContext(Context, Options)
                 {
                     ProtocolMessage = message
@@ -504,6 +511,12 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
             }
             ticket = authenticationValidatedContext.AuthenticationTicket;
 
+            if (Options.SaveTokensAsClaims)
+            {
+                // Persist the tokens extracted from the token response.
+                SaveTokens(ticket.Principal, tokenEndpointResponse.ProtocolMessage, saveRefreshToken: true);
+            }
+
             if (Options.GetClaimsFromUserInfoEndpoint)
             {
                 Logger.LogDebug(Resources.OIDCH_0040_Sending_Request_UIEndpoint);
@@ -548,7 +561,27 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
                 {
                     return null;
                 }
+                message = authorizationCodeReceivedContext.ProtocolMessage;
                 ticket = authorizationCodeReceivedContext.AuthenticationTicket;
+
+                if (Options.SaveTokensAsClaims)
+                {
+                    // TODO: call SaveTokens with the token response and set
+                    // saveRefreshToken to true when the hybrid flow is fully implemented.
+                    SaveTokens(ticket.Principal, message, saveRefreshToken: false);
+                }
+            }
+            // Implicit Flow
+            else
+            {
+                if (Options.SaveTokensAsClaims)
+                {
+                    // Note: don't save the refresh token when it is extracted from the authorization
+                    // response, since it's not a valid parameter when using the implicit flow.
+                    // See http://openid.net/specs/openid-connect-core-1_0.html#Authentication
+                    // and https://tools.ietf.org/html/rfc6749#section-4.2.2.
+                    SaveTokens(ticket.Principal, message, saveRefreshToken: false);
+                }
             }
 
             return ticket;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -89,6 +89,12 @@ public OpenIdConnectHandler(HttpClient backchannel)
                     message.PostLogoutRedirectUri = Options.PostLogoutRedirectUri;
                 }
 
+                var ticket = await HandleAuthenticateOnceAsync();
+                if (ticket != null)
+                {
+                    message.IdTokenHint = ticket.Principal?.FindFirst(OpenIdConnectParameterNames.IdToken)?.Value;
```
PinpointTownes: @leastprivilege I know you're massively using the `RedirectToIdentityProvider` event to set the `id_token_hint` in your samples. Any objection if I add it directly to the OIDC handler?

leastprivilege: Doesn't this have a pretty strong assumption how the original id token was preserved.  Isn't the Katana middleware using the bootstrap context for the same purpose?

leastprivilege: maybe I am missing the context

leastprivilege: There are btw situations where you don't want to send the id_token on logout - I had a couple of cases where it was too big for the query string and IE bombed.

PinpointTownes: > Doesn't this have a pretty strong assumption how the original id token was preserved. Isn't the Katana middleware using the bootstrap context for the same purpose?

Yep, but this feature is disabled by default (`TokenValidationParameters.SaveSigninToken` is `false`) and we'd have to parse/validate the id_token a second time since it'd be stored in the BC as a string (https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs#L639).

PinpointTownes: > There are btw situations where you don't want to send the id_token on logout - I had a couple of cases where it was too big for the query string and IE bombed.

True. Maybe a new option - `FlowIdentityTokenHint` - would help mitigating this issue?

FYI: this middleware now supports POST logout requests via an auto-post form, which solves this kind of issues. But since it's not supported everywhere, I guess the additional option is worth it.

leastprivilege: OK. Sounds good.

As long as the RedirectToIdentityProvider notification is not going away (for both directions).

PinpointTownes: Actually, `RedirectToIdentityProvider` is already gone... replaced by `RedirectToAuthenticationEndpoint` and `RedirectToEndSessionEndpoint` :smile:

leastprivilege: What's the latter?

PinpointTownes: It's like the old `RedirectToIdentityProvider`, but it's only invoked for signout.

leastprivilege: got it.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -549,6 +561,25 @@ public OpenIdConnectHandler(HttpClient backchannel)
                     return null;
                 }
                 ticket = authorizationCodeReceivedContext.AuthenticationTicket;
+
```
Tratcher: `message = authenticationValidatedContext.ProtocolMessage;` since you're using it after the event.

PinpointTownes: Okay. Note that this change will be useless whence you implement the hybrid flow, since you'll have to use the token response and not the authorization response at L. 569.

Tratcher:
Just the one minor comment. Then :shipit:

PinpointTownes:
@Tratcher should I add the option mentioned in https://github.com/aspnet/Security/pull/463#discussion_r39830105 to help mitigating the URL length issues mentioned by @leastprivilege?

Tratcher:
How big is the hint?

Note the hint probably won't flow if you're using separate app cookies.

PinpointTownes:
> How big is the hint?

The hint corresponds to the identity token returned by the authorization server during the authorization response or the token response, so potentially big.

> Note the hint probably won't flow if you're using separate app cookies.

Instead, should we call the authentication manager to resolve the identity returned by the cookie middleware?

Tratcher:
> Instead, should we call the authentication manager to resolve the identity returned by the cookie middleware?
You can try, but the app identity is unlikely to maintain the hint. Now that I think about it, HandleAuthenticateOnceAsync is unlikely to ever provide a result in SignOut.

PinpointTownes:
> You can try, but the app identity is unlikely to maintain the hint.

It works.

> Now that I think about it, HandleAuthenticateOnceAsync is unlikely to ever provide a result in SignOut.

Yeah, I realized that too when asking the question :+1: 
When I tested this feature, I was calling `SignOutAsync` from the callback path, where `HandleAuthenticateOnceAsync` still provides a value. I replaced it by a much more realistic test and updated the code in the handler:

```csharp
app.Map("/signout", map => {
    map.Run(async context => {
        var properties = new AuthenticationProperties {
            RedirectUri = "/"
        };

        await context.Authentication.SignOutAsync(OpenIdConnectDefaults.AuthenticationScheme, properties);
    });
});
```

Tratcher:
Let's skip the extra option for now, there are other ways it can be dealt with (e.g. in the redirect event, disabling SaveTokensAsClaims, etc.). The hint still won't work with app cookies, but I think only the dev could make that work in the redirect event.
:shipit:

Tratcher:
Rebased and merged.

poke:
Just as a big FYI to anyone who is using an on-premise AD FS for OpenID Connect, and as a personal reminder (because this was quite painful to figure out…):

You need to **enable `OpenIdConnectOptions.SaveTokens`** if you want the sign-out process to work properly. **Otherwise ADFS will refuse to redirect back** to the configured `SignedOutRedirectUri` and you are stuck on the ADFS logout page.

Apparently, this was changed somewhat recently (probably a few months ago, because it definitely was working for me before). When handling the OAuth logout, ADFS will now verify the logout URL against the configured redirect URIs. In order to do that, it will require the client id. Unfortunately, the only way it can receive the client id is as part of the `id_token_hint`. So we cannot just pass a `client_id` argument but instead have to pass the `id_token_hint` which also requires us to save the token. Exactly that is what this PR is about, so we just have to enable `SaveTokens` in order to allow ADFS to redirect back.

PR #464, opened 2015-09-18
Remove IOptions from middlewares, use instances directly

HaoK:
Fixes https://github.com/aspnet/Security/issues/431

- Preserved UseXyz(options => { }) to simply be sugar for new XyzOptions() and running the lambda
- Also removed dead sync codepaths from ClaimsTransformation

Note: this is a breaking change, current Startups that are configuring auth in ConfigureServices will all be broken (silently)

cc @divega @Tratcher


src/Microsoft.AspNet.Authentication.Facebook/FacebookAppBuilderExtensions.cs
```diff
@@ -18,10 +17,22 @@ public static class FacebookAppBuilderExtensions
         /// </summary>
         /// <param name="app">The <see cref="IApplicationBuilder"/> passed to the configure method.</param>
         /// <returns>The updated <see cref="IApplicationBuilder"/>.</returns>
-        public static IApplicationBuilder UseFacebookAuthentication([NotNull] this IApplicationBuilder app, Action<FacebookOptions> configureOptions = null)
+        public static IApplicationBuilder UseFacebookAuthentication([NotNull] this IApplicationBuilder app, FacebookOptions options)
```
PinpointTownes: The `options` parameter is not guarded against null values (nor decorated with `NotNull`) but `FacebookMiddleware`'s `options` parameter has `NotNull` applied. Is it worth making everything consistent?
The same remark seems to apply to all the authentication middleware.

src/Microsoft.AspNet.Authentication.Facebook/FacebookAppBuilderExtensions.cs
```diff
@@ -18,10 +17,22 @@ public static class FacebookAppBuilderExtensions
...
+            return app.UseMiddleware<FacebookMiddleware>(options);
+        }
+
+        /// <summary>
+        /// Authenticate users using Facebook.
+        /// </summary>
+        /// <param name="app">The <see cref="IApplicationBuilder"/> passed to the configure method.</param>
+        /// <param name="configureOptions">Configures the options.</param>
+        /// <returns>The updated <see cref="IApplicationBuilder"/>.</returns>
+        public static IApplicationBuilder UseFacebookAuthentication([NotNull] this IApplicationBuilder app, Action<FacebookOptions> configureOptions)
```
PinpointTownes: `[NotNull]` on the `configureOptions` since a null value will cause a NRE?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectExtensions.cs
```diff
@@ -18,10 +18,15 @@ public static class OpenIdConnectExtensions
...
         /// <param name="options">Options which control the processing of the OpenIdConnect protocol and token validation.</param>
         /// <returns>The application builder</returns>
-        public static IApplicationBuilder UseOpenIdConnectAuthentication(this IApplicationBuilder app, Action<OpenIdConnectOptions> configureOptions = null)
+        public static IApplicationBuilder UseOpenIdConnectAuthentication(this IApplicationBuilder app, Action<OpenIdConnectOptions> configureOptions)
         {
-            return app.UseMiddleware<OpenIdConnectMiddleware>(
-                 new ConfigureOptions<OpenIdConnectOptions>(configureOptions ?? (o => { })));
+
+            var options = new OpenIdConnectOptions();
+            if (configureOptions != null)
```
PinpointTownes: This null check doesn't seem consistent with the other extensions.

src/Microsoft.AspNet.Authentication/ClaimsTransformer.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using System;
+using System.Security.Claims;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.Authentication
+{
+    public class ClaimsTransformer : IClaimsTransformer
+    {
+        public Func<ClaimsPrincipal, Task<ClaimsPrincipal>> TransformDelegate { get; set; }
```
PinpointTownes: Glad to see you're merging the two delegates! :+1: 

That said, this name is a bit weird. Why not `OnTransform` to match the pattern used for the events?

src/Microsoft.AspNet.Authentication.Facebook/FacebookAppBuilderExtensions.cs
HaoK: I'd rather null guard the configureOptions, which I was doing everywhere else, but yeah I missed this one :)

PinpointTownes:
nit: don't forget to bring back the delegate-style `app.UseOAuthAuthentication` extension :+1:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectExtensions.cs
HaoK: This will be the new pattern with NotNull dead, might as well just allow it and no-op

PinpointTownes: Nice. This remark was just there to point out OAuth2/FB/Google/MSFT don't have the same null check :smile:

src/Microsoft.AspNet.Authentication/ClaimsTransformer.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public class ClaimsTransformer : IClaimsTransformer
+    {
+        public Func<ClaimsPrincipal, Task<ClaimsPrincipal>> OnTransform { get; set; }
+
+        public virtual Task<ClaimsPrincipal> TransformAsync(ClaimsPrincipal principal)
+        {
+            if (OnTransform != null)
+            {
+                return OnTransform(principal);
```
PinpointTownes: `return OnTransform?.Invoke(principal) ?? Task.FromResult(principal);` :trollface:

HaoK: Well won't this prevent the delegate from return a null principal?  At least before the transform could transform into null

PinpointTownes: Heh no, it would just prevent the delegate from returning a null `Task` :smile:

HaoK: Ah yes, these are async only now...fair enough

muratg:
`NotNull` is going away, so good to add proper null checks. :+1:

Though per agreed upon rules, plural of `middleware` is `middlesware` not `middlewares` :trollface:

samples/OpenIdConnectSample/Startup.cs
```diff
@@ -14,11 +13,7 @@ public class Startup
     {
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddAuthentication();
-            services.Configure<SharedAuthenticationOptions>(options =>
-            {
-                options.SignInScheme = CookieAuthenticationDefaults.AuthenticationScheme;
-            });
+            services.AddAuthentication(options => options.SignInScheme = CookieAuthenticationDefaults.AuthenticationScheme);
```
Tratcher: Just reading this I have no idea what it does. At least change options to sharedOptions.

Tratcher:
:shipit:
Ping @pranavkm when you're done so he can remove NotNull.

HaoK:
Cleaned stuff up, added back some `[NotNull]` so the script will automatically add arguments for the Options.  Removed IOptions from the POCO implementations

HaoK:
081577e4f4996082a4fc28ef77c84f54c7d04183

Issue #323, opened 2015-07-01
The ClaimsPrincipal created by the Authentication Layer is not preserved as created to the Application Layer.


brentschmaltz:
The ClaimsPrincipal that is created by an Authentication Middleware is not preserved as created. Users who are developing using CustomPrincipals will loose any context.

leastprivilege:
Custom principals should not be supported. Identities are made up of a collection of claims - and nothing else. Any OOP sugar you want to add to it should be done at the application level - not the framework level.

brentschmaltz:
@leastprivilege Some users have made use of custom principals, I am curious as to why you think CustomPrincipals are not important.

I am bringing up for discussion that after an AuthenticationHandler has created a CP and returned an AuthenticationTicket, the runtime breaks apart the AuthenticationTicket and merges the contents of the CP into a new CP, sometimes dropping things.

Assuming the AuthenticationHandler is authoritative for the identity (understanding the procotol), I am questioning why the layers above should modify the identity.

PinpointTownes:
Personally, I've always considered that creating extension methods around `ClaimsPrincipal`/`ClaimsIdentity` was easier and more flexible when everything they were doing was formatting a specific claim value (e.g `GetUserId()`, `GetAccessToken()`), but this is a fact we can't deny: a non-negligible number of users WANT to create their custom `ClaimsPrincipal` (the question has been asked many times on JabbR), and that's why I'd tend to agree with @brentschmaltz.

There are currently two main friction points:
* `SecurityHelper.MergeUserPrincipal` (https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication/SecurityHelper.cs#L22): I guess we would have to update it to avoid creating a new `ClaimsPrincipal`. It should probably reuse `existingPrincipal` and directly call `existingPrincipal.AddIdentities(additionalPrincipal.Identities)`.

* `TicketSerializer` (https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication/DataHandler/Serializer/TicketSerializer.cs): probably the most complicated part. Maybe we should simply add new protected virtual methods to allow modifying the logic used to serialize/deserialize `ClaimsPrincipal`.

On the other hand, I'm convinced we should continue exposing `ClaimsPrincipal` in all the authentication APIs. Adding generics here would be really messy. Users wanting to use a custom `ClaimsPrincipal` will have to "pay" for it, by casting their principals :smile: 

[...]

Issue #158, opened 2015-02-19
Switch to ClaimsPrincipal.WriteTo when available


HaoK:

PinpointTownes:
@brentschmaltz I just took a look at the new `ClaimsIdentity.WriteTo` and `ClaimsIdentity.Initialize` methods on Reference Source and I have the weird sensation that serialization and deserialization formats differ: if `WriteTo` supports the `HasLabel`, `Actor` and `UserData`, `Initialize` doesn't seem to have any code to handle them at the deserialization stage. Am I missing something obvious? :scream: 

http://referencesource.microsoft.com/#mscorlib/system/security/claims/ClaimsIdentity.cs,1157
http://referencesource.microsoft.com/#mscorlib/system/security/claims/ClaimsIdentity.cs,1076

PinpointTownes:
Fears confirmed: this new stuff doesn't work AT ALL, even with simple identities that expose no `Actor` and have no `Properties`:

> L'exception System.IO.EndOfStreamException s'est produite
  HResult=-2147024858
  Message=Impossible de lire au-delà de la fin du flux.
  Source=mscorlib
  StackTrace:
       à System.IO.BinaryReader.ReadByte()
       à System.IO.BinaryReader.Read7BitEncodedInt()
       à System.IO.BinaryReader.ReadString()
       à System.Security.Claims.Claim.Initialize(BinaryReader reader, ClaimsIdentity subject)
       à System.Security.Claims.Claim..ctor(BinaryReader reader, ClaimsIdentity subject)
       à System.Security.Claims.ClaimsIdentity.Initialize(BinaryReader reader)
       à System.Security.Claims.ClaimsIdentity..ctor(BinaryReader reader)
       à System.Security.Claims.ClaimsPrincipal.CreateClaimsIdentity(BinaryReader reader)
       à System.Security.Claims.ClaimsPrincipal.Initialize(BinaryReader reader)
       à System.Security.Claims.ClaimsPrincipal..ctor(BinaryReader reader)

@brentschmaltz I'm curious. Have these new APIs been really tested? :smile:

Eilon:
Looks like we already did this another way.

brentschmaltz:
I'll take a peak.

MonDeveloper:
+1 
EndOfStreamException is thrown every time you try to call the ctor with the same data obtained by the WriteTo.
WEIRD !

flouk19:
@MonDeveloper 
[...]

PR #465, opened 2015-09-18
Refactor TicketSerializer/PropertiesSerializer and add ClaimsIdentity.Actor/Claim.Properties support

PinpointTownes:
Fixes https://github.com/aspnet/Security/issues/324, partially fixes https://github.com/aspnet/Security/issues/323 and deprecates https://github.com/aspnet/Security/issues/158.

@Eilon any chance we could include this PR in beta8 release?

/cc @HaoK @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Authentication/DataHandler/PropertiesSerializer.cs
+++ b/src/Microsoft.AspNet.Authentication/DataHandler/PropertiesSerializer.cs
@@ -13,8 +13,10 @@ namespace Microsoft.AspNet.Authentication
     {
         private const int FormatVersion = 1;
 
+        public static PropertiesSerializer Default { get; } = new PropertiesSerializer();
+
         [SuppressMessage("Microsoft.Usage", "CA2202:Do not dispose objects multiple times", Justification = "Dispose is idempotent")]
-        public byte[] Serialize(AuthenticationProperties model)
+        public virtual byte[] Serialize(AuthenticationProperties model)
         {
             using (var memory = new MemoryStream())
             {
@@ -28,7 +30,7 @@ namespace Microsoft.AspNet.Authentication
         }
 
         [SuppressMessage("Microsoft.Usage", "CA2202:Do not dispose objects multiple times", Justification = "Dispose is idempotent")]
-        public AuthenticationProperties Deserialize(byte[] data)
+        public virtual AuthenticationProperties Deserialize(byte[] data)
         {
             using (var memory = new MemoryStream(data))
             {
@@ -39,26 +41,29 @@ namespace Microsoft.AspNet.Authentication
             }
         }
 
-        public static void Write([NotNull] BinaryWriter writer, [NotNull] AuthenticationProperties properties)
+        public virtual void Write([NotNull] BinaryWriter writer, [NotNull] AuthenticationProperties properties)
         {
             writer.Write(FormatVersion);
             writer.Write(properties.Items.Count);
-            foreach (var kv in properties.Items)
+
+            foreach (var item in properties.Items)
             {
-                writer.Write(kv.Key);
-                writer.Write(kv.Value);
+                writer.Write(item.Key ?? string.Empty);
+                writer.Write(item.Value ?? string.Empty);
             }
         }
 
-        public static AuthenticationProperties Read([NotNull] BinaryReader reader)
+        public virtual AuthenticationProperties Read([NotNull] BinaryReader reader)
         {
             if (reader.ReadInt32() != FormatVersion)
             {
                 return null;
             }
-            int count = reader.ReadInt32();
+
+            var count = reader.ReadInt32();
             var extra = new Dictionary<string, string>(count);
-            for (int index = 0; index != count; ++index)
+
+            for (var index = 0; index != count; ++index)
             {
                 string key = reader.ReadString();
                 string value = reader.ReadString();
--- a/src/Microsoft.AspNet.Authentication/DataHandler/TicketSerializer.cs
+++ b/src/Microsoft.AspNet.Authentication/DataHandler/TicketSerializer.cs
@@ -11,15 +11,18 @@ namespace Microsoft.AspNet.Authentication
 {
     public class TicketSerializer : IDataSerializer<AuthenticationTicket>
     {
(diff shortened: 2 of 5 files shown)


src/Microsoft.AspNet.Authentication/DataHandler/PropertiesSerializer.cs
```diff
@@ -39,26 +41,29 @@ public AuthenticationProperties Deserialize(byte[] data)
...
         {
             writer.Write(FormatVersion);
             writer.Write(properties.Items.Count);
-            foreach (var kv in properties.Items)
+
+            foreach (var item in properties.Items)
             {
-                writer.Write(kv.Key);
-                writer.Write(kv.Value);
+                writer.Write(item.Key ?? string.Empty);
```
PinpointTownes: Fixes the bug seen in https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/issues/126.

PinpointTownes:
@brentschmaltz question: should we also support `ClaimsIdentity.Label` serialization/deserialization?

src/Microsoft.AspNet.Authentication/DataHandler/TicketSerializer.cs
```diff
@@ -36,98 +39,177 @@ public virtual AuthenticationTicket Deserialize(byte[] data)
...
+                identity.BootstrapContext = reader.ReadString();
             }
 
-            var properties = PropertiesSerializer.Read(reader);
-            return new AuthenticationTicket(new ClaimsPrincipal(identities), properties, authenticationScheme);
+            // Determine whether the identity
+            // has an actor identity attached.
+            if (reader.ReadBoolean())
+            {
+                identity.Actor = ReadIdentity(reader);
```
PinpointTownes: Note: `ClaimsIdentity.Actor` is guarded against circular references, so there's no risk to end up in an endless recursive call :smile:

src/Microsoft.AspNet.Authentication/DataHandler/TicketSerializer.cs
```diff
@@ -36,98 +39,177 @@ public virtual AuthenticationTicket Deserialize(byte[] data)
...
+        public virtual void WriteClaim([NotNull] BinaryWriter writer, [NotNull] Claim claim)
+        {
+            WriteWithDefault(writer, claim.Type, claim.Subject?.NameClaimType ?? ClaimsIdentity.DefaultNameClaimType);
+            writer.Write(claim.Value);
+            WriteWithDefault(writer, claim.ValueType, ClaimValueTypes.String);
+            WriteWithDefault(writer, claim.Issuer, ClaimsIdentity.DefaultIssuer);
+            WriteWithDefault(writer, claim.OriginalIssuer, claim.Issuer);
+
+            // Write the number of properties contained in the claim.
+            writer.Write(claim.Properties.Count);
```
PinpointTownes: `Claim.Properties` is not supposed to be null, but old versions of <Mono 4.0 were known to not initialize this property. I chatted with @davidfowl and he agreed it was not worth it.

Eilon:
@Tratcher thoughts about taking this?

src/Microsoft.AspNet.Authentication/DataHandler/DataSerializers.cs
```diff
@@ -9,8 +9,8 @@ public static class DataSerializers
     {
         static DataSerializers()
         {
-            Properties = new PropertiesSerializer();
-            Ticket = new TicketSerializer();
+            Properties = PropertiesSerializer.Default;
```
HaoK: I think I nuked these statics in commit today

PinpointTownes: Yes, you ruined my PR :cry: 

Bad Hao, bad, bad, bad :smile:

PinpointTownes:
Rebased, but the build failed because of this super annoying `Invalid chunked data` error (that has nothing to do with this PR). I'll restart it tomorrow.

PinpointTownes:
@brentschmaltz last call: do you think we should also serialize ClaimsIdentity.Label?

src/Microsoft.AspNet.Authentication/DataHandler/TicketSerializer.cs
```diff
@@ -36,98 +39,177 @@ public virtual AuthenticationTicket Deserialize(byte[] data)
             }
         }
 
-        public static void Write([NotNull] BinaryWriter writer, [NotNull] AuthenticationTicket model)
+        public virtual void Write([NotNull] BinaryWriter writer, [NotNull] AuthenticationTicket ticket)
```
Tratcher: Why are any of these read/write methods public? At best they'd be protected.

Tratcher: If it's just for tests then they should be protected internal.

PinpointTownes: I haven't changed the visibility modifier of these methods, but this is something we can consider.

That said, note that `PropertiesSerializer.Write/Read` is internally used by the Twitter middleware.
If you want to make everything protected internal, we'll have to add a `InternalsVisibleToAttribute` for both the test project and the Twitter provider.

Tratcher: Ah, I missed the Twitter cross-over. That's pretty sloppy, but it doesn't look easy to fix. Disregard.

PinpointTownes: We can keep Write/Read public and make WriteIdentity/WriteClaim/ReadIdentity/ReadClaim protected if you want.

Tratcher: That's better than nothing.

PinpointTownes: Done.

Tratcher:
:shipit:

Tratcher:
Rebased and merged.

PR #482, opened 2015-09-22
First pass Security API Reviews

HaoK:
Fairly straight forward changes first

- Nuked ITextEncoder, moved to a single Base64UrlEncoder static (GenerateCorrelationId uses this as well)
- Standardized events removing Apply, Response etc.
- Renamed JwtBearer => OpenIdConnectBearer
- Smushed ReturnEndpoint => SigningInContext, switched events to use this directly where applicable
- Misc other cleaning up from API review

@Tratcher


src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
```diff
@@ -21,22 +21,49 @@ public class CookieAuthenticationEvents : ICookieAuthenticationEvents
...
         /// A delegate assigned to this property will be invoked when the related method is called
         /// </summary>
-        public Func<CookieResponseSignOutContext, Task> OnResponseSignOut { get; set; } = context => Task.FromResult(0);
+        public Func<CookieSigningOutContext, Task> OnSigningOut { get; set; } = context => Task.FromResult(0);
 
         /// <summary>
         /// A delegate assigned to this property will be invoked when the related method is called
         /// </summary>
-        public Func<CookieApplyRedirectContext, Task> OnApplyRedirect { get; set; } = context =>
+        public Func<CookieRedirectContext, Task> OnRedirectToReturnUrl { get; set; } = context =>
```
Tratcher: nit: since all of the redirects have the same signature you could just make one and assign it to all of them.

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
```diff
@@ -21,22 +21,49 @@ public class CookieAuthenticationEvents : ICookieAuthenticationEvents
...
+        {
+            context.Response.Redirect(context.RedirectUri);
+            return Task.FromResult(0);
+        };
+
+        /// <summary>
+        /// A delegate assigned to this property will be invoked when the related method is called
```
Eilon: These doc comments are pretty weak :smile: At the very least add a period to make them sentences. But if these are to have doc comments at all they might as well say what the "related method" is...

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieSigningInContext.cs
```diff
@@ -8,9 +8,9 @@
 namespace Microsoft.AspNet.Authentication.Cookies
 {
     /// <summary>
-    /// Context object passed to the ICookieAuthenticationProvider method ResponseSignIn.
+    /// Context object passed to the ICookieAuthenticationEvents method ResponseSignIn.
```
Eilon: Need to change `ResponseSignIn` in the comment too. Might be other files that need this rename in the docs.

src/Microsoft.AspNet.Authentication.Cookies/Events/ICookieAuthenticationEvents.cs
```diff
@@ -23,25 +23,43 @@ public interface ICookieAuthenticationEvents
...
         /// <summary>
-        /// Called when a Challenge, SignIn, or SignOut causes a redirect in the cookie middleware
+        /// Called when a SignOut causes a redirect in the cookie middleware
         /// </summary>
         /// <param name="context">Contains information about the event</param>
-        Task ApplyRedirect(CookieApplyRedirectContext context);
+        Task RedirectToLogout(CookieRedirectContext context);
+
+        /// <summary>
+        /// Called when a SignIn causes a redirect in the cookie middleware
```
Eilon: End docs with a period.

src/Microsoft.AspNet.Authentication.OpenIdConnectBearer/project.json
```diff
@@ -1,6 +1,6 @@
 {
     "version": "1.0.0-*",
-    "description": "ASP.NET 5 middleware that enables an application to receive a Jwt bearer token.",
+    "description": "ASP.NET 5 middleware that enables an application to receive a OpenIdConnect bearer token.",
```
Eilon: ... *an* OpenIdConnect...

src/Microsoft.AspNet.Authentication/DataHandler/TextEncoder.cs
```diff
@@ -2,18 +2,17 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Authentication
```
Eilon: Should this move to a `.Internal` namespace? User code should never call this, right?

src/Microsoft.AspNet.Authentication/Events/SigningInContext.cs
```diff
@@ -25,6 +28,13 @@ public abstract class ReturnEndpointContext : EndpointContext
         public ClaimsPrincipal Principal { get; set; }
         public AuthenticationProperties Properties { get; set; }
 
+        public bool IsRequestCompleted { get; private set; }
+
+        public void RequestCompleted()
+        {
+            IsRequestCompleted = true;
```
Eilon: Set this inline on the property, and remove the `private set;` (it's implied in C# 6).

src/Microsoft.AspNet.Authentication/SubjectPublicKeyInfoAlgorithm.cs
```diff
@@ -1,15 +0,0 @@
...
-
-
-namespace Microsoft.AspNet.Authentication
-{
-    /// <summary>
-    /// The algorithm used to generate the subject public key information blob hashes.
-    /// </summary>
-    public enum SubjectPublicKeyInfoAlgorithm
-    {
-        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "Sha", Justification = "It is correct.")] Sha1,
```
Eilon: Literally the worst enum ever. Buh-bye!

test/Microsoft.AspNet.Authentication.Test/project.json
```diff
@@ -6,7 +6,7 @@
     "Microsoft.AspNet.Authentication.Cookies": "1.0.0-*",
     "Microsoft.AspNet.Authentication.Facebook": "1.0.0-*",
     "Microsoft.AspNet.Authentication.Google": "1.0.0-*",
-    "Microsoft.AspNet.Authentication.JwtBearer": "1.0.0-*",
+    "Microsoft.AspNet.Authentication.OpenIdConnectBearer": "1.0.0-*",
```
Eilon: Gotta sort these now :smile:

Eilon:
Noiice, just some random comments here and there. The new event names are sooo much nicer!

src/Microsoft.AspNet.Authentication.OpenIdConnectBearer/OpenIdConnectBearerOptions.cs
```diff
@@ -35,24 +35,24 @@ public JwtBearerOptions() : base()
         public string Authority { get; set; }
 
         /// <summary>
-        /// Gets or sets the audience for any received JWT token.
+        /// Gets or sets the audience for any received OpenIdConnect token.
```
Tratcher: nit: it's still a JWT token. line 41 too.

src/Microsoft.AspNet.Authentication.OpenIdConnectBearer/OpenIdConnectBearerOptions.cs
```diff
@@ -9,19 +9,19 @@
...
+    public class OpenIdConnectBearerOptions : AuthenticationOptions
     {
         /// <summary>
         /// Creates an instance of bearer authentication options with default values.
         /// </summary>
-        public JwtBearerOptions() : base()
+        public OpenIdConnectBearerOptions() : base()
         {
-            AuthenticationScheme = JwtBearerDefaults.AuthenticationScheme;
+            AuthenticationScheme = OpenIdConnectBearerDefaults.AuthenticationScheme;
```
Tratcher: It's misleading that this constant is used for both AuthenticationScheme and Challenge. I'd recommend two different constants, even if they had the same value.

src/Microsoft.AspNet.Authentication/Events/SigningInContext.cs
Tratcher: @Eilon, this isn't a constructor even if it does look like one.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectHandlerTests.cs
```diff
@@ -26,8 +26,8 @@ namespace Microsoft.AspNet.Authentication.Tests.OpenIdConnect
     /// </summary>
     public class OpenIdConnectHandlerTests
     {
-        private const string nonceForJwt = "abc";
-        private static SecurityToken specCompliantJwt = new JwtSecurityToken("issuer", "audience", new List<Claim> { new Claim("iat", EpochTime.GetIntDate(DateTime.UtcNow).ToString()), new Claim("nonce", nonceForJwt) }, DateTime.UtcNow, DateTime.UtcNow + TimeSpan.FromDays(1));
+        private const string nonceForOpenIdConnect = "abc";
```
Tratcher: nit: still JWTs here.

Tratcher:
:shipit:

src/Microsoft.AspNet.Authentication/Events/SigningInContext.cs
Eilon: Oh indeed! Well, my `private set;` remark still holds.

Eilon: Can we change the name of this method to have a nice verb in it then? Or do we really need this method at all?

Tratcher: It's doing very much the same thing as BaseControlContext.HandleResponse() https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication/Events/BaseControlContext.cs#L31

Eilon: Yeah those are all kind of gross :smile: I'm ok keeping them as-is - I get why we're using this pattern. It's more of a "command" as opposed to a traditional method.

samples/CookieSessionSample/MemoryCacheSessionStore.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace CookieSessionSample
 {
-    public class MemoryCacheSessionStore : IAuthenticationSessionStore
+    public class MemoryCacheSessionStore : ITicketStore
```
PinpointTownes: Is it worth modifying the name of this sample class to match the new pattern?

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationOptions.cs
```diff
@@ -146,7 +146,7 @@ public string CookieName
         /// An optional container in which to store the identity across requests. When used, only a session identifier is sent
         /// to the client. This can be used to mitigate potential problems with very large identities.
         /// </summary>
-        public IAuthenticationSessionStore SessionStore { get; set; }
+        public ITicketStore SessionStore { get; set; }
```
PinpointTownes: Have you considered renaming this property?

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
PinpointTownes: Ideally, those events shouldn't have any inline behavior. Everything should be handled at the handler level if `context.HandledResponse` is false.

samples/CookieSessionSample/MemoryCacheSessionStore.cs
Eilon: :+1:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -734,7 +734,7 @@ private void GenerateCorrelationId([NotNull] AuthenticationProperties properties
 
             var nonceBytes = new byte[32];
             CryptoRandom.GetBytes(nonceBytes);
-            var correlationId = TextEncodings.Base64Url.Encode(nonceBytes);
+            var correlationId = Base64UrlTextEncoder.Encode(nonceBytes);
```
PinpointTownes: FYI, Wilson has a similar method.

PinpointTownes:
It looks fine :+1: 

That said, I'm still not convinced `OpenIdConnectBearer` is the best name for the bearer middleware: many developers mix up the different security middleware and the new name doesn't seem to improve this situation.

Of course, I realize why you're tempted to rename it - since it has cool (but optional) OIDC metadata features, but even semantically, it's a bit weird. According to the specs, it's not named "OIDC bearer token" but "OAuth2 bearer": https://tools.ietf.org/html/rfc6750

Why not simply `TokenAuthenticationMiddleware` or `TokenValidationMiddleware`?

leastprivilege:
+10 for TokenValidationMiddleware

everything else is madness.

Eilon:
Because of the OIDC tie-in we wanted to include that in the name. Do you feel it's particularly misleading?

leastprivilege:
I missed most parts of the conversation.

But what I really want is a "TokenValidationMiddleware" - with 

* extensibility on how to get the token from the request (e.g. header vs query string - default to authZ header with *Bearer*)(
* extensibility how to validate the token (default to JWT)
* option to auto-configure via OIDC disco doc

Tratcher:
Yes we know people want a generic TokenValidationMiddleware, but this isn't it. We want to change the naming so we can leave room for the generic one.

leastprivilege:
It seems during that time you discussed the naming . the generic one could have been coded already :p

leastprivilege:
OIDC itself does not specify an access token. So OIDC in the name is just plain confusing and wrong.

JwtBearerTokenValidationMiddleware ?

and in "v2" add the extensibility - the name can stay the same.

HaoK:
So do you guys prefer the old name JwtBearer?  We can stick with that if it's preferred

leastprivilege:
JwtBearer is better than anything Oidc ..

NuGet.config
```diff
@@ -1,8 +1,8 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
PinpointTownes: 'zat you BOM?

src/Microsoft.AspNet.Authentication.OAuth/Events/IOAuthEvents.cs
```diff
@@ -16,19 +16,19 @@ public interface IOAuthEvents
...
         /// <returns>A <see cref="Task"/> representing the completed operation.</returns>
-        Task ReturnEndpoint(OAuthReturnEndpointContext context);
+        Task SigningIn(SigningInContext context);
 
         /// <summary>
         /// Called when a Challenge causes a redirect to the authorize endpoint.
         /// </summary>
         /// <param name="context">Contains redirect URI and <see cref="AuthenticationProperties"/> of the challenge.</param>
-        Task ApplyRedirect(OAuthApplyRedirectContext context);
+        Task RedirectToAuthorizationEndpoint(OAuthRedirectToAuthorizationContext context);
```
PinpointTownes: Did you mean OAuthRedirectToAuthorization**Endpoint**Context?

PinpointTownes:
The naming pattern of all these context classes seems a bit inconsistent: some are prefixed with Cookie or OAuth while the other ones have no prefix (OIDC for instance).

Concerning bearer middleware's name, I think I also prefer JwtBearer.

HaoK:
Ok  maybe we should go back to JwtBearer. Yeah I'm not a big fan of the context names right now either. I will take a look at those in the next pr. I'm planning on trying to refactor things there a bit anyways

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
HaoK: These have been here since katana I think, I went and added .'s for the ones in this file, but probably we need to do a doc comment pass at some point

src/Microsoft.AspNet.Authentication/DataHandler/TextEncoder.cs
HaoK: We can if you want, but we don't have anything else in an Internal namespace right now, we would probably move more things there.  Its called by OAuth and OpenId/Twitter today, so its actually getting used.  I was going to make these just private statics before I saw that

src/Microsoft.AspNet.Authentication/Events/SigningInContext.cs
HaoK: You can't remove a private set unless its the ctor that's setting it.  No private set, caused a compile error when trying to set it in the method...

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationOptions.cs
HaoK: Yeah we talked about that, but @Tratcher preferred the Session name to signal intent/usage for the property.

src/Microsoft.AspNet.Authentication.OAuth/Events/IOAuthEvents.cs
HaoK: I was leaning towards going through and removing Endpoint on all of the methods/API except for the Options, since it doesn't appear to add that much meaning and that name is HUMONGOUS.

samples/CookieSessionSample/Startup.cs
```diff
@@ -20,11 +20,11 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
         {
             loggerfactory.AddConsole(LogLevel.Information);
 
-            app.UseCookieAuthentication(options =>
+            app.UseCookieAuthentication((System.Action<CookieAuthenticationOptions>)(options =>
```
PinpointTownes: Outch, it hurts :smile:

src/Microsoft.AspNet.Authentication.OAuth/Events/IOAuthEvents.cs
PinpointTownes: Hum... IMHO, `RedirectToAuthorization` doesn't make much sense :cry:

HaoK: The top choices are:
`RedirectToAuthorization`
`AuthorizationRedirect`
`RedirectToAuthorizationEndpoint`

HaoK: We can revisit this in the next PR as well when I do a pass on all of the context/events

samples/CookieSessionSample/Startup.cs
HaoK: That is pretty nasty, fixed :)

src/Microsoft.AspNet.Authentication.OAuth/Events/IOAuthEvents.cs
Eilon: I really prefer having a proper action verb in the name because it's asking someone to do something.

src/Microsoft.AspNet.Authentication/DataHandler/TextEncoder.cs
Eilon: I want to hide this type *somewhere* - whether you want to mark it as `internal` or move it to a `.Internal` namespace, I'm fine either way.

src/Microsoft.AspNet.Authentication/Events/SigningInContext.cs
Eilon: *DUH* of course I knew that :smile:

Eilon:
Re: OIDCBearer, JwtBearer, JwtOIDCBearerAuthBearerJWTOpenTokenBearerOAuth I don't care just pick something :smile:  :smile:

HaoK:
Going to push what we have, I'll keep this PR open and fix things still open here in the next iteration PR

HaoK:
crap forgot to squash :/

PinpointTownes:
If you're quick enough to rewrite the git history, no one will realize it :smile:

HaoK:
Yeah but I thought that wasn't supposed to be done on public branches like dev :)

PinpointTownes:
Force pushing on "public branches" is evil, but having such an ugly history is much more terrible :smile: 

Plague or Cholera? :smile:

HaoK:
I was somewhat reasonable in my commit messages so I think we'll just have to live with a stain on this day in the history of the repo...

PinpointTownes:
Well, as long as no one clones/forks this repository or rebases his fork on your bad commits, rewriting the history is harmless.

Talking about the git history, a little birdie told me it was **really** time to remove your old branches :smile: 

image

HaoK:
Yeah I'm terrible at this, I have probably 5-10 branches in most repos :)  I'll do a nuke pass today

Issue #454, opened 2015-09-16
Enable dnxcore50 in Microsoft.AspNet.Authentication.Test


jmloeffler:
In #420 I identified a number of items that are incompatible with `dnxcore50`.  We should address these so that the project supports coreclr.

* remaining Moq reference in OpenIdConnectMiddlewareTest
* XmlTextWriter reference in CookieMiddlewareTest
* Uri.GetLeftPart(...) in FacebookMiddlewareTest, GoogleMiddlewareTest, and MicrosoftAccountMiddlewareTest
* UriPartial in FacebookMiddlewareTest, GoogleMiddlewareTest, and MicrosoftAccountMiddlewareTest

Eilon:
@jmloeffler I'm asking @Tratcher to follow up on anything remaining in this PR. Sorry for the delay in getting back to you!

jmloeffler:
No problem.  I think people were waiting on me anyway.  There was some concern over one of the test changes I made and I got a bit discouraged about the whole thing when I saw this commit af14c99b465ef79bea85edd38421f7d08f7d371e.  But I'm happy to put more time into this one if having dnxcore support in the test projects is still desired.

Eilon:
Yes, we definitely want the tests to run on `dnxcore50`!

Tratcher:
https://github.com/aspnet/Security/commit/42cba79e01872a69165ffc957ddcb7d79adf63da

PR #483, opened 2015-09-23
Enable core clr in all test projects

jmloeffler:
I think I was able to get `dnxcore50` supported in all of the test projects.  I'm not sure if everything is exactly identical as it was before but the tests do still pass.  In particular, I wasn't sure about the `XmlTextWriter` replacement and the way I removed Moq from `OpenIdConnectMiddlewareTests`

#454


test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectMiddlewareTests.cs
```diff
@@ -151,20 +150,17 @@ public class OpenIdConnectMiddlewareTests
...
-            var mockOpenIdConnectMessage = new Mock<OpenIdConnectMessage>();
-            mockOpenIdConnectMessage.Setup(m => m.CreateAuthenticationRequestUrl()).Returns(ExpectedAuthorizeRequest);
-            mockOpenIdConnectMessage.Setup(m => m.CreateLogoutRequestUrl()).Returns(ExpectedLogoutRequest);
             options.AutomaticAuthentication = true;
             options.Events = new OpenIdConnectEvents()
             {
                 OnRedirectToAuthenticationEndpoint = (context) =>
                 {
-                    context.ProtocolMessage = mockOpenIdConnectMessage.Object;
+                    context.ProtocolMessage = new OpenIdConnectMessage(ExpectedAuthorizeRequest);
```
Eilon: I'm not sure how that message type works, but this certainly *looks* like it's different behavior. The mock was returning these request objects only when certain methods were called.

test/Microsoft.AspNet.Authentication.Test/Cookies/CookieMiddlewareTests.cs
```diff
@@ -1023,14 +1023,9 @@ private static void Describe(HttpResponse res, AuthenticateContext result)
             {
                 xml.Add(result.Properties.Select(extra => new XElement("extra", new XAttribute("type", extra.Key), new XAttribute("value", extra.Value))));
             }
-            using (var memory = new MemoryStream())
```
davidfowl: Why this change?

test/Microsoft.AspNet.Authentication.Test/Cookies/CookieMiddlewareTests.cs
```diff
@@ -1023,14 +1023,9 @@ private static void Describe(HttpResponse res, AuthenticateContext result)
...
-            {
-                using (var writer = new XmlTextWriter(memory, Encoding.UTF8))
-                {
-                    xml.WriteTo(writer);
-                }
-                res.Body.Write(memory.ToArray(), 0, memory.ToArray().Length);
-            }
+
+            var xmlBytes = Encoding.UTF8.GetBytes(xml.ToString());
```
Eilon: This looks fine to me. It strikes me as something where either it totally works or it totally doesn't...

test/Microsoft.AspNet.Authentication.Test/Cookies/CookieMiddlewareTests.cs
jmloeffler: It did not appear that there was support for XmlTextWriter in dnxcore50.  I needed to replace it with something that was compatible.  Is there a better way?  I wasn't sure about these.  The same code is duplicated in multiple places, but that is a whole other issue.

test/Microsoft.AspNet.Authentication.Test/OpenIdConnect/OpenIdConnectMiddlewareTests.cs
jmloeffler: It is most certainly different but seems to get the same result.  I suspect that I'm abusing some internal behavior of OpenIdConnectMessage, but without a mocking framework to directly replace the return values from those methods I was at a loss for what to do.  This seemed to make the endpoint behave the same way in the tests so I went with it.

I heard that a supported version of moq was coming, but is there an alternative in the meantime?

Eilon: You could create a "manual mock" by deriving from `OpenIdConnectMessage`. In that derived constructor you could take in these two values as parameters, and then override the message's virtual methods to return the instances passed in the constructor. That's effectively what Moq is doing.

And yes, a compatible version of Moq is coming, but not soon enough :smile:

Tratcher:
@jmloeffler your git history was really messy so I recreated the changes locally and merged them. Thanks!

jmloeffler:
Sorry about that @Tratcher.  I appreciate the extra work though.  I'm sure it's not your favorite part of the job having to re-implement somebody else's changes.  Any tips on keeping the history clean?  Once I submitted the PR I didn't see anything I could do aside from keep merging in upstream/dev and correcting things along the way.

In any case, sorry about the additional headache.

Tratcher:
Rebase and force push is the best way to keep a PR branch current. It was pretty easy to replicate you're changes, you'd already solved all the problems.

PR #490, opened 2015-09-23
Rename Jwt events

HaoK:
cc @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Authentication.JwtBearer/Events/IJwtBearerEvents.cs
+++ b/src/Microsoft.AspNet.Authentication.JwtBearer/Events/IJwtBearerEvents.cs
@@ -36,6 +36,6 @@ namespace Microsoft.AspNet.Authentication.JwtBearer
         /// <summary>
         /// Invoked to apply a challenge sent back to the caller.
         /// </summary>
-        Task SetAuthenticationHeader(SetAuthenticationHeaderContext context);
+        Task Challenge(JwtBearerChallengeContext context);
     }
 }
--- a/src/Microsoft.AspNet.Authentication.JwtBearer/Events/IJwtBearerEvents.cs
+++ b/src/Microsoft.AspNet.Authentication.JwtBearer/Events/IJwtBearerEvents.cs
@@ -21,21 +21,21 @@ namespace Microsoft.AspNet.Authentication.JwtBearer
         /// <summary>
         /// Invoked when a protocol message is first received.
         /// </summary>
-        Task MessageReceived(MessageReceivedContext context);
+        Task ReceivingToken(ReceivingTokenContext context);
 
         /// <summary>
         /// Invoked with the security token that has been extracted from the protocol message.
         /// </summary>
-        Task SecurityTokenReceived(SecurityTokenReceivedContext context);
+        Task ReceivedToken(ReceivedTokenContext context);
 
         /// <summary>
         /// Invoked after the security token has passed validation and a ClaimsIdentity has been generated.
         /// </summary>
-        Task SecurityTokenValidated(SecurityTokenValidatedContext context);
+        Task ValidatedToken(ValidatedTokenContext context);
 
         /// <summary>
         /// Invoked to apply a challenge sent back to the caller.
         /// </summary>
-        Task ApplyChallenge(AuthenticationChallengeContext context);
+        Task SetAuthenticationHeader(SetAuthenticationHeaderContext context);
     }
 }
--- a/src/Microsoft.AspNet.Authentication.JwtBearer/Events/JwtBearerEvents.cs
+++ b/src/Microsoft.AspNet.Authentication.JwtBearer/Events/JwtBearerEvents.cs
@@ -23,22 +23,22 @@ namespace Microsoft.AspNet.Authentication.JwtBearer
         /// <summary>
         /// Invoked when a protocol message is first received.
         /// </summary>
-        public Func<MessageReceivedContext, Task> OnMessageReceived { get; set; } = context => Task.FromResult(0);
+        public Func<ReceivingTokenContext, Task> OnMessageReceived { get; set; } = context => Task.FromResult(0);
 
         /// <summary>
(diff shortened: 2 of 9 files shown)


src/Microsoft.AspNet.Authentication.JwtBearer/Events/JwtBearerEvents.cs
```diff
@@ -23,22 +23,22 @@ public class JwtBearerEvents : IJwtBearerEvents
         /// <summary>
         /// Invoked when a protocol message is first received.
         /// </summary>
-        public Func<MessageReceivedContext, Task> OnMessageReceived { get; set; } = context => Task.FromResult(0);
+        public Func<ReceivingTokenContext, Task> OnMessageReceived { get; set; } = context => Task.FromResult(0);
```
Tratcher: You didn't rename the On* delegates.

src/Microsoft.AspNet.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -26,10 +26,10 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
             try
             {
                 // Give application opportunity to find from a different location, adjust, or reject token
-                var messageReceivedContext = new MessageReceivedContext(Context, Options);
+                var messageReceivedContext = new ReceivingTokenContext(Context, Options);
```
Tratcher: Need to rename the variables too.

src/Microsoft.AspNet.Authentication.JwtBearer/Events/JwtBearerEvents.cs
HaoK: Another reminder of why I wish we only had either the delegate or the function :)

HaoK:
Updated renaming events and vars

Tratcher:
:shipit:

muratg:
Why is Travis failing? Hopefully not related to this change, but could you rerun to be sure?

src/Microsoft.AspNet.Authentication.JwtBearer/Events/IJwtBearerEvents.cs
```diff
@@ -21,21 +21,21 @@ public interface IJwtBearerEvents
...
         /// Invoked after the security token has passed validation and a ClaimsIdentity has been generated.
         /// </summary>
-        Task SecurityTokenValidated(SecurityTokenValidatedContext context);
+        Task ValidatedToken(ValidatedTokenContext context);
 
         /// <summary>
         /// Invoked to apply a challenge sent back to the caller.
         /// </summary>
-        Task ApplyChallenge(AuthenticationChallengeContext context);
+        Task SetAuthenticationHeader(SetAuthenticationHeaderContext context);
```
PinpointTownes: I'm not totally convinced by this new name, which seems to imply we can only modify the headers from this event.

One of the most common use cases of this event is to write an error directly to the response body (+ the standard WWW-Authenticate header) to make authentication errors easier to diagnose.

(sample: http://graph.facebook.com/me)

src/Microsoft.AspNet.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -26,26 +26,26 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
...
                 }
 
                 // If application retrieved token from somewhere else, use that.
-                token = messageReceivedContext.Token;
+                token = receivingTokenContext.Token;
 
                 if (string.IsNullOrEmpty(token))
                 {
-                    string authorization = Request.Headers["Authorization"];
+                    var authorization = Request.Headers["Authorization"];
```
PinpointTownes: Are you sure you can remove `string` here? `Request.Headers`'s indexer returns a `StringValues`, which is not implicitly converted to a `string` if you use `var`.

PinpointTownes: Ah, obviously not, since that's why the CI build failed :smile:

HaoK: Doh. Do you have alternate name for the name of the event?

PinpointTownes: IMO, `ApplyChallenge` (or `ApplyingChallenge`) was fine but IIRC, @Eilon was not a huge fan.

HaoK: Yeah I'd rather err on the side of being explicit but not all encompassing. Since at a minimum this event should set the header.

Eilon: If this is one where it actually does something (as opposed to merely notifying), then a "command" verb is indeed better (i.e. `ApplyXyz`). But if it's more of a notification (FYI this happened, that happened, etc.) then it should be something like `ApplyingXyz`/`AppliedXyz`.

HaoK: I wasn't opposed to challenge so much as the apply. 

How about a generic sounding: Challenge?

Eilon: If it's actually doing the challenge, then yeah, just `Challenge` I guess is fine.

HaoK: Ok Challenge it is, now matches the comments again which I forgot to change :)

PinpointTownes: :shipit:

HaoK:
852f44a3691c02c379b586fe7ef3dc0cb4cea6ab

Issue #147, opened 2015-02-03
Change UseTokenLifetime default to false in OIDC Middleware


briandonahue:
I find this really confusing.  If you've configured your cookies auth MW to have a timeout with sliding expiration, and are using OIDC, by default your cookies settings overwritten by the OIDC Identity Token's lifetime.  It seems to me this would be much clearer to default to false.  But maybe there is a valid reason?

Praburaj:
The token issued by the identity provider carries a `expires at` field. When this is true (which is by default) the cookie issued after successful authentication aligns with this timeout which makes sense as we honor the expiry time of the IDP. When this is false it takes what ever configurations on the cookie middleware.

brockallen:
I think the complaint is that with OIDC, id_tokens tend to use a very short expiration (like 5-15 minutes) which is just enough time to deliver the id_token to the client. It's then up to the client to determine how long to issue a cookie for based upon the authentication. Some apps might choose 1h, others 20 mins sliding expiration.

Point being, it's somewhat painful to have to configure the expiration on the cookie middleware and then configure a different flag on the OIDC middleware to to actually honor the settings configured on the cookie middleware.

briandonahue:
:+1:

briandonahue:
I am new to OIDC and relatively new to OWIN, but everything I heard was that id token is independent of client auth cookies/lifetimes, so I spent two painful days trying to figure out exactly what was going on in the MW pieces that was causing the auth session to expire, and finally found this.  Fully admitting it may not take *everyone* as long as it took me :) I still think it is very confusing that this flag basically negates cookies config settings by default.

Eilon:
We believe the default of `true` is overall the safest:

* When using ASP.NET Identity, it will replace the cookie with its own cookie that has its own expiration rules
* When using OIDC is the primary authentication, the value of `true` is presumably what people want.

brockallen:
> When using OIDC is the primary authentication, the value of true is presumably what people want.

I think the 3 people here on this issue were disagreeing with that default. Default to `true` is what might work best with Azure AD, but it somewhat shields developers (in a bad way) from thinking about the issue and deciding for themselves. I guess since the issue is closed, it's moot.

briandonahue:
For posterity, what I found most confusing about this was that I can set a cookie lifetime in my cookie options:
```
  app.UseCookieAuthentication(new CookieAuthenticationOptions
    {
      AuthenticationType = "Cookies",
      ...
[...]

PR #492, opened 2015-09-24
Make OIDC UseTokenLifetime false by default.

Tratcher:
#147 @HaoK @Eilon 
/cc: @leastprivilege @briandonahue @brockallen


Eilon:
:shipit:

PR #495, opened 2015-09-25
Add RemoteAuthenticationHandler base (support for error handling)

HaoK:
cc @Tratcher


HaoK:
Not sure where best to integrate this with ODIC yet.  Also still need to see if this makes sense in Cookies anywhere

src/Microsoft.AspNet.Authentication/AuthenticationOptions.cs
```diff
@@ -42,5 +43,11 @@ public string AuthenticationScheme
         /// Additional information about the authentication type which is made available to the application.
         /// </summary>
         public AuthenticationDescription Description { get; set; } = new AuthenticationDescription();
+
+        /// <summary>
+        /// The request path within the application's base path where the user-agent will be returned
+        /// when a failure occurs.
+        /// </summary>
+        public PathString ErrorHandlerPath { get; set; }
```
PinpointTownes: I'm not sure this property is general enough to put in the base `AuthenticationOptions`.

HaoK: Where else could it live if we want the common error handler logic in the base handler?

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -92,6 +92,31 @@ protected string BuildRedirectUri(string targetPath)
             return Request.Scheme + "://" + Request.Host + OriginalPathBase + targetPath;
         }
 
+        /// <summary>
+        /// Handles errors by redirecting to an error handler path if configured.  Otherwise sets the status code to 500.
+        /// </summary>
+        /// <param name="error"></param>
+        /// <returns></returns>
+        protected virtual Task HandleErrorAsync(ErrorContext error)
```
PinpointTownes: Have you considered returning `Task<bool>` to control whether the request should continue to the next middleware (to display a nice error page via MVC) or stop immediately (to return a 302 response for instance)?

HaoK: Yeah that's definitely one issue right now and related to the greater problem is a lot of the errors are in HandleAuthenticate which currently returns a Ticket and has no flow control...  So I couldn't use this new error behavior there at all

HaoK: But actually re the error page, just throw an uncaught exception to trigger the error page.  That's the new pattern for that.  We are nuking all the try/catch alls and using the error middleware to catch all those.

PinpointTownes: > Yeah that's definitely one issue right now and related to the greater problem is a lot of the errors are in HandleAuthenticate which currently returns a Ticket and has no flow control... So I couldn't use this new error behavior there at all

Yup, a little refactoring will be needed to fix that since we have no way to control the execution flow from `HandleAuthenticateAsync`. The approach you used is probably the only/best one: calling `HandleErrorAsync` from `InvokeAsync`/`InvokeReturnPathAsync` when the returned ticket is null. That said, flowing the exact "error message" in this approach is a bit hard.

A related thing is to decide whether we should throw an exception from `HandleAuthenticateAsync` or return null to indicate that authentication failed. In ASOS, I opted for the second approach but I have no strong opinion on that: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/blob/vNext/src/AspNet.Security.OpenIdConnect.Server/OpenIdConnectServerHandler.cs#L54-L89. Opting for exceptions would allow catching them in `InvokeReturnPathAsync` and call `HandleErrorAsync` with the appropriate exception/message.

(related issue: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/issues/138#issuecomment-141719610)

src/Microsoft.AspNet.Authentication/AuthenticationOptions.cs
PinpointTownes: The question is: do we want to have that in the base handler? This error handling is only applicable to authentication handlers implementing `InvokeAsync` (since you can't control the flow elsewhere) and I'm not sure all middleware will implement it.

HaoK: Yeah well the other option is to introduce a base class for all of the Invoke style middlewares (not cookie/bearer) basically.  I've been leaning towards that anyways, I'm not opposed to taking a stab at that and moving this error handling into the new common base.  We need to rename InvokeAsync anyways, that name is terrible.

HaoK: Perhaps RedirectingAuthenticationMiddleware as the name of the new shared base class for OAuth/OIDC/Twitter

PinpointTownes: A new base authentication would be terrible. IMHO, your new `HandleErrorAsync` is short enough to be added to OAuth2/OIDC/Twitter without requiring a base class.

HaoK: Well the idea was to pull the common flows from Twitter/OAuth and OIDC into this base class, not to only add errors :)

PinpointTownes: What do you mean exactly by "common flows"? It sounds like a crazy idea :smile:

HaoK: Yeah you and @Tratcher both think I'm crazy :)

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
Tratcher: I thought OnError was going to be an Event, not a method in the base class.

src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -89,7 +88,8 @@ public OAuthHandler(HttpClient backchannel)
                 var value = query["error"];
                 if (!StringValues.IsNullOrEmpty(value))
                 {
-                    Logger.LogVerbose("Remote server returned an error: " + Request.QueryString);
+                    Logger.LogWarning("Remote server returned an error: " + value);
+
                     // TODO: Fail request rather than passing through?
```
Tratcher: You can remove this TODO, I think that's how it works now.

src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -89,7 +88,8 @@ public OAuthHandler(HttpClient backchannel)
                 var value = query["error"];
                 if (!StringValues.IsNullOrEmpty(value))
                 {
-                    Logger.LogVerbose("Remote server returned an error: " + Request.QueryString);
+                    Logger.LogWarning("Remote server returned an error: " + value);
```
Tratcher: I had expected the new error event to be invoked here where you still had the original error information. That said, it looks like it would be hard to take meaningful action from this point in the call stack, the return value does not give you good control over what happens next. OIDC has some interesting hacks for this issue.

PinpointTownes: Nah, forget OIDC, it doesn't work AT ALL.
In this basic sample, it doesn't redirect the user:

```csharp
OnAuthenticationFailed = context => {
    if (string.Equals(context.ProtocolMessage.Error, "access_denied", StringComparison.Ordinal)) {
        context.HandleResponse();

        context.Response.Redirect("/");
    }

    return Task.FromResult<object>(null);
}
```

https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/blob/vNext/samples/Mvc/Mvc.Client/Startup.cs#L63-L74

Tratcher: Oh? What does it do instead?

PinpointTownes: When calling `Redirect`, both the status code and the `Location` header are correctly updated, but since the execution is not stopped (despite the `context.HandleResponse()`), the rest of the pipeline is invoked and the default 404 handler modifies the 302 response: you end up getting a 404 response with a `Location` header that has no effect :smile:

PinpointTownes: As observed by @sunsided, trying to write to the response stream is even worse, since you get a "headers already sent" exception when the default 404 handler tries to update the response code: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/issues/138#issuecomment-141719610

Tratcher: Hmm, this worked in Katana. It looks like the following code was removed from InvokeReplyPathAsync:
                if (ticket.Properties.Dictionary.TryGetValue(HandledResponse, out value) && value == "true")
                {
                    return true;
                }
Which was set by GetHandledResponseTicket().
http://katanaproject.codeplex.com/SourceControl/latest#src/Microsoft.Owin.Security.OpenIdConnect/OpenidConnectAuthenticationHandler.cs

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L1128-L1165
You could still cobble something together by returning a non-null AuthTicket with information in the properties, then deal with it in AuthenticationCompleted.

Needs work.

PinpointTownes: Yeah, I could probably use `AuthenticationCompleted`, but it feels far too hacky to use in a sample.

HaoK: Should we try to enable better control flow at this time as part of these changes?  This behavior was exactly what I ran into as well (and why I moved it to a dedicated method)

PinpointTownes: Yeah, that would be great! But I guess it's too late to include these changes in beta8?

HaoK: Yeah this PR already was too big for beta 8, will go into RC

HaoK:
Updated with major changes:

- Depends on HttpAbstractions: PR https://github.com/aspnet/HttpAbstractions/pull/421
- Removed bools from protected Handle methods and use context.IsResponseCompleted
- HandleAuthenticate now returns AuthenticateResult which contains Ticket and ErrorContext
- ErrorContext contains an error message and optional error uri for redirect

HaoK:
Updated with a stab at moving InvokeReplyPath to a new RemoteAuthenticationHandler base for OAuth and Twitter, I also sucked some of the common options into here as well like (BackChannel/CallbackPath).  I also moved SigningIn to this base class/options, but I'm not even sure we really need this event anymore...

src/Microsoft.AspNet.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -150,23 +159,25 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
...
+                    };
                 }
 
                 if (authenticationFailedContext.Skipped)
                 {
-                    return null;
+                    return new AuthenticateResult();
                 }
 
                 throw;
```
PinpointTownes: I'm not opposed to re-throwing exceptions by default. But they MUST be caught somewhere: we can't afford returning a 500 response for a simple invalid/expired token.

HaoK:
New iteration changes:
RemoteAuthorizationOptions is now is the base class for OAuth/Twitter/OpenIdConnectOptions.  It has a separate IRemoteEvents property which contains the Error and SigningIn events which are only invoked via the newly renamed HandleRemoteCallbackAsync.  The old handler specific events are unchanged and continue to live in their stand alone Events property.

HandleErrorAsync is a no-op in the base AuthHandler, and overridden in RemoteHandler to invoke the Error event on RemoteOptions.RemoteEvents.

src/Microsoft.AspNet.Authentication.JwtBearer/JwtBearerHandler.cs
HaoK: Make sure we don't forget about this.  Once the higher level changes are agreed on, the fun process of going through each of the specific error/exception paths needs to be reviewed :/

src/Microsoft.AspNet.Authentication/AuthenticationOptions.cs
HaoK: @PinpointTownes see new RemoteAuthenticationHandler shared base class, @Tratcher has grudgingly come around to its value...  standardizes Invoke path for OAuth/Twitter/OIDC

HaoK:
Update: 
- Only RemoteAuthenticationHandler/Invoke code paths use and inspect the ErrorContext.
- No events for TicketReceived/RemoteError, as that would pose issues with the derived handlers having to cast, or using new.  (I think we should just nuke all of the event classes and switch to simple delegates for each on the options directly).
- Nuked generics off of the base context types, instead each group (OAuth/Twitter/OIDC/Cookies) defines the correctly typed TOption lower down in their respective places.

TBD: Renaming OIDC events, and adding control flow to cookies to see if BaseContext can be merged with BaseControlContext

PinpointTownes:
Nice :smile: 

> @PinpointTownes see new RemoteAuthenticationHandler shared base class, @Tratcher has grudgingly come around to its value... standardizes Invoke path for OAuth/Twitter/OIDC

Actually, it doesn't look that bad but... :smile: 

> No events for TicketReceived/RemoteError, as that would pose issues with the derived handlers having to cast, or using new. (I think we should just nuke all of the event classes and switch to simple delegates for each on the options directly).

Outch, it's a horrible pattern that will kill hundreds of poor kittens :cry: 
If this is the price to pay for having a new `RemoteAuthenticationHandler`, I'm not sure it's really worth it: events are for advanced flow control. Merging them with the options would create far too noise and make the real events harder to discover.

> Nuked generics off of the base context types, instead each group (OAuth/Twitter/OIDC/Cookies) defines the correctly typed TOption lower down in their respective places.

Curious: what's the benefit? :open_mouth: 

> Convenience methods of AuthenticationResult.Success/Failed

Oh yeah, it's much nicer :+1: 

That said, I know wonder why we need this class at all, since it simply exposes the ticket and an exception... why not simply returning the ticket and throwing exceptions? :smile: 
Also, I'm not sure how I feel when I see one of the `Failed` methods creating an exception on-the-fly without ever throwing it.

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieSignedInContext.cs
```diff
@@ -33,6 +33,8 @@ public class CookieSignedInContext : BaseContext<CookieAuthenticationOptions>
             Properties = properties;
         }
 
+        public new CookieAuthenticationOptions Options { get { return (CookieAuthenticationOptions)base.Options; } }
```
PinpointTownes: Is this needed?

src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -30,131 +29,65 @@ public OAuthHandler(HttpClient backchannel)
...
-        protected override async Task<AuthenticationTicket> HandleAuthenticateAsync()
-        {
-            AuthenticationProperties properties = null;
-            try
+            if (tokens == null || string.IsNullOrEmpty(tokens.AccessToken))
             {
-                var query = Request.Query;
-
-                // TODO: Is this a standard error returned by servers?
-                var value = query["error"];
```
PinpointTownes: Looks like you removed this thing, which may potentially hide why the authentication failed.
Ideally, it would be nice to log `error`, `error_description` and `error_uri` (when populated by the authorization server).

(see https://tools.ietf.org/html/rfc6749#section-4.1.2.1 for more info about `error`, `error_description` and `error_uri`)

src/Microsoft.AspNet.Authentication/AuthenticationTicket.cs
```diff
@@ -45,5 +45,10 @@ public AuthenticationTicket(ClaimsPrincipal principal, AuthenticationProperties
         /// Additional state values for the authentication session.
         /// </summary>
         public AuthenticationProperties Properties { get; private set; }
+
+        /// <summary>
+        /// Holds error information caused by authentication.
+        /// </summary>
+        public ErrorContext Error { get; set; }
```
PinpointTownes: Hum, not sure about that. `AuthenticationTicket` is supposed to be fully serializable but this type has a name that may suggest it's not really meant to be persisted :smile: 

Is it even needed, since `AuthenticateResult` has an `Exception` property?

src/Microsoft.AspNet.Authentication/RemoteAuthenticationOptions.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// <summary>
+        /// Get or sets the text that the user can display on a sign in user interface.
+        /// </summary>
+        public string DisplayName
+        {
+            get { return Description.DisplayName; }
+            set { Description.DisplayName = value; }
+        }
+
+        public Func<ErrorContext, Task> OnRemoteError { get; set; } = context => Task.FromResult(0);
```
PinpointTownes: :fearful:

src/Microsoft.AspNet.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -150,23 +151,21 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
...
+                    return AuthenticateResult.Success(authenticationFailedContext.AuthenticationTicket);
                 }
-
                 if (authenticationFailedContext.Skipped)
                 {
-                    return null;
+                    return AuthenticateResult.Success(ticket: null);
                 }
 
                 throw;
```
PinpointTownes: Looks like this PR doesn't solve (yet) the most important problem: making all the authentication middleware consistent when returning an error/throwing an exception and avoid surfacing exceptions inappropriately.

Sadly, this middleware will continue returning a 500 error for an expired access token :sob:

HaoK: Yeah once everyone is more or less happy with the overall design. I will make the final pass through every middleware to address each specific error flow. I updated a few already.

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieSignedInContext.cs
HaoK: Leftover turds from previous iterations, nuking thanks :)

src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
HaoK: Okay I wasn't sure this was something we still needed, but looks like it is, I'll resurrect it

src/Microsoft.AspNet.Authentication/AuthenticationTicket.cs
HaoK: Yeah this is another orphan from a previous iteration, nuking it

HaoK:
So the other option which I had previously was keeping the base remote events in a separate IRemoteEvents property on the base options so they don't interfere with the actual provider events.  

options.RemoteEvents = new RemoteEvents() { OnError = ... }; (note I've come around fully to @Tratcher's view that the events should just be simple delegates without methods/interface)

src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
HaoK: I added back error, we should file a new issue if we want to consider supporting the other fields, this PR is big enough with me slipping in any MORE additional features :)

HaoK:
Updated: AuthenticationResult.Sucess to throw if a null ticket is provided, so the only time Ticket should be null is via the Failed method

src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
PinpointTownes: Done: https://github.com/aspnet/Security/issues/512

src/Microsoft.AspNet.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -118,18 +118,19 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
...
-                            return null;
+                            return AuthenticateResult.Success(ticket: null);
                         }
 
-                        return ticket;
+                        return AuthenticateResult.Success(ticket);
                     }
                 }
 
+                // REVIEW: this maybe return an error instead?
```
Tratcher: Agreed.

HaoK:
Introduced new HandleRemoteAuthenticateAsync to RemoteAuthenticationHandler and moved all the Invoke path logic there (old HandleAuthenticate method now returns a failed authenticated result).  We should consider if its more appropriate to use a no-op successful empty ticket like before

HaoK:
@Tratcher Updated with new iteration based on what we decided on:

Put events back into the base Event property and use it as a new base

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -257,7 +273,10 @@ protected virtual Task HandleSignInAsync(SignInContext context)
                 await PriorHandler.SignOutAsync(context);
             }
         }
-
+        /// <summary>
+        /// </summary>
+        /// <param name="context"></param>
+        /// <returns>True if no other handlers should be called</returns>
```
Tratcher: Remove empty docs, doesn't return anything.

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -238,6 +250,10 @@ protected Task<AuthenticationTicket> HandleAuthenticateOnceAsync()
             }
         }
 
+        /// <summary>
+        /// </summary>
+        /// <param name="context"></param>
+        /// <returns>True if no other handlers should be called</returns>
```
Tratcher: Remove empty docs, doesn't return anything.

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -267,10 +286,10 @@ protected virtual Task HandleSignOutAsync(SignOutContext context)
         /// </summary>
         /// <param name="context"></param>
         /// <returns>True if no other handlers should be called</returns>
```
Tratcher: Remove empty docs, doesn't return anything.

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -280,15 +299,14 @@ protected virtual Task<bool> HandleForbiddenAsync(ChallengeContext context)
...
```
Tratcher: etc.

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -296,27 +314,37 @@ protected virtual Task<bool> HandleUnauthorizedAsync(ChallengeContext context)
                 {
                     case ChallengeBehavior.Automatic:
                         // If there is a principal already, invoke the forbidden code path
-                        var ticket = await HandleAuthenticateOnceAsync();
-                        if (ticket?.Principal != null)
+                        var result = await HandleAuthenticateOnceAsync();
+
+                        //if (result?.Error != null)
```
Tratcher: ?

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -296,27 +314,37 @@ protected virtual Task<bool> HandleUnauthorizedAsync(ChallengeContext context)
...
+                        //    if (result.Error.IsRequestComplete)
+                        //    {
+                        //        context.CompleteRequest();
+                        //    }
+                        //}
+
+                        if (result?.Ticket?.Principal != null)
                         {
-                            handled = await HandleForbiddenAsync(context);
+                            await HandleForbiddenAsync(context);
```
Tratcher: goto case ChallengeBehavior.Forbidden;

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -296,27 +314,37 @@ protected virtual Task<bool> HandleUnauthorizedAsync(ChallengeContext context)
...
                     case ChallengeBehavior.Forbidden:
-                        handled = await HandleForbiddenAsync(context);
+                        await HandleForbiddenAsync(context);
                         break;
                 }
                 context.Accept();
             }
 
-            if (!handled && PriorHandler != null)
+            if (PriorHandler != null)
```
Tratcher: So there's no longer a mechanism to end chaining? Was that OK because any conflicting chaining would be the result of a configuration error? e.g. Basic and Cookies being set to automatic?

src/Microsoft.AspNet.Authentication/AuthenticationOptions.cs
Tratcher: This isn't needed anymore, correct?

HaoK: Yep nuking

src/Microsoft.AspNet.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        protected virtual async Task<bool> HandleRemoteCallbackAsync()
+        {
+            var authResult = await HandleRemoteAuthenticateAsync();
+            if (authResult == null || !authResult.Succeeded)
+            {
+                var errorContext = new ErrorContext(Context, 
+                    authResult?.Error ?? new Exception("Invalid return state, unable to redirect."));
+
+                Logger.LogInformation("Error from RemoteAuthentication: "+errorContext.Error.Message);
```
Tratcher: nit spacing.

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
HaoK: Blah, no this is a regression, we used to be able to rely on the Control flow in ChallengeContext, but we reverted those.  So our options are to bring control flow back in ChallengeContext for this, or go with our sort of lame book return again.  I prefer putting the control flow into ChallengeContext and checking that.  Thoughts?

HaoK: And I will write a test for this scenario :)

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -101,43 +101,28 @@ private Task<AuthenticationTicket> EnsureCookieTicket()
...
-                }
-
-                var context = new CookieValidatePrincipalContext(Context, ticket, Options);
-                await Options.Events.ValidatePrincipal(context);
-
-                if (context.Principal == null)
-                {
-                    return null;
-                }
+                return AuthenticateResult.Failed("No ticket.");
```
Tratcher: We will eventually need better descriptions for these.

HaoK: Yeah next iteration when I go through all of the cases, we can tweak all of the errors there

src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -30,131 +29,71 @@ public OAuthHandler(HttpClient backchannel)
...
-            await Options.Events.SigningIn(context);
+            var code = query["code"];
+            var state = query["state"];
 
-            if (context.SignInScheme != null && context.Principal != null)
+            properties = Options.StateDataFormat.Unprotect(state);
+            if (properties == null)
             {
-                await Context.Authentication.SignInAsync(context.SignInScheme, context.Principal, context.Properties);
+                return AuthenticateResult.Failed("State not found.");
```
Tratcher: Or unprotect failed (and returned null). "The oauth state was missing or invalid."

src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -174,10 +113,12 @@ public OAuthHandler(HttpClient backchannel)
             requestMessage.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
             requestMessage.Content = requestContent;
             var response = await Backchannel.SendAsync(requestMessage, Context.RequestAborted);
-            response.EnsureSuccessStatusCode();
-            var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
-
-            return new OAuthTokenResponse(payload);
+            if (response.IsSuccessStatusCode)
```
Tratcher: We'll need to be able to surface errors on this code path (e.g. else).

Tratcher: I recommend just reverting this for now.

src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -233,7 +173,7 @@ protected override Task HandleSignInAsync(SignInContext context)
             throw new NotSupportedException();
         }
 
-        protected override Task<bool> HandleForbiddenAsync(ChallengeContext context)
+        protected override Task HandleForbiddenAsync(ChallengeContext context)
         {
             throw new NotSupportedException();
```
Tratcher: You don't need these overrides anymore, correct? They're handled in the new base class.

src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
HaoK: How about we add an error string field to the OAuthToken response class? Looks like today we just use null token response to signal error.

Tratcher: No, it used to throw. I don't know why the caller was checking for null.

src/Microsoft.AspNet.Authentication.OAuth/OAuthOptions.cs
```diff
@@ -42,54 +47,26 @@ public class OAuthOptions : AuthenticationOptions
...
+        /// Gets or sets the <see cref="IOAuthEvents"/> used to handle authentication events.
         /// </summary>
-        public string DisplayName
+        public new IOAuthEvents Events
         {
-            get { return Description.DisplayName; }
-            set { Description.DisplayName = value; }
+            get
+            {
+                return (IOAuthEvents)base.Events;
```
Tratcher: nit: formatting

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -305,14 +302,14 @@ public OpenIdConnectHandler(HttpClient backchannel)
         /// </summary>
         /// <returns>An <see cref="AuthenticationTicket"/> if successful.</returns>
         /// <remarks>Uses log id's OIDCH-0000 - OIDCH-0025</remarks>
-        protected override async Task<AuthenticationTicket> HandleAuthenticateAsync()
+        protected override async Task<AuthenticateResult> HandleRemoteAuthenticateAsync()
         {
             Logger.LogDebug(Resources.OIDCH_0000_AuthenticateCoreAsync, this.GetType());
 
             // Allow login to be constrained to a specific path. Need to make this runtime configurable.
             if (Options.CallbackPath.HasValue && Options.CallbackPath != (Request.PathBase + Request.Path))
```
Tratcher: Isn't this handled in the base class now?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -373,13 +369,14 @@ public OpenIdConnectHandler(HttpClient backchannel)
...
                     Logger.LogError(Resources.OIDCH_0005_MessageStateIsInvalid);
-                    return null;
+                    return AuthenticateResult.Failed(Resources.OIDCH_0005_MessageStateIsInvalid);
                 }
 
                 // if any of the error fields are set, throw error null
                 if (!string.IsNullOrEmpty(message.Error))
                 {
                     Logger.LogError(Resources.OIDCH_0006_MessageContainsError, message.Error, message.ErrorDescription ?? "ErrorDecription null", message.ErrorUri ?? "ErrorUri null");
+                    // REVIEW: throw really?
```
Tratcher: No, use the new error mechanic.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
HaoK: Yeah this method won't get invoked unless it matches the callback path so I will nuke this check

src/Microsoft.AspNet.Authentication.OAuth/OAuthOptions.cs
HaoK: I'm not seeing the formatting issue

Tratcher: It used to be single-line.
```
get { return Description.DisplayName; }  
set { Description.DisplayName = value; }  
```

HaoK: Got it, I prefer that too :)  I'll revert

src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
HaoK: So you want this to just throw again?  But we removed all the guards catching the exception.

HaoK: As things stand if this returns null, we just immediately fail with Access token not found in OAuthHandler

Tratcher: I'd rather throw until we design a better error flow. This is the kind of error you need details for to debug.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
```diff
@@ -136,7 +106,17 @@ public string DisplayName
         /// <summary>
         /// Gets or sets the <see cref="IOpenIdConnectEvents"/> to notify when processing OpenIdConnect messages.
         /// </summary>
-        public IOpenIdConnectEvents Events { get; set; } = new OpenIdConnectEvents();
+        public new IOpenIdConnectEvents Events
+        {
+            get
+            {
```
Tratcher: nit: formatting

src/Microsoft.AspNet.Authentication.Twitter/TwitterOptions.cs
```diff
@@ -82,7 +44,17 @@ public string DisplayName
         /// <summary>
         /// Gets or sets the <see cref="ITwitterEvents"/> used to handle authentication events.
         /// </summary>
-        public ITwitterEvents Events { get; set; }
+        public new ITwitterEvents Events
+        {
+            get
+            {
+                return (ITwitterEvents)base.Events;
```
Tratcher: nit: formatting

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
Tratcher: Discussed offline, we need to do a deeper review of the control flow here.

src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
HaoK: My main point is we didn't throw before (we turned them into no-op successful no tickets) and we had test coverage for these scenarios, so unless we want to turn all of them into 500s, this was closest to the old behavior (at least for tests).  Basically I replaced try/catch return empty result, with return null instead of throw, and Fail

Tratcher: Yes, removing the catch-alls was expected to uncover suppressed exceptions and destabilize things a bit. Dealing with those kinds of issues can wait until after this PR. We'll want to look a them on a case-by-case basis. In the meantime if there are test issues we can update them or disable them.

The change above re-suppresses an error that we were trying to uncover and makes it harder to notice and handle properly in the future.

HaoK: Ok updated, with new uncaught throw behavior

HaoK:
Updated with the new iteration:
- Split AutomaticAuthentication into AutomaticChallenge / AutomaticAuthenticate
- Brought back bool control flow for Unauthorized/Challenge
- OAuth throws with no catch which is a temporary behavior change until we revisit error handling in the next PR

@Tratcher

samples/SocialSample/Startup.cs
```diff
@@ -29,7 +29,7 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
 
             app.UseCookieAuthentication(options =>
             {
-                options.AutomaticAuthentication = true;
+                options.AutomaticAuthenticate = true;
```
Tratcher: This one should also use AutomaticChallenge. See https://github.com/aspnet/Security/blob/haok9-25/errors/samples/SocialSample/Startup.cs#L225

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -62,8 +63,8 @@ protected string CurrentUri
         /// <param name="options">The original options passed by the application control behavior</param>
         /// <param name="context">The utility object to observe the current request and response</param>
         /// <param name="logger">The logging factory used to create loggers</param>
-        /// <returns>async completion</returns>
-        public async Task InitializeAsync(TOptions options, HttpContext context, ILogger logger, IUrlEncoder encoder)
+        /// <returns>True if initialization was successful</returns>
+        public async Task<bool> InitializeAsync(TOptions options, HttpContext context, ILogger logger, IUrlEncoder encoder)
```
Tratcher: It doesn't look like InitializeAsync is ever overridden. The flow control comes from HandleRequestAsync insteadl. I think you can revert this.

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -139,7 +141,7 @@ protected virtual Task FinishResponseAsync()
 
         private async Task HandleAutomaticChallengeIfNeeded()
         {
-            if (!ChallengeCalled && Options.AutomaticAuthentication && Response.StatusCode == 401)
+            if (!ChallengeCalled && Options.AutomaticAuthenticate && Response.StatusCode == 401)
```
Tratcher: AutomaticChallenge

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -184,35 +186,46 @@ public void GetDescriptions(DescribeSchemesContext describeContext)
             }
         }
 
-        public bool ShouldHandleScheme(string authenticationScheme)
+        public bool ShouldHandleScheme(string authenticationScheme, bool handleAutomatic)
         {
             return string.Equals(Options.AuthenticationScheme, authenticationScheme, StringComparison.Ordinal) ||
-                (Options.AutomaticAuthentication && string.IsNullOrEmpty(authenticationScheme));
+                (handleAutomatic && string.Equals(authenticationScheme, AuthenticationManager.AutomaticScheme));
```
Tratcher: Ordinal

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -279,7 +286,6 @@ protected virtual Task<bool> HandleForbiddenAsync(ChallengeContext context)
         /// changing the 401 result to 302 of a login page or external sign-in location.)
         /// </summary>
         /// <param name="context"></param>
-        /// <returns>True if no other handlers should be called</returns>
```
Tratcher: I guess you can put this comment back too.

src/Microsoft.AspNet.Authentication/AuthenticationOptions.cs
```diff
@@ -27,11 +27,16 @@ public string AuthenticationScheme
...
-        /// alter 401 Unauthorized responses going out. If false the authentication middleware will only provide
+        /// If true the authentication middleware alter the request user coming in. If false the authentication middleware will only provide
         /// identity and alter responses when explicitly indicated by the AuthenticationScheme.
         /// </summary>
-        public bool AutomaticAuthentication { get; set; }
+        public bool AutomaticAuthenticate { get; set; }
+
+        /// <summary>
+        /// If true the authentication middleware should handle automatic challenge.
+        /// If false the authentication middleware will only provide identity and alter responses when explicitly indicated by the AuthenticationScheme.
```
Tratcher: Remove the `provide identity` part.

src/Microsoft.AspNet.Authentication/AuthenticationOptions.cs
```diff
@@ -27,11 +27,16 @@ public string AuthenticationScheme
         }
 
         /// <summary>
-        /// If true the authentication middleware alter the request user coming in and
-        /// alter 401 Unauthorized responses going out. If false the authentication middleware will only provide
+        /// If true the authentication middleware alter the request user coming in. If false the authentication middleware will only provide
         /// identity and alter responses when explicitly indicated by the AuthenticationScheme.
```
Tratcher: Remove the `alter responses` part

src/Microsoft.AspNet.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -0,0 +1,103 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Http.Features.Authentication;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNet.Authentication
+{
+    public abstract class RemoteAuthenticationHandler<TOptions> : AuthenticationHandler<TOptions> where TOptions : RemoteAuthenticationOptions
```
Tratcher: TOptions is never actually used in this file, can you remove it?

Tratcher: Nevermind.

src/Microsoft.AspNet.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -0,0 +1,103 @@
...
+    {
+        public override async Task<bool> HandleRequestAsync()
+        {
+            if (Options.CallbackPath.HasValue && Options.CallbackPath == Request.Path)
```
Tratcher: CallbackPath is a required parameter for these middleware to work. We could move this HasValue check to the middleware constructor.

src/Microsoft.AspNet.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -0,0 +1,103 @@
...
+                    return false;
+                }
+
+                Context.Response.StatusCode = 500;
+                return true;
+            }
+
+            // We have a ticket if we get here
+            var ticket = authResult.Ticket;
+            var context = new TicketReceivedContext(Context, ticket)
```
Tratcher: TicketReceivedContext could take a RemoteAuthenticationOptions. Then you wouldn't need SignInScheme.

src/Microsoft.AspNet.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -0,0 +1,103 @@
...
+                Logger.LogVerbose("The SigningIn event returned Handled.");
+                return true;
+            }
+            else if (context.Skipped)
+            {
+                Logger.LogVerbose("The SigningIn event returned Skipped.");
+                return false;
+            }
+
+            // REVIEW: SignInScheme can be null if Shared SignInScheme and its not set on options, we could throw in Ctor?
```
Tratcher: Agreed, SignInScheme should be required in the constructor.

src/Microsoft.AspNet.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -0,0 +1,103 @@
...
+            if (context.SignInScheme != null && context.Principal != null)
+            {
+                await Context.Authentication.SignInAsync(context.SignInScheme, context.Principal, context.Properties);
+            }
+
+            if (context.ReturnUri != null)
```
Tratcher: I think a ReturnUri would be required at this point. I don't see why you would let the request continue executing after this.

Tratcher:
Almost done :watch:

src/Microsoft.AspNet.Authentication.Cookies/Events/BaseCookieContext.cs
```diff
@@ -0,0 +1,28 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Security.Claims;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Authentication;
+
+namespace Microsoft.AspNet.Authentication.Cookies
+{
+    public class BaseCookieContext : BaseContext
```
PinpointTownes: Curious: why do you need this?

src/Microsoft.AspNet.Authentication.JwtBearer/JwtBearerHandler.cs
PinpointTownes: `AuthenticateResult.Failed`

src/Microsoft.AspNet.Authentication.OAuth/Events/OAuthCreatingTicketContext.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.AspNet.Authentication.OAuth
     /// <summary>
     /// Contains information about the login session as well as the user <see cref="System.Security.Claims.ClaimsIdentity"/>.
     /// </summary>
-    public class OAuthCreatingTicketContext : BaseContext<OAuthOptions>
+    public class OAuthCreatingTicketContext : BaseContext
```
PinpointTownes: Same remark. The generic `BaseContext` was really convenient.

src/Microsoft.AspNet.Authentication.Twitter/TwitterOptions.cs
```diff
@@ -82,7 +44,11 @@ public string DisplayName
         /// <summary>
         /// Gets or sets the <see cref="ITwitterEvents"/> used to handle authentication events.
         /// </summary>
-        public ITwitterEvents Events { get; set; }
+        public new ITwitterEvents Events
```
PinpointTownes: This pattern reminds me the mess we removed from the OAuth2 handlers :smile:

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
HaoK: Good one, uncovered a lot of tests using the old one too since they all got renamed together :(

src/Microsoft.AspNet.Authentication/RemoteAuthenticationHandler.cs
HaoK: Fair enough I'll add the check to each middleware

src/Microsoft.AspNet.Authentication/RemoteAuthenticationHandler.cs
HaoK: This prevents them from changing what sign in scheme to use if we did this.  Although I'm not sure if that's really a scenario we want to support

src/Microsoft.AspNet.Authentication/RemoteAuthenticationHandler.cs
HaoK: Done, interestingly enough twitter already required this, added same check to OIDC / OAuth

src/Microsoft.AspNet.Authentication/RemoteAuthenticationHandler.cs
HaoK: I'm not sure I follow, so if no ReturnUri is found we do what?  Raise an error?

src/Microsoft.AspNet.Authentication/RemoteAuthenticationHandler.cs
HaoK: Just a minor comment, looks like we made it a bit harder to configure social auth + cookies to just work.  In the old days when Facebook + Cookies would work well without specifying auth schemes since Facebook would signin empty and cookies was automatic and pick it up.

Now we basically always require something like:
```
 services.AddAuthentication(options => options.SignInScheme = CookieAuthenticationDefaults.AuthenticationScheme);
```

HaoK: And if you forget that, you get SignInScheme is required which isn't really clear that they need to add the cookies auth scheme...  ah well, pretty sure we will get issues in the future about this :)

src/Microsoft.AspNet.Authentication.Cookies/Events/BaseCookieContext.cs
HaoK: We nuked Options/Generics off of the base contexts, and just added them all back in the specific OAuth/Cookies/Facebook contexts themselves.  Let us get rid of the TOption generics on the base context classes

```
        public CookieAuthenticationOptions Options { get; }
```

src/Microsoft.AspNet.Authentication.Twitter/TwitterOptions.cs
HaoK: Yep we brought back Event inheritance (but without generics), but since it allowed me to actually make Twitter/OIDC/OAuth share a base class and same Invoke path code, I think its worth it now :)

src/Microsoft.AspNet.Authentication.OAuth/Events/OAuthCreatingTicketContext.cs
HaoK: I know, it was a tradeoff with getting rid of generics which let us somewhat cleanly have the events derive from RemoteAuthenticationEvents

HaoK:
Updated, any last requests?  I'll push this tomorrow in the afternoon, MVC/Identity changes look to be fairly small surprisingly :)

HaoK:
409b50269a9d85618a31f84a3ffc2868abb3eeea will address any remaining issues in next iteration/PR

src/Microsoft.AspNet.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -0,0 +1,101 @@
...
+                Logger.LogVerbose("The SigningIn event returned Handled.");
+                return true;
+            }
+            else if (context.Skipped)
+            {
+                Logger.LogVerbose("The SigningIn event returned Skipped.");
+                return false;
+            }
+
+            if (context.Principal != null)
```
Tratcher: I'm still trying to understand how you would get to this point without a principal or ReturnUri. I would expect all of those cases to be error conditions handled above.

HaoK: The event can simply null out the principal

Tratcher: That would mean you're using the context property values as a form of flow control. Since we have real flow control now I don't think we should support the old hacky approach.

HaoK: Make the context properties read only for Ticket/Principal then?  I'm fine with that :)

Tratcher: No, we still want to allow people to replace it, but I don't think we should support null values.

HaoK: The problem is we allow AuthenticationTickets with null principals.  So we can't easily block this unless we change that.

eriksendc:
Hi All,

Before posting a new issue I wanted to see if perhaps this issue was about or related to an issue I'm having. I'm on RC1, When a user selects to authenticate w/Facebook, and they put in their user name and password correctly, but clicks the Cancel button on the "this is what the app is requesting" confirmation, then the user's browser is redirected to 

https://localhost:44300/signin-facebook?error=access_denied&error_code=200&error_description=Permissions+error&error_reason=user_denied&state=CfDJ8BCOQ9uBPU9Cu4ITtOhOeD-LHOt2vhRmJu5_xD4rvL3zm4ka-ak_iYgNuIJOOjjMsO1Fg5p5Tb8Fll6CJ064ysTkRrC46xaeG_vu5sFQdpPKjjGPokfsPnndPsB84CnDmRIzjwLeGgYzbESqjjGyAFUmSACDAYLo9LmmA8HJNPcdJIxTvcrepdr0GfE2DhJtP_D0Z7wNxyzJw1Jf8pGprHP8P4Q130hpmFzOLA2UCwjXYPxijlOP9LpY6Pgz66PKVJt_nO-6e0yA7aStVJTCOH8ic-I2xRguFAWXpfoim8wlqaizDupXFBBgeBtw9RvopkxKENby2N89okirRZSFK6FTLZJaG7RKxEWv_aUZskoFKmrKrkZDm0Ur3DyWnH53fQ#_=_

Control doesn't seem to be flowing back to my ASP.NET web application (I never get back anything in any /Account/xxx account controller method.

Wanted to report as an issue if it isn't one yet, but didn't want to add a new issue if this has already been fixed.

Thanks,
-Brian Eriksen

PinpointTownes:
@eriksendc hey,

There's actually a ticket tracking that: https://github.com/aspnet/Security/issues/710.

eriksendc:
Thx!

Issue #485, opened 2015-09-23
Add a RequireHttps property when migrating to Wilson beta8


PinpointTownes:
In a recent change in Wilson, `HttpDocumentRetriever` was updated to reject non-HTTPS metadata addresses. Overriding the default configuration manager to allow HTTP for testing scenarios is possible but a real PITA from a usability perspective: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/blob/vNext/samples/Mvc/Mvc.Server/Startup.cs#L50-L57

I suggest adding a new `RequireHttps` (or `AllowInsecureHttp`) property to both the OIDC and the bearer middleware to control whether the instantiated configuration manager should support HTTP or not.

This property could also be used to reject non-HTTPS addresses earlier in the middleware itself, specially since there's a bug in Wilson that hides the exception thrown when you try to use a HTTP endpoint and makes this new requirement really hard to discover: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/253.

@Eilon is migrating to Wilson beta8 planed for ASP.NET 5 beta8?

leastprivilege:
I really don't understand why this HTTPs by default requirement was added - there are some valid scenarios where HTTP is OK - and it is up to the developer to set the right URL.

lol for `AllowInsecureHttp` ;)

Please don't start to do the same for redirect_uri or authority.

leastprivilege:
but anyways - please make that at least easily configurable.

Eilon:
Yes, ASP.NET beta8 will use Wilson beta8. If this is what I think it is, I recall seeing some spec verbiage around the requirement of HTTPS.

@brentschmaltz @Tratcher @blowdart do any of you recall what this was about?

blowdart:
Pinpoint pretty much explains it. The OIDC spec requires HTTPS return points. That's a PITA for dev. So add a "DisableHttpsRequirement" property. This would then be checked when the URLs are configured, and an exception thrown as appropriate to the settings.

Eilon:
@Tratcher can you take a look at this?

stephenstroud:
+1

brentschmaltz:
@PinpointTownes @leastprivilege @blowdart This is a 'secure by default' change. Unlike, fed metadata which can be signed, normally OIDC metadata is not signed. Hence, https required by default. I agree we need to make it easier to turn off.

PR #498, opened 2015-09-28
Upate to IdentityModel beta8 dependencies.

Tratcher:
#493 
@muratg @tushargupta51 @brentschmaltz

### DIFF
--- a/samples/OpenIdConnectSample/Startup.cs
+++ b/samples/OpenIdConnectSample/Startup.cs
@@ -6,6 +6,7 @@ using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Authentication;
 using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.Logging;
+using Microsoft.IdentityModel.Protocols.OpenIdConnect;
 
 namespace OpenIdConnectSample
 {
@@ -28,8 +29,10 @@ namespace OpenIdConnectSample
             app.UseOpenIdConnectAuthentication(options =>
             {
                 options.ClientId = "63a87a83-64b9-4ac1-b2c5-092126f8474f";
+                options.ClientSecret = "Yse2iP7tO1Azq0iDajNisMaTSnIDv+FXmAsFuXr+Cy8="; // for code flow
                 options.Authority = "https://login.windows.net/tratcheroutlook.onmicrosoft.com";
                 options.RedirectUri = "http://localhost:42023";
+                options.ResponseType = OpenIdConnectResponseTypes.Code;
             });
 
             app.Run(async context =>
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
@@ -320,7 +320,7 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
                 // response_mode=query (explicit or not) and a response_type containing id_token
                 // or token are not considered as a safe combination and MUST be rejected.
                 // See http://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#Security
-                if (!string.IsNullOrEmpty(message.IdToken) || !string.IsNullOrEmpty(message.Token))
+                if (!string.IsNullOrEmpty(message.IdToken) || !string.IsNullOrEmpty(message.AccessToken))
                 {
                     Logger.LogError("An OpenID Connect response cannot contain an identity token " +
                                     "or an access token when using response_mode=query");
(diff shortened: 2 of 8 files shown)


src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -498,7 +505,19 @@ public OpenIdConnectHandler(HttpClient backchannel)
 
             ticket = ValidateToken(tokenEndpointResponse.ProtocolMessage.IdToken, message, properties, validationParameters, out jwt);
 
-            ValidateOpenIdConnectProtocol(null, message);
+            var nonce = jwt?.Payload.Nonce;
```
muratg: Can `PayLoad be null?

muratg:
:shipit:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: Shouldn't be.

brentschmaltz: Correct, Payload is never null.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -648,6 +679,12 @@ public OpenIdConnectHandler(HttpClient backchannel)
...
             user = userInformationReceivedContext.User;
 
+            Options.ProtocolValidator.ValidateUserInfoResponse(new OpenIdConnectProtocolValidationContext()
+            {
+                ClientId = Options.ClientId,
+                ProtocolMessage = message,
+            });
+
             var identity = (ClaimsIdentity)ticket.Principal.Identity;
             var subjectClaimType = identity.FindFirst(ClaimTypes.NameIdentifier);
```
brentschmaltz: This is a potential problem since there could be some mapping going on here. If users mapped 'sub' to something else in the ClaimsIdentity in the ticket. We should open a RC tracking issue.

samples/OpenIdConnectSample/Startup.cs
```diff
@@ -28,8 +29,10 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
             app.UseOpenIdConnectAuthentication(options =>
             {
                 options.ClientId = "63a87a83-64b9-4ac1-b2c5-092126f8474f";
+                options.ClientSecret = "Yse2iP7tO1Azq0iDajNisMaTSnIDv+FXmAsFuXr+Cy8=";
```
PinpointTownes: nit: what's the point of setting `client_secret` if you're also using `response_type=id_token`?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -648,6 +679,12 @@ public OpenIdConnectHandler(HttpClient backchannel)
             ticket = userInformationReceivedContext.AuthenticationTicket;
             user = userInformationReceivedContext.User;
 
+            Options.ProtocolValidator.ValidateUserInfoResponse(new OpenIdConnectProtocolValidationContext()
+            {
+                ClientId = Options.ClientId,
+                ProtocolMessage = message,
```
tushargupta51: This is not correct. You need to set 2 properties:
1. UserInfoEndpointResponse: which is of type JwtSecurityToken
2. ValidatedIdToken: which is the validated token that is used to create the identity (JwtSecurityToken corresponding to message.idToken

All that validator does is match the "sub" claim. Also, you can get rid of the "sub" claim matching logic below.

samples/OpenIdConnectSample/Startup.cs
Tratcher: so I can toggle back and forth between code and idtoken for manual testing.

PinpointTownes: Understood :smile: 

Is it worth adding a comment indicating that `client_secret` is not used when specifying `response_type=id_token`/`response_type=id_token token`?

Tratcher:
Updated

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
PinpointTownes: > 1. UserInfoEndpointResponse: which is the user info endpoint response and of type JwtSecurityToken. The endpoint response will become the JwtPayload of the token.

Interesting. Curious: when the userinfo response is not signed nor encrypted, a simple JSON object is returned and you have to manually "create" a JWT for `ValidateUserInfoResponse` to work correctly. Have you considered accepting a `JObject`?

PinpointTownes:
@Tratcher what's the plan for https://github.com/aspnet/Security/issues/485? Migrating to Wilson beta8 without providing a simple property to enable non-HTTPS scenarios will be a PITA.

Tratcher:
@PinpointTownes #485 is scheduled for rc1, we're out of time for beta8.

PinpointTownes:
#fail. Have you at least considered adding a warning somewhere indicating that non-HTTPS endpoints are not supported and how HTTP can be manually enabled?

Tratcher:
We digress. Lets go back to #485.

PR #503, opened 2015-09-29
OIDC RequireHttpsMetadata

Tratcher:
#485 
Something like this @PinpointTownes?
@muratg @Eilon @brentschmaltz @tushargupta51 @blowdart 

(I'll do Bearer once we agree on the fix).

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
@@ -93,6 +93,12 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
         public bool GetClaimsFromUserInfoEndpoint { get; set; }
 
         /// <summary>
+        /// Gets or sets if HTTPS is required for the metadata address or authority.
+        /// The default is true. This should be disabled only in development environments.
+        /// </summary>
+        public bool RequireHttpsMetadata { get; set; } = true;
+
+        /// <summary>
         /// Gets or sets the discovery endpoint for obtaining metadata
         /// </summary>
         public string MetadataAddress { get; set; }
--- a/src/Microsoft.AspNet.Authentication.JwtBearer/JwtBearerMiddleware.cs
+++ b/src/Microsoft.AspNet.Authentication.JwtBearer/JwtBearerMiddleware.cs
@@ -79,11 +79,17 @@ namespace Microsoft.AspNet.Authentication.JwtBearer
                         Options.MetadataAddress += ".well-known/openid-configuration";
                     }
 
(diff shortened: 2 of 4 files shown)


PinpointTownes:
:shipit: it looks fine.

3 remarks/questions though:
* I think it would be appropriate to update the OIDC middleware to throw an exception when a non-HTTPS authority/metadata endpoint is used without `RequireHttpsMetadata` being set to `false` directly at the startup time (the document retriever will only throw an exception during the challenge, it's too late IMHO).

* Should we automatically set it to `false` on production/testing environments, similarly to what you've done in Hosting to make the development stage smoother?

* Should we generalize this property to related things like the token request or the callback stage to make sure the OIDC middleware doesn't send insecure requests and doesn't listen on non-HTTPS endpoints?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
```diff
@@ -106,6 +106,12 @@ public string DisplayName
         public string ClientSecret { get; set; }
 
         /// <summary>
+        /// Gets or sets if https is required for the metadata address or authority.
```
Eilon: Make `HTTPS` uppercase in the comment.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
```diff
@@ -106,6 +106,12 @@ public string DisplayName
...
+        /// The default is true. This should only be disabled in development environments.
```
Eilon: Grammar police: `This should be disabled only in development environments.` :monkey_face:

PinpointTownes:
Is this still an option we want to add?

Tratcher:
Yes, I'll get back to it this week.

Tratcher:
Rebased and updated.
@PinpointTownes I think there's a big difference between requiring metadata retrieval to use https vs requiring your site to run on https. I have almost never needed to get my metadata from an http source, but I always develop my sites using http because https is harder to set up locally.

PinpointTownes:
> @PinpointTownes I think there's a big difference between requiring metadata retrieval to use https vs requiring your site to run on https.

Not from a security perspective. @brentschmaltz claimed that this change was to make the OIDC stuff "secure by default". If only the initial stage requires HTTPS, then it's not really "secure by default", as the whole authentication flow (authorization response, token request/response) would need to use HTTPS to avoid disclosing critical data like client credentials or access/identity/refresh tokens.

> I have almost never needed to get my metadata from an http source, but I always develop my sites using http because https is harder to set up locally.

I guess that's because you're not using your own authorization/identity server, but I bet that most IdentityServer3/ASOS users will develop their authorization server using HTTP.

PinpointTownes:
Thanks for adding the early check :+1:

HaoK:
:shipit:

Issue #506, opened 2015-10-01
Update the call to OpenIdConnectProtocolValidator.ValidateUserInfoEndpointResponse


tushargupta51:
There will a small OM change in wilson: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/263. Update the API to react to that.

brentschmaltz:
Change is in: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/commit/2b2969149fb73f0cdde07a2064eaa2f8744fd312

Eilon:
Need to update Wilson dependencies to RC1.

PR #517, opened 2015-10-13
Update to Rc1 IdentityModel, update ValidateUserInfoEndpointResponse

Tratcher:
#506 @tushargupta51 @brentschmaltz @HaoK 
Also update the sample to work with HttpPlatformHandler

### DIFF
--- a/samples/OpenIdConnectSample/project.json
+++ b/samples/OpenIdConnectSample/project.json
@@ -2,19 +2,19 @@
     "dependencies": {
         "Microsoft.AspNet.Authentication.Cookies": "1.0.0-*",
         "Microsoft.AspNet.Authentication.OpenIdConnect": "1.0.0-*",
-        "Microsoft.AspNet.DataProtection": "1.0.0-*",
-        "Microsoft.AspNet.Server.IIS": "1.0.0-*",
+        "Microsoft.AspNet.IISPlatformHandler": "1.0.0-*",
+        "Microsoft.AspNet.Server.Kestrel": "1.0.0-*",
         "Microsoft.AspNet.Server.WebListener": "1.0.0-*",
-        "Microsoft.Extensions.Logging.Console": "1.0.0-*",
-        "Microsoft.AspNet.Server.Kestrel": "1.0.0-*"
+        "Microsoft.Extensions.Logging.Console": "1.0.0-*"
     },
     "frameworks": {
         "dnx451": { },
         "dnxcore50": { }
     },
     "commands": {
-        "web": "Microsoft.AspNet.Hosting server=Microsoft.AspNet.Server.WebListener server.urls=http://localhost:42023",
-        "kestrel": "Microsoft.AspNet.Hosting --server Microsoft.AspNet.Server.Kestrel --server.urls http://localhost:42023"
+        "web": "Microsoft.AspNet.Server.Kestrel",
+        "kestrel": "Microsoft.AspNet.Server.Kestrel --server.urls http://localhost:42023",
+        "weblistener": "Microsoft.AspNet.Server.WebListener --server.urls http://localhost:42023"
     },
     "webroot": "wwwroot"
 }
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
@@ -659,8 +659,21 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
             var responseMessage = await Backchannel.SendAsync(requestMessage);
             responseMessage.EnsureSuccessStatusCode();
             var userInfoResponse = await responseMessage.Content.ReadAsStringAsync();
-            var userInfoEndpointJwt = new JwtSecurityToken(userInfoResponse);
-            var user = JObject.Parse(userInfoResponse);
+            JObject user;
+            var contentType = responseMessage.Content.Headers.ContentType;
+            if (contentType.MediaType.Equals("application/json", StringComparison.OrdinalIgnoreCase))
+            {
+                user = JObject.Parse(userInfoResponse);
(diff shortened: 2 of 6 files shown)


samples/OpenIdConnectSample/project.json
```diff
@@ -2,19 +2,19 @@
...
     "frameworks": {
         "dnx451": { },
         "dnxcore50": { }
     },
     "commands": {
-        "web": "Microsoft.AspNet.Hosting server=Microsoft.AspNet.Server.WebListener server.urls=http://localhost:42023",
-        "kestrel": "Microsoft.AspNet.Hosting --server Microsoft.AspNet.Server.Kestrel --server.urls http://localhost:42023"
+        "web": "Microsoft.AspNet.Server.Kestrel",
+        "kestrel": "Microsoft.AspNet.Server.Kestrel --server.urls http://localhost:42023",
+        "weblistenter": "Microsoft.AspNet.Server.WebListener --server.urls http://localhost:42023"
```
PinpointTownes: `weblistenter` -> `weblistener`

PinpointTownes:
nit: there's a typo in the title: `IdentityMode` -> `IdentityModel` :trollface:

Tratcher:
Updated

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -669,7 +669,6 @@ public OpenIdConnectHandler(HttpClient backchannel)
             var responseMessage = await Backchannel.SendAsync(requestMessage);
             responseMessage.EnsureSuccessStatusCode();
             var userInfoResponse = await responseMessage.Content.ReadAsStringAsync();
-            var userInfoEndpointJwt = new JwtSecurityToken(userInfoResponse);
             var user = JObject.Parse(userInfoResponse);
```
tushargupta51: What happens if userInfoResponse is a jwt?

Tratcher: When would it be a JWT? And would the content-type be different?

tushargupta51: If the response is signed, then it will be a jwt. content-type will be application/jwt. http://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse

Tratcher: Is that currently supported by AAD? Or some other server I can test against?

tushargupta51: @Tratcher I and @brentschmaltz talked about this and we were thinking if the protocol validator should take care of the claims merging so that the handler does not need to worry about the format of the response, which is what we wanted when we changed the type from JwtSecurityToken to string. On your end, you need to update the call to validate method to pass one extra parameter: ticket.Principal.Identity and also update the UserInfoReceivedContext to take 'userinfoendpointresponse' instead of 'user'. Validate method will return the identity with merged claims.

Tratcher: No, we want UserInfoReceivedContext to take a JObject so the developer doesn't have to deal with the same problem. I have a fix but I don't have a good way of testing it. Are there any IDPs we use that support JWT responses here?

Tratcher: Updated for reference.

tushargupta51: @Tratcher I am aware of any IDPs that returns jwt. For testing, I crafted up a JWT. If we are going for your solution, then it makes sense to remove that logic from protocol validator, otherwise it will be redundant. Instead, it can take the 'user' object.

brentschmaltz:
@Tratcher At line 736, we  should not be creating a new ticket, just return the existing.
return new AuthenticationTicket(new ClaimsPrincipal(identity), ticket.Properties, ticket.AuthenticationScheme);

Tratcher:
Updated. @tushargupta51 I've done some manual testing to verify the code that converts from JWTs to JObjects.

tushargupta51:
@Tratcher looks good to me. :shipit:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -659,8 +659,21 @@ public OpenIdConnectHandler(HttpClient backchannel, IHtmlEncoder htmlEncoder)
...
+            JObject user;
+            var contentType = responseMessage.Content.Headers.ContentType;
+            if (contentType.MediaType.Equals("application/json", StringComparison.OrdinalIgnoreCase))
+            {
+                user = JObject.Parse(userInfoResponse);
+            }
+            else if (contentType.MediaType.Equals("application/jwt", StringComparison.OrdinalIgnoreCase))
+            {
+                var userInfoEndpointJwt = new JwtSecurityToken(userInfoResponse);
+                user = JObject.FromObject(userInfoEndpointJwt.Payload);
```
PinpointTownes: Curious: will this throw if the JWT userinfo response is encrypted?

Tratcher: @PinpointTownes Have you seen any servers that actually send application/jwt? It's pretty hard to validate otherwise.

PinpointTownes: Nope, sorry. Signing/encrypting the userinfo response is extremely unusual in practice :smile:

PR #520, opened 2015-10-15
Moving AllowAnonymous attribute from MVC

ajaybhargavb:
https://github.com/aspnet/Mvc/issues/3246
@rynowak @Eilon

### DIFF
--- a/src/Microsoft.AspNet.Authorization/AllowAnonymousAttribute.cs
+++ b/src/Microsoft.AspNet.Authorization/AllowAnonymousAttribute.cs
@@ -1,0 +1,12 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Authorization/AllowAnonymousAttribute.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNet.Authorization
+{
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
```
rynowak: Is method right here? shouldn't this allow class also?

ajaybhargavb: It allows class and method.

rynowak:
:shipit:

PR #521, opened 2015-10-15
Behavior tweaks, cleanup OIDC options

HaoK:
cc @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectMiddleware.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectMiddleware.cs
@@ -91,10 +91,7 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
                 throw new ArgumentException("Options.SignInScheme is required.");
             }
 
-            if (Options.HtmlEncoder == null)
-            {
-                Options.HtmlEncoder = services.GetHtmlEncoder();
-            }
+            HtmlEncoder = services.GetHtmlEncoder();
 
             if (Options.StateDataFormat == null)
             {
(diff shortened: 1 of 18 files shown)


src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectMiddleware.cs
```diff
@@ -91,10 +91,7 @@ public class OpenIdConnectMiddleware : AuthenticationMiddleware<OpenIdConnectOpt
                 throw new ArgumentException("Options.SignInScheme is required.");
             }
 
-            if (Options.HtmlEncoder == null)
-            {
-                Options.HtmlEncoder = services.GetHtmlEncoder();
-            }
+            HtmlEncoder = services.GetHtmlEncoder();
```
Tratcher: If HtmlEncoder it's not coming from options anymore then you can directly inject it via the constructor params like IUrlEncoder.

Tratcher:
:shipit:

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectMiddleware.cs
HaoK: Fair enough I'll make that change.  Slightly unrelated question, do you think we need to all of these duplicate null arguement checks?  My preference is to nuke them from the child ctors if the base ctor is already doing the checks.  We have so many overloads that each are doing the same null check since the NotNull tool put them EVERYWHERE.

HaoK:
835903892771b9e833985c43bc3b967f66f31892

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectMiddleware.cs
Tratcher: good point, you don't need to null check stuff that's directly passed to base, that with execute first anyways.

Issue #263, opened 2015-05-18
[Cookies] Consume ITlsTokenBindingFeature


Tratcher:
ITlsTokenBindingFeature represents a new TLS extension where clients can use their TPM to generate a stable machine ID and include that on each TLS handshake. The server can then use this to bind encrypted cookies to that machine ID (across TLS sessions), preventing them from being exported by malware for use on another device.  WebListener, IIS, and IE currently support this on Win10+.

Logic: Per request, if ITlsTokenBindingFeature is present and GetProvidedTokenBindingId returns a value, base64 encode the bytes into a string and call `IDataProtector.CreateProtector` to create a sub-dataprotector that is client specific. Note this should only be used to encrypt cookies that are `Secure`/https only. You wouldn't be able to decrypt the cookie if it came back over a plain http connection.

muratg:
Who'll do this? @HaoK or @Tratcher?

HaoK:
I don't have much context to this one, perhaps this one would be better for @Tratcher

PR #523, opened 2015-10-16
Consume ITlsTokenBindingFeature in CookieAuthMiddleware.

Tratcher:
#263 @HaoK @PinpointTownes

I also manually verified it works with IE (Win10) and WebListener.

### DIFF
--- a/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -8,6 +8,7 @@ using System.Security.Claims;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Authentication;
+using Microsoft.AspNet.Http.Features;
 using Microsoft.AspNet.Http.Features.Authentication;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
@@ -45,7 +46,7 @@ namespace Microsoft.AspNet.Authentication.Cookies
                 return null;
             }
 
-            var ticket = Options.TicketDataFormat.Unprotect(cookie);
+            var ticket = Options.TicketDataFormat.Unprotect(cookie, GetTlsTokenBinding());
             if (ticket == null)
             {
                 Logger.LogWarning(@"Unprotect ticket failed");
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -357,5 +358,20 @@ private static bool IsHostRelative(string path)
...
         }
+
+        private string GetTlsTokenBinding()
+        {
+            string userTokenBinding = null;
+            var tokenBinding = Context.Features.Get<ITlsTokenBindingFeature>();
+            if (tokenBinding != null)
+            {
+                var binding = tokenBinding.GetProvidedTokenBindingId();
+                if (binding != null)
```
HaoK: nit: collapse outer layer of if/null checks?
```
var binding = Context.Features.Get<ITlsTokenBindingFeature>()?.GetProvidedTokenBindingId();
```

HaoK: Could then also follow it up with no ifs and removing the local 
```
return binding == null ? null : Convert.ToBase64String(binding);
```

Tratcher: well if you want to be all fancy .... :-)

src/Microsoft.AspNet.Authentication/DataHandler/SecureDataFormat.cs
```diff
@@ -20,15 +18,30 @@ public SecureDataFormat(IDataSerializer<TData> serializer, IDataProtector protec
...
-            string protectedText = Base64UrlTextEncoder.Encode(protectedData);
-            return protectedText;
+
+            var protector = _protector;
+            if (!string.IsNullOrEmpty(purpose))
+            {
+                protector = protector.CreateProtector(purpose);
+            }
+
+            byte[] protectedData = protector.Protect(userData);
```
HaoK: nit: var police are coming...

src/Microsoft.AspNet.Authentication/DataHandler/SecureDataFormat.cs
```diff
@@ -42,7 +55,13 @@ public TData Unprotect(string protectedText)
...
                 }
 
-                byte[] userData = _protector.Unprotect(protectedData);
+                var protector = _protector;
+                if (!string.IsNullOrEmpty(purpose))
+                {
+                    protector = protector.CreateProtector(purpose);
+                }
+
+                byte[] userData = protector.Unprotect(protectedData);
```
HaoK: nit^3: var^2

HaoK:
If you say it works, I guess I have to believe you :)  :shipit:

PinpointTownes:
It looks really nice :+1:

PinpointTownes:
Curious: are there any downsides using this TLS binding token stuff?

Tratcher:
There may be some perf overhead, but I was promised it would be minimal. We'll find out...

Issue #418, opened 2015-08-27
Change log messages to use string literals instead of resources


Eilon:
Logs aren't localized.

Eilon:
https://github.com/aspnet/Security/commit/121e6891e71801ed54da8637aa79fb86c1d0c384

Issue #414, opened 2015-08-27
Remove log codes from log messages


Eilon:
Need to be consistent in all the middlesware.

Eilon:
https://github.com/aspnet/Security/commit/121e6891e71801ed54da8637aa79fb86c1d0c384

PR #525, opened 2015-10-17
Remove log codes from exception/log messages; don't localize logs

Eilon:
https://github.com/aspnet/Security/issues/414 and https://github.com/aspnet/Security/issues/418

Also started putting in event ids for logs.

@Tratcher @HaoK

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
@@ -107,12 +107,12 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
                 await Options.Events.RedirectToEndSessionEndpoint(redirectContext);
                 if (redirectContext.HandledResponse)
                 {
-                    Logger.LogVerbose("RedirectToEndSessionEndpoint.HandledResponse");
+                    Logger.LogVerbose(1, "RedirectToEndSessionEndpoint.HandledResponse");
                     return;
                 }
                 else if (redirectContext.Skipped)
                 {
-                    Logger.LogVerbose("RedirectToEndSessionEndpoint.Skipped");
+                    Logger.LogVerbose(2, "RedirectToEndSessionEndpoint.Skipped");
                     return;
                 }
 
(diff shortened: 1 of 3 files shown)


Eilon:
I focused on re-working the existing code without any significant changes. I didn't focus on whether the log *level* was right, or whether the messages were super useful. I did, however, correct smaller clarity issues in the messages.

Tratcher:
:shipit:

Eilon:
@Tratcher thanks!

HaoK:
Do we need to worry about log event ids colliding from base classes?  Also looks a bit weird seeing random ints in the code like this.

Eilon:
Hence my question to a bunch of folks yesterday. If each class uses its own type name as the category name then conflicts won't happen.

As far as ints being weird, well, they're arbitrary numbers, so they're always weird :smile: But they're unique within a category name (unless I messed one up!).

HaoK:
Yeah this just feels a bit weird and error prone for the future, its not too bad right now with most of the classes having < 5 log sites, but I can imagine if we start getting having some files with 30+ logger calls, adding a new log code later on is going to result in a struggle to find what's the appropriate nextId(), especially if you add it in the middle, then file order is broken and anarchy ensues :)

Eilon:
Yeah these are all good points. Clearly the event ids should be the SHA hash if the call site or something :smile:

HaoK:
Yeah if we just want it to always be unique, seems better to just automagically make it so...

Eilon:
It has to be two things:

1. Unique
2. Stable

Solving 1. is easy, but solving 2. is difficult.

HaoK:
Sounds like something like how we name our unit tests is appropriate maybe?
"AuthFailureCuzNoClaim"
"AuthSuccess"
"CookieFailedValidationNoTicket"

Eilon:
The codes are ints, not strings.

There's a category name, for which solving uniqueness is fairly easy: namespace+type name.

There's also an event id, which is an int, and has to be unique within a category name.

HaoK:
Yeah so can't we just use hash code for a logical string to make it stable and uniquish?

Eilon:
Well that's not stable across CLR versions either :frowning:

HaoK:
We could write our own stable helper hasher then? `GenerateEventId("ZeroIsSoUgly")`

Eilon:
Honestly I'm not sure that's much better. There's another proposal floating about that's suggesting using opaque strings, which while they certainly don't guarantee uniqueness either, are perhaps easier to deal with.

Tratcher:
How about a pattern like this for bookkeeping?
```c#
internal static class FacebookEventIds
{
  public const int EventAbc1 = 1;
  public const int EventDef2 = 2;
}
```
I think you could do something similar with an `enum FacebookEventIds : int`.

HaoK:
I like this more than the random ints sprinkled around.  Reminds me of the pattern Eilon just nuked with error codes :)

Eilon:
Ah yeah sure I just looked at some EF code and I think they have a class with identifiers for all these.

Logged https://github.com/aspnet/Security/issues/532 and assigned to myself.

Issue #491, opened 2015-09-24
Add Logging to Authn and Authz


rynowak:
From offline logging discussion:

- Add Verbose logging where it can be useful for debugging
- Time box to 2 dev-days maximum
- Use EventIds at each logging statement
- Update any existing logging to include an EventId

Relevant logging points we want to include for sure:
```
    Authn:
        -Identity Authenticated - Information
        -SignIn/SighOut - Information
        -Challenge - Information
        -Access Denied - Information

    Authz:
        - Granted/Denied - Information
```

muratg:
FYI @rynowak

muratg:
FYI @Tratcher

muratg:
@HaoK Will you be able to tackle this early in RC?

HaoK:
Yep

Eilon:
We have security repo triage this afternoon.

HaoK:
e0464c950853909aa3491c65b519db47bb22679f

PR #530, opened 2015-10-19
Add some basic logging for AuthN/AuthZ

HaoK:
Fixes https://github.com/aspnet/Security/issues/491

@Tratcher @Eilon @divega 

Only added the important top level logging needed for the review later this week

### DIFF
--- a/src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
@@ -103,6 +103,7 @@ namespace Microsoft.AspNet.Authentication
                 var ticket = result?.Ticket;
                 if (ticket?.Principal != null)
                 {
+                    Logger.LogInformation(0, $"HttContext.User merged via AutomaticAuthentication from authenticationScheme: {Options.AuthenticationScheme}.");
                     Context.User = SecurityHelper.MergeUserPrincipal(Context.User, ticket.Principal);
                 }
             }
@@ -208,11 +209,13 @@ namespace Microsoft.AspNet.Authentication
                     var ticket = result?.Ticket;
                     if (ticket?.Principal != null)
                     {
+                        Logger.LogInformation(1, $"AuthenticationScheme: {Options.AuthenticationScheme} was successfully authenticated.");
                         context.Authenticated(ticket.Principal, ticket.Properties.Items, Options.Description.Items);
                         handled = true;
                     }
                     else
                     {
+                        Logger.LogVerbose(2, $"AuthenticationScheme: {Options.AuthenticationScheme} was not authenticated.");
                         context.NotAuthenticated();
                     }
                 }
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -240,6 +243,7 @@ protected Task<AuthenticateResult> HandleAuthenticateOnceAsync()
             if (ShouldHandleScheme(context.AuthenticationScheme, handleAutomatic: false))
             {
                 SignInAccepted = true;
+                Logger.LogInformation(3, $"AuthenticationScheme: {Options.AuthenticationScheme} signed in.");
                 await HandleSignInAsync(context);
```
Tratcher: It's better to log afterwards in case there are exceptions.

HaoK:
Updated moving logging after method calls

divega:
LGTM. CC @rynowak and @loudej who I think can look at this from a wider logging perspective.

src/Microsoft.AspNet.Authorization/DefaultAuthorizationService.cs
```diff
@@ -33,7 +50,17 @@ public DefaultAuthorizationService(IOptions<AuthorizationOptions> options, IEnum
...
-            return authContext.HasSucceeded;
+
+            if (authContext.HasSucceeded)
+            {
+                _logger.LogInformation(0, $"Authorization was successful for user: {user?.Identity?.Name}.");
+                return true;
+            }
+            else
+            {
+                _logger.LogInformation(0, $"Authorization failed for user: {user?.Identity?.Name}.");
```
Tratcher: 0?

HaoK: Grrr, copy paste fail, yeah event ids, manually assigning them for now, 0 and 1 :)

Tratcher:
:shipit:

src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs
```diff
@@ -104,6 +104,7 @@ protected string CurrentUri
                 if (ticket?.Principal != null)
                 {
                     Context.User = SecurityHelper.MergeUserPrincipal(Context.User, ticket.Principal);
+                    Logger.LogInformation(0, $"HttContext.User merged via AutomaticAuthentication from authenticationScheme: {Options.AuthenticationScheme}.");
```
Eilon: You know, come to think of it, I think these aren't supposed to use string interpolation or format strings.

Logging has its own format stuff for structured data, something like:

```c#
Logger.LogInformation(0, "HttContext.User merged via AutomaticAuthentication from authenticationScheme: {AuthenticationScheme}.", Options.AuthenticationScheme);
```

Note that there's no string formatting going on here, but Logging will see the `{foo}` in the string and create an associated structured data element with `AuthenticationScheme = Options.AuthenticationScheme`.

I missed this in my earlier logging change.

HaoK: Ah that's right, removed all interpolation

HaoK:
e0464c950853909aa3491c65b519db47bb22679f

Issue #494, opened 2015-09-24
resource-based AuthorizationHandler is invoked for other resource types.


MikeWasson:
I have two authZ handlers that derive from `AuthorizationHandler<TRequirement, TResource>` with different resource types:
 
    services.AddInstance<IAuthorizationHandler>(new FooAuthorizationHandler());
    services.AddInstance<IAuthorizationHandler>(new BarAuthorizationHandler());
 
The AuthorizationService runs both handlers, passing in null when the resource type does not match. Is this the intended behavior? If so, I need to check for null resource and exit.

HaoK:
defc9faac06958b89ad3bf8fa28a135265cef46f

PR #533, opened 2015-10-20
Don't call resource handlers without required resource

HaoK:
Fixes https://github.com/aspnet/Security/issues/494

cc @divega @blowdart


HaoK:
defc9faac06958b89ad3bf8fa28a135265cef46f

PR #534, opened 2015-10-20
DI Aware AuthZ Requirement changes

HaoK:


HaoK:
@divega @blowdart Updated with the proposed DI aware IRequirements.  It causes some significant changes to some of the surface area.

Before: IRequirements were always instances, now we need to register factory methods to produce these, and only create them at the very end.  As a result all of the old collections of Requirements are now internal/hidden, as they have turned into Func of ISP => IRequirements.  Other than that the concepts are mostly the same, except we also had to drop any extension methods that directly authorized against Policies as well since they also no longer have instances.

I think these changes are relatively harmless in terms of usage...thoughts?

blowdart:
I'm ok with that.

HaoK:
Decided against this for now

PR #536, opened 2015-10-20
Update samples with error handling

HaoK:
@Tratcher

### DIFF
--- a/samples/SocialSample/Startup.cs
+++ b/samples/SocialSample/Startup.cs
@@ -2,15 +2,18 @@ using System.Linq;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Security.Claims;
+using System.Threading.Tasks;
 using Microsoft.AspNet.Authentication.Cookies;
 using Microsoft.AspNet.Authentication.Google;
 using Microsoft.AspNet.Authentication.MicrosoftAccount;
 using Microsoft.AspNet.Authentication.OAuth;
+using Microsoft.AspNet.Authentication.Twitter;
 using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Authentication;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.WebEncoders;
 using Newtonsoft.Json.Linq;
 
 namespace CookieSample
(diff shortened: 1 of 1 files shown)


samples/SocialSample/Startup.cs
```diff
@@ -58,6 +61,17 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
             {
                 options.ClientId = "560027070069-37ldt4kfuohhu3m495hk2j4pjp92d382.apps.googleusercontent.com";
                 options.ClientSecret = "n2Q-GEw9RQjzcRbU3qhfTj8f";
+                options.Events = new OAuthEvents()
+                {
+                    OnRemoteError = ctx =>
```
PinpointTownes: Meh, I thought we now had a property to automatically redirect the user in this case? :confused:

Tratcher: We opted not to have a default redirect yet as it would most likely cause a hard to understand 404. It's also not clear how many of these error text messages should be leaked to the client.

HaoK: We can certainly consider adding sugar later, this is mostly just the most basic building of an error event to start.

Tratcher:
:shipit:

samples/SocialSample/Startup.cs
PinpointTownes: Hum, understood. Are there plans to make this error path better in the future? Looks like throwing exceptions and letting them surface up to the (optional) exception handler middleware might be a better approach.

Tratcher: There's a difference between exceptions and protocol errors. Unexpected exceptions will already bubble up, protocol errors need more direct handling (though you could choose to throw in OnRemoteError).

HaoK: That's always the fallback option, this is supposed to only be for any specific targetted exceptions that can be special case handled.  We haven't fully worked through those yet, but yes the idea is a few common error paths should be handled somewhat more gracefully than today (which is an admittedly low bar :)

HaoK: That's an interesting thought, maybe we should just have OnRemoteError throw with the message by default?

Tratcher: I guess that would result in the same 500 most of the time if you don't want to handle those cases. Put it here? https://github.com/aspnet/Security/blob/c14119b612151d0d528b8cdb2bfaa2f11f55ef40/src/Microsoft.AspNet.Authentication/RemoteAuthenticationHandler.cs#L39

HaoK: Yeah seems like a better flow than just a flat 500

HaoK:
b09777fe386251f0077a5f8820d1d80007008df7

Issue #455, opened 2015-09-16
Consider requiring callback path


Tratcher:
OIDC is the only middleware with an optional callback path, and it's not set by default.

HandleAuthenticateAsync:
`if (Options.CallbackPath.HasValue && Options.CallbackPath != (Request.PathBase + Request.Path))`

This results in it reading every query string or form body and trying to authenticate.

WsFed needed this this because it could accept logins without first sending a challenge. OIDC doesn't support this because State, Nonce, correlation id, etc. are required.

The only other argument I remember for this design was that developers may not control the value, it may be dicated by the IDP. Even in that case, they should be able to set the path to whatever the IDP requires (usually `/`).

Consider making callback path required and provide a default value.

leastprivilege:
Shouldn't the callback not just default to the redirect URI? IOW - the MW should only listen to the redirect URI (then callback path is redundant)

Eilon:
Let's require it and have a default value of `/oidc` (or similar).

brentschmaltz:
@leastprivilege are you suggesting Options.RedirectUri? If so, we need to account for DefaultToCurrentUriOnRedirect (if we keep this property #478).

leastprivilege:
Either Callback path is not needed (when you have a single middleware of the same type) - or you need to set it because you have e.g. two OIDC MWs (or WS-Fed) - in this case the Callback is always the same as the redirect URI (or wreply). 

That's why I don't see the purpose of this property at all.

Am I missing something?

Tratcher:
I think it's the redirect uri that we should consider removing. We only had both because it was said IDPs had very strict requirements for it, but we've observed similar requirements for the OAuth providers and they haven't had any trouble using just callback path and generating the redirect uri on demand.

In the worst case if the default generation isn't adequate then we have events you can hook to modify it as needed.

leastprivilege:
but redirect URI is part of the protocol - callback path a historical leftover because the ws-fed MW "listened" to all incoming requests (to not break backwards compat with WIF FAM).

HaoK:
@Tratcher Do we have consensus on what the plan for this bug is?  Callback logic has been pushed into the new RemoteAuthentication base so OIDC is now consistent with the other remote middlewares

Tratcher:
[...]

PR #540, opened 2015-10-22
Remove RedirectUri from OIDC, use CallbackPath.

Tratcher:
#455 @HaoK @brentschmaltz @tushargupta51 @PinpointTownes @leastprivilege
From what we can tell RedirectUri can be correctly generated in most scenarios, and if you really need to override it you can do so in the events. This brings OIDC in line with the other middleware and requires less configuration parameter.

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
@@ -192,7 +192,7 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
             {
                 ClientId = Options.ClientId,
                 IssuerAddress = _configuration?.AuthorizationEndpoint ?? string.Empty,
-                RedirectUri = Options.RedirectUri,
+                RedirectUri = BuildRedirectUri(Options.CallbackPath),
                 Resource = Options.Resource,
                 ResponseType = Options.ResponseType,
                 Scope = string.Join(" ", Options.Scope)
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -192,7 +192,7 @@ public OpenIdConnectHandler(HttpClient backchannel, IHtmlEncoder htmlEncoder)
             {
                 ClientId = Options.ClientId,
                 IssuerAddress = _configuration?.AuthorizationEndpoint ?? string.Empty,
-                RedirectUri = Options.RedirectUri,
+                RedirectUri = Request.Scheme + "://" + Request.Host + OriginalPathBase + Options.CallbackPath,
```
HaoK: use the helper: BuildRedirectUri(Options.CallbackPath)

HaoK:
One minor thing then :shipit:

i4Contractor:
I am upgrading an old application to use the .net core. There is a reverse proxy on another server that redirects to this server. Without RedirectUri, how the heck am I supposed to change where this is redirected too, it should be redirected to the reverse proxies' hostname instead of this server's. You mention that this can be overridden in the events, I have played around with these for quite sometime and don't understand what I am supposed to do to fix this. Any help would be appreciated @Tratcher

Tratcher:
@i4Contractor Link generation for reverse proxies is a standard scenario. We're working on the docs here: https://github.com/aspnet/Docs/issues/2384.

PR #541, opened 2015-10-22
Switch to throwing instead of just setting 500

HaoK:
cc @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
@@ -7,6 +7,7 @@ using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Security.Claims;
 using System.Security.Cryptography;
+using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Authentication;
@@ -125,8 +126,17 @@ namespace Microsoft.AspNet.Authentication.OAuth
             }
             else
             {
-                return OAuthTokenResponse.Failed("OAuth token endpoint failure: "+response.StatusCode);
+                var error = "OAuth token endpoint failure: " + await Display(response);
+                return OAuthTokenResponse.Failed(new Exception(error));
             }
+        }
+
+        private static async Task<string> Display(HttpResponseMessage response)
+        {
+            var output = new StringBuilder();
+            output.Append("Headers: " + response.Headers.ToString() + ";");
+            output.Append("Body: " + await response.Content.ReadAsStringAsync() + ";");
+            return output.ToString();
         }
 
         protected virtual async Task<AuthenticationTicket> CreateTicketAsync(ClaimsIdentity identity, AuthenticationProperties properties, OAuthTokenResponse tokens)
--- a/src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
@@ -134,6 +134,7 @@ namespace Microsoft.AspNet.Authentication.OAuth
         private static async Task<string> Display(HttpResponseMessage response)
         {
             var output = new StringBuilder();
+            output.Append("Status: " + response.StatusCode + ";");
             output.Append("Headers: " + response.Headers.ToString() + ";");
             output.Append("Body: " + await response.Content.ReadAsStringAsync() + ";");
             return output.ToString();
--- a/src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
@@ -61,7 +61,13 @@ namespace Microsoft.AspNet.Authentication.OAuth
             }
 
             var tokens = await ExchangeCodeAsync(code, BuildRedirectUri(Options.CallbackPath));
-            if (tokens == null || string.IsNullOrEmpty(tokens.AccessToken))
+
+            if (tokens.Error != null)
+            {
+                return AuthenticateResult.Failed(tokens.Error);
+            }
+
+            if (string.IsNullOrEmpty(tokens.AccessToken))
             {
                 return AuthenticateResult.Failed("Failed to retrieve access token.");
             }
@@ -115,11 +121,11 @@ namespace Microsoft.AspNet.Authentication.OAuth
             if (response.IsSuccessStatusCode)
             {
                 var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
-                return new OAuthTokenResponse(payload);
+                return OAuthTokenResponse.Success(payload);
             }
             else
             {
-                return null;
+                return OAuthTokenResponse.Failed("OAuth token endpoint failure: "+response.StatusCode);
             }
         }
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNet.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -36,8 +36,7 @@ public abstract class RemoteAuthenticationHandler<TOptions> : AuthenticationHand
                     return false;
                 }
 
-                Context.Response.StatusCode = 500;
-                return true;
+                throw errorContext.Error;
```
Tratcher: This could already be a caught exception so re-throwing it will reset the call stack. Do you need to wrap it?

test/Microsoft.AspNet.Authentication.Test/project.json
```diff
@@ -10,6 +10,7 @@
     "Microsoft.AspNet.Authentication.MicrosoftAccount": "1.0.0-*",
     "Microsoft.AspNet.Authentication.OpenIdConnect": "1.0.0-*",
     "Microsoft.AspNet.Authentication.Twitter": "1.0.0-*",
+    "Microsoft.AspNet.Diagnostics": "1.0.0-*",
```
Tratcher: Sorry, you can't add this dependency in this repo.

test/Microsoft.AspNet.Authentication.Test/Google/GoogleMiddlewareTests.cs
```diff
@@ -403,23 +403,10 @@ public class GoogleMiddlewareTests
 
             var state = stateFormat.Protect(properties);
 
-            await Assert.ThrowsAsync<HttpRequestException>(() => server.SendAsync(
+            var transaction = await server.SendAsync(
```
Tratcher: You can just check that the inner exception is the one you threw.

test/Microsoft.AspNet.Authentication.Test/project.json
HaoK: Lame... ok so I guess we just blow up in our tests then?  At least I verified that things work :)

src/Microsoft.AspNet.Authentication/RemoteAuthenticationHandler.cs
HaoK: Not sure I follow, the errors in context today aren't actually usually caught, we are constructing the exceptions mostly as opposed to catching and stuffing them in here...

Tratcher: Not usually no, but the certainly can be. We don't want to loose that information if it's there.

HaoK: Ok so we should just wrap it inside of another exception then?

Tratcher: Yes, though I couldn't say what type. Aggregate? InvalidOperation?

HaoK: Went with aggregate since it doesn't do anything really except wrap

HaoK:
Updated, and removed EnsureStatusCode and replaced with a generic error

src/Microsoft.AspNet.Authentication.OAuth/OAuthTokenResponse.cs
```diff
@@ -16,10 +16,26 @@ public OAuthTokenResponse(JObject response)
...
+        {
+            return new OAuthTokenResponse(error);
+        }
+
         public JObject Response { get; set; }
         public string AccessToken { get; set; }
         public string TokenType { get; set; }
         public string RefreshToken { get; set; }
         public string ExpiresIn { get; set; }
+        public string Error { get; set; }
```
Tratcher: Should Error be an Exception like on AuthenticateResult?

src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -113,9 +118,15 @@ public OAuthHandler(HttpClient backchannel)
...
-            var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
-            return new OAuthTokenResponse(payload);
+            if (response.IsSuccessStatusCode)
+            {
+                var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
+                return OAuthTokenResponse.Success(payload);
+            }
+            else
+            {
+                return OAuthTokenResponse.Failed("OAuth token endpoint failure: "+response.StatusCode);
```
Tratcher: nit: spacing.

Tratcher: Would it make sense to capture any more data here? response.ToString? response.Body.ReadAsStringAsync()? I don't see how you'd debug this with just the status code.

src/Microsoft.AspNet.Authentication.OAuth/OAuthTokenResponse.cs
PinpointTownes: Yeah, this looks weird. `error` is a standard OAuth2 parameter, but this property is just a user-defined exception.

src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
PinpointTownes: I guess `response.Body.ReadAsStringAsync()` would be a good candidate for advanced logging data.

HaoK: When I looked at this with google at least, the body was empty, but we can do something more by default sure.  response body?

PinpointTownes: Ideally, we should determine on the failure path if the response is a JSON payload and try to extract `error`, `error_description` and `error_uri` to log them.

Tratcher: Dumping the whole headers and body is a good catch-all.

HaoK:
Updated: string => Exception for error, we now dump out the headers and content in the error response.  This should capture the json somewhat.  We could special case a json payload if we wanted still.

src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
PinpointTownes: Curious: why not throwing an exception here?
IMHO, this code path is "exceptional" enough to deserve an exception :smile:

Tratcher: We want a graceful path to the Error event without doing a global try/catch

HaoK: Yeah we are just deferring the throw, by default if no error handler handles it, we end up throwing from the RemoteAuthenticationHandler code

PinpointTownes: Your graceful path seems really fragile. I guess even a simple transient network error could throw an exception that would escape your graceful path.

Tratcher: The graceful path is for known error conditions. Yes, unkown errors will escape.

HaoK: The graceful path is indeed very narrow :)  Basically we only wanted to flow targeted errors, so for this particular method, we are only trying to expose a way to do something for a bad oauth token response, nothing more.

Tratcher:
:shipit:

HaoK:
238fdf24e8b0e3c5e84cd8dc7f7324502a439429

Issue #508, opened 2015-10-06
Cookies: HTTP 401 always turns into 302


kpocza:
When using cookie authentication the HTTP 401 response always turns into HTTP 302 Redirect to Accounts/Login.
I setup the authentication this way:

```c#
            app.UseCookieAuthentication(options => 
            {
                options.CookieHttpOnly = false;
                options.ExpireTimeSpan = System.TimeSpan.FromMinutes(60);
                options.SlidingExpiration = true;
                options.CookieName = "ThriotMgmtAuth";
                options.AutomaticAuthentication = true;
            });
```

I've tried to set the LoginPath to empty, since the engine redirects only when there is a valid login url. However it's already empty.

As I see the CookieAuthenticationMiddleware class updates the LoginPath with the default value when it's empty.

I think this is not the right behavior in every situations. In my case I have a cookie-protected API and call it from AngularJS from a browser besides other places.

Further details:
https://github.com/kpocza/thriot/tree/master/Service/Management/Thriot.Management.WebApi

PinpointTownes:
(duplicate of https://github.com/aspnet/Security/issues/505)

+1. I completely understand it's a way to deter us from using cookies authentication for Web API endpoints, but not being able to disable these redirects is evil.

/cc @HaoK

kpocza:
I've created a workaround for beta7.

New middleware based on the built-in cookieauth middleware:
+++
    public class WorkaroundCookieAuthenticationMiddleware : CookieAuthenticationMiddleware
    {
        public WorkaroundCookieAuthenticationMiddleware(RequestDelegate next, IDataProtectionProvider dataProtectionProvider, ILoggerFactory loggerFactory, IUrlEncoder urlEncoder, IOptions<CookieAuthenticationOptions> options, ConfigureOptions<CookieAuthenticationOptions> configureOptions) : 
            base(next, dataProtectionProvider, loggerFactory, urlEncoder, options, configureOptions)
        {
            ((CookieAuthenticationNotifications) Options.Notifications).OnApplyRedirect = context =>
            {
                context.Response.StatusCode = 401;
            };
        }
    }
+++

Registration:
+++ 
app.UseMiddleware<Thriot.Management.WebApi.Auth.WorkaroundCookieAuthenticationMiddleware>(
                new ConfigureOptions<CookieAuthenticationOptions>(options =>
                {
                    options.CookieHttpOnly = false;
                    options.ExpireTimeSpan = System.TimeSpan.FromMinutes(60);
[...]

PR #543, opened 2015-10-22
Disable redirects for Ajax requests

HaoK:
cc @Tratcher @blowdart

### DIFF
--- a/src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
+++ b/src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
@@ -37,16 +37,73 @@ namespace Microsoft.AspNet.Authentication.Cookies
         /// <summary>
         /// A delegate assigned to this property will be invoked when the related method is called.
         /// </summary>
-        public Func<CookieRedirectContext, Task> OnRedirect { get; set; } = context =>
+        public Func<CookieRedirectContext, Task> OnRedirectToLogin { get; set; } = context =>
         {
-            context.Response.Redirect(context.RedirectUri);
+            if (IsAjaxRequest(context.Request))
+            {
+                context.Response.Headers["Location"] = context.RedirectUri;
+                context.Response.StatusCode = 401;
+            }
+            else
+            {
+                context.Response.Redirect(context.RedirectUri);
+            }
+            return Task.FromResult(0);
+        };
+
+        /// <summary>
+        /// A delegate assigned to this property will be invoked when the related method is called.
+        /// </summary>
+        public Func<CookieRedirectContext, Task> OnRedirectToAccessDenied { get; set; } = context =>
+        {
+            if (IsAjaxRequest(context.Request))
+            {
+                context.Response.Headers["Location"] = context.RedirectUri;
+                context.Response.StatusCode = 403;
+            }
+            else
+            {
+                context.Response.Redirect(context.RedirectUri);
+            }
+            return Task.FromResult(0);
+        };
+
+        /// <summary>
+        /// A delegate assigned to this property will be invoked when the related method is called.
+        /// </summary>
+        public Func<CookieRedirectContext, Task> OnRedirectToLogout { get; set; } = context =>
+        {
+            if (IsAjaxRequest(context.Request))
+            {
+                context.Response.Headers["Location"] = context.RedirectUri;
+            }
+            else
+            {
+                context.Response.Redirect(context.RedirectUri);
+            }
+            return Task.FromResult(0);
+        };
+
+        /// <summary>
+        /// A delegate assigned to this property will be invoked when the related method is called.
+        /// </summary>
+        public Func<CookieRedirectContext, Task> OnRedirectToReturnUrl { get; set; } = context =>
+        {
+            if (IsAjaxRequest(context.Request))
+            {
+                context.Response.Headers["Location"] = context.RedirectUri;
+            }
+            else
+            {
+                context.Response.Redirect(context.RedirectUri);
+            }
             return Task.FromResult(0);
         };
 
         private static bool IsAjaxRequest(HttpRequest request)
         {
-            return request?.Query["X-Requested-With"] == "XMLHttpRequest" ||
-                request?.Headers["X-Requested-With"] == "XMLHttpRequest";
+            return request.Query["X-Requested-With"] == "XMLHttpRequest" ||
+                request.Headers["X-Requested-With"] == "XMLHttpRequest";
         }		
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


Tratcher:
#508. @PinpointTownes

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
```diff
@@ -42,6 +43,28 @@ public class CookieAuthenticationEvents : ICookieAuthenticationEvents
             return Task.FromResult(0);
         };
 
+        private static bool IsAjaxRequest(HttpRequest request)
+        {		
+            var query = request.Query;		
+            if (query != null)		
+            {		
+                if (query["X-Requested-With"] == "XMLHttpRequest")
```
Tratcher: These null checks shouldn't be needed. Worse case you can combine them into a single if.

HaoK: Indeed, collapsed into one return, I just lifted this code from the old stuff as is before :)

PinpointTownes:
Nice.

Could we also remove these checks and move the default values to the options class? https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationMiddleware.cs#L65-L76

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
```diff
@@ -71,24 +94,73 @@ public class CookieAuthenticationEvents : ICookieAuthenticationEvents
...
+        }
 
         /// <summary>
         /// Implements the interface method by invoking the related delegate method.
         /// </summary>
         /// <param name="context">Contains information about the event</param>
-        public virtual Task RedirectToLogin(CookieRedirectContext context) => OnRedirect(context);
+        public virtual Task RedirectToLogin(CookieRedirectContext context)
+        {
+            if (IsAjaxRequest(context.Request))
```
Tratcher: This isn't the right place for a default, it's very difficult to override. It should go into OnRedirect.

Why isn't there a different OnRedirect* for each of these?

HaoK: I believe you suggested to collapse them all into one in the last PR for this :)  They started all different

Tratcher: Really? Then it was a terrible idea, nobody uses these directly. Maybe I meant that the same func could be assigned to each of them.

HaoK: That said unless we had a central redirect place that took an optional UseThisStatusCodeForAjax parameter, its always going to be a bit meh with 4 different redirect events :/

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
```diff
@@ -71,24 +78,73 @@ public class CookieAuthenticationEvents : ICookieAuthenticationEvents
         /// Implements the interface method by invoking the related delegate method.
         /// </summary>
         /// <param name="context">Contains information about the event</param>
-        public virtual Task RedirectToLogout(CookieRedirectContext context) => OnRedirect(context);
+        public virtual Task RedirectToLogout(CookieRedirectContext context)
+        {
+            if (IsAjaxRequest(context.Request))
+            {
+                context.Response.Headers["Location"] = context.RedirectUri;
+                context.Response.StatusCode = 200;
```
Tratcher: No need to set the status code for logout.

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
```diff
@@ -71,24 +78,73 @@ public class CookieAuthenticationEvents : ICookieAuthenticationEvents
...
         /// Implements the interface method by invoking the related delegate method.
         /// </summary>
         /// <param name="context">Contains information about the event</param>
-        public virtual Task RedirectToReturnUrl(CookieRedirectContext context) => OnRedirect(context);
+        public virtual Task RedirectToReturnUrl(CookieRedirectContext context)
+        {
+            if (IsAjaxRequest(context.Request))
+            {
+                context.Response.Headers["Location"] = context.RedirectUri;
+                context.Response.StatusCode = 200;
```
Tratcher: No need to set the status code here.

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
HaoK: Okay I'll move these to a new OnXyz for each redirect.  Still seems a bit meh that each one will need to be overriden

Tratcher: Yeah, but without individual On's you still don't know which redirect is happening.

PinpointTownes: Ideally, why not moving these default "behaviors" to the handler itself, just like the other events classes?

Tratcher: You could do that if you used BaseControlContext.

PinpointTownes: Sure. `CookieRedirectContext` seems a good candidate for that :smile:

Tratcher:
@PinpointTownes Removing those checks would break it unless we also added a "don't redirect if these are null" logic. Using the events is a better way to take control.

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
HaoK: If we don't set this, the status code appears to stay 401 for the logout test

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
HaoK: Eh cookies is the only one that isn't currently using control context, I thought it was intentional, but I can't remember anymore honestly :)

PinpointTownes:
@Tratcher IMHO, having to use the events to simply avoid the default redirects is really annoying. There are definitely cases where you don't want to automagically replace 401 responses by 302 redirects.

HaoK:
Do we just want a global options switch for cookies to turn off all redirects?  Options.DoNotRedirect = true?

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
Tratcher: why, who set it?

Tratcher:
We already have it, Options.AutomaticChallenge = false.

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
```diff
@@ -42,6 +43,12 @@ public class CookieAuthenticationEvents : ICookieAuthenticationEvents
             return Task.FromResult(0);
         };
 
+        private static bool IsAjaxRequest(HttpRequest request)
+        {
+            return request?.Query["X-Requested-With"] == "XMLHttpRequest" ||
```
PinpointTownes: Curious: why would `request` be null?

Tratcher: Don't bother with null checks here, these are core APIs.

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieAuthenticationEvents.cs
HaoK: Event was wired up to login not return url, least the tests are catching stuff :)  Fixed and updated

Tratcher:
:shipit:

HaoK:
1d2c6ba1223748df02a879622e8deb6ac23db5b7

PR #547, opened 2015-10-23
React to break change in StringValues

rynowak:


src/Microsoft.AspNet.Authentication.Cookies/ChunkingCookieManager.cs
```diff
@@ -87,7 +87,7 @@ public string GetRequestCookie(HttpContext context, string key)
                 for (var chunkId = 1; chunkId <= chunksCount; chunkId++)
                 {
                     var chunk = requestCookies[key + "C" + chunkId.ToString(CultureInfo.InvariantCulture)];
-                    if (chunk == null)
```
rynowak: Oops - tests caught this one. This should be `Count == 0`

Tratcher:
:shipit: when you fix it.

rynowak:
Looks like this is already fixed. Thanks @HaoK

HaoK:
No I mistakenly updated something else thinking it was related.  We still need this, looks good :shipit:

rynowak:
Ok, pushing it then.

PR #562, opened 2015-11-03
Forbid now includes ReturnUrl

HaoK:
cc @Tratcher


src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -327,14 +327,9 @@ private static bool IsHostRelative(string path)
 
         protected override async Task<bool> HandleForbiddenAsync(ChallengeContext context)
```
Tratcher: Shouldn't this completely mimic HandleUnauthorizedAsync? e.g. allowing an alternate redirect uri, etc..

HaoK: Sure why not, updated

Tratcher:
:shipit:

HaoK:
dc6e916bd465628c3e6441925cdb16239a0b82b9

PR #563, opened 2015-11-03
Replace Success with Failed for Skips

HaoK:
Success with null ticket throws ArgumentNull...

cc @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Authentication.JwtBearer/JwtBearerHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.JwtBearer/JwtBearerHandler.cs
@@ -37,7 +37,7 @@ namespace Microsoft.AspNet.Authentication.JwtBearer
 
                 if (receivingTokenContext.Skipped)
                 {
-                    return AuthenticateResult.Success(ticket: null);
+                    return AuthenticateResult.Failed("Skipped.");
                 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -37,7 +37,7 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
 
                 if (receivingTokenContext.Skipped)
                 {
-                    return AuthenticateResult.Success(ticket: null);
+                    return AuthenticateResult.Failed("Skipped.");
```
Tratcher: Skipped != Failed. Skipped is supposed to mean "This wasn't for me, try the next middleware". You may need to promote Skipped into AuthenticateResult to make that work.

HaoK: Failed doesn't actually mean anything. It's just syntax sugar to make the code easier to read. It just allows a null ticket :). We can rename the static helpers to align with this intent.

HaoK:
Replaced by https://github.com/aspnet/Security/pull/568

PR #568, opened 2015-11-05
Add explicit skip

HaoK:


HaoK:
cc @Tratcher

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -336,7 +336,7 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
                 }
                 else if (messageReceivedContext.Skipped)
                 {
-                    return AuthenticateResult.Success(ticket: null);
+                    return AuthenticateResult.Skipped();
```
Tratcher: For this to work you need to update https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication/RemoteAuthenticationHandler.cs#L25 as well, correct?

src/Microsoft.AspNet.Authentication/AuthenticateResult.cs
```diff
@@ -54,5 +54,9 @@ public static AuthenticateResult Failed(string errorMessage)
             return new AuthenticateResult() { Error = new Exception(errorMessage) };
         }
 
+        public static AuthenticateResult Skipped()
```
Tratcher: Add a WasSkipped property that checks Ticket == null && Error == null?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
HaoK: Why?  What does it mean to skip and produce no ticket?  That's still an error wrt to the base... We have nothing to return

HaoK: I still don't understand what it means for everything to skip and noone produces a ticket...that feels like an error to me :)

PinpointTownes:
Could we also get a `Failed` overload taking no parameter? It would be used to indicate that authentication failed while bypassing the automatic logging (to log the error separately, for instance).

Tratcher:
Replaced

PR #570, opened 2015-11-05
Fixed formatting and removed a comment that broke policheck

ajaybhargavb:
@pranavkm

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
@@ -359,7 +359,6 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
                 // if any of the error fields are set, throw error null
                 if (!string.IsNullOrEmpty(message.Error))
                 {
-                    // REVIEW: this error formatting is pretty nuts
                     Logger.LogError(13, "Message contains error: '{0}', error_description: '{1}', error_uri: '{2}'.", message.Error, message.ErrorDescription ?? "ErrorDecription null", message.ErrorUri ?? "ErrorUri null");
                     return AuthenticateResult.Failed(new OpenIdConnectProtocolException(string.Format(CultureInfo.InvariantCulture, Resources.MessageContainsError, message.Error, message.ErrorDescription ?? "ErrorDecription null", message.ErrorUri ?? "ErrorUri null")));
                 }


pranavkm:
cc @Eilon

pranavkm:
:shipit: once @Eilon's happy

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -359,9 +359,21 @@ public OpenIdConnectHandler(HttpClient backchannel, IHtmlEncoder htmlEncoder)
                 // if any of the error fields are set, throw error null
                 if (!string.IsNullOrEmpty(message.Error))
                 {
-                    // REVIEW: this error formatting is pretty nuts
-                    Logger.LogError(13, "Message contains error: '{0}', error_description: '{1}', error_uri: '{2}'.", message.Error, message.ErrorDescription ?? "ErrorDecription null", message.ErrorUri ?? "ErrorUri null");
-                    return AuthenticateResult.Failed(new OpenIdConnectProtocolException(string.Format(CultureInfo.InvariantCulture, Resources.MessageContainsError, message.Error, message.ErrorDescription ?? "ErrorDecription null", message.ErrorUri ?? "ErrorUri null")));
+                    Logger.LogError(
+                        13,
```
Tratcher: :-1:

Tratcher:
```
Logger.LogError(13, "Message contains error: '{0}', error_description: '{1}', error_uri: '{2}'.",
    message.Error, message.ErrorDescription ?? "ErrorDecription null", message.ErrorUri ?? "ErrorUri null");
```

ajaybhargavb: This goes beyond our per line character limit.

Tratcher: What limit? https://github.com/aspnet/Home/wiki/Engineering-guidelines#coding-guidelines. 
This format is inconsistent with the rest of the project and a poor use of space in this context. If you really want to cut it down then make it three lines, but not 6.

ajaybhargavb: We generally try to keep it under ~120 characters per line to make it easier to review in Github. Also I see this format used in many other places in this project. Eg: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectMiddleware.cs#L104

ajaybhargavb: But I do agree that it is a poor use of space in this context. I just did this to keep the format consistent.

HaoK: This ain't MVC, please don't make it ugly :)

ajaybhargavb: Haha. I guess I'll close this PR then? I've already added an exclusion for policheck.

Tratcher: No, please fix the offending comment, but leave the unrelated code alone.

ajaybhargavb: Done.

PinpointTownes: FWIW, why not removing the (horrible) `"ErrorDecription null"` and `"ErrorUri null"` default values? IMHO, they don't make much sense in a "semantic log"... and it would help reducing the length of this line :smile:

HaoK: Yep that was the original motivation for the un polycheck friendly comment. That logging should be simplified...a lot.

ajaybhargavb:
b05f8771ae4f6fcc4e3a6fded04046eedfb59e12

PR #573, opened 2015-11-09
Fix some authentication unit tests failing on Linux and OS X.

cesarbs:
It looks like HttpClient on Windows extracts the base exception from an AggregateException in a response, while Mono doesn't. So we need to look at the base exception of whatever exception is returned when an error is expected in a response. This works for non-aggregate exceptions because GetBaseException() returns the exception itself in that case.

cc @Tratcher


Tratcher:
@HaoK

cesarbs:
Ping.

cesarbs:
@troydai

troydai:
Looks good to me. :shipit:

HaoK:
Sorry missed this, yeah :shipit:

Issue #512, opened 2015-10-09
Update the OAuth2 handler to log error_description and error_uri when receiving an error


PinpointTownes:
https://github.com/aspnet/Security/pull/495#discussion_r41671366

FYI: the OIDC handler already does that: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L382-L383

brentschmaltz:
As long as the handler is sure the message is for itself this is fine, otherwise handlers will get in each others face. I have seen handlers where the param 'f' => failure.

PinpointTownes:
@brentschmaltz no risk: unlike the OIDC middleware (though it may change in the future: https://github.com/aspnet/Security/issues/455), the OAuth2 base middleware only listens on its own callback path, so unless you use the same one for multiple handlers, you shouldn't see any issue :+1:

HaoK:
620622f260ba235c864d753262c0e6e923db93a5

PR #578, opened 2015-11-16
Add error_description/uri info to OAuthHandler

HaoK:
cc @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
@@ -38,7 +38,20 @@ namespace Microsoft.AspNet.Authentication.OAuth
             var error = query["error"];
             if (!StringValues.IsNullOrEmpty(error))
             {
-                return AuthenticateResult.Failed(error);
+                var errorMessage = new StringBuilder();
+                errorMessage.Append(error);
+                var errorDescription = query["error_description"];
+                if (!StringValues.IsNullOrEmpty(errorDescription))
+                {
+                    errorMessage.Append(";Description=" + errorDescription);
+                }
+                var errorUri = query["error_uri"];
+                if (!StringValues.IsNullOrEmpty(errorUri))
+                {
+                    errorMessage.Append(";Uri=" + errorUri);
+                }
+
+                return AuthenticateResult.Failed(errorMessage.ToString());
             }
 
             var code = query["code"];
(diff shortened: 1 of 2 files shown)


HaoK:
Fixes https://github.com/aspnet/Security/issues/512

PinpointTownes:
Curious: is this the final design for logging error messages? IMHO, it feels very weird compared to the great semantic logging you're adding everywhere. Why not logging an error with the (optional) error_description and error_uri and just return a minimal `AuthenticateResult`?

FWIW, we started using this pattern in the OAuth2 social providers that override `ExchangeCodeAsync` and it really seems clearer: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/76c146f3bde4d2de73e34f61ecac22cd764eb981/src/AspNet.Security.OAuth.Fitbit/FitbitAuthenticationHandler.cs#L73-L79

HaoK:
The original motivation for these error handling changes was to actually enable the ability to do something other than just log.  Just logging should be more or less unaffected, we can certainly tweak the format of the logging, my hope was to push the logging into the base RemoteHandler which logs all errors at Info already today.

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication/RemoteAuthenticationHandler.cs#L28

HaoK:
So logging and then returning an error will result in double logging of the errors...

src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -38,7 +38,20 @@ public OAuthHandler(HttpClient backchannel)
             var error = query["error"];
             if (!StringValues.IsNullOrEmpty(error))
             {
-                return AuthenticateResult.Failed(error);
+                var errorMessage = new StringBuilder();
+                errorMessage.Append(error);
+                var errorDescription = query["error_description"];
+                if (!StringValues.IsNullOrEmpty(errorDescription))
+                {
+                    errorMessage.Append(";Description=" + errorDescription);
```
Tratcher: I'm amused that you use both `Append` and `+` on the same line.

Tratcher:
:shipit:

src/Microsoft.AspNet.Authentication.OAuth/OAuthHandler.cs
HaoK: Oops, switched to Append :)

HaoK:
620622f260ba235c864d753262c0e6e923db93a5

Issue #190, opened 2015-03-15
Need to do a doc pass for new AuthZ/AuthN changes


HaoK:

HaoK:
@blowdart Do you want to take a stab at this and help validate that our design is sane and understandable by someone that didn't write the code :)

HaoK:
38e89d498da65ceb603b3b19e5c8fb64e2986ca4

PR #579, opened 2015-11-17
Include AuthProperties in OIDC RedirectContext

HaoK:
cc @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -304,10 +304,7 @@ namespace Microsoft.AspNet.Authentication.Cookies
                 if (!StringValues.IsNullOrEmpty(redirectUri)
                     && IsHostRelative(redirectUri))
                 {
-                    var redirectContext = new CookieRedirectContext(Context, Options, redirectUri)
-                    {
-                        Properties = properties
-                    };
+                    var redirectContext = new CookieRedirectContext(Context, Options, redirectUri, properties);
                     await Options.Events.RedirectToReturnUrl(redirectContext);
                 }
             }
@@ -329,13 +326,14 @@ namespace Microsoft.AspNet.Authentication.Cookies
 
         protected override async Task<bool> HandleForbiddenAsync(ChallengeContext context)
         {
-            var returnUrl = new AuthenticationProperties(context.Properties).RedirectUri;
+            var properties = new AuthenticationProperties(context.Properties);
+            var returnUrl = properties.RedirectUri;
             if (string.IsNullOrEmpty(returnUrl))
             {
                 returnUrl = OriginalPathBase + Request.Path + Request.QueryString;
             }
             var accessDeniedUri = Options.AccessDeniedPath + QueryString.Create(Options.ReturnUrlParameter, returnUrl);
-            var redirectContext = new CookieRedirectContext(Context, Options, BuildRedirectUri(accessDeniedUri));
+            var redirectContext = new CookieRedirectContext(Context, Options, BuildRedirectUri(accessDeniedUri), properties);
             await Options.Events.RedirectToAccessDenied(redirectContext);
             return true;
         }
@@ -347,14 +345,15 @@ namespace Microsoft.AspNet.Authentication.Cookies
                 throw new ArgumentNullException(nameof(context));
             }
 
-            var redirectUri = new AuthenticationProperties(context.Properties).RedirectUri;
+            var properties = new AuthenticationProperties(context.Properties);
+            var redirectUri = properties.RedirectUri;
             if (string.IsNullOrEmpty(redirectUri))
             {
                 redirectUri = OriginalPathBase + Request.Path + Request.QueryString;
             }
 
             var loginUri = Options.LoginPath + QueryString.Create(Options.ReturnUrlParameter, redirectUri);
-            var redirectContext = new CookieRedirectContext(Context, Options, BuildRedirectUri(loginUri));
+            var redirectContext = new CookieRedirectContext(Context, Options, BuildRedirectUri(loginUri), properties);
             await Options.Events.RedirectToLogin(redirectContext);
             return true;
 
--- a/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -186,7 +186,7 @@ namespace Microsoft.AspNet.Authentication.Cookies
                     cookieValue,
                     cookieOptions);
 
-                await ApplyHeaders(shouldRedirectToReturnUrl: false);
+                await ApplyHeaders(shouldRedirectToReturnUrl: false, properties: ticket.Properties);
             }
         }
 
@@ -261,7 +261,7 @@ namespace Microsoft.AspNet.Authentication.Cookies
 
             // Only redirect on the login path
             var shouldRedirect = Options.LoginPath.HasValue && OriginalPath == Options.LoginPath;
-            await ApplyHeaders(shouldRedirect);
+            await ApplyHeaders(shouldRedirect, signedInContext.Properties);
         }
 
         protected override async Task HandleSignOutAsync(SignOutContext signOutContext)
@@ -277,6 +277,7 @@ namespace Microsoft.AspNet.Authentication.Cookies
             var context = new CookieSigningOutContext(
                 Context,
                 Options,
+                new AuthenticationProperties(signOutContext.Properties),
                 cookieOptions);
 
             await Options.Events.SigningOut(context);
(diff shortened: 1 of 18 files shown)


src/Microsoft.AspNet.Authentication.OpenIdConnect/Events/RedirectContext.cs
```diff
@@ -12,9 +13,12 @@ namespace Microsoft.AspNet.Authentication.OpenIdConnect
     /// </summary>
     public class RedirectContext : BaseOpenIdConnectContext
     {
-        public RedirectContext(HttpContext context, OpenIdConnectOptions options)
+        public RedirectContext(HttpContext context, OpenIdConnectOptions options, AuthenticationProperties properties)
```
Tratcher: Don't put it in the constructor, the property is already settable.

Tratcher:
This ask is generally applicable to all middleware events that can be triggered from Challenge, SignIn, SignOut, etc.. No?

HaoK:
Looks like that's only one other OIDC event, the AuthorizationCodeReceivedContext

src/Microsoft.AspNet.Authentication.OpenIdConnect/Events/RedirectContext.cs
HaoK: Updated, to use property

HaoK:
I guess we can flow it everywhere that we have properties, can add it to the ValidatedContext too

HaoK:
Worth considering moving this to the base context, as there's only three context's that don't set this now

Tratcher:
How about the non-OIDC middleware?

HaoK:
OAuth looks to already have these in the base events, JWT most likely has the same issues I can update that too

HaoK:
Changes like this really make me think the events stuff is way too complicated :)

PinpointTownes:
Definitely a must :clap:

Tratcher:
CookieRedirectContext?

HaoK:
Updated Cookies, added Properties to SigningOutContext and the CookieRedirectContext.  We use the auth ticket's properties for the FinishResponse method since we don't have any other source for the properties there...

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -279,6 +279,7 @@ private CookieOptions BuildCookieOptions()
             var context = new CookieSigningOutContext(
                 Context,
                 Options,
+                new AuthenticationProperties(signOutContext.Properties),
```
PinpointTownes: Any reason to flow the properties using a constructor parameter here and a property everywhere else?

Personally, I'd use a parameter everywhere (and probably make the `Properties` property get-only).

HaoK: Dev style mostly, I prefer ctor.  I think @Tratcher preferred properties.  I can make them consistent and ctor required everywhere I guess.  Chris thoughts?

src/Microsoft.AspNet.Authentication.Cookies/Events/CookieRedirectContext.cs
```diff
@@ -29,5 +30,7 @@ public CookieRedirectContext(HttpContext context, CookieAuthenticationOptions op
         /// </summary>
         [SuppressMessage("Microsoft.Design", "CA1056:UriPropertiesShouldNotBeStrings", Justification = "Represents header value")]
         public string RedirectUri { get; set; }
+
+        public AuthenticationProperties Properties { get; set; }
```
PinpointTownes: nit: no documentation love for these properties in this PR? :trollface:

HaoK: My plan was to wait and do a full pass, lots of doc comments are missing/out of date currently in this repo...  This issue is tracking all up https://github.com/aspnet/Security/issues/190

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
Tratcher: Property

HaoK: Property made sense when they are optional, if we are making them required, i think it belongs in the ctor.

HaoK: compile errors >> null ref exceptions...

PinpointTownes: Internal then? Because having a settable property worries me a bit.

(and it wouldn't be very consistent: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Cookies/Events/CookieSignedInContext.cs#L49)

Tratcher: No, it's not internal. We don't care if anyone changes it, we won't flow it.

HaoK: Why is this one second class compared to the other ctor arguments which could be left as get/set properties as well

PinpointTownes: No, it's not internal, it's private and set from the constructor :trollface: 
Hence my remark.

> We don't care if anyone changes it, we won't flow it.

Not sure why you want to let the user think he's allowed to replace the instance, then.

HaoK: CookieSigningInContext for example takes several properties in the ctor and has get/set for each, I would think its either or, public set (no ctor), or private set in ctor, with strong preference towards private/ctor IMO

Tratcher: Required params go in the constructor (especially if they're passed to a base constructor). optional params use properties.

HaoK: Exactly, the point of this change was to make AuthProperties required no?

PinpointTownes: Curious: why do you consider them as optional?

Tratcher: It's extra data, one less thing to set up for tests.

HaoK: I'm not sure that's a valid argument right now, since we don't test any of these contexts anyways :)  I'd say making them not settable at all, is actually less surface area in the long term anyways...

muratg:
@HaoK @Tratcher is this ready?

HaoK:
Just a minor remaining stylistic/consistency thing I need to hash out with @Tratcher but this is basically done yes...

muratg:
Cool. I think @Tratcher will be out till January. Let's get it in soon after that. :)

HaoK:
Discussed with @Tratcher he is ok if we go with required ctor as long as there is only one ctor per context

muratg:
:+1:

HaoK:
Updated, promoted AuthenticationProperties to a required ctor argument in contexts where its used.  I left the Cookie properties settable as those are intended to be used to replace them in some place.  But I tried to make them get only everywhere else

src/Microsoft.AspNet.Authentication.Google/GoogleHandler.cs
```diff
@@ -31,12 +31,8 @@ public GoogleHandler(HttpClient httpClient)
...
 
             var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
-            
-            var context = new OAuthCreatingTicketContext(Context, Options, Backchannel, tokens, payload)
-            {
-                Properties = properties,
-                Principal = new ClaimsPrincipal(identity)
-            };
+
+            var context = new OAuthCreatingTicketContext(new ClaimsPrincipal(identity), properties, Context, Options, Backchannel, tokens, payload);
```
Tratcher: This constructor is getting a bit out of hand.

HaoK: Yeah its seems big, but I would assume all of these are required for OAuth ticket contexts?  Making them properties doesn't really change anything except moving them into a { } block and making it possible to forget to set one...

Tratcher:
:shipit:

src/Microsoft.AspNet.Authentication.Google/GoogleHandler.cs
PinpointTownes: We could replace the `ClaimsPrincipal` and `AuthenticationProperties` arguments by a single `AuthenticationTicket`.

PinpointTownes: Also, since the last parameter is optional, you could replace it a by property.

HaoK: Doesn't really seem worth the additional churn at this point as it doesn't seem to buy us that much does it?

HaoK: The user overload ctor?  Sure I can fix that

Tratcher: Actually, passing in a ticket would make sense, that's the ultimate output at the end of this method.

HaoK: I'll do that in a followup PR then, as its basically moving the creation of the ticket from the method CreateTicketAsync into whatever is provided in the context

HaoK:
3257a82367e946a351cc1eaa02a9bb5dae2a432b

Issue #574, opened 2015-11-12
Consider adding a sample using the JWT authentication


janpieterz:
I've been trying to implement this and am currently basing the implementation on some loose issues that show code of people implementing this. It would be very helpful if either the docs or this repository had a valid way of handling JWT tokens.

I realize it's probably an extensive topic, but a minimal implementation could be linking to a couple of good explanatory articles and then pointing towards one sample project that has this working.

brentschmaltz:
@janpieterz  https://github.com/AzureADSamples have many samples showing how to validate a JWT.

Issue #555, opened 2015-10-28
Stop using AuthenticateResult.Success(ticket: null)


PinpointTownes:
The whole new `AuthenticateResult.Success(ticket: null)` is used in multiple places (e.g https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L397 or https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L437)

Sadly, this doesn't seem to be a legal operation, as `Success` throws an exception when `ticket` is null: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication/AuthenticateResult.cs#L40

/cc @HaoK @Tratcher

HaoK:
Yeah we should definitely be phasing those out...

HaoK:
In the OIDC case, what does it actually mean to Skip a message, is that really a failure now in the new sense that no ticket is produced?

HaoK:
Note: JwtBearer also has this usage still.

PinpointTownes:
Ahah, I just had the same concern when updating ASOS. I simply decided to return null :smile: 

> In the OIDC case, what does it actually mean to Skip a message, is that really a failure now in the new sense that no ticket is produced?

I'm really tempted to say no. Actually, it wouldn't be an issue if the authenticate result wasn't a binary thing.

The cookies middleware suffers from the same symptom: I don't think returning a "failure" is semantically appropriate when no cookie was received: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs#L110

HaoK:
Well we don't actually treat a "Failure" any differently than a success.  It just means there was no Ticket.  Its up to the caller to use the exception if desired.  In the cookies case, it doesn't care about errors, so it just calls context.Failed instead of the old NotAuthenticated, which is pretty much the same thing in that you get a result with no ticket.  Perhaps Success/Failure isn't the right names.  Its basically "Ticket/NoTicket" which is the definition for success/failure now...

HaoK:
Previously we used to just return null or non null, which maps to Success/Failure, except we used to have a 'success' with no ticket which was a bit ambiguous.

sunsided:
[...]

Issue #411, opened 2015-08-25
Revisit the OIDC/OAuth2 bearer middleware to stop re-throwing exceptions for invalid tokens


PinpointTownes:
By default, both the OIDC and the OAuth2 bearer middleware throw a bunch of security exceptions when a token is invalid. These exceptions are caught but the catch handler always re-throw them if you don't override the `AuthenticationFailed` notification and explicitly call `HandleResponse()` or `SkipToNextMiddleware()`, which is - obviously - a very bad thing, as you don't want to get a 500 response - or worse, an error page - for an invalid token.

In this case, 401 is the only response you want: https://tools.ietf.org/html/rfc6750#section-3.1

Tratcher:
RE: https://github.com/aspnet/Security/issues/55

PinpointTownes:
Alzheimer... :scream: 

Let's close this one then :smile:

PinpointTownes:
Reopening this issue, since the bearer middleware still needs to be reworked to avoid throwing exceptions for invalid tokens, despite https://github.com/aspnet/Security/issues/55 being merged.

/cc @HaoK

PinpointTownes:
Any chance we could prioritize this item?

Tratcher:
I'd rather get a TryValidateToken from IdentityModel. @brentschmaltz?

muratg:
@brentschmaltz can we stop throwing and return false?

HaoK:
Please file an issue

blowdart:
I'm actually OK with SecureDataFormat swallowing exceptions.

Tratcher:
Yes, but I suppose it could also use a Try pattern as failures are expected.

louislewis2:
Guys, having an issue like this one present for RC seems terrible. We are in preview with customers and due to this error, we are receiving a lot of flack. Can we please prioritise this fix?

PR #584, opened 2015-11-24
Add a JwtBearer sample.

Tratcher:
#574 @HaoK @brentschmaltz @muratg 
Liberally borrowed the SPA code from https://github.com/Azure-Samples/active-directory-angularjs-singlepageapp

This will give us somewhere to work on #411 and #555 via multi-tenant JWT.

### DIFF
--- a/samples/JwtBearerSample/Startup.cs
+++ b/samples/JwtBearerSample/Startup.cs
@@ -1,0 +1,104 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.Hosting;
+using Microsoft.AspNet.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Newtonsoft.Json.Linq;
+
+namespace JwtBearerSample
+{
+    public class Startup
+    {
+        // Shared between users in memory
+        public IList<Todo> Todos { get; } = new List<Todo>();
+
+        // This method gets called by the runtime. Use this method to add services to the container.
+        // For more information on how to configure your application, visit http://go.microsoft.com/fwlink/?LinkID=398940
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddAuthentication();
+        }
+
+        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
+        public void Configure(IApplicationBuilder app)
+        {
+            // Simple error page to avoid a repo dependency.
+            app.Use(async (context, next) =>
+            {
+                try
+                {
+                    await next();
+                }
+                catch (Exception ex)
+                {
+                    if (context.Response.HasStarted)
+                    {
+                        throw;
+                    }
+                    context.Response.StatusCode = 500;
+                    await context.Response.WriteAsync(ex.ToString());
+                }
+            });
+
+            app.UseIISPlatformHandler();
+
+            app.UseDefaultFiles();
+            app.UseStaticFiles();
+
+            app.UseJwtBearerAuthentication(options =>
+            {
+                options.AutomaticAuthenticate = true;
+                options.AutomaticChallenge = true;
+                // You also need to update /wwwroot/app/scripts/app.js
+                options.Authority = "https://login.windows.net/tratcheroutlook.onmicrosoft.com";
+                options.Audience = "63a87a83-64b9-4ac1-b2c5-092126f8474f";
+            });
+
+            // [Authorize] would usually handle this
+            app.Use(async (context, next) =>
+            {
+                // Use this if options.AutomaticAuthenticate = false
+                // var user = await context.Authentication.AuthenticateAsync(JwtBearerDefaults.AuthenticationScheme);
+
+                var user = context.User; // We can do this because of  options.AutomaticAuthenticate = true; above.
+                if (user?.Identity?.IsAuthenticated ?? false)
+                {
+                    await next();
+                }
+                else
+                {
+                    // We can do this because of options.AutomaticChallenge = true; above
+                    await context.Authentication.ChallengeAsync();
+                }
+            });
+
+            // MVC would usually handle this:
+            app.Map("/api/TodoList", todoApp =>
+            {
+                todoApp.Run(async context =>
+                {
+                    var response = context.Response;
+                    if (context.Request.Method.Equals("POST", System.StringComparison.OrdinalIgnoreCase))
+                    {
+                        var reader = new StreamReader(context.Request.Body);
+                        var body = await reader.ReadToEndAsync();
+                        var obj = JObject.Parse(body);
+                        var todo = new Todo() { Description = obj["Description"].Value<string>(), Owner = context.User.Identity.Name };
+                        Todos.Add(todo);
+                    }
+                    else
+                    {
+                        response.ContentType = "application/json";
+                        var json = JToken.FromObject(Todos);
+                        await response.WriteAsync(json.ToString());
+                    }
+                });
+            });
+        }
+
+        // Entry point for the application.
+        public static void Main(string[] args) => WebApplication.Run<Startup>(args);
+    }
+}
--- a/samples/JwtBearerSample/wwwroot/App/Scripts/app.js
+++ b/samples/JwtBearerSample/wwwroot/App/Scripts/app.js
@@ -1,0 +1,28 @@
+﻿'use strict';
+angular.module('todoApp', ['ngRoute','AdalAngular'])
+.config(['$routeProvider', '$httpProvider', 'adalAuthenticationServiceProvider', function ($routeProvider, $httpProvider, adalProvider) {
+
+    $routeProvider.when("/Home", {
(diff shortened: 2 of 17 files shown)


Tratcher:
@halter73 Read some JS please.

samples/JwtBearerSample/wwwroot/App/Scripts/adal.js
```diff
@@ -0,0 +1,1060 @@
...
+    // y could be 1000, 1001, 1010, 1011 since most significant two bits needs to be 10
+    // y values are 8, 9, A, B
+    var guidHolder = 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx';
+    var hex = '0123456789abcdef';
+    var r = 0;
+    var guidResponse = "";
+    for (var i = 0; i < 36; i++) {
+        if (guidHolder[i] !== '-' && guidHolder[i] !== '4') {
+            // each x and y needs to be random
+            r = Math.random() * 16 | 0;
```
halter73: Math.random (at least in chrome) might not be the best thing to use for generating globally unique identifiers. Here's a recent blog post about it:

https://medium.com/@betable/tifu-by-using-math-random-f1c308c4fd9d#.ey40psqsl

In the blog post, the generated a "unique" id with supposedly 132 bits of entropy, but saw real world collisions because of Chromes PRNG.

Can we generate the GUID on the server instead of using Math.random?

samples/JwtBearerSample/wwwroot/App/Scripts/adal-angular.js
```diff
@@ -0,0 +1,320 @@
...
+        localStorage = localStorageInj;
+        document = documentInj;
+        Math = MathInj; // jshint ignore:line
+        angular = angularInj;
+        AuthenticationContext = adalInj;
+    };
+}
+
+(function () {
+    // ============= Angular modules- Start =============
```
halter73: Might be better to have someone like @DamianEdwards look at the angular stuff.

samples/JwtBearerSample/wwwroot/App/Scripts/app.js
```diff
@@ -0,0 +1,28 @@
...
+        templateUrl: "/App/Views/UserData.html",
+    }).otherwise({ redirectTo: "/Home" });
+
+    adalProvider.init(
+        {
+            instance: 'https://login.microsoftonline.com/', 
+            tenant: 'tratcheroutlook.onmicrosoft.com',
+            clientId: '63a87a83-64b9-4ac1-b2c5-092126f8474f',
+            extraQueryParameter: 'nux=1',
+            // cacheLocation: 'localStorage', // enable this for IE, as sessionStorage does not work for localhost.
```
halter73: Why not just default to localStorage if it's available?

samples/JwtBearerSample/Startup.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            app.UseDefaultFiles();
+            app.UseStaticFiles();
+
+            app.UseJwtBearerAuthentication(options =>
+            {
+                options.AutomaticAuthenticate = true;
+                options.AutomaticChallenge = true;
+                // You also need to update /wwwroot/app/scripts/app.js
+                options.Authority = "https://login.windows.net/tratcheroutlook.onmicrosoft.com";
```
halter73: > tratcheroutlook

Is this the type of thing we want in a sample?

samples/JwtBearerSample/Startup.cs
```diff
@@ -0,0 +1,102 @@
...
+                options.Audience = "63a87a83-64b9-4ac1-b2c5-092126f8474f";
+            });
+
+            // [Authorize] would usually handle this
+            app.Use(async (context, next) =>
+            {
+                // var user = await context.Authentication.AuthenticateAsync(JwtBearerDefaults.AuthenticationScheme);
```
halter73: If we are including commented code in a sample, I think we should at least explain why it's commented.

halter73:
:watch:

Tratcher:
@DamianEdwards would you like to sanity check some Angular?

samples/JwtBearerSample/wwwroot/App/Scripts/adal.js
Tratcher: This is from the ADAL SDK. I'll avoid modifying their files.

samples/JwtBearerSample/Startup.cs
Tratcher: All the security samples are like this. We could extract it to conifig, but we'd still want some sample values.

halter73: I see that it's explained in the end-of-line comment below. I think I would clarify by moving that comment up and explaining this is what you do if `options.AutomaticAuthenticate` isn't true.

Tratcher: https://github.com/aspnet/Security/issues/593

samples/JwtBearerSample/wwwroot/App/Scripts/adal.js
halter73: OK. So we're not modifying adal-angular.js either?

Tratcher: Not if we can avoid it.

Tratcher:
Minor update.

samples/JwtBearerSample/wwwroot/App/Scripts/adal.js
halter73: Maybe it's better just to include these using [the Bower package or from the CDN](https://github.com/AzureAD/azure-activedirectory-library-for-js).

PinpointTownes:
Curious: will users be able to use this sample if they don't have an Azure account?

Tratcher:
@PinpointTownes Not as is, the client uses the azure SDK to get the bearer token. To move to a different IDP you need a different client SDK.

samples/JwtBearerSample/Startup.cs
```diff
@@ -0,0 +1,104 @@
...
+            app.UseDefaultFiles();
+            app.UseStaticFiles();
+
+            app.UseJwtBearerAuthentication(options =>
+            {
+                options.AutomaticAuthenticate = true;
+                options.AutomaticChallenge = true;
+                // You also need to update /wwwroot/app/scripts/app.js
+                options.Authority = "https://login.windows.net/tratcheroutlook.onmicrosoft.com";
+                options.Audience = "63a87a83-64b9-4ac1-b2c5-092126f8474f";
```
PinpointTownes: Is that your `client_id` here? Technically, it's not wrong, but using the `client_id` for the `audience`/`resource` parameter is semantically odd. Instead, consider using a value that describes the resource server (i.e this server app) like its URL.

samples/JwtBearerSample/wwwroot/App/Scripts/app.js
```diff
@@ -0,0 +1,28 @@
...
+    }).when("/TodoList", {
+        controller: "todoListCtrl",
+        templateUrl: "/App/Views/TodoList.html",
+        requireADLogin: true,
+    }).when("/UserData", {
+        controller: "userDataCtrl",
+        templateUrl: "/App/Views/UserData.html",
+    }).otherwise({ redirectTo: "/Home" });
+
+    adalProvider.init(
```
PinpointTownes: Consider setting `loginResource` with a value matching the `audience` registered in the JWT bearer options :clap:

samples/JwtBearerSample/Startup.cs
Tratcher: The client ID seems to be the only value that works with my AAD setup.

Tratcher:
:up: :date:

samples/JwtBearerSample/wwwroot/App/Scripts/app.js
Tratcher: loginResource defautls to clientId: https://github.com/AzureAD/azure-activedirectory-library-for-js/blob/master/lib/adal.js#L154

samples/JwtBearerSample/Startup.cs
PinpointTownes: Have you registered a new web API? (https://github.com/Azure-Samples/active-directory-dotnet-daemon#register-the-todolistservice-web-api)

samples/JwtBearerSample/wwwroot/App/Scripts/app.js
PinpointTownes: Sure, but using a `client_id` corresponding to a JS façade to describe a resource server is extremely weird. I don't even know why it's tolerated.

samples/JwtBearerSample/Startup.cs
Tratcher: Hmm, no luck.

HaoK:
Sample looks fine to me, we should also lift some of the flows directly into unit tests assuming they aren't already there (which would let us drop all of the JS that's in the sample) :shipit:

PR #585, opened 2015-11-25
Handle validation failures from ValidateToken.

Tratcher:
#411 @HaoK @PinpointTownes @tushargupta51 @brentschmaltz 
RE: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/312

ISecurityTokenValidator.ValidateToken throws exceptions for any failure such as expired tokens, invalid tokens, etc.. Since these kind of failures will be common we should handle them gracefully.

For AutomaticAuthenticate=true this will raise the AuthenticationFailed event but not throw. It will appear as if there were no credentials provided and the authorization will usually fail with a 401.

The advanced HttpContext.Authentication.AuthenticateAsync API can return the detailed errors.

Would this also make sense for other auth middleware? I'm not sure it would be as useful because they operate primarily with immediate redirects, not out-of-band communication like Bearer.

### DIFF
--- a/src/Microsoft.AspNet.Authentication.JwtBearer/JwtBearerHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.JwtBearer/JwtBearerHandler.cs
@@ -2,8 +2,10 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.IdentityModel.Tokens;
 using System.Linq;
+using System.Security.Claims;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Http.Authentication;
 using Microsoft.AspNet.Http.Features.Authentication;
@@ -34,7 +36,6 @@ namespace Microsoft.AspNet.Authentication.JwtBearer
                 {
                     return AuthenticateResult.Success(receivingTokenContext.AuthenticationTicket);
                 }
-
                 if (receivingTokenContext.Skipped)
                 {
                     return AuthenticateResult.Success(ticket: null);
@@ -76,7 +77,6 @@ namespace Microsoft.AspNet.Authentication.JwtBearer
                 {
                     return AuthenticateResult.Success(receivedTokenContext.AuthenticationTicket);
                 }
-
                 if (receivedTokenContext.Skipped)
                 {
                     return AuthenticateResult.Success(ticket: null);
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -103,12 +103,37 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
...
+
+                            if (validationFailures == null)
+                            {
+                                validationFailures = new List<Exception>(1);
+                            }
+                            validationFailures.Add(ex);
+                            continue;
+                        }
+
+                        Logger.LogInformation("Successfully validated the token");
```
PinpointTownes: Curious: isn't this log message a bit redundant with the other messages that are logged when the user is signed in?

PinpointTownes:
Better late than never, thanks! :shipit:

src/Microsoft.AspNet.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -103,12 +103,37 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
...
+                        try
+                        {
+                            principal = validator.ValidateToken(token, validationParameters, out validatedToken);
+                        }
+                        catch (Exception ex)
+                        {
+                            Logger.LogInformation("Failed to validate the token: " + token, ex);
+
+                            // Refresh the configuration for exceptions that may be caused by key rollovers. The user can also request a refresh in the event.
+                            if (Options.RefreshOnIssuerKeyNotFound && ex.GetType().Equals(typeof(SecurityTokenSignatureKeyNotFoundException)))
```
brentschmaltz: For multi IDP's, it may be the case that the token should not have been processed. Refreshing here doesn't seem optimal.

Tratcher: Yes, we need to do a comprehensive design for multi-tenancy.

src/Microsoft.AspNet.Authentication.JwtBearer/JwtBearerHandler.cs
Tratcher: Where are there other sign-in logs for the bearer token flow?

PinpointTownes: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication/AuthenticationHandler.cs#L213

Issue #565, opened 2015-11-04
Update facebook provider to v2.5 APIs


Tratcher:
We're currently using v2.2 which will be supported until March 25, 2017, where v2.5 will be supported until at least until April 2018.

https://developers.facebook.com/docs/apps/changelog

RE: https://github.com/aspnet/Security/issues/92

In the process we should verify we're exposing the needed APIs for these scenarios: https://github.com/aspnet/Security/issues/365,

Tratcher:
"[Oauth Access Token] Format - The response format of https://www.facebook.com/v2.3/oauth/access_token returned when you exchange a code for an access_token now return valid JSON instead of being URL encoded. The new format of this response is {“access_token”: <TOKEN>, “token_type”:<TYPE>, “expires_in”:<TIME>}. We made this update to be compliant with section 5.1 of RFC 6749."

digiseller:
facebook oauth2 email problem fix

http://stackoverflow.com/a/32636149/4594225

alisherdavronov:
facebook oauth2 email problem fix

http://stackoverflow.com/a/32636149/4594225

Issue #435, opened 2015-09-04
Can't get email claim from Facebook


MikeWasson:
Using beta7 packages. In my startup:

            services.Configure<FacebookAuthenticationOptions>(options =>
            {
                options.AppId = Configuration["Authentication:Facebook:AppId"];
                options.AppSecret = Configuration["Authentication:Facebook:AppSecret"];
                options.Scope.Add("email");
            });

However, inside AccountController.ExternalLoginCallback, I get back these claim types in the ExternalPrincipal:

- "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier"
- "urn:facebook:name"
- "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name"

In the Fiddler traffic, I can see that it’s calling the FB Graph API and getting:

    {"name":"xxx","id":"xxxx"}

If I modify the request to ask for email and replay it, then I do get the email, so I think my scope is correct

    GET https://graph.facebook.com/v2.2/me?fields=id,name,email&access_token=blah

Response:

    {"id":"xxx","name":"xxxx","email":"xxx"}

Tratcher:
Related?
https://github.com/aspnet/Security/issues/365
https://github.com/aspnet/Security/pull/366

MikeWasson:
I don't think so? It's sending a well-formed request to FB:

    https://graph.facebook.com/v2.2/me?access_token=blah&appsecret_proof=blah

But I don't see the middleware asking for email anywhere, and FB is just returning id and name. I'm not familiar enough with FB graph API to know if FB should be returning email by default, or if the middleware needs to ask for it specifically.

Eilon:
Need to try to repro this. It might be due to the user marking their email address as hidden data, so even if you request it, it won't be available.

AlexTo:
Yes I have the same issue, can't get email from Facebook if I just add

services.Configure<FacebookAuthenticationOptions>(options =>
        {
            options.AppId = Configuration["Authentication:Facebook:AppId"];
            options.AppSecret = Configuration["Authentication:Facebook:AppSecret"];
            options.Scope.Add("email");
        });

I could see that Facebook asks me for permission to provide email to the webapp but I just don't have the claim in info.ExternalPrincipal

hajjat:
Hi, I'm hitting the same issue. Were you guys able to find any workarounds? Thanks!

Jayman1305:
[...]

PR #598, opened 2015-12-02
Update facebook APIs to v2.5.

Tratcher:
#565, #435
1. They now use a standard JSON instead of a custom Form response from the TokenEndpoint.
2. Added support for requesting non-default Fields like e-mail.
@HaoK
/cc: @MikeWasson @AlexTo @hajjat @Jayman1305 @chris02031 @dotnetshadow


muratg:
Looks good

:shipit:

PR #610, opened 2015-12-09
Make SkipToNextMiddleware work on events.

Tratcher:
#555 @HaoK @PinpointTownes

### DIFF
--- a/samples/JwtBearerSample/Startup.cs
+++ b/samples/JwtBearerSample/Startup.cs
@@ -5,6 +5,7 @@ using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Hosting;
 using Microsoft.AspNet.Http;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Net.Http.Headers;
 using Newtonsoft.Json.Linq;
 
 namespace JwtBearerSample
@@ -91,6 +92,7 @@ namespace JwtBearerSample
                     else
                     {
                         response.ContentType = "application/json";
+                        response.Headers[HeaderNames.CacheControl] = "no-cache";
                         var json = JToken.FromObject(Todos);
                         await response.WriteAsync(json.ToString());
                     }
--- a/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -10,7 +10,6 @@ using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Authentication;
 using Microsoft.AspNet.Http.Features;
 using Microsoft.AspNet.Http.Features.Authentication;
-using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
 using Microsoft.Net.Http.Headers;
 
@@ -26,31 +25,30 @@ namespace Microsoft.AspNet.Authentication.Cookies
         private DateTimeOffset? _renewIssuedUtc;
         private DateTimeOffset? _renewExpiresUtc;
         private string _sessionKey;
-        private Task<AuthenticationTicket> _cookieTicketTask;
+        private Task<AuthenticateResult> _readCookieTask;
 
-        private Task<AuthenticationTicket> EnsureCookieTicket()
+        private Task<AuthenticateResult> EnsureCookieTicket()
         {
             // We only need to read the ticket once
-            if (_cookieTicketTask == null)
+            if (_readCookieTask == null)
             {
-                _cookieTicketTask = ReadCookieTicket();
+                _readCookieTask = ReadCookieTicket();
             }
-            return _cookieTicketTask;
+            return _readCookieTask;
         }
 
-        private async Task<AuthenticationTicket> ReadCookieTicket()
+        private async Task<AuthenticateResult> ReadCookieTicket()
         {
             var cookie = Options.CookieManager.GetRequestCookie(Context, Options.CookieName);
             if (string.IsNullOrEmpty(cookie))
             {
-                return null;
+                return AuthenticateResult.Skip();
             }
 
             var ticket = Options.TicketDataFormat.Unprotect(cookie, GetTlsTokenBinding());
             if (ticket == null)
             {
-                Logger.LogWarning(@"Unprotect ticket failed");
-                return null;
+                return AuthenticateResult.Fail("Unprotect ticket failed");
             }
 
             if (Options.SessionStore != null)
@@ -58,15 +56,13 @@ namespace Microsoft.AspNet.Authentication.Cookies
                 var claim = ticket.Principal.Claims.FirstOrDefault(c => c.Type.Equals(SessionIdClaim));
                 if (claim == null)
                 {
-                    Logger.LogWarning(@"SessionId missing");
-                    return null;
+                    return AuthenticateResult.Fail("SessionId missing");
                 }
                 _sessionKey = claim.Value;
                 ticket = await Options.SessionStore.RetrieveAsync(_sessionKey);
                 if (ticket == null)
                 {
-                    Logger.LogWarning(@"Identity missing in session store");
-                    return null;
+                    return AuthenticateResult.Fail("Identity missing in session store");
                 }
             }
 
@@ -80,7 +76,7 @@ namespace Microsoft.AspNet.Authentication.Cookies
                 {
                     await Options.SessionStore.RemoveAsync(_sessionKey);
                 }
-                return null;
+                return AuthenticateResult.Fail("Ticket expired");
             }
 
             var allowRefresh = ticket.Properties.AllowRefresh ?? true;
@@ -99,23 +95,23 @@ namespace Microsoft.AspNet.Authentication.Cookies
             }
 
             // Finally we have a valid ticket
-            return ticket;
+            return AuthenticateResult.Success(ticket);
         }
 
         protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
         {
-            var ticket = await EnsureCookieTicket();
-            if (ticket == null)
+            var result = await EnsureCookieTicket();
+            if (!result.Succeeded)
             {
-                return AuthenticateResult.Failed("No ticket.");
+                return result;
             }
 
-            var context = new CookieValidatePrincipalContext(Context, ticket, Options);
+            var context = new CookieValidatePrincipalContext(Context, result.Ticket, Options);
             await Options.Events.ValidatePrincipal(context);
 
             if (context.Principal == null)
             {
-                return AuthenticateResult.Failed("No principal.");
+                return AuthenticateResult.Fail("No principal.");
             }
 
             if (context.ShouldRenew)
@@ -196,7 +192,8 @@ namespace Microsoft.AspNet.Authentication.Cookies
 
         protected override async Task HandleSignInAsync(SignInContext signin)
         {
-            var ticket = await EnsureCookieTicket();
+            // Process the request cookie to initialize members like _sessionKey.
+            var result = await EnsureCookieTicket();
             var cookieOptions = BuildCookieOptions();
 
             var signInContext = new CookieSigningInContext(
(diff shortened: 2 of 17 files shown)


samples/JwtBearerSample/Startup.cs
```diff
@@ -91,6 +91,7 @@ public void Configure(IApplicationBuilder app)
                     else
                     {
                         response.ContentType = "application/json";
+                        response.Headers["Cache-Control"] = "no-cache";
```
PinpointTownes: nit: no `HeaderNames` love? :heart:

src/Microsoft.AspNet.Authentication.JwtBearer/Events/JwtBearerEvents.cs
```diff
@@ -36,13 +36,9 @@ public class JwtBearerEvents : IJwtBearerEvents
...
         /// <summary>
-        /// Invoked to apply a challenge sent back to the caller.
+        /// Invoked before a challenge is sent back to the caller.
         /// </summary>
-        public Func<JwtBearerChallengeContext, Task> OnChallenge { get; set; } = context =>
-        {
-            context.HttpContext.Response.Headers.Append("WWW-Authenticate", context.Options.Challenge);
-            return Task.FromResult(0);
-        };
+        public Func<JwtBearerChallengeContext, Task> OnChallenge { get; set; } = context => Task.FromResult(0);
```
PinpointTownes: Yeaaaaah! :clap:

src/Microsoft.AspNet.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -201,8 +202,20 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
...
+                return true;
+            }
+            if (eventContext.Skipped)
+            {
+                return false;
+            }
+
             Response.StatusCode = 401;
-            await Options.Events.Challenge(new JwtBearerChallengeContext(Context, Options));
+            Response.Headers.Append("WWW-Authenticate", Options.Challenge);
```
PinpointTownes: `HeaderNames.WWWAuthenticate`?

src/Microsoft.AspNet.Authentication/AuthenticateResult.cs
```diff
@@ -44,6 +46,11 @@ public static AuthenticateResult Success(AuthenticationTicket ticket)
             return new AuthenticateResult() { Ticket = ticket };
         }
 
+        public static AuthenticateResult Skip()
+        {
+            return new AuthenticateResult() { Skipped = true };
+        }
+
         public static AuthenticateResult Failed(Exception error)
```
PinpointTownes: Should `Failed` be renamed to `Fail` to match `Skip`'s form? (FYI: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/issues/173#issuecomment-155238696)

PinpointTownes: Also, having a parameterless `Fail`/`Failed` overload would be really useful: https://github.com/aspnet/Security/pull/568#issuecomment-162691988

PinpointTownes:
Amazing :shipit:

PinpointTownes:
I suppose this PR replaces https://github.com/aspnet/Security/pull/568?

PinpointTownes:
On a related note, it might be a good idea to determine when we're supposed to return `null` (e.g https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs#L46) and when using `AuthenticateResult.Skip` is preferable.

Tratcher:
Hmm, EnsureCookieTicket returns AuthTicket, not AuthResult. We could change it to AuthResult and use Skip and Fail more. Then when you get to https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs#L110 it could already have a better error than "No Ticket".

Tratcher:
Updated.

src/Microsoft.AspNet.Authentication/AuthenticateResult.cs
```diff
@@ -44,12 +44,17 @@ public static AuthenticateResult Success(AuthenticationTicket ticket)
             return new AuthenticateResult() { Ticket = ticket };
         }
 
-        public static AuthenticateResult Failed(Exception error)
+        public static AuthenticateResult Skip()
+        {
+            return new AuthenticateResult() { Skipped = true };
+        }
+
+        public static AuthenticateResult Fail(Exception error)
```
HaoK: Maybe we should pair the naming of Fail/Error better?  Error => FailureException maybe?

Tratcher: Fail & Failure?

HaoK: Or maybe rename Fail to Error, with the property ErrorException is maybe better?

HaoK: Sure lets go with Fail/Failure

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -26,31 +25,30 @@ internal class CookieAuthenticationHandler : AuthenticationHandler<CookieAuthent
...
         }
 
-        private async Task<AuthenticationTicket> ReadCookieTicket()
+        private async Task<AuthenticateResult> ReadCookieTicket()
         {
             var cookie = Options.CookieManager.GetRequestCookie(Context, Options.CookieName);
             if (string.IsNullOrEmpty(cookie))
             {
-                return null;
+                return AuthenticateResult.Fail($"'{Options.CookieName}' cookie not found.");
```
PinpointTownes: Is this really a failure? Logging a message when the user is not authenticated might be a bit excessive.

https://github.com/aspnet/Security/issues/555#issuecomment-152033087

HaoK: We don't actually log cookie failures at all, that behavior only is part of Remote based Auth Handlers.  The 'failures' in cookie are on the result that no one looks at, so they are silently ignored right now.  You would be able to see them if you called Authenticate using the context overload that returned the Result.

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -99,23 +95,23 @@ private Task<AuthenticationTicket> EnsureCookieTicket()
...
 
         protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
         {
-            var ticket = await EnsureCookieTicket();
-            if (ticket == null)
+            var result = await EnsureCookieTicket();
+            if (result.Error != null)
             {
-                return AuthenticateResult.Failed("No ticket.");
+                return result;
```
HaoK: Perhaps we should LogVerbose the error?

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
PinpointTownes: Doesn't this PR now log the authentication errors?

https://github.com/aspnet/Security/pull/610/files#diff-4e1d744cdbfdc6fa13724d45717a666bR105

HaoK: Ah I missed that, yeah that's going to be way too much noise, we can't log info every time a cookie isn't found.  Its not really an error, unlike the Remote ones which are special failures.

HaoK: Maybe its not worth switching cookie to use AuthResult, and we should just explicitly choose what to log since we actively do not want to LogInfo in many situations.

PinpointTownes: Yep. BTW, the same remark applies to the JWT bearer middleware, that returns a failed result when there's no Authorization header (or when the token cannot be extracted, but it's less annoying): https://github.com/aspnet/Security/pull/610/files#diff-9ef5d7d4acb4674e3ac4c81c994a6c12R56.

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -58,15 +56,13 @@ private Task<AuthenticationTicket> EnsureCookieTicket()
...
-                    return null;
+                    return AuthenticateResult.Fail("SessionId missing");
                 }
                 _sessionKey = claim.Value;
                 ticket = await Options.SessionStore.RetrieveAsync(_sessionKey);
                 if (ticket == null)
                 {
-                    Logger.LogWarning(@"Identity missing in session store");
-                    return null;
+                    return AuthenticateResult.Fail("Identity missing in session store");
```
HaoK: We went from Warning to Info as part of this change too right?  I don't remember what the guidance is for which we should use

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
HaoK: Yeah I think those are the valid 'success' null ticket use cases

HaoK: And not Failure(no error)

Tratcher: How about using AuthResult.Skip here? All the others were previously logged.

PinpointTownes: :+1: for `Skip` or a new `Ignore` result;

HaoK: What does Skip mean in this context?  Skip to next middleware?

HaoK: Perhaps for clarity, Skip should be renamed Ignore which seems clearer to me...

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
Tratcher: I think these were too high. Cookies and cache entries expire all the time.

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
Tratcher: I'd rather keep Skip in line with the SkipToNextMiddleware event API. In this context your cookie middleware was skipped because there wasn't an auth cookie.

HaoK:
Looks like the only sticking point left is the logging, since I think before, even though we returned Failure("no Ticket") that wasn't getting logged by default...  I don't think we want to log "No Ticket" for every unauthenticated request

Tratcher:
Updated to use Skip for non-starter scenarios.  Note the logging changes are only for AutomaticAuthenticate middleware, and I don't think we'll end up logging much that wasn't being logged before.

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
PinpointTownes: > Perhaps for clarity, Skip should be renamed Ignore which seems clearer to me...

:+1: for `Ignore`.

HaoK:
RE Skip/SkipToMiddleware, could consider Ignore/IgnoreMiddleware to remain consistent but be more clear...

Tratcher:
I don't think it's a meaningful enough distinction to warrant more churn.

HaoK:
Well we are promoting Skip to something more visible now, as one of the 3 things your auth middleware can do...

Success/Fail/Skip??

HaoK:
But yeah its fine for this iteration :shipit:

PinpointTownes:
And my parameterless `Fail`? :cry:

HaoK:
Are you not able to make up an error message? :)

PinpointTownes:
> Are you not able to make up an error message? :)

I'm not that stupid :trollface: 

It's just that there are cases where I prefer logging a complete error trace (with values attached), and not just a stupid message hacked with `string.Format`.

Tratcher:
I think you can just use Skip for that.

PinpointTownes:
Technically yes. Semantically, it would be a bit weird. But heh... :smile:

HaoK:
Yeah basically how I think of these 3 states is...

Success : good have ticket
Skip : nothing
Fail : Exception that gets logged

Issue #593, opened 2015-11-30
[Samples] Extract IDP settings to config / user-secrets


Tratcher:
Currently all the samples hardcode app specific IDs like clientId, client secret, etc..
```
            app.UseOpenIdConnectAuthentication(options =>
            {
                options.ClientId = "63a87a83-64b9-4ac1-b2c5-092126f8474f";
                options.ClientSecret = "Yse2iP7tO1Azq0iDajNisMaTSnIDv+FXmAsFuXr+Cy8="; // for code flow
                options.Authority = "https://login.windows.net/tratcheroutlook.onmicrosoft.com";
                options.ResponseType = OpenIdConnectResponseTypes.Code;
                options.GetClaimsFromUserInfoEndpoint = true;
            });
```
Consider moving these to config / user-secrets. Compare to:
https://github.com/Azure-Samples/active-directory-dotnet-webapp-openidconnect/blob/master/WebApp-OpenIDConnect-DotNet/App_Start/Startup.Auth.cs#L42

PR #612, opened 2015-12-10
Convert samples to use UserSecrets.

Tratcher:
#593 @HaoK @PinpointTownes @muratg

### DIFF
--- a/samples/JwtBearerSample/Startup.cs
+++ b/samples/JwtBearerSample/Startup.cs
@@ -4,6 +4,7 @@ using System.IO;
 using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Hosting;
 using Microsoft.AspNet.Http;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Net.Http.Headers;
 using Newtonsoft.Json.Linq;
@@ -12,6 +13,16 @@ namespace JwtBearerSample
 {
     public class Startup
     {
+        public Startup()
+        {
+            Configuration = new ConfigurationBuilder()
+                .AddEnvironmentVariables()
+                .AddUserSecrets()
+                .Build();
+        }
+
+        public IConfiguration Configuration { get; set; }
+
         // Shared between users in memory
         public IList<Todo> Todos { get; } = new List<Todo>();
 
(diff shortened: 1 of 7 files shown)


PinpointTownes:
Personally, I'm not fundamentally opposed to using the user secret thingy, but you need to add a README somewhere with all the `client_id/client_secret` or nobody will be able to use the samples without registering an app with the social providers.

Tratcher:
How many of them work for you today? Most of them only work with my login because the associated apps are in dev mode. You need your own configurations anyways.

PinpointTownes:
The last time I tried, Google/Twitter/GitHub all worked like a charm, but Facebook and MSFT (but this one is terribly painful...). I'll give them another try to confirm.

PinpointTownes:
Confirmed, only FB and MSFT don't work. If you want to avoid hardcoding the client credentials, why not using a JSON config file and adding a comment indicating that the user secret store is recommended on non-sample apps?

samples/JwtBearerSample/Startup.cs
```diff
@@ -53,8 +64,8 @@ public void Configure(IApplicationBuilder app)
                 options.AutomaticAuthenticate = true;
                 options.AutomaticChallenge = true;
                 // You also need to update /wwwroot/app/scripts/app.js
-                options.Authority = "https://login.windows.net/tratcheroutlook.onmicrosoft.com";
-                options.Audience = "63a87a83-64b9-4ac1-b2c5-092126f8474f";
+                options.Authority = Configuration["jwt:authority"];
+                options.Audience = Configuration["jwt:audiance"];
```
PinpointTownes: `jwt:audiance` -> `jwt:audience`

samples/JwtBearerSample/Startup.cs
```diff
@@ -12,6 +13,16 @@ namespace JwtBearerSample
...
     {
+        public Startup()
+        {
+            Configuration = new ConfigurationBuilder()
+                .AddEnvironmentVariables()
+                .AddUserSecrets()
+                .Build();
+        }
+
+        public IConfiguration Configuration { get; set; }
```
PinpointTownes:
```csharp
public IConfiguration Configuration { get; } = new ConfigurationBuilder()
    .AddEnvironmentVariables()
    .AddUserSecrets()
    .Build();
```

:trollface:

Eilon:
@blowdart this good?

blowdart:
Sure

samples/JwtBearerSample/Startup.cs
Tratcher: I'm going to leave it because it's closer to the template.

Tratcher:
Updated.

Issue #423, opened 2015-09-02
Support distributed sign-out


Tratcher:
When another site triggers a sign-out at the IDP, the client may be redirected to us with a sign-out message. We need to detect sign-out messages and allow the application to take meaningful action. The default could be just calling signout with the SignInScheme, but you also need to allow for the case where Identity has issued their own cookie.

PinpointTownes:
Definitely a must-have :+1:

Tratcher:
Design:
- The browser sends a GET to an address pre-configured in the IDP.
- The middleware listens for GETs on this exact path. There is no default value, so if it's not set then the feature is disabled.
- The AAD implementation does not currently include any validation parameters, but there will eventually be some validation required on the request. @blowdart, is this going to be a problem?
- Fire a new SignOutReceived event. There can be a default implementation that calls SignOut(options.SignInScheme), but this will only work in single-cookie configurations. If the app is configured with a separate application cookie (like in the templates) it will need to hook this notification, verify the user is even signed-in with OIDC, and perform any sign-out actions.
- What should the response look like? an empty 200?

PinpointTownes:
Heh, no.

There's an official draft for this feature, don't try to reinvent the wheel: http://openid.net/specs/openid-connect-backchannel-1_0.html :smile: 
Of course it's a draft, but it's probably better than trying to follow the AAD-specific implementation.

> The browser sends a GET to an address pre-configured in the IDP.
The middleware listens for GETs on this exact path. There is no default value, so if it's not set then the feature is disabled.

Only POST is allowed (with `application/x-www-form-urlencoded`): http://openid.net/specs/openid-connect-backchannel-1_0.html#BCRequest

> The AAD implementation does not currently include any validation parameters, but there will eventually be some validation required on the request. @blowdart, is this going to be a problem?

Unless you want to expose all the ASP.NET 5 apps to the superlogout.com syndrome... yeah :smile: 
The current draft adds a new "logout token" to mitigate this risk: http://openid.net/specs/openid-connect-backchannel-1_0.html#LogoutToken

@Eilon if you think this draft is still too young (and it is), please postpone this feature rather than trying to implement an AAD-specific feature.

PinpointTownes:
I updated my post to include the older `logout` draft: http://openid.net/specs/openid-connect-logout-1_0.html#RPLogout. 

[...]

PR #613, opened 2015-12-11
Add OIDC remote sign-out to the sample.

Tratcher:
#423 @brentschmaltz @tushargupta51 @PinpointTownes @muratg @HaoK @Eilon 

This is one way to implement http://openid.net/specs/openid-connect-logout-1_0.html as an application endpoint. Incorporating this into the middleware has a number of problems:
A) it's another endpoint to listen on.
B) the spec is still pretty new, it's unclear how much it may change.
C) to avoid xsrf issues you need to flow and verify the session id. The middleware could only do this for the simple OIDC + cookies scenario. If you used Identity you'd have to manually flow the session id and implement your own endpoint anyways.
D) an app probably wants to do other things when the user logs out like clear session state, etc..

I recommend leaving it as a sample for now.

### DIFF
--- a/samples/OpenIdConnectSample/Startup.cs
+++ b/samples/OpenIdConnectSample/Startup.cs
@@ -4,6 +4,7 @@ using Microsoft.AspNet.Authentication.OpenIdConnect;
 using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Authentication;
+using Microsoft.AspNet.Http.Features.Authentication;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.IdentityModel.Protocols.OpenIdConnect;
@@ -39,6 +40,23 @@ namespace OpenIdConnectSample
 
             app.Run(async context =>
             {
+                var authContext = new AuthenticateContext(CookieAuthenticationDefaults.AuthenticationScheme);
+                await context.Authentication.AuthenticateAsync(authContext);
+                if (context.Request.Path == new PathString("/remote-sign-out"))
+                {
+                    var session = authContext.Properties[OpenIdConnectSessionProperties.SessionState];
+                    var sid = context.Request.Query["sid"];
+                    if (string.Equals(sid, session, System.StringComparison.Ordinal))
+                    {
+                        await context.Authentication.SignOutAsync(CookieAuthenticationDefaults.AuthenticationScheme);
+                        await context.Response.WriteAsync("Signed out remotely.");
+                        return;
+                    }
+                    context.Response.StatusCode = 400;
+                    await context.Response.WriteAsync("Missing or mismatched sid.");
+                    return;
+                }
+
                 if (!context.User.Identities.Any(identity => identity.IsAuthenticated))
                 {
                     await context.Authentication.ChallengeAsync(OpenIdConnectDefaults.AuthenticationScheme, new AuthenticationProperties { RedirectUri = "/" });
@@ -49,7 +67,17 @@ namespace OpenIdConnectSample
                 }
 
                 context.Response.ContentType = "text/plain";
-                await context.Response.WriteAsync("Hello Authenticated User");
+                await context.Response.WriteAsync("Hello Authenticated User\r\n");
+                foreach (var claim in context.User.Claims)
+                {
+                    await context.Response.WriteAsync($"{claim.Type}, {claim.Value}\r\n");
+                }
+                foreach (var property in authContext.Properties)
+                {
+                    await context.Response.WriteAsync($"{property.Key}, {property.Value}\r\n");
+                }
+
+                await context.Response.WriteAsync($"Logout: /remote-sign-out?sid={authContext.Properties[OpenIdConnectSessionProperties.SessionState]}\r\n");
             });
         }
     }


samples/OpenIdConnectSample/Startup.cs
```diff
@@ -39,6 +40,23 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
 
             app.Run(async context =>
             {
+                var authContext = new AuthenticateContext(CookieAuthenticationDefaults.AuthenticationScheme);
+                await context.Authentication.AuthenticateAsync(authContext);
+                if (context.Request.Path == new PathString("/remote-sign-out"))
+                {
+                    var session = authContext.Properties[OpenIdConnectSessionProperties.SessionState];
+                    var sid = context.Request.Query["sid"];
+                    if (string.Equals(sid, session, System.StringComparison.Ordinal))
```
PinpointTownes: https://github.com/aspnet/Security/issues/423#issuecomment-163794309

IMHO, this comparison will unlikely work everywhere, since there's absolutely no obligation for an OP to use the same values for `sid` and `session_state`.

Identity Server, for instance, derives the session state from the session id (IIRC, it concatenates the session id and the client id and adds some salt) so you can't compare both values.

I guess it's probably not a big deal since it's just an AAD sample, but if you wanted it to be a general guidance, it would be something to fix.

/cc @leastprivilege

brockallen: IIRC the spec says what those are supposed to be. Maybe AAD does it differently then the spec?

PinpointTownes: The specs are rather permissive as the only requirement is to include the client identifier, the origin URL and the initial state in the session state :smile: 

> **sid (session ID)**

> The sid (session ID) Claim used in ID Tokens and as a logout_uri parameter has the following definition:

> OPTIONAL. String identifier for a Session. This represents a Session of an OP at an RP to a User Agent or device for a logged-in End-User. Its contents are unique to the OP and opaque to the RP. It MUST have sufficient entropy to prevent collisions between Session IDs generated by different OPs and to prevent it from being guessed by potential attackers. Its syntax is the same as an OAuth 2.0 Client Identifier.

http://openid.net/specs/openid-connect-logout-1_0.html

> **session_state**

> Session State. JSON [RFC7159] string that represents the End-User's login state at the OP. It MUST NOT contain the space (" ") character. This value is opaque to the RP. This is REQUIRED if session management is supported.

> The Session State value is initially calculated on the server. The same Session State value is also recalculated by the OP iframe in the browser client. The generation of suitable Session State values is specified in Section 4.2, and is based on a salted cryptographic hash of Client ID, origin URL, and OP browser state. For the origin URL, the server can use the origin URL of the Authentication Response, following the algorithm specified in Section 4 of RFC 6454 [RFC6454].

> // Here, the session_state is calculated in this particular way,
// but it is entirely up to the OP how to do it under the
// requirements defined in this specification.

http://openid.net/specs/openid-connect-session-1_0.html

brockallen: Sure, but as you point out they're not the same thing (except in AAD I guess where the spec is a more of a mild suggestion on how to do things). Point being, the check above shouldn't really work. You need to keep the sid from the original id_token and not the session state from the authorization response.

PinpointTownes: Agreed. The weird thing is that it works with AAD (but maybe it's not that weird? :smile:)

Tratcher: Note I haven't actually figured out how to test this against AAD, this is just what @brentschmaltz said should work.

PinpointTownes: Heh, maybe it's more reasonable to postpone this sample then?

Tratcher: We needed to start this discussion. If this is not resolved next week the feature will be cut.

brockallen: Well first thing to find out: does AAD issue a sid claim in the id_token?

Tratcher: @brentschmaltz? I don't see one making it through to the ClaimsIdentity.

brentschmaltz: If the 'sid' claim is not there, it is not being sent. 
The only session AAD sends is the one attached with the signin response, which is used for RP initiated signout. I suspect it should work with AAD, but in general will not for all IDP's.

The correct thing is the 'sid' claim which should be available if logout_session_supported is true which it part of discovery.

brockallen: Right, so where does the `sid` claim come from? It's not necessarily the same thing as the session_state in the authorization response. Claims, from the client's perspective, are obtained in the id_token, no? Is ADD expecting a client to use the session_state value for this?

brentschmaltz: the contract is: if discover.logout_session_supported==true, then a 'sid' claim should show up in the id_token. I agree that session_state is a different thing, however I suspect the that it may be the same. It's confusing that the two 'logouts' have a different model for the session identifier, found in two different places.
My take is that AAD is not supporting the 'sid' yet. We added a property to support this in OIDCConfiguration - HttpLogoutSupported, I think RC1. Not sure, but it is there now. You can just check the metadata.

brockallen: sure, session_state can be the same as the sid, but a client should not have to know the particularities of the implementation and do things that aren't described in the spec (such as using the session_state instead of the [absent] sid for this type of check). 

i guess what's confusing to me is why doesn't AAD just do an AAD-specific MW and leave the protocol focused MW in the ASP.NET team's hands? AAD has already created AAD-specific libraries with ADAL, so why not this too? or perhaps that would expose too many of these lax protocol compliance issues in AAD. these are the sorts of issues OIDC was meant to fix with OAuth2. or maybe it's the fault of the spec authors all being too lax given all the vendors' influence.

i know this isn't your your fault @brentschmaltz, but i'm just expressing frustration that these AAD specifics seem to crop up frequently in what's purported to be a protocol library. and i apologize to @Tratcher for hijacking his PR with my off topic comments.

PinpointTownes: So we're all reviewing an AAD sample that relies on a feature that AAD doesn't support? #YOLO

@brockallen nah, [let's integrate ADAL directly in the OIDC middleware] (https://github.com/aspnet/Security/pull/258#discussion_r30331659) :trollface:

brentschmaltz: @brockallen I think you are correct, the 'session_state' belongs in a AAD-specific MW. We should just rely on the 'sid' claim. We shouldn't muddy the waters.

Tratcher:
HttpLogoutSupported = false. Closing this PR for now as we're lacking an implantation for interop testing.

leastprivilege:
Maybe you should use an openid connect spec compliant and certified server to test with - 

you can find a selection here: http://openid.net/certification/

just sayin' ;p

PR #629, opened 2015-12-22
Stardardizing middleware to use ConfigureOption lambda

JunTaoLuo:
Standardizing ConfigureOptions to lambda https://github.com/aspnet/Coherence-Signed/issues/144

### DIFF
--- a/src/Microsoft.AspNet.Authentication.Facebook/FacebookAppBuilderExtensions.cs
+++ b/src/Microsoft.AspNet.Authentication.Facebook/FacebookAppBuilderExtensions.cs
@@ -15,27 +15,6 @@ namespace Microsoft.AspNet.Builder
         /// Adds the <see cref="FacebookMiddleware"/> middleware to the specified <see cref="IApplicationBuilder"/>, which enables Facebook authentication capabilities.
         /// </summary>
         /// <param name="app">The <see cref="IApplicationBuilder"/> to add the middleware to.</param>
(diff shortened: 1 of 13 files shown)


src/Microsoft.AspNet.Authentication.Facebook/FacebookAppBuilderExtensions.cs
```diff
@@ -44,13 +23,15 @@ public static IApplicationBuilder UseFacebookAuthentication(this IApplicationBui
             {
                 throw new ArgumentNullException(nameof(app));
             }
-
-            var options = new FacebookOptions();
-            if (configureOptions != null)
+            if (configureOptions == null)
             {
-                configureOptions(options);
+                throw new ArgumentNullException(nameof(configureOptions));
```
JunTaoLuo: I've made `configureOptions` explicitly required but should we allow null configure options? I've seen both patterns being used. Some places even default `configureOptions` to null so that it's an optional parameter.

Tratcher: Only if the default options are meaningful. That's not the case for any of the security middleware. Also, don't use opational params, add an extra overload.

Tratcher: Well, maybe for cookies...

Tratcher:
:shipit:

Issue #627, opened 2015-12-21
The values of Roles property in AuthorizeAttribute aren't trimmed


phenixdotnet:
Hi.

Actually in the asp.net 5 (1.0.0-rc1-final) the values of the Roles property in Authorize attribute aren't trimmed before using it.
This lead to the situation where if for any reason you have this code:
```csharp

[Authorize(Roles="user, admin")]
public class MyController : Controller
{
}

```
Users with the admin role will never be authorized because the role name will be evaluated to " admin" and not "admin".

I can work on a patch and submit it if you want.

Regards

HaoK:
Sure, if you want to submit a PR to fix this, that would be great :)

phenixdotnet:
Hi.

PR submitted (#634)

Regards

HaoK:
a8ef4df32a6ce8f47da882d34685768dafbe7c8e

PR #634, opened 2015-12-28
Fix missing Trim in Roles and Schemes split

phenixdotnet:
PR for issue #627

### DIFF
--- a/src/Microsoft.AspNet.Authorization/AuthorizationPolicy.cs
+++ b/src/Microsoft.AspNet.Authorization/AuthorizationPolicy.cs
@@ -85,26 +85,22 @@ namespace Microsoft.AspNet.Authorization
                     policyBuilder.Combine(policy);
                     useDefaultPolicy = false;
                 }
-                var rolesSplit = authorizeAttribute.Roles?.Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
+                var rolesSplit = authorizeAttribute.Roles?.Split(',');
                 if (rolesSplit != null && rolesSplit.Any())
                 {
-                    for (int i = 0; i < rolesSplit.Length; ++i)
-                    {    
-                        rolesSplit[i] = rolesSplit[i].Trim(); 
-                    }
+                    var trimmedRolesSplit = rolesSplit.Where(r => !string.IsNullOrWhiteSpace(r)).Select(r => r.Trim());
 
-                    policyBuilder.RequireRole(rolesSplit.Where(r => !string.IsNullOrWhiteSpace(r)));
+                    policyBuilder.RequireRole(trimmedRolesSplit);
                     useDefaultPolicy = false;
                 }
-                var authTypesSplit = authorizeAttribute.ActiveAuthenticationSchemes?.Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries);
+                var authTypesSplit = authorizeAttribute.ActiveAuthenticationSchemes?.Split(',');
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Authorization/AuthorizationPolicy.cs
```diff
@@ -88,6 +88,9 @@ public static AuthorizationPolicy Combine(AuthorizationOptions options, IEnumera
                 var rolesSplit = authorizeAttribute.Roles?.Split(',');
                 if (rolesSplit != null && rolesSplit.Any())
                 {
+                    for (int i = 0; i < rolesSplit.Length; ++i)
+                        rolesSplit[i] = rolesSplit[i]?.Trim();
```
HaoK: nit: coding style requires { } even around one liners, same for below

src/Microsoft.AspNet.Authorization/AuthorizationPolicy.cs
```diff
@@ -96,7 +99,9 @@ public static AuthorizationPolicy Combine(AuthorizationOptions options, IEnumera
                 {
                     foreach (var authType in authTypesSplit)
                     {
-                        policyBuilder.AuthenticationSchemes.Add(authType);
+                        if (string.IsNullOrEmpty(authType))
```
HaoK: Instead of this explicit check for empty, can you use the the overoad of Split which takes in `StringSplitOptions.RemoveEmptyEntries`

test/Microsoft.AspNet.Authorization.Test/AuthorizationPolicyFacts.cs
```diff
@@ -67,5 +67,39 @@ public void CanReplaceDefaultPolicy()
...
+            var combined = AuthorizationPolicy.Combine(options, attributes);
+
+            Assert.True(combined.Requirements.Any(r => r is RolesAuthorizationRequirement));
+            var rolesAuthorizationRequirement = combined.Requirements.OfType<RolesAuthorizationRequirement>().First();
+            Assert.Equal(2, rolesAuthorizationRequirement.AllowedRoles.Count());
+            Assert.True(rolesAuthorizationRequirement.AllowedRoles.Any(r => r.Equals("r1")));
+            Assert.True(rolesAuthorizationRequirement.AllowedRoles.Any(r => r.Equals("r2")));
+        }
+
+        [Fact]
```
HaoK: Can you also add/or update one of these test for the an empty scheme/role?
Roles = "r1,,,,r2", AUthenticationSchemes = "a1,,,a2"

phenixdotnet:
I added 2 tests with the roles or authentication scheme empty or with white space.
I have to keep the ```if(string.IsNullOrWhiteSpace())```and use the ```StringSplitOptions.RemoveEmptyEntries``` to remove the empty or white space entries.

HaoK:
Okay so if we still need the NullOrWhiteSpace check, there's no real value in the RemoveEmptyEntries option, sorry about that, can you revert that back to the simple Split and other than that this looks good, I'll merge this in next week...  Thanks!

phenixdotnet:
Ok no prob :).

I will do that ASAP and let you known.

phenixdotnet:
Rework done.
One last things, I used linq to filter roles (````roles.Where(r => !string.IsNullOrWhiteSpace(r)).Select(r => r => r.Trim());````)
Not sure if it's the best way to do this.

HaoK:
Looks good, I'll get this in next week, thanks again!

phenixdotnet:
You're welcome :)

Issue #611, opened 2015-12-09
Split Microsoft.Owin.Security.Cookies.Interop into 2 packages


PinpointTownes:
We'd like to use the ticket serializer and the data protector adapter in the OWIN/Katana version of ASOS, but having a dependency on the cookies middleware is something we'd like to avoid. Separating the cookies extension from the rest of the package would be ideal.

https://github.com/aspnet/Security/pull/408#discussion_r38157266

https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/issues/188#issuecomment-163291304

If you agreed with the general idea, I (or @MonkeyJamboree) would be happy to submit a PR. I guess simply creating a new `Microsoft.Owin.Security.Interop` package containing the ticket serializer and the data protector would be the best way to go.

/cc @Eilon @HaoK @Tratcher

Eilon:
We are not planning on doing this; we recommend just copying the code as needed.

PinpointTownes:
TBH, this doesn't seem like a good idea, as the ticket serializer adapter must exactly match the one used by ASP.NET 5, which is an easy task when using an official package but rather fragile when rolling your own version.

PinpointTownes:
@Eilon if you don't like the idea of having a new package, maybe we could simply make the new `Microsoft.Owin.Security.Interop` I guess a sources package?

MonkeyJamboree:
@Eilon out of curiosity, why is copying the code recommended over actually separating out `Microsoft.Owin.Security.Interop` into its own package? It seems to me since the namespace and code is assumed to be for interoperability, it should be consumable by others easily. If not, is there a package that's explicitly for this purpose instead? Creating unneeded dependencies in code or duplicating code are usually are not good things.

Tratcher:
This package was built specifically for interop with Microsoft.Owin.Security.Cookies. 

One way to pull the  src is via git submodule.

PinpointTownes:
> This package was built specifically for interop with Microsoft.Owin.Security.Cookies.

True, but the ticket serializer and the data protector shim are not specific to the cookies middleware. Sounds like a bad layering IMHO.

blowdart:
The ticket serializer was for Forms auth. Which became cookie auth, Arguably it is for the cookie middleware.

Data Protection has a 4.5.2 implementation which you can bring in as a normal nuget package.

PinpointTownes:
[...]

PR #641, opened 2016-01-04
Remove Microsoft.Owin.Security.Cookies.Interop and introduce Microsoft.Owin.Security.Interop

PinpointTownes:
Fixes https://github.com/aspnet/Security/issues/611.

Note: making the new `Microsoft.Owin.Security.Cookies.Interop` a sources package was not possible, as they are not supported by .csproj projects (using a project.json-compatible-only thing would defeat the whole purpose of an interop package).

/cc @Eilon @Tratcher


global.json
```diff
@@ -1,3 +1,4 @@
 {
+    "packages": "packages",
```
PinpointTownes: Ignore that, I'll remove it when we're done with the main changes (I didn't want the CI dev packages to be added to my global packages folder).

src/Microsoft.Owin.Security.Interop/AspNetTicketSerializer.cs
```diff
@@ -15,7 +15,7 @@ public class AspNetTicketSerializer : IDataSerializer<AuthenticationTicket>
         private const string DefaultStringPlaceholder = "\0";
         private const int FormatVersion = 5;
 
-        public static TicketSerializer Default { get; } = new TicketSerializer();
+        public static AspNetTicketSerializer Default { get; } = new AspNetTicketSerializer();
```
PinpointTownes: Exposing a property using the vNext serializer was wrong, I took the liberty to fix it :smile:

test/Microsoft.Owin.Security.Cookies.Interop.Test/CookiesInteropTests.cs
```diff
@@ -11,103 +10,24 @@
...
 using Microsoft.AspNet.Hosting;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Owin;
-using Microsoft.Owin.Security.Cookies;
-using Microsoft.Owin.Security.Cookies.Interop;
 using Microsoft.Owin.Testing;
 using Owin;
 using Xunit;
 
-namespace Microsoft.AspNet.CookiePolicy.Test
```
PinpointTownes: Probably a bad copy/paste that deserved to be fixed :smile:

src/Microsoft.AspNet.Authentication/DataHandler/TicketDataFormat.cs
```diff
@@ -7,7 +7,8 @@ namespace Microsoft.AspNet.Authentication
 {
     public class TicketDataFormat : SecureDataFormat<AuthenticationTicket>
     {
-        public TicketDataFormat(IDataProtector protector) : base(new TicketSerializer(), protector)
+        public TicketDataFormat(IDataProtector protector)
+            : base(TicketSerializer.Default, protector)
```
PinpointTownes: I updated this constructor to use the static instance (that was not used at all in the current code base :weary:)

HaoK: Maybe we should go the other way and just nuke the static that wasn't used...  I recently nuked all the statics in the current bits, this might be a leftover...  I don't think its very compelling as a static...

PinpointTownes: I'm fine removing them, but note that we won't be able to remove `PropertiesSerializer.Default` without a more important refactoring (as it's used by `TicketSerializer` and `RequestTokenSerializer`).

src/Microsoft.Owin.Security.Interop/project.json
```diff
@@ -0,0 +1,18 @@
...
+    "compilationOptions": {
+        "warningsAsErrors": true,
+        "keyFile": "../../tools/Key.snk"
+    },
+    "description": "A compatibility layer for sharing authentication tickets between Microsoft.Owin.Security and Microsoft.AspNet.Authentication.",
+    "dependencies": {
+    },
+    "frameworks": {
+        "net451": {
+            "dependencies": {
```
Tratcher: Move these dependencies to the root dependencies node.

Tratcher:
This reduces the Microsoft.Owin.Security.Cookies.Interop package to a single extension method: UseCookieAuthentication. At that point we should just move it to a sample. @HaoK, @muratg, @Eilon?

HaoK:
I'm fine with removing the cookie dependency for now, I'm still working through the full migration end to end anyways, we can nuke the extension and the old package for now, and I'll just add the extension when I get the full migration checked in.  The only thing that depends on this is an identity interop sample which isn't part of the build so we should be safe nuking the extension for now...

src/Microsoft.AspNet.Authentication/DataHandler/TicketDataFormat.cs
HaoK: I think we addressed that by just newing up a new instance instead of using the static when we removed them in the current bits.

src/Microsoft.Owin.Security.Interop/project.json
PinpointTownes: Sure (this was just a copy/paste from the existing project.json)

src/Microsoft.AspNet.Authentication/DataHandler/TicketDataFormat.cs
PinpointTownes: Not for `PropertiesSerializer`:

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Twitter/Messages/RequestTokenSerializer.cs#L98

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication/DataHandler/TicketSerializer.cs#L66

HaoK: Okay then lets just leave that one for now

PinpointTownes: To remove all the statics, we could flow the properties serializer as a constructor dependency.

PinpointTownes:
I'm fine removing `Microsoft.Owin.Security.Cookies.Interop` if you think it's not useful enough.

src/Microsoft.Owin.Security.Interop/project.json
PinpointTownes: Fixed (in both projects)

HaoK:
Yeah you can go ahead and delete it with this PR, and I'll add it back in some form in the future

PinpointTownes:
I removed `Microsoft.Owin.Security.Cookies.Interop` and `Microsoft.Owin.Security.Cookies.Interop.Test` but I kept the cookies interop tests (now part of `Microsoft.Owin.Security.Interop.Test`).

Tratcher:
:shipit: with the usual squash&rebase cleanup

Tratcher:
Rebased and merged.

Issue #628, opened 2015-12-21
Authentication cookie is badly renewed when the security stamp has been validated


fredgate:
The authentication cookie can be renewed in these two cases :
- the elapsed expiration time is more halfway expiration window
- the security stamp has been validated and a new identity is rebuilt

In the FinishResponseAsync method of the CookieAuthenticationHandler class (namepsace Microsoft.AspNet.Authentication.Cookies), a new ticket is built with optionally setting the properties IssuedUtc and ExpiresUtc.

When the sliding expiration is refreshed, the Issued and expires properties are well marked as to be updated (ReadCookieTicket method)
```
if (timeRemaining < timeElapsed)
{
  _shouldRenew = true;
  _renewIssuedUtc = currentUtc;
  var timeSpan = expiresUtc.Value.Subtract(issuedUtc.Value);
  _renewExpiresUtc = currentUtc.Add(timeSpan);
}
```
But when the stamp validator set a new identity and indicates that the cookie should be renew, the handler does not define that the Issued property should be refreshed. I think that the HandleAuthenticateAsync method should be fixed with the line added below 
```
var context = new CookieValidatePrincipalContext(Context, ticket, Options);
await Options.Events.ValidatePrincipal(context);

if (context.Principal == null)
{
  return AuthenticateResult.Failed("No principal.");
}

if (context.ShouldRenew)
{
  _shouldRenew = true;
  // Add this line ?
  _renewIssuedUtc = Options.SystemClock.UtcNow;
}
```
The problem is actually that when the stamp was validated once, it will then be for each next request, requiring to query the database calling the FindByIdAsync method of the UserStore.

HaoK:
5bcc1bfc260978b0bea9b981056dd0beb31df400

PR #645, opened 2016-01-06
Need to update renewIssueEtc when renewing cookie

HaoK:
cc @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -117,6 +117,7 @@ namespace Microsoft.AspNet.Authentication.Cookies
             if (context.ShouldRenew)
             {
                 _shouldRenew = true;
+                _renewIssuedUtc = Options.SystemClock.UtcNow;
             }
 
             return AuthenticateResult.Success(new AuthenticationTicket(context.Principal, context.Properties, Options.AuthenticationScheme));
--- a/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -37,6 +37,33 @@ namespace Microsoft.AspNet.Authentication.Cookies
             return _readCookieTask;
         }
 
+        private void CheckForRefresh(AuthenticationTicket ticket)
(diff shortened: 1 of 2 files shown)


HaoK:
Fixes https://github.com/aspnet/Security/issues/628

I'm not sure if we need to handle sliding expiry as well here...

Tratcher:
I think you need to handle sliding refresh.

https://github.com/aspnet/Security/blob/haok/1-6-issue/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs#L85-L86
If they access the site every 30 seconds and issuedUtc is updated each time, then timeElapsed will always be 30s and timeRemaining will keep shrinking. The sliding expiration won't be renewed until you're within 30s, and if you miss a request in that window then your cookie suddenly expires.

HaoK:
Yeah I thought as much, this cookie refresh/expires code is pretty complicated as its written :/

HaoK:
I think I will try to refactor the cookie refresh/renew logic into some sane methods so this is actually readable and sane...

HaoK:
Updated:
- Moved logic for refresh into two helpers
- We request refresh if explicitly asked for via context.ShouldRenew, or sliding check
- If we renew, we ALWAYS set expires/issues UTC (is that correct/safe?)

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -37,6 +37,33 @@ private Task<AuthenticateResult> EnsureCookieTicket()
...
+                if (timeRemaining < timeElapsed)
+                {
+                    RequestRefresh(ticket);
+                }
+            }
+        }
+
+        private void RequestRefresh(AuthenticationTicket ticket)
+        {
+            _shouldRenew = true;
```
Tratcher: Any idea why the variables all say renew and not refresh? Are they used for anything else?

Tratcher:
:shipit: and consider updating the naming.

src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationHandler.cs
HaoK: Not used anywhere else I don't think, I'll rename to refresh, seems fine

HaoK:
5bcc1bfc260978b0bea9b981056dd0beb31df400

PR #646, opened 2016-01-06
Updating to new options pattern

JunTaoLuo:
Updating using the design in https://github.com/aspnet/Options/issues/105. Any thoughts?

### DIFF
--- a/src/Microsoft.AspNet.Authentication.Cookies/CookieAppBuilderExtensions.cs
+++ b/src/Microsoft.AspNet.Authentication.Cookies/CookieAppBuilderExtensions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using Microsoft.AspNet.Authentication.Cookies;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNet.Builder
 {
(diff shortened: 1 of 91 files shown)


src/Microsoft.AspNet.Authentication.Cookies/CookieAppBuilderExtensions.cs
```diff
@@ -23,30 +24,7 @@ public static IApplicationBuilder UseCookieAuthentication(this IApplicationBuild
...
-            {
-                throw new ArgumentNullException(nameof(configureOptions));
-            }
-
-            var options = new CookieAuthenticationOptions();
-            configureOptions(options);
-
-            return app.UseMiddleware<CookieAuthenticationMiddleware>(options);
+            return app.UseCookieAuthentication();
         }
```
Tratcher: All the other middleware also need this parameter-less extension to pull options from DI.

src/Microsoft.AspNet.Authentication.Cookies/CookieAppBuilderExtensions.cs
```diff
@@ -22,31 +23,8 @@ public static IApplicationBuilder UseCookieAuthentication(this IApplicationBuild
...
-            if (configureOptions == null)
-            {
-                throw new ArgumentNullException(nameof(configureOptions));
-            }
-
-            var options = new CookieAuthenticationOptions();
-            configureOptions(options);
-
-            return app.UseMiddleware<CookieAuthenticationMiddleware>(options);
+
```
HaoK: This is reopening the pit of doom again with respect to cookies, since unlike most other middleware cookies will show up more than once in the pipeline and this overload only works against the options from DI.  Hrm...

HaoK:
Overall its not too bad, I definitely agree that having the new makes it clear that its an instance, making the additional UseCookies/Facebook/Google() that takes no arguments hopefully a little less confusing.  

Are we going to update templates/samples to prefer configuration of middleware options in ConfigureServices instead of inline in Configure across the board?  

Moving configuration of the middleware options get us closer to my dream where ConfigureServices does everything, including building the app pipeline similar to how options get configured :)

Tratcher:
Looks good.

Tratcher:
:shipit:

PR #653, opened 2016-01-14
OIDC cleanup, feedback

Tratcher:
- Remove broken test classes. These were failing silently.
- Allow post logout redirect uri to be relative.
- Allow sharing the callback path. E.g. fall-through. Default to true.
- Change CallbackPath default to "/".
- Use the token issuer for claims
- Default to AutomaticChallenge
- Remove unused Audiance

@brentschmaltz @tushargupta51 @HaoK @PinpointTownes


Tratcher:
@vibronet

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -345,6 +363,10 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
                 if (string.IsNullOrEmpty(message.State))
                 {
                     // This wasn't a valid ODIC message, it may not have been intended for us.
```
Eilon: Typo in the above comment: should be OIDC.

PinpointTownes:
Curious: what's the reason behind these two changes?

> Allow sharing the callback path. E.g. fall-through. Default to true.
Change CallbackPath default to "/".

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -345,6 +363,10 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
+                    if (Options.ShareCallbackPath)
+                    {
+                        return AuthenticateResult.Skip();
+                    }
                     Logger.LogDebug(11, "message.State is null or empty.");
```
brentschmaltz: Consistent logging? Sometimes before, sometimes after skipping.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -594,7 +625,7 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
                     // response, since it's not a valid parameter when using the implicit flow.
                     // See http://openid.net/specs/openid-connect-core-1_0.html#Authentication
                     // and https://tools.ietf.org/html/rfc6749#section-4.2.2.
-                    SaveTokens(ticket.Principal, message, saveRefreshToken: false);
+                    SaveTokens(ticket.Principal, message, jwt.Issuer, saveRefreshToken: false);
```
brentschmaltz: Is jwt.Issuer used to set the Claim.Issuer? if so we may want to think about a different value so users can distinguish.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -308,6 +316,11 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
                 // See http://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#Security
                 if (!string.IsNullOrEmpty(message.IdToken) || !string.IsNullOrEmpty(message.AccessToken))
                 {
+                    if (Options.ShareCallbackPath)
```
PinpointTownes: This extra check is here to ensure authorization responses using an unsafe `response_mode`/`response_type` combination (i.e an identity or an access token flowed using the query string) are always rejected, as it's necessarily the result of a non-standard OIDC server doing crazy things (personally, I'd even be tempted to add a check for `refresh_token`).

Since it's a serious protocol exception, I'm not sure skipping the request is the right thing to do.

src/Microsoft.AspNet.Authentication/RemoteAuthenticationOptions.cs
```diff
@@ -50,7 +50,7 @@ public string DisplayName
 
         /// <summary>
         /// Defines whether access and refresh tokens should be stored in the
-        /// <see cref="ClaimsPrincipal"/> after a successful authentication.
+        /// <see cref="ClaimsPrincipal"/> after a successful authorization with the remote provider.
```
PinpointTownes: nit: why replacing `authorization` by `authentication` here? The remote middleware stuff is generic enough not to be tied to OAuth2 (which is the only authorization protocol implemented by ASP.NET 5).

Tratcher:
@PinpointTownes It's a compromise between the old Katana behavior (and common usage beyond that) that listened to everything, and the new infrastructure that encourages you to have a dedicated path.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: I don't want too log much noise when we're pretty sure it wasn't our message. I'll cross-check with Katana.

PinpointTownes:
FWIW, this new `ShareCallbackPath` looks absolutely horrifying and rather fragile.

For instance, what happens if I register 2 instances of the OIDC middleware and that they both skip an authorization response missing the `state` parameter. AFAICT, this error will never be logged, unless I explicitly set `ShareCallbackPath` on the second instance.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: @vibronet explicitly recommended this. Follow up with him and get back to me.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: But this early in the request it can be a false positive, any query containing the key id_token. Either way we won't process the message. Disabling ShareCallbackPath will get you better error handling though.

Tratcher:
I think you're mixing two scenarios. Yes, invalid messages are harder to detect with ShareCallbackPath, but we want to avoid logging a misleading error for all non-OIDC requests. We'll have to find a balance.

HaoK:
Maybe this should be off by default since more often than not, there is not a callback path being shared correct?

PinpointTownes:
> It's a compromise between the old Katana behavior (and common usage beyond that) that listened to everything, and the new infrastructure that encourages you to have a dedicated path.

The old behavior was absolutely bad, why do you want to re-introduce it? Personally, I've never seen a serious implementation recommending to use the root path as the callback URL :smile:

Eilon:
@PinpointTownes we feel that the supportability of both new apps and for migrating existing apps would be better by defaulting to the root of the app, and allowing fall-through in case the request is not for auth. It ties in better with existing guidance on setting up this kind of auth, and works better with the Azure AD configuration experience as well.

Another scenario we think this would better support is IDP-initiated requests, where there's no extra bounce between the IDP, the client, and the server, there could be a direct request to a resource and we'd want to have the right infrastructure available to support capturing those requests (though we admittedly don't yet support that scenario).

Though several of us would agree that there are advantages to having a distinct URI (and that's why we had that code until now), we're concerned it would be too restrictive in the future, and more difficult to support as well.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
```diff
@@ -190,5 +183,13 @@ public new IOpenIdConnectEvents Events
         /// This is disabled by default.
         /// </summary>
         public bool UseTokenLifetime { get; set; }
+
+        /// <summary>
+        /// Indicates if requests to the CallbackPath may also be for other components. If enabled the middleware will pass
+        /// requests through that do not contain OpenIdConnect authentication responses. Disabling this and setting the
+        /// CallbackPath to a dedicated endpoint may provide better error handling.
+        /// This is enabled by default
```
Eilon: End sentence with a period

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
```diff
@@ -190,5 +183,13 @@ public new IOpenIdConnectEvents Events
...
+        /// </summary>
+        public bool ShareCallbackPath { get; set; } = true;
```
Eilon: Would like to see a new name for this. The "share" part is irking me.

Maybe `IgnoreNonAuthenticationRequests`? Kind of a mouthful but I think it better expresses the intent. So if not this exact wordy name, maybe something similar?

HaoK: Ignore/SkipUnrecognizedRequests would be what I understood the functionality actually does. Not sure if that's better though

PinpointTownes:
> @PinpointTownes we feel that the supportability of both new apps and for migrating existing apps would be better by defaulting to the root of the app, and allowing fall-through in case the request is not for auth. It ties in better with existing guidance on setting up this kind of auth, and works better with the Azure AD configuration experience as well.

So your plan is to punish all the poor people using standard servers just to make the Azure AD guys happy? If there's something wrong with the Azure AD implementation, it's up to the AD team to fix it :smile: 

Curious: why not creating an Azure AD-specific OIDC middleware, as suggested by @leastprivilege recently?

> Another scenario we think this would better support is IDP-initiated requests, where there's no extra bounce between the IDP, the client, and the server, there could be a direct request to a resource and we'd want to have the right infrastructure available to support capturing those requests (though we admittedly don't yet support that scenario).

This PR is all about OIDC, right? Not SAML, WS-Fed or any other thing? :smile: 
Unless a whole new draft/RFC was just posted (maybe I'm living in a cave? :smile:), there's absolutely no such thing in OIDC, so I guess it will be some non-standard stuff...

> Though several of us would agree that there are advantages to having a distinct URI (and that's why we had that code until now), we're concerned it would be too restrictive in the future, and more difficult to support as well.

What prevents you from setting the callback path to `PathString.Empty` to override the default value? If you're concerned about discoverability, why not reintroducing `RedirectUri` and set `CallbackPath` accordingly?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
Eilon: @HaoK yeah sure maybe something like that.

Tratcher:
I'm trying to avoid reintroducing RedirectUri (even though AAD wants it) and deriving CallbackPath from it because it is error prone. RedirectUri is absolute, where CallbackPath is relative to the current PathBase and we can't determine that at startup.

Tratcher:
@vibronet, care to comment?

Tratcher:
Minor updates after comparing to the Katana flow.

vibronet:
Comments, mostly for @PinpointTownes 
- Ignoring the "serious implementation" zing - all defaults in Katana and VS2013/2015 use the app root as the default. The implementation before this PR made it impossible to use the root, which would have forced anybody who wants to upgrade to renegotiate their app registration with their provider. Having object model details have an impact on the protocol level is bad, and requiring customers to renegotiate their app registration at the IdP would be unacceptable. It cannot be harder to move from ASP.NET 4.6 to 5 than it would be to move from 4.6 to Node or Ruby, which is what would happen if the root would be disallowed as RedirectUri. In any case, my understanding is that you are free to override it with any URL of your choice any time you want.
- Can you enumerate the "standard servers" you are referring to? Using the root URL does not appear to be violating any standard. Besides of the formal compliance, things also seem to work in practice. There is an enormous amount of apps currently using Azure AD today, and a large number of apps using pre-OpenId Connect authentication strategies which get back their tokens on the application's root. They are all in production, happily producing business. Can you substantiate what you are referring to with "something wrong with the Azure AD implementation"? 
- Re: IdP initiated flow, Chris is referring to http://openid.net/specs/openid-connect-core-1_0.html#ThirdPartyInitiatedLogin. This was more of a discussion about possible problems that might arise by not listening for auth messages for all possible app URLs at once, as we did in the past, and the IdP initiated flow was one example we threw around when considering possible problematic scenarios. Going back to the spec, for the way it works in OIDC it appears that won't be a problem. But again, the point wasn't about the root as the returnUri - it was about listening only to one URL, regardless of what such URL is.
- I agree it would be great to have back RedirectUri. I would like it back mostly for usability purposes (the instructions for using the middleware are much simpler if the developer just needs to paste in a value he/she gets from the IdP, without havijg to do path math in his/her head) and consistency with both Katana and other auth components on other platforms (AFAIK they all admit passing the absolute RedirectUri).

PinpointTownes:
> Ignoring the "serious implementation" zing.

Don't take it personally, the reference to "serious implementations" was only to mention that other well-known OIDC client implementations didn't use the app root as the default callback path, it had nothing to do with Azure AD (e.g, phpOIDC only accepts the hardcoded `/callback` path)

> The implementation before this PR made it impossible to use the root, which would have forced anybody who wants to upgrade to renegotiate their app registration with their provider. Having object model details have an impact on the protocol level is bad, and requiring customers to renegotiate their app registration at the IdP would be unacceptable.

Impacting the performance and logging experience for all the users of the OIDC middleware just to support app-root callbacks seems much worse than forcing people to explicitly turn a setting to `true` or to update the callback path in their app registration, IMHO. I'm not fundamentally opposed to the passthrough mode offered by this PR, but I'm convinced that enabling it by default is not a good idea.

> Can you enumerate the "standard servers" you are referring to? Using the root URL does not appear to be violating any standard. Besides of the formal compliance, things also seem to work in practice. There is an enormous amount of apps currently using Azure AD today, and a large number of apps using pre-OpenId Connect authentication strategies which get back their tokens on the application's root.

I've never said using the app root was not standard, just that it has practical implications we can't ignore in a web server environment: if this PR is merged, the OIDC middleware will be **triggered for every GET or POST call made to the app root** (which is a hot path for many applications) and allocate useless `OpenIdConnectMessage` instances even if the request doesn't correspond to an OIDC response.

This new `ShareCallbackPath` also makes the default logging experience horrible, since invalid OIDC responses are skipped and never logged: with `ShareCallbackPath` being `true` by default, you'll never know that the OIDC response lacked the `state` parameter or that it was invalid (which may indicate a XSRF attack).

> They are all in production, happily producing business. Can you substantiate what you are referring to with "something wrong with the Azure AD implementation"?

If Azure AD only supports or suggests using app-root callback paths, then you should probably change that for the practical reasons we're discussing: having to change the client implementation of a standard protocol to match what a single server implementation does seems wrong, IMHO. On the other hand, simply updating the app registration in Azure AD doesn't really seem like a big deal, specially when migrating an existing app to vNext, whose migration path is generally not trivial and involves many breaking changes.

> Re: IdP initiated flow, Chris is referring to http://openid.net/specs/openid-connect-core-1_0.html#ThirdPartyInitiatedLogin. This was more of a discussion about possible problems that might arise by not listening for auth messages for all possible app URLs at once, as we did in the past, and the IdP initiated flow was one example we threw around when considering possible problematic scenarios.

Third-party initialization doesn't change anything to the fact the client application is still responsible for preparing the authorization request (with the mandatory `redirect_uri`): the third party has no influence on the callback path the client should or must use.

Going back to the spec, for the way it works in OIDC it appears that won't be a problem. But again, the point wasn't about the root as the returnUri - it was about listening only to one URL, regardless of what such URL is.

Note that this PR doesn't allow listening on every possible URL, it just makes the app-root the default callback path and the callback endpoint passthrough by default.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
brentschmaltz: @Tratcher @vibronet I pinged vibronet on a Slack message.

vibronet: Hey Brent, the idea here is that given that we do verify that those claims are coming from the issuer, we should reflect that in the resulting Claim entries. The mechanics through which we obtained those claims should not make a difference form code that simply consumes the claim values and need to know whether those claims are backed by the issuer.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -308,6 +316,11 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
                 // See http://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#Security
                 if (!string.IsNullOrEmpty(message.IdToken) || !string.IsNullOrEmpty(message.AccessToken))
                 {
+                    if (Options.SkipUnrecognizedRequests)
```
brentschmaltz: This is a hard call. We want to discourage this and we should fail for OIDC. When would it be considered OK to process this message?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -344,7 +362,11 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
                 // Fail if state is missing, it's required for the correlation id.
                 if (string.IsNullOrEmpty(message.State))
                 {
-                    // This wasn't a valid ODIC message, it may not have been intended for us.
+                    // This wasn't a valid OIDC message, it may not have been intended for us.
+                    if (Options.SkipUnrecognizedRequests)
+                    {
+                        return AuthenticateResult.Skip();
+                    }
                     Logger.LogDebug(11, "message.State is null or empty.");
```
brentschmaltz: Consider moving the Logger.LogDebug before returning.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
brentschmaltz: @vibronet @Tratcher I understand that logically the issuer of the artifacts is the issuer. However, it is not true that the IDP issued those claims. Setting the issuer == 'iss' is implying the claims arrived inside a signed token. Don't know what the harm is, but we are deviating from what arrived. I am not aware how the consumer will be consume these claims.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -721,38 +748,38 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
                 identity.AddClaim(new Claim(OpenIdConnectParameterNames.IdToken, message.IdToken,
-                                            ClaimValueTypes.String, Options.ClaimsIssuer));
+                                            ClaimValueTypes.String, issuer));
             }
 
             if (saveRefreshToken && !string.IsNullOrEmpty(message.RefreshToken))
             {
                 identity.AddClaim(new Claim(OpenIdConnectParameterNames.RefreshToken, message.RefreshToken,
-                                            ClaimValueTypes.String, Options.ClaimsIssuer));
+                                            ClaimValueTypes.String, issuer));
```
brentschmaltz: OIDC should never have a RefreshToken.

PinpointTownes: Look closer, there's a `saveRefreshToken` boolean determining whether storing a refresh token is legal or not.
It's set to `false` when receiving an authorization response and to `true` when receiving a token response.

brentschmaltz: Why is this code even here? If a refreshtoken is in the message, something is mixed up.

PinpointTownes: Heh... because we want to store the `refresh_token` returned in the token response in the code/hybrid flow?...

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
vibronet: I might be misunderstanding the matter here. My understanding is that the token WAS issued by the IDP, and if we are executing this code it means that we validated it as such. The fact that we obtained it via backchannel rather than front channel should be immaterial, *as long as we did validate the token and its origin as being the IDP*. Is there anything inaccurate in this?

src/Microsoft.AspNet.Authentication/RemoteAuthenticationOptions.cs
brentschmaltz: While our model allows dynamic setting of redirecturi it does not allow for dynamically accepting a uri. Every IDP requires registering a redirecturi, is there a place where the MW can examine the 'path' a message arrived and decide if it should proceed?

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: In pass through mode you have to err on the side of assuming the message does not belong to you. This is exactly the problem with pass through mode, it makes detecting protocol errors much more difficult.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: This says you have to request them separately, no? http://openid.net/specs/openid-connect-core-1_0.html#RefreshTokens

PinpointTownes:
I'm really surprised nobody mentioned the security implications of this PR: if the `state` is missing or invalid, the OIDC middleware will skip the request if `ShareCallbackPath = true`.

In this case, the rest of the pipeline will be invoked and since the default callback path corresponds to the root path, there's a high chance the request will be handled by MVC (in practice, to render a view): the query string will be flowed as-is and will be potentially leaked in the referrer when clicking a link or rendering an external image. And if there's an authorization code in the query string... it will be terribad.

src/Microsoft.AspNet.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
PinpointTownes: No, a refresh token may be returned in the initial token response (see http://openid.net/specs/openid-connect-core-1_0.html#TokenResponse for a non-normative sample).

Tratcher:
Updated. I've reverted the CallbackPath change and disabled SkipUnrecognizedRequests by default. I'm going to push this in the morning and then we'll do some app building to verify the usability and how this interacts with redirecturi.

PR #655, opened 2016-01-15
Delegate => AssertionRequirement sugar

HaoK:
Tweaks RequireDelegate a bit to make it simplier:
RequireAssertion(authContext => bool) is used to register a handler
directly without any of the mechanical methods/classes.

Intended to make it pretty easy to do simple validations against the context.User ClaimsPrincipal

cc @blowdart @divega


src/Microsoft.AspNet.Authorization/Infrastructure/AssertionRequirement.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace Microsoft.AspNet.Authorization.Infrastructure
+{
+    public class AssertionRequirement : AuthorizationHandler<AssertionRequirement>, IAuthorizationRequirement
+    {
+        /// <summary>
+        /// Function that is called to handle this requirement
+        /// </summary>
+        public Func<AuthorizationContext, bool> Handler { get; }
+
+        public AssertionRequirement(Func<AuthorizationContext, bool> handleMe)
```
davidfowl: handleMe? Lolz

src/Microsoft.AspNet.Authorization/Infrastructure/AssertionRequirement.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            Handler = handleMe;
```
davidfowl: Check for null.

src/Microsoft.AspNet.Authorization/Infrastructure/AssertionRequirement.cs
```diff
@@ -0,0 +1,28 @@
...
+        }
+
+        protected override void Handle(AuthorizationContext context, AssertionRequirement requirement)
+        {
+            if (Handler(context))
```
davidfowl: What should happen if this throws?

src/Microsoft.AspNet.Authorization/Infrastructure/AssertionRequirement.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// Function that is called to handle this requirement
+        /// </summary>
+        public Func<AuthorizationContext, bool> Handler { get; }
+
+        public AssertionRequirement(Func<AuthorizationContext, bool> handleMe)
+        {
+            Handler = handleMe;
+        }
+
+        protected override void Handle(AuthorizationContext context, AssertionRequirement requirement)
```
davidfowl: What about HandleAsync?

src/Microsoft.AspNet.Authorization/Infrastructure/AssertionRequirement.cs
HaoK: We don't have any try/catch guards today when running AuthZ handlers everywhere else, they just become uncaught exceptions.  

I don't think this one should be special as a result, I think blowing up is fine, they can always wrap it themselves

src/Microsoft.AspNet.Authorization/Infrastructure/AssertionRequirement.cs
HaoK: We don't have any sugar for adding an async assert today.  I'm not opposed to adding support for a `Func<AuthorizationContext, Task<bool>>` but it will be a bit messier with two of these...  @blowdart thoughts?

blowdart:
How does this look for end users? i.e. what are they writing in code?

HaoK:
Something like this where you replace the context => true with something less trivial :)

```
                services.AddAuthorization(options =>
                {
                    options.AddPolicy("Basic", policy => policy.RequireAssertion(context => true));
                });
```

We could yet one more layer of sugar on top to support registering a policy consisting entirely of this one func as well:

```
   services.AddAuthorization(options => {
      options.AddAssertionPolicy("Assertion1", context => context.User.Identity.HasClaim("SuperUser"))
   }
```

divega:
@HaoK FWIW, that extra layer of sugar looks much nicer to me, but I feel that I lack enough context in this area. Did you decide on handling async lambdas?

HaoK:
I'll merge what I have and file bugs for the extra layer of sugar and async sugar...

https://github.com/aspnet/Security/issues/657

HaoK:
2e1a8b31cd0c520ffd69ba7818313ae9a9f688a4

Issue #647, opened 2016-01-07
Consider moving GenerateCorrelationId and ValidateCorrelationId to RemoteAuthenticationHandler


PinpointTownes:
Currently, both `OAuthHandler` and `OpenIdConnectHandler` rely on private or protected methods to generate and validate anti-XSRF tokens.

To make things DRYier, standardize the two implementations and allow `RemoteAuthenticationHandler` inheritors to use correlation checks without having to embed their own methods, moving `GenerateCorrelationId` and `ValidateCorrelationId` to `RemoteAuthenticationHandler` would be ideal.

(FYI, I had to copy these methods for the generic OpenID2 middleware: https://github.com/aspnet-contrib/AspNet.Security.OpenId.Providers/blob/dev/src/AspNet.Security.OpenId/OpenIdAuthenticationHandler.cs)

/cc @HaoK @Tratcher

PR #669, opened 2016-01-23
Move GenerateCorrelationId and ValidateCorrelationId to RemoteAuthenticationHandler

PinpointTownes:
Fixes https://github.com/aspnet/Security/issues/647.

/cc @Eilon @Tratcher please don't merge it yet, there are a few things I'm not totally happy with.


src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -99,5 +107,69 @@ protected override Task<bool> HandleForbiddenAsync(ChallengeContext context)
...
+            var cookieOptions = new CookieOptions
+            {
+                HttpOnly = true,
+                Secure = Request.IsHttps
+            };
+            Response.Cookies.Delete(cookieName, cookieOptions);
+
+            if (!string.Equals(correlationCookie, CorrelationMarker, StringComparison.Ordinal))
+            {
+                Logger.LogWarning(28, "{0} correlation cookie and state property mismatch.", cookieName);
```
PinpointTownes: This log warning is misleading as the actual check is made earlier, at line 153 (it checks whether a cookie whose name corresponds to the nonce value exists).

This check only ensures the cookie contains "N" :see_no_evil:

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -169,7 +166,12 @@ public OAuthHandler(HttpClient backchannel)
                 throw new ArgumentNullException(nameof(context));
             }
 
-            var properties = new AuthenticationProperties(context.Properties);
+            var properties = new AuthenticationProperties(context.Properties)
+            {
+                // Review: should we add an option to configure the state lifetime?
+                ExpiresUtc = DateTimeOffset.UtcNow + TimeSpan.FromHours(1)
```
PinpointTownes: Note: a 1-hour lifetime is probably a bit excessive, but it's the default value used by IdentityModel for the OIDC middleware, so I've reused it consistency.

global.json
```diff
@@ -1,3 +1,4 @@
 {
+    "packages": "packages",
```
PinpointTownes: (I'll remove that when we're done with the other changes)

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
Eilon: BTW you can call `DateTimeOffset.UtcNow.AddHours(1)`.

PinpointTownes: Will do. I'll also use the `SystemClock` property added by @Tratcher in his recent PR to both the OAuth2 and the OIDC middleware.

PinpointTownes: One question remains: should we make it configurable?

PinpointTownes: Updated to use a 15mn lifetime (I just rememberd the OIDC options explicitly replace the default lifetime :smile:)

PinpointTownes:
Updated. I fixed the unit tests (it's much easier when you have the packages :smile:)

The `ChallengeSettingState` test still fails. I guess it's caused by a subtle difference in the way `HttpClient` and `UrlEncoder` (used by the fake properties serializer) encode the comma:

`.expires%20Tue,%252026%2520Jan%25202016%252016%253A35%253A26%2520GMT`

`.expires%20Tue%2C%252026%2520Jan%25202016%252016%253A35%253A26%2520GMT`

This test is really cryptic for me. I'm not sure how I'm supposed to fix it...

Tratcher:
The test formatter was making assumptions about the data. I was able to get it working by changing AuthenticationPropertiesFormatterKeyValue from
```
-                sb.Append(encoder.Encode(item.Key) + " " + encoder.Encode(item.Value) + " ");
to
+                sb.Append(Uri.EscapeDataString(item.Key) + " " + Uri.EscapeDataString(item.Value) + " ");
```
See https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/src/Microsoft.IdentityModel.Protocols/AuthenticationProtocolMessage.cs#L113
Give that a try.

PinpointTownes:
@Tratcher thanks for the tip! :smile: 

Sadly, I can't test it locally as the xUnit runner no longer works on my machine, even with the latest packages/DNX runtime. Let's see what the CI says :smile:

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -177,7 +172,12 @@ public OAuthHandler(HttpClient backchannel)
                 throw new ArgumentNullException(nameof(context));
             }
 
-            var properties = new AuthenticationProperties(context.Properties);
+            var properties = new AuthenticationProperties(context.Properties)
+            {
+                // Review: should we add an option to configure the state lifetime?
+                ExpiresUtc = Options.SystemClock.UtcNow.AddMinutes(15)
```
PinpointTownes: So, should we make it configurable?

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
PinpointTownes: Options:
* Change the way we generate the cookie to store the generated bits in the cookie value instead of the name (or a few bits in the name and the rest in the value).
* Remove the check, as it's basically useless.
* Move the log message to the appropriate place (i.e L. 158)

@Tratcher @Eilon any preference?

Tratcher: - Don't change the name/format.
- The check is valid, though I don't know what would ever cause the value to be wrong. It's closer to an assert. 
- The log on 158 is adequate for that check.
- Just reword this log to say "The correlation cookie value '{0}' did not match the expected value '{1}'".

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
Tratcher: Yes. It didn't used to expire at all before. `TimeSpan? RoundTripTimeout { get; set; } = TimeSpan.FromMinutes(15);`?

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -2,15 +2,24 @@
...
 using Microsoft.AspNetCore.Http.Features.Authentication;
+using Microsoft.AspNetCore.Http.Authentication;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Authentication
 {
     public abstract class RemoteAuthenticationHandler<TOptions> : AuthenticationHandler<TOptions> where TOptions : RemoteAuthenticationOptions
     {
+        private const string CorrelationPrefix = ".AspNetCore.Correlation.";
+        private const string CorrelationProperty = ".xsrf";
```
HaoK: could make this internal so you don't have so many hardcoded .xsrf's in the tests

PinpointTownes: Yeah, but then we'd have to move these constants to a different class, as this one is generic (it's rather weird/ugly/cumbersome with generics: `RemoteAuthenticationHandler<RemoteAuthenticationOptions>.CorrelationProperty` :trollface:)

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
PinpointTownes: The property name is rather clear (though maybe not specific enough?), but I'm not sure we should make it nullable (which would result in a session cookie). At least, not without a warning indicating that setting it to null may result in a terrible cookie flooding :smile:

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
HaoK: Don't we have a Constants internal class already?
>

PinpointTownes: AFAICT, nope, no constants class for `Microsoft.AspNetCore.Authentication` (the OAuth2 project has one but this PR removes it as it's now useless).

If you think it's worth it, I'm fine creating one and marking `Microsoft.AspNetCore.Authentication` as a friend assembly to make it accessible from the unit tests project :smile:

HaoK: Nah if we don't have a constants class its not worth adding it for this, hardcoding is fine :)

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
Tratcher: There wasn't at timeout before in OAuth, but I guess there isn't really a scenario for long timeouts here.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -184,7 +184,10 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
             // order for local RedirectUri
             // 1. challenge.Properties.RedirectUri
             // 2. CurrentUri if RedirectUri is not set)
-            var properties = new AuthenticationProperties(context.Properties);
+            var properties = new AuthenticationProperties(context.Properties)
+            {
+                ExpiresUtc = Options.SystemClock.UtcNow.Add(Options.CorrelationLifetime)
```
Tratcher: There's no reason for the correlation lifetime to be independent of the nonce lifetime is there?

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationOptions.cs
```diff
@@ -56,6 +56,11 @@ public string DisplayName
         /// </summary>
         public bool SaveTokensAsClaims { get; set; }
 
+        /// <summary>
+        /// Gets or sets the lifetime of the correlation cookie used to prevent cross-site scripting attacks.
```
Tratcher: Yes, but what this really means is that this is the time limit for completing the authentication flow.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: Ah, you put CorrelationLifetime on RemoteAuthenticationOptions rather than just OAuthOptions. Hmm..

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationOptions.cs
Tratcher: That's why I suggested the ridiculous name that I did. Any ideas for a property name that conveys this? There's no reason for this timeout usage to be restricted to the correlation cookie. Some derivations of RemoteAuthenticationHandler might use mechanics other than correlation cookies. E.g. how about applying this to Twitter's State cookie?

HaoK: What was the name you suggested?  Maybe something generic like RemoteAuthenticationTimeout then?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
PinpointTownes: Yep.

In practice, I can't think of a scenario where you'd like to have different lifetimes for the nonce cookie and the correlation cookie (aka `state`).

That said, `state` and `nonce` are really two different things and `nonce` is an OIDC-specific thing, while the antiforgery/correlation cookie is more general and can be applied to any protocol (even to OpenID2, that doesn't support the `state` parameter). In OIDC, using a `nonce` is not mandatory with the code flow and - unlike the correlation check - can be disabled.

So maybe it's a good thing to have a different option for the antiforgery cookie.

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationOptions.cs
PinpointTownes: If you think `RAT` is clearer, I'm fine :smile: 

I updated the Twitter handler to use the new property. TBH, I'd really like to update it to also use `GenerateCorrelationId`/`ValidateCorrelationId` like the other providers (instead of the custom `RequestToken` and the associated serializer), but I guess it's a bit late for that.

Tratcher: `RemoteAuthenticationTimeout` is good enough.

Yeah, now's not the time to try re-writing Twitter. I think most of that junk was required as part of the OAuth1 or Twitter spec anyways.  Adding the timeout should be sufficient.

PinpointTownes: Updated to use the name you suggested :clap:

Tratcher:
:shipit: Squash and rebase?

PinpointTownes:
@Tratcher sure. I'll run a few tests to ensure nothing breaks and squash it :smile: 

Any idea why the CI builds fail so hard? I guess it's caused by the DNX -> CLI migration?

HaoK:
:fire:  :fire: :fire: :fire: :fire: That pretty much sums up what running tests locally looks like right now as well :fire: :fire:

Tratcher:
Yeah, the CLI conversion is causing lots of builds to fail. I can still test locally running build.cmd. Just squash and re-base and I'll run a final test pass.

PinpointTownes:
Rebased and squashed :smile: 

I tried the social sample with the Google and GitHub and the OIDC sample with OpenIddict and both worked as intended. I was not able to run the tests locally, tho.

Tratcher:
Tested, rebased, and merged.

PR #671, opened 2016-01-25
Remove unused dependency

davidfowl:
- The authorization stack doesn't depend on AspNetCore at all really

This brings up an interesting question, why is it tied to ASP.NET at all? Shouldn't the AuthorizationContext flow the HttpContext as well? 

/cc @HaoK @blowdart @Tratcher @lodejard

### DIFF
--- a/src/Microsoft.AspNetCore.Authorization/project.json
+++ b/src/Microsoft.AspNetCore.Authorization/project.json
@@ -10,12 +10,15 @@
     "keyFile": "../../tools/Key.snk"
   },
   "dependencies": {
-    "Microsoft.AspNetCore.Http.Features": "1.0.0-*",
     "Microsoft.Extensions.Logging.Abstractions": "1.0.0-*",
     "Microsoft.Extensions.Options": "1.0.0-*"
   },
   "frameworks": {
-    "net451": {},
-    "dotnet5.4": {}
+    "net451": { },
+    "dotnet5.4": {
+      "dependencies": {
+        "System.Security.Claims": "4.0.1-*"
+      }
+    }
   }
 }


blowdart:
Design decision. We really want to underscore you make authn decisions based on an identity, nothing more (if you can).

leastprivilege:
If you need the context - you could inject it.

blowdart:
Well you don't need to inject it, MVC will add it as part of it's authorization context, it's just not part of the base one.

davidfowl:
> If you need the context - you could inject it.

Yea, but using the IHttpContextAccessor is kinda gross. It also implies that you could use this outside of http requests and the "current" http context would be null in this case. Also, it's not added by default in RC2, it needs to be added to the container manually (and it's not free either, there is a performance slight penalty when you using it).

> Well you don't need to inject it, MVC will add it as part of it's authorization context, it's just not part of the base one.

MVC won't be the only thing using this system, also MVC passes its `AuthorizationContext` as the resource.

Bottom line is that there's no real **AspNet** in this package name. Maybe it should be **Microsoft.Extensions.Authorization** (but then again who would want to use this in a console application :wink: )

leastprivilege:
Yea, but using the IHttpContextAccessor is kinda gross.

How would I get to the context and the current User when using Mvc.Core (e.g. for web apis)

davidfowl:
> How would I get to the context and the current User when using Mvc.Core (e.g. for web apis). Since there is no controller base class.

There's 2 options:

1. Full Poco

```C#
public class MyController
{
   [ActionContext]
   public ActionContext ActionContext { get; set; }
  
   public HttpContext => ActionContext.HttpContext;
}
```

1b (This might have changed in RC2)


```C#
public class MyController
{
   [ActionContext]
   public ControllerContext ControllerContext { get; set; }
  
   public HttpContext => ControllerContext.HttpContext;
}
```

2. Use the new lightweight controller base.

```C#
public class MyController : ControllerBase
{
   ...
}
```

leastprivilege:
ok. thanks!

leastprivilege:
> Bottom line is that there's no real AspNet in this package name. Maybe it should be Microsoft.Extensions.Authorization (but then again who would want to use this in a console application :wink: )

Well - aren't asp.net applications all console applications ;) ??

I don't really care in which namespace or assemly the API lives - but yes - strictly speaking this is asp.net independent.

davidfowl:
> Well - aren't asp.net applications all console applications ;) ??

Yes, but not all console applications are ASP.NET applications.

> I don't really care in which namespace or assemly the API lives - but yes - strictly speaking this is asp.net independent.

It's not though.

leastprivilege:
So why are we having this discussion?

davidfowl:
The discussion is on the wrong thing, it should be on this issue https://github.com/aspnet/Security/issues/726 :wink:

PR #672, opened 2016-01-25
Add default fields to Facebook

mikes-gh:
- Remove username field  from helper and handler as this is no longer returned by the Facebook API as of v 2.0 onwards.

see https://developers.facebook.com/docs/apps/changelog#v2_0_graph_api

- Add a set of default fields to the request if none are supplied.  This gives parity with Google.

This means we can simply use
```
    app.UseFacebookAuthentication(new FacebookOptions
            {
                AppId = "122334455666";
                AppSecret = "9876543";
            });
```

And be done.


mikes-gh:
cc @Tratcher

mikes-gh:
OK I just saw I removed email by mistake in first commit.

Issue #670, opened 2016-01-24
Authorization infrastructure does not handle "per action permissions" use case well


kwaclaw:
In WebApi 2 we used an AuthorizationFlterAttribute to implement the following permissions based concept:
* Each controller action has a permission code associated.
* Roles are associated with a set of permissions. This set can change over time.
* A user has one or more roles.
* Authorization is determined by checking if any of the user's roles is associated with the action's permission code.

This was easy to use by the developer. He/she only had to decorate an action with the corresponding attribute.

It seems the new policy based approach in ASP.NET 5 makes it hard to implement his concept, because there is no way for the Handle method of the AuthorizationHandler to include the permission code in its arguments. The only way I could solve this problem was to create one new policy for each unique permission code. Since there is a large number of actions, I now have to create an equally large number of policies. It would not be a huge change to the infrastructure to be able to attach extra information to the attribute and pass it through to the handler in the AuthorizationContext, or even pass the attribute through itself. For example, the handler code could then look like this:

	protected override void Handle(AuthorizationContext context, PermissionRequirement requirement) {
		var userRoles = context.User.FindAll(System.Security.Claims.ClaimTypes.Role);
		var permissionsAttribute = context.Attribute as PermissionAttribute;
		if (permissionsAttribute == null)
			return;
		if (myPermissionsChecker.RolesHavePermission(userRoles, permissionsAttribute.Permission)) {
			context.Succeed(requirement);
		}
	}

This would only require a single policy, and the developer would not be burdened with having to create a new policy everytime a new controller action is added.

brockallen:
This might help (in RC2): https://github.com/aspnet/Security/issues/630

kwaclaw:
From where is the resource picked up? I don't see a change to the AuthorizeAttribute to include  a Resource property.

brockallen:
You would pass it explicitly. It's the imperative approach. You're right that it's harder to from declaratively.

kwaclaw:
I see. I would prefer the declarative option though.

davidfowl:
When you're using MVC, the resource by default is the filter's authorization context. You can use that to get information about the action being executed and that's how you can discover the attribute:

```C#
public class Startup
{
    public void ConfigureServices(IServiceCollection services)
    {
        services.AddMvc();

        services.AddAuthorization(options =>
        {
            options.AddPolicy("Permissions", builder =>
            {
                builder.Requirements.Add(new PermissionsRequirement());
            });
        });

[...]

PR #674, opened 2016-01-25
Add IAuthorizationPolicyProvider

HaoK:
Allows for dynamic policy creation
(Should we rename AuthorizationOptions => AuthorizationPolicy[Provider]Options now that its moved into a different service?)

See https://github.com/aspnet/Security/issues/670

cc @davidfowl @blowdart @divega

### DIFF
--- a/src/Microsoft.AspNetCore.Authorization/IAuthorizationPolicyProvider.cs
+++ b/src/Microsoft.AspNetCore.Authorization/IAuthorizationPolicyProvider.cs
@@ -1,0 +1,18 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Authorization
+{
+    /// <summary>
+    /// A type which can provide a <see cref="AuthorizationPolicy"/> for a particular name.
+    /// </summary>
+    public interface IAuthorizationPolicyProvider
+    {
+        /// <summary>
+        /// Gets a <see cref="AuthorizationPolicy"/> from the given <paramref name="policyName"/>
+        /// </summary>
+        /// <param name="policyName"></param>
+        /// <returns></returns>
+        AuthorizationPolicy GetPolicy(string policyName);
+    }
+}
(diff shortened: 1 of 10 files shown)


blowdart:
AuthorizationPolicyOptions sounds fine to me, but I'll need you to help me doc the changes so we can have them available when RC2 drops.

HaoK:
Ok AuthorizationOptions => AuthorizationPolicyOptions

PinpointTownes:
nit: `AuthorizationPolicyOptions` is fine, but not consistent with the CORS options:  https://github.com/aspnet/CORS/blob/dev/src/Microsoft.AspNetCore.Cors/CorsOptions.cs#L12 :smile:

src/Microsoft.AspNetCore.Authorization/IAuthorizationPolicyProvider.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// <summary>
+    /// A type which can provide a <see cref="AuthorizationPolicy"/> for a particular name.
+    /// </summary>
+    public interface IAuthorizationPolicyProvider
+    {
+        /// <summary>
+        /// Gets a <see cref="AuthorizationPolicy"/> from the given <paramref name="policyName"/>
+        /// </summary>
+        /// <param name="policyName"></param>
+        /// <returns></returns>
```
davidfowl: Should probably be async.

davidfowl:
Another crazy question, why not support multiple?

HaoK:
@davidfowl RE multiple providers @divega brought this up yesterday too when we reviewed this,  what would be a rationale behavior, return the first hit?  We felt multiple providers aren't really necessary since its pretty easy to extend/reuse/wrap the default implementation and use it as a fallback.

@PinpointTownes RE AuthorizationPolicyOptions, I wouldn't be surprised if the CorsOptions was named that way due to the fact following how AuthorizationOptions was named :), so there is a slight consistency issue, but I don't think that's enough to reason to not clarify the name (assuming we all agree its moar better)

src/Microsoft.AspNetCore.Authorization/IAuthorizationPolicyProvider.cs
HaoK: Yeah I saw Cors was async, sure I will update to async

davidfowl:
> @PinpointTownes RE AuthorizationPolicyOptions, I wouldn't be surprised if the CorsOptions was named that way due to the fact following how AuthorizationOptions was named :), so there is a slight consistency issue, but I don't think that's enough to reason to not clarify the name (assuming we all agree its moar better)

Not sure, we may want to expand the options on there one day as we add new features. It might not be policy specific. The PolicyMap might just be one of the things on that type. I wouldn't rename it.

HaoK:
Fair enough, reverting back to AuthorizationOptions and I'll merge

HaoK:
00ceeb026262bc848404d056416709dc306a57df

PR #675, opened 2016-01-25
Add async overload for AuthZ RequireAssertion

HaoK:
cc @divega @davidfowl @blowdart


HaoK:
80dc5759cc1f9cd447724ff068cc8253fe7b2b1c

PR #677, opened 2016-01-26
Add core fields to Facebook

mikes-gh:
- Remove username field  from helper and handler.

This field is no longer returned by the Facebook API as of v 2.0 onwards.
see https://developers.facebook.com/docs/apps/changelog#v2_0_graph_api

- Add a set of default fields to the request if none are supplied. 

This gives parity with Microsoft.AspNetCore.Authentication.Google.
It allows the most common scenario with minimal configuration which will greatly benefit new users.
The change also has no impact on existing code.

Using the following minimal configuration
```
    app.UseFacebookAuthentication(new FacebookOptions
            {
                AppId = "122334455666";
                AppSecret = "9876543";
            });
```

Gives the following results from the Social Sample
```
Hello Joe Blogs
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier: 999999999999
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress: joe@blogs.co.uk
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name: Joe Blogs
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname: Joe
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/surname: Blogs
urn:facebook:link: https://www.facebook.com/app_scoped_user_id/1234567678/
Logout
```

cc @Tratcher @HaoK

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
@@ -60,6 +60,18 @@ namespace Microsoft.AspNetCore.Authentication.Facebook
                 identity.AddClaim(new Claim(ClaimTypes.Name, name, ClaimValueTypes.String, Options.ClaimsIssuer));
             }
 
+            var givenName = FacebookHelper.GetGivenName(payload);
+            if (!string.IsNullOrEmpty(givenName))
+            {
+                identity.AddClaim(new Claim(ClaimTypes.GivenName, givenName, ClaimValueTypes.String, Options.ClaimsIssuer));
+            }
+
+            var surname = FacebookHelper.GetFamilyName(payload);
+            if (!string.IsNullOrEmpty(surname))
+            {
+                identity.AddClaim(new Claim(ClaimTypes.Surname, surname, ClaimValueTypes.String, Options.ClaimsIssuer));
+            }
+
             var link = FacebookHelper.GetLink(payload);
             if (!string.IsNullOrEmpty(link))
             {
--- a/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
@@ -12,6 +12,7 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http.Authentication;
 using Microsoft.AspNetCore.WebUtilities;
 using Newtonsoft.Json.Linq;
+using System.Linq;
 
 namespace Microsoft.AspNetCore.Authentication.Facebook
 {
@@ -31,7 +32,7 @@ namespace Microsoft.AspNetCore.Authentication.Facebook
             }
             if (Options.Fields.Count > 0)
             {
-                endpoint = QueryHelpers.AddQueryString(endpoint, "fields", string.Join(",", Options.Fields));
+                endpoint = QueryHelpers.AddQueryString(endpoint, "fields", string.Join(",", Options.Fields.Distinct()));
             }
 
             var response = await Backchannel.GetAsync(endpoint, Context.RequestAborted);
--- a/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
@@ -48,12 +48,6 @@ namespace Microsoft.AspNetCore.Authentication.Facebook
                 identity.AddClaim(new Claim(ClaimTypes.NameIdentifier, identifier, ClaimValueTypes.String, Options.ClaimsIssuer));
             }
 
-            var userName = FacebookHelper.GetUserName(payload);
-            if (!string.IsNullOrEmpty(userName))
-            {
-                identity.AddClaim(new Claim(ClaimsIdentity.DefaultNameClaimType, userName, ClaimValueTypes.String, Options.ClaimsIssuer));
-            }
-
             var email = FacebookHelper.GetEmail(payload);
             if (!string.IsNullOrEmpty(email))
             {
@@ -63,13 +57,7 @@ namespace Microsoft.AspNetCore.Authentication.Facebook
             var name = FacebookHelper.GetName(payload);
             if (!string.IsNullOrEmpty(name))
             {
-                identity.AddClaim(new Claim("urn:facebook:name", name, ClaimValueTypes.String, Options.ClaimsIssuer));
-
-                // Many Facebook accounts do not set the UserName field.  Fall back to the Name field instead.
-                if (string.IsNullOrEmpty(userName))
-                {
-                    identity.AddClaim(new Claim(identity.NameClaimType, name, ClaimValueTypes.String, Options.ClaimsIssuer));
-                }
+                identity.AddClaim(new Claim(ClaimTypes.Name, name, ClaimValueTypes.String, Options.ClaimsIssuer));
             }
 
             var link = FacebookHelper.GetLink(payload);
--- a/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
@@ -49,10 +49,70 @@ namespace Microsoft.AspNetCore.Authentication.Facebook
                 identity.AddClaim(new Claim(ClaimTypes.NameIdentifier, identifier, ClaimValueTypes.String, Options.ClaimsIssuer));
             }
 
+            var ageRangeMin = FacebookHelper.GetAgeRangeMin(payload);
+            if (!string.IsNullOrEmpty(ageRangeMin))
+            {
+                identity.AddClaim(new Claim("urn:facebook:age_range_min", ageRangeMin, ClaimValueTypes.String, Options.ClaimsIssuer));
+            }
+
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Authentication.Facebook/FacebookMiddleware.cs
```diff
@@ -71,10 +71,30 @@ public class FacebookMiddleware : OAuthMiddleware<FacebookOptions>
             {
                 throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(Options.AppId)));
             }
+
             if (string.IsNullOrEmpty(Options.AppSecret))
             {
                 throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.Exception_OptionMustBeProvided, nameof(Options.AppSecret)));
             }
+
+            if (Options.Scope.Count == 0)
```
Tratcher: These should both be updated to use the new pattern from https://github.com/aspnet/Security/pull/680

mikes-gh:
@Tratcher @HaoK
Should be good to go now.
All tested using social sample.

Tratcher:
This is going in the right direction.  It's good to support more fields if they're returned and I think we should support all of the `core` fields defined in https://developers.facebook.com/docs/graph-api/reference/user.

That said, I don't think most of these should be enabled by default. Adding support in FacebookHandler and FacebookHelper removes most of the developer burden, but they should set their scopes and fields themselves. Restrict the default scope to `email` and fields to `name` and `email`.

mikes-gh:
Sure no problem?

I think we should keep public_profile though

This is what Facebook says about it

-On the web, public_profile is implied with every request and isn't required, although the best practice is to declare it. On iOS and Android, you must manually request it as part of your login flow.

see
https://developers.facebook.com/docs/facebook-login/permissions#reference-public_profile

mikes-gh:
Actually name is part of public_profile so we definitely need it.

Tratcher:
Fair.

mikes-gh:
Im going to break this down into 3 PRs.

-Remove username field from Facebook
-Support all the core Facebook fields
-Add default scopes and fields to Facebook.

Default scopes email and public_profile
Default fields email and name
This current PR is getting a bit messy now.

Does that sound good to you before I start?

Tratcher:
No need to create separate PRs, you can just keep pushing to this branch to update this one.

mikes-gh:
OK will do

mikes-gh:
Hopefully this should be good to go

Social sample tests
```
       app.UseFacebookAuthentication(new FacebookOptions
            {
                AppId = Configuration["facebook:appid"],
                AppSecret = Configuration["facebook:appsecret"],
            });
```
gives

```
Hello Joe Peter Blogs
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier: 99999999999999
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress: joe@blogs.co.uk
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name: Joe Peter Blogs
Logout
```

``` 
          app.UseFacebookAuthentication(new FacebookOptions
            {
                AppId = Configuration["facebook:appid"],
                AppSecret = Configuration["facebook:appsecret"],
                Scope = {"user_birthday","user_location"},
                Fields = { "age_range", "birthday","email", "first_name", "gender", "last_name", "link", "location","locale", "middle_name" , "name" ,"timezone" } 
            });
```

gives

```
Hello Joe Peter Blogs
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier: 999999999999
urn:facebook:age_range_min: 21
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/dateofbirth: 01/01/56
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress: joe@blogs.co.uk
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname: Joe
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/gender: male
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/surname: Blogs
urn:facebook:link: https://www.facebook.com/app_scoped_user_id/9999999999999/
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/stateorprovince: London, United Kingdom
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/locality: en_GB
urn:facebook:middle_name: Peter
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name: Joe Peter Blogs
urn:facebook:timezone: 0
Logout
```

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
```diff
@@ -31,7 +32,7 @@ public FacebookHandler(HttpClient httpClient)
             }
             if (Options.Fields.Count > 0)
             {
-                endpoint = QueryHelpers.AddQueryString(endpoint, "fields", string.Join(",", Options.Fields));
+                endpoint = QueryHelpers.AddQueryString(endpoint, "fields", string.Join(",", Options.Fields.Distinct()));
```
Tratcher: If you're going to do options validation like calling `Distinct()` it should be done once in the middleware constructor, not on every request.

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
```diff
@@ -78,6 +97,36 @@ public FacebookHandler(HttpClient httpClient)
                 identity.AddClaim(new Claim("urn:facebook:link", link, ClaimValueTypes.String, Options.ClaimsIssuer));
             }
 
+            var location = FacebookHelper.GetLocation(payload);
+            if (!string.IsNullOrEmpty(location))
+            {
+                identity.AddClaim(new Claim(ClaimTypes.StateOrProvince, location, ClaimValueTypes.String, Options.ClaimsIssuer));
```
Tratcher: StateOrProvince? I don't think you have any guarantee about what's in this field.

Tratcher:
:+1: Only minor comments left.

mikes-gh:
I double checked and it does fall over for duplicate fields

Options is read only in Middleware constructor so I have used a HashSet for the fields.
That obviously only implements ICollection but the lack of an indexer shouldn't be an issue here.

However I am open to suggestions if you don't feel this is appropriate.

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookOptions.cs
```diff
@@ -57,6 +60,6 @@ public string AppSecret
         /// The list of fields to retrieve from the UserInformationEndpoint.
         /// https://developers.facebook.com/docs/graph-api/reference/user
         /// </summary>
-        public IList<string> Fields { get; }
+        public ICollection<string> Fields { get; } = new HashSet<string>();
```
PinpointTownes: Nice! :smile: 

Should we also update `OAuthOptions.Scope`/`OpenIdConnectOptions.Scope` to use a set instead of a good old list? (and why not using `ISet<string>`?)

mikes-gh: Do you see senarios other than add remove count for each?
IColection seems to provide enough but I'll see what @Tratcher thinks. It's an easy change.
Re scope yes I agree but you'll have to have that discussion with @Tratcher

PinpointTownes: No particular scenario in mind. But exposing `ISet` in the public contract makes it clear that the collection cannot have duplicate items (which is not something `IList`/ `ICollection` imply).

mikes-gh: Let's see what @Tratcher  has to say.  I think your right though ISet<T> would give a better indication to the user and if we are happy that it will always be a hash set then that's ok..

Tratcher: The only difference with IList was that it guaranteed ordering, though I don't know if that will matter to any of the providers.

As a guideline an API should return the most general thing it can so that in the future you can return something more specific without breaking people. In this case I'd recommend ICollection over ISet as you don't actually need any of the ISet APIs, you just want it for descriptiveness sake. Doc comments should be sufficient explanation for now.

You can do Scope too if you want, but start a separate PR since it doesn't overlap with any of this work.

Tratcher:
:shipit: I'll merge it tomorrow.

mikes-gh:
:smile: Appreciate your patience.  I noticed I was pushing to dev :unamused: so Ill wait for the merge then do the scope changes in new PR once I've sorted out my branches.

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookOptions.cs
PinpointTownes: > The only difference with IList was that it guaranteed ordering, though I don't know if that will matter to any of the providers.

[The order doesn't matter](https://tools.ietf.org/html/rfc6749#section-3.3) and in practice, I've never seen a provider breaking the spec here (well, except for the separator... :smile:):

> The value of the scope parameter is expressed as a list of space-delimited, case-sensitive strings. The strings are defined by the authorization server.  If the value contains multiple space-delimited strings, **their order does not matter**, and each string adds an additional access range to the requested scope.

Tratcher:
Rebased, squashed, and merged.

Issue #614, opened 2015-12-13
Discussion for Scope being a list and not a string?


brockallen:
Was there ever any discussion as to why `Scope` on the OIDC middleware is a collection and not a string? `ResponseType` is still a string. It feels cumbersome. Anyone else feel the same way?

PinpointTownes:
No discussion, but since [I'm the guy who replaced the `Scope` property by a list] (https://github.com/aspnet/Security/pull/276/files#r31257639), I can shed some light on the subject :smile: 

The most important reason was for consistency: the OAuth2 generic middleware used a list of strings and the OIDC client middleware a single string. I opted for `IList<string>` because it was the most appropriate choice for the OAuth2 generic middleware, whose inheritors have to deal with non-standard implementations (Facebook, we hate you for using coma-separated scopes... :smile:). Using a list hides how the `scope` is finally formatted by the specific implementation and make the way users deal with `scope` consistent across the social providers.

brockallen:
Well, screw all of that, IMO. The katana OAuth2 MW is dead. And you know exactly what the OIDC and OAuth2 specs say for the scope. If the aberrant implementations needs something different, then do it in their own implementation. 

The job of frameworks is to make the life of the consumer easier, not the life of the implementer.

PinpointTownes:
> Well, screw all of that, IMO. The katana OAuth2 MW is dead.

There was no OAuth2 generic middleware in Katana, so not sure to understand what you mean.

> And you know exactly what the OIDC and OAuth2 specs say for the scope.

Both specs only state that `scope` is space-separated string, they don't mention anything about how we should implement it internally.

> If the aberrant implementations needs something different, then do it in their own implementation.

That's exactly what we do in the virtual `FormatScope` method. But it seems you totally missed my point. If we used a list, consumers would have to use different scopes patterns for Facebook and Google:

```csharp
options.Scope = "scope1,scope2"; // Facebook, Reddit.
options.Scope = "scope1 scope2"; // Google (and the other standard providers)
```

Using a list abstracts the way the scope is formatted:

```csharp
options.Scope.Add("scope1");
options.Scope.Add("scope2");
```

brockallen:
> There was no OAuth2 generic middleware in Katana, so not sure to understand what you mean.

I was referring to the Katana OAuth2 authorization server middleware. Is that not what you mentioned above when you said that the point was for consistency? Perhaps we're not talking about the same thing.

Other complaints:

[...]

PR #680, opened 2016-01-27
Standardize the pattern for scope defaults.

Tratcher:
#614 @PinpointTownes @brockallen @HaoK @Eilon 

Google and MSA require scopes by default so we provide a few. However they also send an ok error message if you clear the default scopes so we're not checking for missing scopes in the constructor.

OIDC already works like this.


HaoK:
:shipit:

PinpointTownes:
:shipit:

Issue #92, opened 2014-11-13
Verify external providers are using latest APIs


Tratcher:
RTM milestone.

External login providers routinely update their APIs and deprecate the old ones. Before we ship RTM we need to verify that each of our implementations is targeting the latest provider API versions to ensure our packages have the longest useful lifespan.

Example: #85

brentschmaltz:
What is the definition for an 'External login provider'?

Praburaj:
Like Google, Facebook, Twitter, MicrosoftAccount etc.

Tratcher:
FYI: Facebook is up to v2.4: https://developers.facebook.com/blog/post/2015/07/08/graph-api-v2.4/
https://katanaproject.codeplex.com/workitem/417
https://github.com/aspnet/Security/issues/365

muratg:
We'll attempt to do this during RC1 sign-offs.

muratg:
Let's have a look in RC2.

Tratcher:
MSA: https://github.com/aspnet/Security/pull/691

HaoK:
https://github.com/aspnet/Security/pull/829

HaoK:
FB to v2.6 https://github.com/aspnet/Security/pull/828

HaoK:
Twitter looks fine

Tratcher:
Reviving for the next release.

HaoK:
Its the zombie issue, that will never die...

Tratcher:
*ping* @muratg @Eilon reviving this for 2.0 signoff.

Eilon:
@Tratcher - anything we need to do for 2.0 RTM?

Tratcher:
I did a pass recently and filed some non-blocking bugs linked above. Move this to 2.1?

Eilon:
Ok.

blowdart:
Bringing this front and center. Facebook is now at 2.12 and 2.6 will expire in July this year.

Tratcher:
See https://github.com/aspnet/Security/issues/1306 for Facebook

Tratcher:
Facebook and Google fixed for 2.1. Twitter and MSA do not appear to require any updates. Punting this zombie issue to 2.2.0.

PR #691, opened 2016-02-08
Update MicrosoftAccount to use graph API & converged authentication

OsmanMElsayed:
- Time has come to update `Microsoft.AspNetCore.Authentication.MicrosoftAccount` package to target *Microsoft graph API* as *Live Connect API* & *Identity API* are [depreciated](https://msdn.microsoft.com/en-us/library/hh826532.aspx). For this, we need to issue tokens from the **app model v2.0** endpoints which will extend the reach of the package to all **work/school accounts** which is clearly a big added value.

- App model v2.0 has been in public preview since Aug 2015. AFAIK the GA is very close & it is expected to be in Feb & there are [no expected breaking changes](https://azure.microsoft.com/en-us/documentation/articles/active-directory-v2-preview-oidc-changes/#how-often-will-protocol-changes-occur)

- Unfortunately, existing Microsoft account apps are not compatible with app model v2.0 & a new app registration is required. **That's why this update is considered a breaking change**. Check [Restrictions on App  #Registrations](https://azure.microsoft.com/en-us/documentation/articles/active-directory-v2-limitations/#restrictions-on-app-registrations)

#92


Tratcher:
Very promising.

@blowdart Are you aware of this transition for MSA?

blowdart:
Yup. We ought to run this past @brentschmaltz et al.

samples/SocialSample/Startup.cs
```diff
@@ -17,9 +17,14 @@
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using Microsoft.AspNetCore.Server.Kestrel.Filter;
 using Newtonsoft.Json.Linq;
+using System.Security.Cryptography.X509Certificates;
```
Tratcher: These need sorting, with System on top.

samples/SocialSample/project.json
```diff
@@ -8,9 +8,11 @@
     "Microsoft.AspNetCore.DataProtection": "1.0.0-*",
     "Microsoft.AspNetCore.IISPlatformHandler": "1.0.0-*",
     "Microsoft.AspNetCore.Server.Kestrel": "1.0.0-*",
-    "Microsoft.Extensions.Configuration.UserSecrets": "1.0.0-*",
+    "Microsoft.AspNetCore.Server.Kestrel.Https": "1.0.0-*",
+    "Microsoft.Extensions.Configuration.UserSecrets": "1.0.0-*",
     "Microsoft.Extensions.Logging.Console": "1.0.0-*",
-    "Microsoft.NETCore.Platforms": "1.0.1-*"
+    "Microsoft.NETCore.Platforms": "1.0.1-*",
+    "Microsoft.AspNet.FileProviders.Embedded":  "1.0.0-*"
```
Tratcher: sort

src/Microsoft.AspNetCore.Authentication.MicrosoftAccount/MicrosoftAccountHelper.cs
```diff
@@ -34,8 +35,17 @@ public static string GetName(JObject user)
...
                 throw new ArgumentNullException(nameof(user));
             }
+            
+            //displayName is required for Azure AD accounts, but not for Microsoft accounts.
+            //givenName & surename is required for Microsoft accounts, but not for Azure AD accounts
+            //In case of microsoft accounts displayName is null => fallback to givenName + surename
+            var name = user.Value<string>("displayName");
+            if (string.IsNullOrEmpty(name))
+            {
+                name = string.Format("{0} {1}", GetFirstName(user), GetLastName(user));
```
Tratcher: Remove this. We don't want to transform the claims we receive, just pass them through. It's up to apps to make these kind of transformation decisions, especially for something like this can can be culture sensitive.

samples/SocialSample/Startup.cs
```diff
@@ -128,44 +137,32 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
...
                 AuthenticationScheme = "Microsoft-AccessToken",
-                DisplayName = "MicrosoftAccount-AccessToken - Requires project changes",
+                DisplayName = "MicrosoftAccount-AccessToken",
                 ClientId = Configuration["msa:clientid"],
                 ClientSecret = Configuration["msa:clientsecret"],
                 CallbackPath = new PathString("/signin-microsoft-token"),
                 AuthorizationEndpoint = MicrosoftAccountDefaults.AuthorizationEndpoint,
                 TokenEndpoint = MicrosoftAccountDefaults.TokenEndpoint,
-                Scope = { "wl.basic" },
+                Scope = { "openid", "https://graph.microsoft.com/user.read" },
```
Tratcher: why use `openid` and `user.read` here, but only `user.read` in the actual middleware defaults?

src/Microsoft.AspNetCore.Authentication.MicrosoftAccount/MicrosoftAccountHelper.cs
OsmanMElsayed: Yea this is just ugly.
Will add the `GivenName` & `Surname` claims to `ClaimsIdentity` in the next commit.
btw, what is the value of having the `urn:microsoftaccount:*` claims?

samples/SocialSample/Startup.cs
OsmanMElsayed: Yup this has nothing to do with openid stuff. `user.read` is quite enough.
The same applies to https://github.com/aspnet/Security/blob/dev/samples/SocialSample/Startup.cs#L92 ?

Tratcher:
@OsmanMElsayed did you test this with an outlook.com account? I can't get it working, it keeps complaining about redirect uri mismatches no matter how I configure it.

OsmanMElsayed:
@Tratcher I tried it with outlook.com account & it works fine.
outlook-account

OsmanMElsayed:
Hooray!
[app model v2 is GA](https://blogs.technet.microsoft.com/ad/2016/02/23/for-developers-the-first-use-cases-of-the-converged-microsoft-account-and-azure-active-directory-programming-model-are-now-ga/)

Tratcher:
Rebased and merged. The problem I ran into was because my old local usersecrets were overwriting the config.json so I was using the wrong client id.

PR #692, opened 2016-02-11
Integrating Antiforgery with Authentication

rynowak:
This is a rough-cut at what auth+antiforgery might look like.

Some integration tests included as a POC.

I'm not super familiar with the design in this area, so any and all suggestions are great.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -6,6 +6,7 @@ using System;
 using System.Linq;
 using System.Security.Claims;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Antiforgery;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Authentication;
@@ -22,11 +23,18 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
         private const string HeaderValueMinusOne = "-1";
         private const string SessionIdClaim = "Microsoft.AspNetCore.Authentication.Cookies-SessionId";
 
+        private readonly IAntiforgery _antiforgery;
+
         private bool _shouldRefresh;
         private DateTimeOffset? _refreshIssuedUtc;
         private DateTimeOffset? _refreshExpiresUtc;
         private string _sessionKey;
         private Task<AuthenticateResult> _readCookieTask;
+
+        public CookieAuthenticationHandler(IAntiforgery antiforgery)
+        {
+            _antiforgery = antiforgery;
+        }
 
         private Task<AuthenticateResult> EnsureCookieTicket()
         {
@@ -113,12 +121,32 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
             return AuthenticateResult.Success(ticket);
         }
 
+        private bool RequiresAntiforgery()
+        {
+            if (Options.SuppressAntiforgeryValidation)
+            {
+                return false;
+            }
+
+            var method = Context.Request.Method;
+            return
+                !string.Equals(method, "GET", StringComparison.OrdinalIgnoreCase) &&
+                !string.Equals(method, "HEAD", StringComparison.OrdinalIgnoreCase) &&
+                !string.Equals(method, "OPTIONS", StringComparison.OrdinalIgnoreCase) &&
+                !string.Equals(method, "TRACE", StringComparison.OrdinalIgnoreCase);
+        }
+
         protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
         {
             var result = await EnsureCookieTicket();
             if (!result.Succeeded)
             {
                 return result;
+            }
+
+            if (RequiresAntiforgery() && !await _antiforgery.IsRequestValidAsync(Context))
+            {
+                return AuthenticateResult.Fail("No antiforgery token");
             }
 
             var context = new CookieValidatePrincipalContext(Context, result.Ticket, Options);
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -121,21 +121,6 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
             return AuthenticateResult.Success(ticket);
         }
 
-        private bool RequiresAntiforgery()
-        {
-            if (Options.SuppressAntiforgeryValidation)
-            {
-                return false;
-            }
-
-            var method = Context.Request.Method;
-            return
-                !string.Equals(method, "GET", StringComparison.OrdinalIgnoreCase) &&
-                !string.Equals(method, "HEAD", StringComparison.OrdinalIgnoreCase) &&
-                !string.Equals(method, "OPTIONS", StringComparison.OrdinalIgnoreCase) &&
-                !string.Equals(method, "TRACE", StringComparison.OrdinalIgnoreCase);
-        }
-
         protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
         {
             var result = await EnsureCookieTicket();
@@ -144,7 +129,8 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
                 return result;
             }
 
-            if (RequiresAntiforgery() && !await _antiforgery.IsRequestValidAsync(Context))
+            if (!Options.SuppressAntiforgeryValidation && 
+                !await _antiforgery.IsRequestValidAsync(Context))
             {
                 return AuthenticateResult.Fail("No antiforgery token");
             }
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationMiddleware.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Text.Encodings.Web;
+using Microsoft.AspNetCore.Antiforgery;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.DataProtection;
 using Microsoft.AspNetCore.Http;
@@ -13,9 +14,12 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
 {
     public class CookieAuthenticationMiddleware : AuthenticationMiddleware<CookieAuthenticationOptions>
     {
+        private readonly IAntiforgery _antiforgery;
+
         public CookieAuthenticationMiddleware(
             RequestDelegate next,
             IDataProtectionProvider dataProtectionProvider,
+            IAntiforgery antiforgery,
             ILoggerFactory loggerFactory,
             UrlEncoder urlEncoder,
             IOptions<CookieAuthenticationOptions> options)
@@ -25,6 +29,13 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
             {
                 throw new ArgumentNullException(nameof(dataProtectionProvider));
             }
+
+            if (antiforgery == null)
+            {
+                throw new ArgumentNullException(nameof(antiforgery));
+            }
+
+            _antiforgery = antiforgery;
 
             if (Options.Events == null)
             {
@@ -60,7 +71,7 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
 
         protected override AuthenticationHandler<CookieAuthenticationOptions> CreateHandler()
         {
-            return new CookieAuthenticationHandler();
+            return new CookieAuthenticationHandler(_antiforgery);
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
@@ -143,6 +143,13 @@ namespace Microsoft.AspNetCore.Builder
         public ISecureDataFormat<AuthenticationTicket> TicketDataFormat { get; set; }
 
         /// <summary>
+        /// The SuppressAntiforgeryValidation option disables validation of antiforgery (CSRF) tokens for authenticated
+        /// users. By default an antiforgery token must be provided in the request for HTTP methods that can change
+        /// server side state. If an antiforgery token is not provided the user will be considered unauthenticated.
+        /// </summary>
+        public bool SuppressAntiforgeryValidation { get; set; }
+
+        /// <summary>
         /// For testing purposes only.
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationServiceCollectionExtensions.cs
@@ -1,0 +1,55 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Authentication;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    /// <summary>
+    /// Extension methods for setting up cookie authentication services in an <see cref="IServiceCollection" />.
+    /// </summary>
+    public static class CookieAuthenticationServiceCollectionExtensions
+    {
+        /// <summary>
+        /// Adds cookie authentication services to the specified <see cref="IServiceCollection" />. 
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
+        /// <returns>A reference to this instance after the operation has completed.</returns>
+        public static IServiceCollection AddCookieAuthentication(this IServiceCollection services)
+        {
+            if (services == null)
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            services.AddAuthentication();
+            services.AddAntiforgery();
+            return services;
+        }
+
+        /// <summary>
+        /// Adds cookie authentication services to the specified <see cref="IServiceCollection" />. 
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
+        /// <param name="configureOptions">An action delegate to configure the provided <see cref="SharedAuthenticationOptions"/>.</param>
+        /// <returns>A reference to this instance after the operation has completed.</returns>
+        public static IServiceCollection AddCookieAuthentication(this IServiceCollection services, Action<SharedAuthenticationOptions> configureOptions)
+        {
+            if (services == null)
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            if (configureOptions == null)
+            {
+                throw new ArgumentNullException(nameof(configureOptions));
+            }
+
+            services.Configure(configureOptions);
+            services.AddAuthentication();
+            services.AddAntiforgery();
+            return services;
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/project.json
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/project.json
@@ -10,6 +10,7 @@
     "keyFile": "../../tools/Key.snk"
   },
   "dependencies": {
+    "Microsoft.AspNetCore.Antiforgery": "1.0.0-*",
     "Microsoft.AspNetCore.Authentication": "1.0.0-*",
     "Microsoft.Extensions.Options": "1.0.0-*",
     "Microsoft.Extensions.WebEncoders": "1.0.0-*"
--- a/test/Microsoft.AspNetCore.Authentication.Test/Cookies/CookieMiddlewareTests.cs
+++ b/test/Microsoft.AspNetCore.Authentication.Test/Cookies/CookieMiddlewareTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
@@ -10,6 +11,7 @@ using System.Security.Principal;
 using System.Text;
 using System.Threading.Tasks;
 using System.Xml.Linq;
+using Microsoft.AspNetCore.Antiforgery;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.DataProtection;
 using Microsoft.AspNetCore.Hosting;
@@ -25,7 +27,7 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
     public class CookieMiddlewareTests
     {
         [Fact]
-        public async Task NormalRequestPassesThrough()
+        public async Task NormalGetRequestPassesThrough()
         {
             var server = CreateServer(new CookieAuthenticationOptions());
             var response = await server.CreateClient().GetAsync("http://example.com/normal");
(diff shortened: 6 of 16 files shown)


src/Microsoft.AspNetCore.Authentication.Cookies/project.json
```diff
@@ -10,6 +10,7 @@
     "keyFile": "../../tools/Key.snk"
   },
   "dependencies": {
+    "Microsoft.AspNetCore.Antiforgery": "1.0.0-*",
```
rynowak: Not needed, will undo.

src/Microsoft.AspNetCore.Authentication/project.json
```diff
@@ -10,6 +10,7 @@
     "keyFile": "../../tools/Key.snk"
   },
   "dependencies": {
+    "Microsoft.AspNetCore.Antiforgery": "1.0.0-*",
```
rynowak: `.Authentication` is the only package here that has a `AddXyz(...)` so it needs to be a dependency at this level

test/Microsoft.AspNetCore.Authentication.Test/Cookies/CookieMiddlewareTests.cs
```diff
@@ -713,6 +715,155 @@ private Task SignOutAsWrong(HttpContext context)
...
+            Assert.Equal("Alice", FindClaimValue(transaction2, ClaimTypes.Name));
+        }
+
+        [Theory]
+        [InlineData("PUT")]
+        [InlineData("PoST")]
+        [InlineData("delete")]
+        [InlineData("DEBUG")]
+        [InlineData("bleh")]
+        public async Task Cookie_TreatedAsAnonymous_WithoutCSRF(string httpMethod)
```
rynowak: Just a reminder for myself that I'm missing a test for:
- Authenticated + POST + has CSRF
- CSRF protection suppressed by options

test/Microsoft.AspNetCore.Authentication.Test/Cookies/CookieMiddlewareTests.cs
```diff
@@ -1264,11 +1420,13 @@ private static void Describe(HttpResponse res, AuthenticateContext result)
...
             {
-                transaction.SetCookie = transaction.Response.Headers.GetValues("Set-Cookie").SingleOrDefault();
+                transaction.SetCookie = string.Join(", ", transaction.Response.Headers.GetValues("Set-Cookie"));
             }
             if (!string.IsNullOrEmpty(transaction.SetCookie))
             {
-                transaction.CookieNameValue = transaction.SetCookie.Split(new[] { ';' }, 2).First();
+                transaction.CookieNameValue = string.Join(
+                    ", ", 
+                    transaction.Response.Headers.GetValues("Set-Cookie").Select(c => c.Split(new[] { ';' }, 2).First()));
```
rynowak: Needed to make some adjustments here because CSRF adds another cookie.

src/Microsoft.AspNetCore.Authentication/PersistentAuthenticationHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Antiforgery;
+using Microsoft.AspNetCore.Builder;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public abstract class PersistentAuthenticationHandler<TOptions> : AuthenticationHandler<TOptions>
+        where TOptions : AuthenticationOptions
+    {
+        public PersistentAuthenticationHandler(IAntiforgery antiforgery)
```
rynowak: Personal request of @blowdart was that we add a base-class that enforces antiforgery. I'm not 100% sold - we could really just put all of this stuff in the base-class.

I originally started with this class overriding `HandleAuthenticateAsync`, but the problem is that you can circumvent this easily by just not calling the `base.HandleAuthenticateAsync`.

Thoughts?

src/Microsoft.AspNetCore.Authentication/PersistentAuthenticationHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Antiforgery;
+using Microsoft.AspNetCore.Builder;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public abstract class PersistentAuthenticationHandler<TOptions> : AuthenticationHandler<TOptions>
```
Tratcher: why the base class if it's only implemented in cookies?

src/Microsoft.AspNetCore.Authentication/project.json
Tratcher: :-1: We can have other AddXyz's. This only appears to apply to cookies, it shouldn't be pulled in for everything else.

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
```diff
@@ -143,6 +143,13 @@ public string CookieName
         public ISecureDataFormat<AuthenticationTicket> TicketDataFormat { get; set; }
 
         /// <summary>
+        /// The SuppressAntiforgeryValidation option disables validation of antiforgery (CSRF) tokens for authenticated
+        /// users. By default an antiforgery token must be provided in the request for HTTP methods that can change
+        /// server side state. If an antiforgery token is not provided the user will be considered unauthenticated.
+        /// </summary>
+        public bool SuppressAntiforgeryValidation { get; set; }
```
Tratcher: You don't actually check this anywhere.

Tratcher:
I assume you'll have to make similar changes to WebListener and IISPlatformHandler?

src/Microsoft.AspNetCore.Authentication/project.json
rynowak: As discussed earlier, this is for any persistent auth (windows, cookies).

rynowak:
> I assume you'll have to make similar changes to WebListener and IISPlatformHandler?

Yep.

src/Microsoft.AspNetCore.Authentication/PersistentAuthenticationHandler.cs
rynowak: @blowdart wants it to be easy to opt-in to this behavior in a standard way.  As I mentioned elsewhere, I don't really feel like having a base class in this case helps. It's still easy to get wrong.

I'd either put this all on `AuthenticationHandler` or all on the `Cookie...Handler`

src/Microsoft.AspNetCore.Authentication/PersistentAuthenticationHandler.cs
Tratcher: It doesn't make sense if only cookies derives from it.

src/Microsoft.AspNetCore.Authentication/PersistentAuthenticationHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+            {
+                return await Antiforgery.IsRequestValidAsync(Context);
+            }
+
+            return true;
+        }
+
+        protected virtual bool RequiresAntiforgery()
+        {
+            var method = Context.Request.Method;
```
Tratcher: Move this check into Antiforgery? You'll want to re-use it in WebListener and IISPlatformHandler and they don't/can't depend on this repo.

src/Microsoft.AspNetCore.Authentication/project.json
Tratcher: Windows auth doesn't use Microsoft.AspNetCore.Authentication, only Microsoft.AspNetCore.Http.Features.Authentication.IHttpAuthenticationFeature

src/Microsoft.AspNetCore.Authentication/PersistentAuthenticationHandler.cs
rynowak: Yeah I think you're right. This is starting to feel like these pieces need to live there.

PinpointTownes:
All this stuff seems really strange, IMHO. Why not having a separate antiforgery middleware?

src/Microsoft.AspNetCore.Authentication/project.json
davidfowl: Yah bro, layering.

HaoK:
Just so I understand the main scenario/motivation for this change is to make it easy for all of the auth middlewares that support sign in to automatically call anti forgery?

rynowak:
> All this stuff seems really strange, IMHO. Why not having a separate antiforgery middleware?

Otherwise, it's just more to configure.

The need for antiforgery is tied to your choice of auth. You might might have an app that's using cookies + bearer, and antiforgery only applies when you're using cookies.

rynowak:
> Just so I understand the main scenario/motivation for this change is to make it easy for all of the auth middlewares that support sign in to automatically call anti forgery?

Just persistent auth mechanisms (cookie, windows, basic)

src/Microsoft.AspNetCore.Authentication/project.json
rynowak: You guys realize this results in a new `services.AddCookieAuthentication()` being required everywhere right?

rynowak:
Updated

src/Microsoft.AspNetCore.Authentication/PersistentAuthenticationHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+            return
+                !string.Equals(method, "GET", StringComparison.OrdinalIgnoreCase) &&
```
PinpointTownes: So this new thing won't help prevent CSRF attacks pointing to GET endpoints returning JSON arrays (aka the GMail bug)?

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -113,6 +121,21 @@ private void RequestRefresh(AuthenticationTicket ticket)
...
+        private bool RequiresAntiforgery()
+        {
+            if (Options.SuppressAntiforgeryValidation)
+            {
+                return false;
+            }
+
+            var method = Context.Request.Method;
+            return
+                !string.Equals(method, "GET", StringComparison.OrdinalIgnoreCase) &&
```
Tratcher: I thought you were moving this check into anti-forgery?

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationMiddleware.cs
```diff
@@ -26,6 +30,13 @@ public class CookieAuthenticationMiddleware : AuthenticationMiddleware<CookieAut
                 throw new ArgumentNullException(nameof(dataProtectionProvider));
             }
 
+            if (antiforgery == null)
```
Tratcher: It's interesting that this is required even if SuppressAntiforgeryValidation=true.

test/Microsoft.AspNetCore.Authentication.Test/Twitter/TwitterMiddlewareTests.cs
```diff
@@ -184,7 +184,8 @@ private static TestServer CreateServer(TwitterOptions options, Func<HttpContext,
                 })
                 .ConfigureServices(services =>
                 {
-                    services.AddAuthentication();
+                    services.AddAntiforgery();
```
Tratcher: You don't need AddAntiforgery here anymore because it's in AddCookieAuthentication, right?

Tratcher:
:+1: Much cleaner.

blowdart:
@PinpointTownes the reason I like this change is it'll all happen automatically. No need to create a token in the view, and no need to validate it in the controller.

Tratcher:
What does the view end of this look like?

rynowak:
> What does the view end of this look like?

If you mean Razor views... our form taghelper and `@Html.BeginForm(...)` now generate a token by default.

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
rynowak: Yep. This method will move into Antiforgery

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationMiddleware.cs
rynowak: Why? This is a service, either it's available or your have a configuration error.

src/Microsoft.AspNetCore.Authentication/PersistentAuthenticationHandler.cs
rynowak: @blowdart - want to comment? I'm not familiar with this case.

PinpointTownes:
> @PinpointTownes the reason I like this change is it'll all happen automatically. No need to create a token in the view, and no need to validate it in the controller.

The biggest issue I have with this feature is that it gives you the *totally* wrong impression that using cookies authentication with `SuppressAntiforgeryValidation = false` (the default value) is safe.

I think a separate middleware covering all the HTTP verbs (even GET, to prevent data leakage when returning JSON arrays) and returning a 400 error instead of preventing successful authentication would be better and more flexible.

blowdart:
We've never had anything for JSON gets. You can still do it manually as you always have.

As for SupressAntiforgeryValidation = false; and then automatically adding CSRF tokens and validation being unsafe ... why? We are generating and validating by default in this case.

For unauthenticated forms, well you have to add the attribute manually to your controller, just like you always have.

PinpointTownes:
> We've never had anything for JSON gets. You can still do it manually as you always have.

That's the whole point of my remark: why having an automatic feature that covers POST requests but not GET requests? Do you really think that people will ever realize that returning JSON arrays from their GET endpoints is not covered by the automatic anti-XSRF feature of your revamped cookie middleware?

IMHO, it just feels like a limited and wonky protection, that will result in a terrible trap for non-security  experts (but again, that's just my two cents :smile:)

blowdart:
GET should be impotent. CSRF is for non-impotent actions. It's NOT a CSRF concern, it's a potential information leak, but that's not CSRF

PinpointTownes:
> GET should be impotent. CSRF is for non-impotent actions. It's NOT a CSRF concern, it's a potential information leak, but that's not CSRF

Of course [it is](http://jeremiahgrossman.blogspot.fr/2006/01/advanced-web-attack-techniques-using.html). If you don't agree, you still can ping Jeremiah Grossman - [who reported the issue to Google in 2006](http://jeremiahgrossman.blogspot.fr/2007/01/gmail-xsrf-json-call-back-hackery.html) - to tell him he's using the wrong term :smile:

And where have you seen that CSRF was limited to non-idempotent (and not *non-impotent* :laughing:) actions?

blowdart:
So to confirm you're saying is we don't protect against something we've never protected against anyway?

PinpointTownes:
It's never too late to do the right thing :smile: 

And BTW, your sarcastic remark applies to this PR: Katana's cookie middleware had no anti-XSRF feature... so why would you start adding an anti-XSRF feature to the vNext equivalent? :smile:

blowdart:
Because we have the time and a chance to do it.

HaoK:
Changes look good. 

But going back to having a new base class that cookie and jwtbearer and any of the non remote authentication handlers would derive from that contains this functionality is something that I like. I considered adding a base class when we introduced the remote base class but there wasn't any functionality that needed to be shared between these types of handler. Now that we have this (assuming this is something that most non remote authentication handlers do want), I think it's a good reason to introduce a base class.

PinpointTownes:
> So to confirm you're saying is we don't protect against something we've never protected against anyway?

BTW, it's not true: in ASP.NET MVC (3 IIRC), the `Controller.Json` method had been modified to take a `JsonRequestBehavior` parameter to prevent this kind of attack.

> But going back to having a new base class that cookie and jwtbearer and any of the non remote authentication handlers would derive from that contains this functionality is something that I like.

Actually, bearer authentication doesn't need this feature at all, as access tokens are not automatically flowed by browsers (unlike cookies and basic or Windows authentication credentials).

HaoK:
Actually, bearer authentication doesn't need this feature at all, as access tokens are not automatically flowed by browsers (unlike cookies and basic or Windows authentication credentials).

Having potentially 3 other usages of this would still make the base class useful. Jwt could always suppress this behavior by default and still extend the common base class which I would consider naming LocalAuthenticationHandler to mirror RemoteAuthenticatioHandler.

Tratcher:
@HaoK There's still nothing to share in a base class. Windows auth doesn't use any of the M.A.Auth object heirarchy, and as you said it doesn't apply to JWT.

HaoK:
@HaoK<https://github.com/HaoK> There's still nothing to share in a base class. Windows auth doesn't use any of the M.A.Auth object heirarchy, and as you said it doesn't apply to JWT.

Why doesn't/shouldn't/can't the windows auth use our base classes?  I know we discussed this a long time ago but I can't recall the reason anymore. But I guess we can address this in a different pr to consider refactoring things then.

rynowak:
> Why doesn't/shouldn't/can't the windows auth use our base classes?  I know we discussed this a long time ago but I can't recall the reason anymore. But I guess we can address this in a different pr to consider refactoring things then.

I know @blowdart is strongly in favor of this. The trouble that I see with it is that I implemented it and it didn't seem compelling without a bigger refactor. I'd rather proceed with this workitem and follow up on whether or not the layering is appropriate afterwards

rynowak:
> And BTW, your sarcastic remark applies to this PR: Katana's cookie middleware had no anti-XSRF feature... so why would you start adding an anti-XSRF feature to the vNext equivalent?

A big part of why is that the layering wasn't right for this to work, and honestly different teams doing their own thing. As someone else on the team remarked, there's an instinctive reaction to adding this feature, but if it had always been there no one would even blink.

How would you protect your app from CSRF if you're using cookies but not CSRF? Not something we tried to solve in the past.

HaoK:
Yeah we don't have to hold up the pr for the refactor. We can look at a larger refactor after the initial support is in.  I've been looking for a reason to introduce this kind of base class for a long time anyways, and if it gives us a reason to move the windows auth stuff over to using auth base classes as well that's two birds with one stone :)

PinpointTownes:
> How would you protect your app from CSRF if you're using cookies but not MVC?

Well, you've done the biggest part by revamping `IAntiforgery` so I'd just use an antiforgery middleware exposing the "form/header" antiforgery token and validating it when receiving a request. No need for MVC in this case.

Actually, something similar to the excellent CORS model would probably offer the best experience.

rynowak:
Updated - tests won't pass until the change propagates through antiforgery

Tratcher:
:shipit:

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+            return services;
+        }
+
+        /// <summary>
+        /// Adds cookie authentication services to the specified <see cref="IServiceCollection" />. 
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
+        /// <param name="configureOptions">An action delegate to configure the provided <see cref="SharedAuthenticationOptions"/>.</param>
+        /// <returns>A reference to this instance after the operation has completed.</returns>
+        public static IServiceCollection AddCookieAuthentication(this IServiceCollection services, Action<SharedAuthenticationOptions> configureOptions)
```
HaoK: The fact that this takes SharedAuthenticationOptions instead of CookieAuthenticationOptions seems a bit weird to me...

rynowak: Cookie middleware takes `CookieAuthenticationOptions` in its `Use...` method. 

I originally started with `CookieAuthenticationOptions` here, but it seemed like it would do the wrong thing. Let's sync up on this sometime this week.

Tratcher: Yeah, it's a little weird, but it's more correct than using CookieAuthenticationOptions. CookieAuthenticationOptions can't be in DI because there are commonly multiple instances.

PinpointTownes: This extension is horrible. TBH, I can't figure out why it would be so bad to simply reference the antiforgery package from the core authentication package as it's a rather light package and needs few dependencies. 

If the antiforgery block was a bit more flexible about the cookie name, we could even use it in RemoteAuthenticationHandler.

HaoK: Yeah, it's a little weird, but it's more correct than using CookieAuthenticationOptions. CookieAuthenticationOptions can't be in DI because there are commonly multiple instances.


I think we should just drop the parameter. Typically cookies doesn't use the shared options at all so I don't think there's any real loss of usage by requiring an explicit configure if they wanted to configure the shared.

Tratcher: He just renamed this from `AddAuthentication(SharedAuthenticationOptions)` and added anti-forgery. Without it, you have to do:
```
services.AddAntiforgery();
services.AddAuthentication(sharedOptions => ...);
```

HaoK: He just renamed this from AddAuthentication(SharedAuthenticationOptions) and added anti-forgery. Without it, you have to do:

services.AddAntiforgery();
services.AddAuthentication(sharedOptions => ...);

I'm with pinpoint honestly this feels super gross with this new name. I think once we refactor things into a base so this can continue to be called AddAuthentication things will be better.

Tratcher: I don't think that base should happen, but that's a separate discussion.  @rynowak Scrap this part for now and just make people add `services.AddAntiforgery();`

PinpointTownes: Alternatively, if you don't want to update the core package to depend on the antiforgery block, you could consider removing this extension and let the user explicitly call AddAntiforgery. The UseCookieAuthentication could be updated to ensure IAntiforgery is correctly added to the DI container and throw an exception inviting the user to call AddAntiforgery.

Tratcher: Wow, that just tells me that our DI exception messages are terrible.

HaoK: Well I'm sure the default would say something like No service of type IAntiForgery is found, leaving it as an exercise to the user to understand that they need to call AddAntiforgery...

Tratcher: The same as everywhere else. A pattern like this should be used eveywhere or not at all.

PinpointTownes: Well, Identity and MVC already uses marker services to display a better message. We could do a similar thing here.

Tratcher: ?

PinpointTownes: https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNetCore.Identity/BuilderExtensions.cs#L30

Tratcher: @Eilon @muratg should all middleware use this approach to verify their service dependencies? Most don't today.

rynowak:
Closing/Abandoning this PR for now

We talked through the middleware approach and it seems like it offers some significant advantages. Currently working through a PR that will plug in Antiforgery as a middleware/`IAuthenticationHandler` combo.

Issue #690, opened 2016-02-08
Clean up OIDC events


Tratcher:
After the modified control flow in https://github.com/aspnet/Security/pull/683, we need to go back and update the events accordingly.

There have also been requests for the ability to customize / replace the code redemption logic.

PR #695, opened 2016-02-12
Enable custom OIDC authorization code redemption.

Tratcher:
#690 
This allows developers to either customize the request to the token endpoint, or to take complete control and make the request themselves. This gives us enough flexibility to plug in something like ADAL for their advanced redemption and storage scenarios. We'll add an ADAL specific sample as part of a following discussion.
@vibronet @brentschmaltz @tushargupta51 @PinpointTownes @HaoK @muratg

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/AuthorizationCodeReceivedContext.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/AuthorizationCodeReceivedContext.cs
@@ -1,11 +1,12 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Diagnostics.CodeAnalysis;
 using System.IdentityModel.Tokens.Jwt;
+using System.Net.Http;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Authentication;
+using Microsoft.IdentityModel.Protocols.OpenIdConnect;
 
 namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
 {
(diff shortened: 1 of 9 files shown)


HaoK:
Do we already have existing test converge for these code paths?  It might be worth adding a new test demonstrating the new extensibility with test code as well

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -450,16 +450,24 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
                 // Authorization Code or Hybrid flow
                 if (!string.IsNullOrEmpty(authorizationResponse.Code))
                 {
-                    // TODO: Does this event provide any value over AuthorizationResponseReceived or AuthorizationResponseValidated?
+                    // TODO: RedeemingAuthorizationCode?
```
PinpointTownes: Yep. I'm pretty sure I've mentioned it in a previous PR, but `AuthorizationCodeReceived` makes no sense as `AuthorizationResponseReceived` already covers the main use cases.

`RedeemingAuthorizationCode` is much more appropriate name.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/AuthorizationCodeReceivedContext.cs
```diff
@@ -17,29 +18,70 @@ public class AuthorizationCodeReceivedContext : BaseOpenIdConnectContext
...
+        }
+
+        /// <summary>
+        /// Tells the middleware to skip the code redemption process. The developer may have redeemed the code themselves, or
+        /// decided that the redemption was not required. If tokens were retrieved that are needed for further processing then
+        /// call one of the overloads that allows providing tokens. An IdToken is required if one had not been previously received
+        /// in the authorization response. An access token can optionally be provided for the middleware to contact the
+        /// user-info endpoint. Calling this is the same as setting TokenEndpointResponse.
+        /// </summary>
+        public void SkipCodeRedemption(string accessToken)
```
PinpointTownes: Not sure having such an overload is really pertinent as it doesn't allow setting the identity token, which MUST be provided by the authorization server in the token response (in both the code and the implicit flows).

Tratcher:
@HaoK not really, OIDC only has some challenge and signout tests right now. Unfortunately the tooling is in such a state that it's impractical to develop a new set of tests at the moment.

Tratcher:
Updated with an ADAL sample.
Added CodeRedemptionSkipped helper.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/AuthorizationCodeReceivedContext.cs
```diff
@@ -17,29 +18,72 @@ public class AuthorizationCodeReceivedContext : BaseOpenIdConnectContext
...
         }
 
         public AuthenticationProperties Properties { get; set; }
 
         /// <summary>
-        /// Gets or sets the 'code'.
+        /// Gets or sets the <see cref="JwtSecurityToken"/> that was received in the authentication response, if any.
         /// </summary>
-        public string Code { get; set; }
+        public JwtSecurityToken JwtSecurityToken { get; set; }
```
PinpointTownes: Does that correspond to the identity token or the access token, that can possibly be a JWT token?

Tratcher: IdToken. access tokens aren't for interpretation.

PinpointTownes: Then it would be clearer to name this property `IdentityToken` or `IdToken`.

Tratcher: Hmm, `IdToken` is used in the code to refer to the raw text value, e.g. OpenIdConnectMessage.IdToken.  `IdentityToken` isn't used anywhere so I don't know if it would help clarify. `SecurityToken` is used in a few places.

I'm going to leave it for now. There's not a clearly better name and it is still then name carried forward from Katana. You only noticed here because I moved it around in the file.

samples/OpenIdConnect.AzureSample/AuthPropertiesTokenCache.cs
```diff
@@ -0,0 +1,66 @@
+﻿using System;
+using Microsoft.AspNetCore.Http.Authentication;
+using Microsoft.IdentityModel.Clients.ActiveDirectory;
+
+namespace OpenIdConnect.AzureSample
```
brentschmaltz: could you use AzureAdSample?

samples/OpenIdConnect.AzureSample/Startup.cs
```diff
@@ -0,0 +1,157 @@
...
+            });
+
+            var clientId = Configuration["oidc:clientid"];
+            var clientSecret = Configuration["oidc:clientsecret"];
+            var authority = Configuration["oidc:authority"];
+            var resource = "https://graph.windows.net";
+            app.UseOpenIdConnectAuthentication(new OpenIdConnectOptions
+            {
+                ClientId = clientId,
+                ClientSecret = clientSecret, // for code flow
```
brentschmaltz: https://github.com/aspnet/Security/issues/701 opened to cover Certificates and Signed Assertions.

samples/OpenIdConnect.AzureSample/Startup.cs
```diff
@@ -0,0 +1,157 @@
...
+                Authority = authority,
+                ResponseType = OpenIdConnectResponseTypes.CodeIdToken,
+                // GetClaimsFromUserInfoEndpoint = true,
+                Events = new OpenIdConnectEvents()
+                {
+                    OnAuthorizationCodeReceived = async context =>
+                    {
+                        var request = context.HttpContext.Request;
+                        var currentUri = request.Scheme + "://" + request.Host + request.PathBase + request.Path;
```
brentschmaltz: We need to make sure this is the same that was on the request.

Tratcher: It should be because that's how we got the this authorization response.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/AuthorizationCodeReceivedContext.cs
```diff
@@ -17,29 +18,72 @@ public class AuthorizationCodeReceivedContext : BaseOpenIdConnectContext
...
+        /// same as calling SkipCodeRedemption. If set then the middleware will not attempt to redeem the code. An IdToken
+        /// is required if one had not been previously received in the authorization response. An access token is optional
+        /// if the middleware is to contact the user-info endpoint.
         /// </summary>
-        /// <remarks>This is the redirect_uri that was sent in the id_token + code OpenIdConnectRequest.</remarks>
-        [SuppressMessage("Microsoft.Design", "CA1056:UriPropertiesShouldNotBeStrings", Justification = "user controlled, not necessarily a URI")]
-        public string RedirectUri { get; set; }
+        public OpenIdConnectMessage TokenEndpointResponse { get; set; }
+
+        public bool CodeRedemptionSkipped => TokenEndpointResponse != null;
```
brentschmaltz: It may read better if this is named: SkipCodeRedemption. As in the handler the logic currently reads: if (!CodeRedemptionSkipped) -> redeem code...

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -457,9 +456,17 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
                         return result;
                     }
                     authorizationResponse = authorizationCodeReceivedContext.ProtocolMessage;
-                    var code = authorizationCodeReceivedContext.Code;
-
-                    tokenEndpointResponse = await RedeemAuthorizationCodeAsync(code, authorizationCodeReceivedContext.RedirectUri);
+                    properties = authorizationCodeReceivedContext.Properties;
+                    var tokenEndpointRequest = authorizationCodeReceivedContext.TokenEndpointRequest;
+                    // If the developer redeemed the code themselves...
+                    tokenEndpointResponse = authorizationCodeReceivedContext.TokenEndpointResponse;
```
brentschmaltz: tokenEndpointResponse only needs to be set if '! skipped'

vibronet:
The sample looks good to me. Very minor nitpick: To be complete, the cache implementations should also flesh out item deletions (as those don't call any of the *Access notifications) so that they can be reflected in the storage layer. However that really does not change the ASP.NET API surface, hence there's no need to address those in the sample.

samples/OpenIdConnect.AzureSample/Startup.cs
davidfowl: Any reason you don't use the extension methods we have for this?

Tratcher:
Some minor updates

Issue #639, opened 2016-01-03
SaveTokenAsClaim for JwtBearer


leastprivilege:
For consistency - the JwtBearer MW should also have a `SaveTokenAsClaim` setting - this would make it easier to get to the token e.g. for contacting user info in claims transformation - or "delegation"

leastprivilege:
it probably should go on `AuthenticationOptions` and not `RemoteAuthenticationOptions`

Tratcher:
It wouldn't belong on the base AuthenticationOptions because it wouldn't apply to things like Cookies, Basic, etc.. If we do this I'd just duplicate the property for JWT rather than moving it up to AuthOptions.

Eilon:
We're considering putting the data into the `AuthenticationProperties` instead of as claims. Would that make sense here?

leastprivilege:
I thought this was meant to be a convenient way to get to the access token without having to implement an event. The token should be available as a property (or on the protocol message in OIDC) anyways.

Tratcher:
The auth properties flow with the ClaimsPrincipal all the way back to the app, and through the cookie.

PinpointTownes:
@Eilon @Tratcher when adding `SaveTokensAsClaims` to the OIDC middleware, I had considered using `AuthenticationProperties`, as it was the best candidate for that. Sadly, retrieving authentication properties is much harder than retrieving a simple claim (not to mention that it requires using a very strange syntax).

```csharp
var token = context.User.FindFirst("access_token")?.Value;
```

```csharp
var context = new AuthenticateContext("[authentication scheme]");
await context.Authentication.AuthenticateAsync(context);

if (context.Principal == null || context.Properties == null) {
    throw new InvalidOperationException("The request is not authenticated.");
}

string token;
context.Properties.TryGetValue("access_token", out token);
```

Eilon:
Putting on Backlog for now. We will revisit the popular suggestion at a later time. (Popular == not the one I made :smile: )

Tratcher:
@vibronet recommends cutting SaveTokensAsClaims outright on the grounds that the ClaimsPrincipal (and the auth properties) go into the session cookie and that's the wrong place to save this kind of information. Moreover the lifetimes of the access and refresh tokens can be fundamentally different from the login session. Finally if there are additional tokens acquired later they would have to be handled separately. The recommended way to handle the tokens is to get them via the events and store them independently.

PinpointTownes:
I disagree.

[...]

PR #696, opened 2016-02-13
Replace SaveTokensAsClaims with ITokenStore

Tratcher:
#639 
@PinpointTownes @leastprivilege @HaoK @Eilon @vibronet @brentschmaltz 
Here is an initial design on how we might replace SaveTokensAsClaims with something more versatile but not more difficult to use.

Design points:
- Tokens are stored and retrieved per provider, per user.
- Tokens must be updateable, e.g. for when you use a refresh token to update the access token.
- Tokens should be available even with indirection in place. E.g. looking up a local user's associated google tokens.
- We do NOT manage lifetimes, refresh, etc.. The only sure way to know if a token is expired is to try to use it. If it fails you must attempt to refresh it or re-sign-in to get new tokens.
- The store is shared via DI across the various auth middleware and the user code.
- Sugar for common scenarios.

Unaddressed so far:
- JwtBearer, though it could use the same model.
- What store implementations should we provide? At least IMemoryCache, IDistributedCache, and something for Identity.
- How this interfaces with ADAL's storage.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookMiddleware.cs
@@ -22,20 +22,21 @@ namespace Microsoft.AspNetCore.Authentication.Facebook
         /// Initializes a new <see cref="FacebookMiddleware"/>.
         /// </summary>
         /// <param name="next">The next middleware in the HTTP pipeline to invoke.</param>
+        /// <param name="serviceProvider"></param>
         /// <param name="dataProtectionProvider"></param>
         /// <param name="loggerFactory"></param>
         /// <param name="encoder"></param>
         /// <param name="sharedOptions"></param>
         /// <param name="options">Configuration options for the middleware.</param>
-        /// <param name="configureOptions"></param>
         public FacebookMiddleware(
             RequestDelegate next,
+            IServiceProvider serviceProvider,
             IDataProtectionProvider dataProtectionProvider,
             ILoggerFactory loggerFactory,
             UrlEncoder encoder,
             IOptions<SharedAuthenticationOptions> sharedOptions,
             IOptions<FacebookOptions> options)
-            : base(next, dataProtectionProvider, loggerFactory, encoder, sharedOptions, options)
+            : base(next, serviceProvider, dataProtectionProvider, loggerFactory, encoder, sharedOptions, options)
         {
             if (next == null)
             {
--- a/src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
@@ -85,21 +85,21 @@ namespace Microsoft.AspNetCore.Authentication.OAuth
 
             var identity = new ClaimsIdentity(Options.ClaimsIssuer);
 
-            if (Options.SaveTokensAsClaims)
+            var ticket = await CreateTicketAsync(identity, properties, tokens);
+
+            var id = ticket.Principal.FindFirst(ClaimTypes.NameIdentifier)?.Value;
+            if (Options.TokenStore != null && !string.IsNullOrEmpty(id))
             {
-                identity.AddClaim(new Claim("access_token", tokens.AccessToken,
-                                            ClaimValueTypes.String, Options.ClaimsIssuer));
+                Options.TokenStore.Set(Options.AuthenticationScheme, id, "access_token", tokens.AccessToken);
 
                 if (!string.IsNullOrEmpty(tokens.RefreshToken))
                 {
-                    identity.AddClaim(new Claim("refresh_token", tokens.RefreshToken,
-                                                ClaimValueTypes.String, Options.ClaimsIssuer));
+                    Options.TokenStore.Set(Options.AuthenticationScheme, id, "refresh_token", tokens.RefreshToken);
                 }
 
                 if (!string.IsNullOrEmpty(tokens.TokenType))
                 {
-                    identity.AddClaim(new Claim("token_type", tokens.TokenType,
-                                                ClaimValueTypes.String, Options.ClaimsIssuer));
+                    Options.TokenStore.Set(Options.AuthenticationScheme, id, "token_type", tokens.TokenType);
                 }
 
                 if (!string.IsNullOrEmpty(tokens.ExpiresIn))
@@ -108,15 +108,13 @@ namespace Microsoft.AspNetCore.Authentication.OAuth
                     if (int.TryParse(tokens.ExpiresIn, NumberStyles.Integer, CultureInfo.InvariantCulture, out value))
                     {
                         var expiresAt = Options.SystemClock.UtcNow + TimeSpan.FromSeconds(value);
-                        // https://www.w3.org/TR/xmlschema-2/#dateTime
                         // https://msdn.microsoft.com/en-us/library/az4se3k1(v=vs.110).aspx
-                        identity.AddClaim(new Claim("expires_at", expiresAt.ToString("o", CultureInfo.InvariantCulture),
-                                                    ClaimValueTypes.DateTime, Options.ClaimsIssuer));
+                        Options.TokenStore.Set(Options.AuthenticationScheme, id, "expires_at", expiresAt.ToString("o", CultureInfo.InvariantCulture));
                     }
                 }
             }
 
-            return AuthenticateResult.Success(await CreateTicketAsync(identity, properties, tokens));
+            return AuthenticateResult.Success(ticket);
         }
 
         protected virtual async Task<OAuthTokenResponse> ExchangeCodeAsync(string code, string redirectUri)
@@ -162,6 +160,14 @@ namespace Microsoft.AspNetCore.Authentication.OAuth
             var ticket = new AuthenticationTicket(new ClaimsPrincipal(identity), properties, Options.AuthenticationScheme);
             var context = new OAuthCreatingTicketContext(ticket, Context, Options, Backchannel, tokens);
             await Options.Events.CreatingTicket(context);
+
+            if (ticket.Principal.FindFirst(ClaimTypes.NameIdentifier) == null)
+            {
+                // If there aren't any claims at all, at least add a unique id for this login session. This will help associate the
+                // tokens from the token store.
+                identity.AddClaim(new Claim(ClaimTypes.NameIdentifier, Guid.NewGuid().ToString(), ClaimValueTypes.String, Options.ClaimsIssuer));
+            }
+
             return context.Ticket;
         }
 
--- a/src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationOptions.cs
@@ -49,17 +49,15 @@ namespace Microsoft.AspNetCore.Builder
         }
 
         /// <summary>
-        /// Defines whether access and refresh tokens should be stored in the
-        /// <see cref="ClaimsPrincipal"/> after a successful authorization with the remote provider.
-        /// This property is set to <c>false</c> by default to reduce
-        /// the size of the final authentication cookie.
-        /// </summary>
-        public bool SaveTokensAsClaims { get; set; }
-
-        /// <summary>
         /// Gets or sets the time limit for completing the authentication flow (15 minutes by default).
         /// </summary>
         public TimeSpan RemoteAuthenticationTimeout { get; set; } = TimeSpan.FromMinutes(15);
+
+        /// <summary>
+        /// An optional location to store access and refresh tokens after a successful authorization with the remote provider.
+        /// This can also be provided via dependency injection.
+        /// </summary>
+        public ITokenStore TokenStore { get; set; }
 
         public IRemoteAuthenticationEvents Events = new RemoteAuthenticationEvents();
     }
--- a/src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
+++ b/src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
@@ -1,0 +1,103 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Globalization;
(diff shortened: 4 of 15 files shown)


src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationOptions.cs
```diff
@@ -49,17 +49,15 @@ public string DisplayName
...
-        public bool SaveTokensAsClaims { get; set; }
+        public TimeSpan RemoteAuthenticationTimeout { get; set; } = TimeSpan.FromMinutes(15);
 
         /// <summary>
-        /// Gets or sets the time limit for completing the authentication flow (15 minutes by default).
+        /// An optional location to store access and refresh tokens after a successful authorization with the remote provider.
+        /// This can also be provided via dependency injection.
         /// </summary>
-        public TimeSpan RemoteAuthenticationTimeout { get; set; } = TimeSpan.FromMinutes(15);
+        public ITokenStore TokenStore { get; set; }
```
HaoK: You know, there's nothing specific to Remote here, we could move this into the base AuthenticationOptions, and then consolidate the logic in the base AuthenticationMiddleware so every ctor doesn't have to set it.  This would let any auth middleware put things into the TokenStore as well if they wanted to.  It doesn't hurt, and it cleans up the code a bit too...

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookMiddleware.cs
```diff
@@ -22,20 +22,21 @@ public class FacebookMiddleware : OAuthMiddleware<FacebookOptions>
...
         public FacebookMiddleware(
             RequestDelegate next,
+            IServiceProvider serviceProvider,
             IDataProtectionProvider dataProtectionProvider,
             ILoggerFactory loggerFactory,
             UrlEncoder encoder,
             IOptions<SharedAuthenticationOptions> sharedOptions,
             IOptions<FacebookOptions> options)
-            : base(next, dataProtectionProvider, loggerFactory, encoder, sharedOptions, options)
+            : base(next, serviceProvider, dataProtectionProvider, loggerFactory, encoder, sharedOptions, options)
```
HaoK: Instead of doing service locator for the ITokenStore, maybe we should just stick a NoOpTokenStore into DI via TryAdd by default, and always call into the store.

src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+            {
+                throw new ArgumentException("The id claim does not have a value.", nameof(user));
+            }
+            store.Set(authenticationScheme, userId, tokenType, value);
+        }
+
+        public static string GetToken(this AuthenticationManager authManager, string authenticationScheme, string tokenType)
+        {
+            var store = authManager.HttpContext.RequestServices.GetService<ITokenStore>();
+            return store.Get(authenticationScheme, authManager.HttpContext.User, tokenType);
```
HaoK: ITokenStore can be null, need a ? I think

src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        public static void SetToken(this AuthenticationManager authManager, string authenticationScheme, string tokenType, string value)
+        {
+            var store = authManager.HttpContext.RequestServices.GetRequiredService<ITokenStore>();
```
HaoK: ITokenStore is in ApplicationServices no?

Tratcher: HttpContext.ApplicationServices was removed a while ago.

src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
Tratcher: It doesn't actually matter for extension methods, only concrete methods like on line 16.

Tratcher:
Note: depends on https://github.com/aspnet/HttpAbstractions/pull/558

src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+            }
+            var userId = claim.Value;
+            if (string.IsNullOrEmpty(userId))
+            {
+                throw new ArgumentException("The id claim does not have a value.", nameof(user));
+            }
+            store.Set(authenticationScheme, userId, tokenType, value);
+        }
+
+        public static string GetToken(this AuthenticationManager authManager, string authenticationScheme, string tokenType)
```
PinpointTownes: IMHO, these extensions feel a bit artificial as they don't use the authentication APIs but only access the current HTTP context to retrieve the token store from the request services (which it won't work when configuring a token store in the middleware options instead of using the global store from the DI container).

src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
HaoK: Why won't store.Get null ref?

Tratcher: Because extension methods are actually static methods in disguise.

src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
Tratcher: Yes, we picked AuthenticationManager more for organizational purposes. And yes some of these only work with the DI approach. That's why we have the whole sequence of extensions, for each layer you may be working at.

src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
HaoK: Oh right, I forgot that store.Get is now an extension method :)

src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
HaoK: Yeah these extension methods were added purely as sugar for the mainline scenario as
`context.Authentication.GetRefreshToken("Facebook")` reads a whole lot better than:
```
   var tokenStore = context.RequestServices.GetRequiredService<ITokenStore>();
   tokenStore.Get("Facebook", context.User.FindFirst(ClaimsTypes.NameIdenfifier), "refresh_token")
```

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -162,6 +160,14 @@ public OAuthHandler(HttpClient backchannel)
             var ticket = new AuthenticationTicket(new ClaimsPrincipal(identity), properties, Options.AuthenticationScheme);
             var context = new OAuthCreatingTicketContext(ticket, Context, Options, Backchannel, tokens);
             await Options.Events.CreatingTicket(context);
+
+            if (ticket.Principal.FindFirst(ClaimTypes.NameIdentifier) == null)
+            {
+                // If there aren't any claims at all, at least add a unique id for this login session. This will help associate the
+                // tokens from the token store.
+                identity.AddClaim(new Claim(ClaimTypes.NameIdentifier, Guid.NewGuid().ToString(), ClaimValueTypes.String, Options.ClaimsIssuer));
```
PinpointTownes: We're clearly opening Pandora's box here. Generating a unique per-login identifier would be misleading for application code and may result in terrible things like being able to associate a provider with multiple Identity accounts.

src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
davidfowl: I sorta agree with pinpoint. It's. Strange to hang these off the auth manager

Tratcher: Would you rather have them on HttpContext or elsewhere?

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
Tratcher: Yeah, this scenario is pretty strange. Apps using the middleware without at least getting a user id seems to defeat the "Authentication" purpose of the middleware. "This is some google user, though we couldn't tell you which one...".

Other suggestions for associating this psudo-anonymous user and their tokens?

PinpointTownes:
So, with this new design, we're moving from per-session tokens (attached to a specific `AuthenticationTicket`) to per-user(-identifier) tokens, without having a chance to correlate the access/refresh token with the origin session? This may be problematic when trying to use `id_token` as an `id_token_hint` during a logout request as you now have no way to bind a "provider session" with a "local session".

IMHO, storing the tokens in the authentication ticket is really the right thing to do. The more I think about it, the more I'm convinced this interface - well, maybe a more complete version - should be added to Identity. It would be ideally invoked from the account controller, after a complete authentication dance: https://github.com/aspnet/Templates/blob/dev/src/Rules/StarterWeb/IndividualAuth/Controllers/AccountController.cs#L216

That said, I agree claims are not the ideal place to store them (though it's a really convenient one :smile:). Instead, we should store them in the authentication properties and make them easier to retrieve when using `AuthenticationManager` or from a MVC controller.

HaoK:
Chris do you remember if there was any reasons that this wouldn't work in Identity instead?  We can certainly come up with a flow that flows the tokens via the ClaimsIdentity/AuthenticationProperties into the 'External' cookie where identity could update the tokens as part of the normal login flow.  And the tokens could be retrieved via something like `UserManager.GetToken(user, "facebook", "access_token")`.  

Updating the token outside of the normal login flow wouldn't be too bad, it would look something like:

```
  var user = UserManager.FindByLogin("Facebook", "<facebook identity name identifier>");
  UserManager.SetToken(user, "Facebook", "access_token", "<access token value>";
```

vibronet:
One thing to consider is that access and refresh tokens might be acquired for long term and offline use, hence the session might not be the most natural place to keep them. To make an example only at the pattern level (I am NOT referring to the homonym middlewares): if, while accessing Twitter from their web site, I associate my Twitter to Facebook, every time I tweet I also get a FB update with the same text. However Twitter isn't storing FB's access and refresh tokens in the web app session: it stores them deeper in the application, so that when I tweet via the Twitter app from my iPhone I also get corresponding FB updates.
If we store API access token (as opposed to tokens used during sign in) in the session, this very common scenario won't be achievable without unnatural contortions.
Thoughts?

PinpointTownes:
@HaoK :clap: :clap: 

@vibronet I agree that the authentication ticket is not the best place for long-term token storage (Identity would be a much more appropriate store for that).

On the other hand, an authentication ticket is nothing more than a "screenshot" of a web session at a given moment, so IMHO, it makes total sense to store the access/refresh tokens in the authentication properties, as it's just a transcription of how things were when the ticket was issued (and nothing prevents you from dynamically re-issuing an authentication ticket exposing a refreshed token using the `ValidatePrincipal` event).

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
PinpointTownes: Strange but not unusual when you don't need to authenticate the user and just want to access his resources on his behalf (pure authorization scenario).

> Other suggestions for associating this psudo-anonymous user and their tokens?

Well, storing the tokens in the ticket solves this problem :smile:

brockallen:
If an app needs refresh tokens, then why don't you let the app manage those itself? Why hide all of these things? It doesn't help people to understand security any better.

leastprivilege:
My gut feeling is - postpone it to a release after v1. All use cases are doable today using the events - let's see and learn how people are really using it.

Tratcher:
@leastprivilege I agree we're not reaching a consensus readily enough so we should at least temporarily remove the feature.

That said, I don't think we need to wait for feedback from v1 since people have already been dealing with this issue in Katana and we haven't made any significant changes for it in vNext. The common usage I've seen is hook the events and copy the data into claims, then to hook another event in Identity and manually copy the data into the user database. Sub par all around. Have folks seen other usages?

brockallen:
Well, how much better does this interface/abstraction help over someone just doing that manually in the SecurityTokenValidated event?

Tratcher:
The design in this PR makes it so they don't have to hook the events in security or identity. They just inject the service at startup and then retrieve the tokens from it as needed.

brockallen:
And I guess the keys are constants somewhere?

Tratcher:
This abstraction enables putting the tokens all in one place. You could still manually put some of them elsewhere, but I don't think it makes much sense to split them up for any of the sets of tokens we're looking at so far.

HaoK:
I think I've come around to thinking the auth middleware really shouldn't be responsible for storing/managing the lifetime of the tokens, i.e. directly storing them into the token store.

Instead I think the middleware should just expose the tokens as part of the results of Authenticate, since claims aren't appropriate.  Maybe something like this as sugar/convention in the auth properties:

```
    public class AuthenticationToken {
       public string Name { get; set; }
       public string Value { get; set; }
    }

   public static IEnumerable<AuthenticationToken> GetTokens(this AuthenticationProperties prop);
   public static void StoreTokens(this AuthenticationProperties props, IEnumerable<AuthenticationToken> tokens)
```

We can rename SaveTokensAsClaims => SaveTokensInAuthenticationProperties, and 
Identity can introduce something like:
```
   userManager.GetToken(user, "Facebook", "refresh_token");
   userManager.StoreToken(user, "Facebook", "refresh_token", "2359FDVLKDSJF");
```

The controller action that handles handoff between OAuth and the application cookie(i.e. AccountController.ExternalLoginCallback) would always get the tokens off of the AuthenticationProperties and call StoreToken as part of the login flow.

Tratcher:
@HaoK it sounds like it's your turn to send a PR showing how the end-to-end would work.

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
Tratcher: I think it's reasonable to require the authentication middleware to authenticate the user.

HaoK:
First half of the Security changes for storing the tokens in the AuthProperties is here: https://github.com/aspnet/Security/pull/698

Will send out the identity side next

PR #698, opened 2016-02-16
Save tokens in auth properties instead of claims

HaoK:


samples/SocialSample/Startup.cs
```diff
@@ -90,7 +90,7 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
                 AuthorizationEndpoint = GoogleDefaults.AuthorizationEndpoint,
                 TokenEndpoint = GoogleDefaults.TokenEndpoint,
                 Scope = { "openid", "profile", "email" },
-                SaveTokensAsClaims = true
+                SaveTokensInAuthenticationProperties = true
```
Tratcher: SaveTokens. We can say where in the comments.

Tratcher:
What, no usage sample?

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -85,21 +85,19 @@ public OAuthHandler(HttpClient backchannel)
 
             var identity = new ClaimsIdentity(Options.ClaimsIssuer);
 
-            if (Options.SaveTokensAsClaims)
+            if (Options.SaveTokensInAuthenticationProperties)
             {
-                identity.AddClaim(new Claim("access_token", tokens.AccessToken,
-                                            ClaimValueTypes.String, Options.ClaimsIssuer));
+                var toks = new List<AuthenticationToken>();
```
Tratcher: sp, tokens.

Tratcher: At least this way the tokens are grouped together.

HaoK:
Well the usage without identity is going to be pretty gross, so I didn't update the samples, it'd basically be this:

```
                            var authContext = new AuthenticateContext("<the sign in as scheme">);
                            await context.Authentication.AuthenticateAsync(authContext);
                            var tokens = AuthenticationToken.GetTokens(new AuthenticationProperties(authContext.Properties));
```

We definitely need to add some sugar around extracting an individual token

Tratcher:
Please update the sample and propose some sugar. "This is Hard" is one of the main complaints we're trying to address.

HaoK:
Sure, but the difference in the fix is whether its identity that makes it easier or security.  The feedback from the last PR I thought was that the token APIs belong in identity.  I don't think we should try to fix this at the security layer.  I think the basic pieces to build the sugar should be exposed in Security, but its identity that makes the end to end nice.

src/Microsoft.AspNetCore.Authentication/AuthenticationToken.cs
```diff
@@ -0,0 +1,61 @@
...
+namespace Microsoft.AspNetCore.Authentication
+{
+    public class AuthenticationToken
+    {
+        public string Name { get; set; }
+        public string Value { get; set; }
+
+        private static string TokenNamesKey = ".TokenNames";
+
+        public static void StoreTokens(AuthenticationProperties properties, IEnumerable<AuthenticationToken> tokens)
```
Tratcher: It's a little odd to have these as static here. I'd expect them to be extension methods.

Tratcher:
Yes Identity needs to provide support for some of the larger end-to-end scenarios (multiple providers, local account, etc.), but we also need a usable solution that does not require Identity for the basic scenarios (single provider).

src/Microsoft.AspNetCore.Authentication/AuthenticationToken.cs
HaoK: Yeah could go either way here, I left them as helpers for now, extension methods work just as well...

HaoK:
So the whole remote authentication flow is non trivial, but I can certainly add sugar to enable something like this:

```
   httpContext.Authentication.GetToken("Cookies", "access_token")
```

The one bit of weirdness is that the authentication scheme passed in needs to be the SignInAsScheme and not the actual scheme since remote auth schemes do not support Authenticate anymore...

HaoK:
Updated the social sample with usage, I can flip the helpers on AuthenticationToken to extension methods if we want, they do feel a bit clunky this way.  But you get the idea anyways

leastprivilege:
Most of our customers don't and won't use identity. So design the api without identity in mind.

HaoK:
The current direction in this iteration of the PRs is that the AuthN layer exposes a few 'nicer' apis that easily let you extract/store named tokens from AuthenticationProperties.  Its up to the caller to manage these/associate these with users.  Identity will do this, but if you aren't using identity, you or your customers will need to associate/manage the tokens somehow on your own

HaoK:
Basically this PR just exposes a simpler way to get a token out of facebook/google, via.
`AuthenticationToken.GetToken(context, "Cookie", "access_token")` or 'context.Authentication.GetToken("Cookie", "access_token")`.  What you do with this token isn't really addressed

samples/SocialSample/Startup.cs
```diff
@@ -318,6 +319,13 @@ dnx web
                 {
                     await context.Response.WriteAsync(claim.Type + ": " + claim.Value + "<br>");
                 }
+
+                await context.Response.WriteAsync("Tokens:<br>");
+                
+                await context.Response.WriteAsync("Access Token: " + await AuthenticationToken.GetTokenAsync(context, CookieAuthenticationDefaults.AuthenticationScheme, "access_token") + "<br>");
+                await context.Response.WriteAsync("Refresh Token: " + await AuthenticationToken.GetTokenAsync(context, CookieAuthenticationDefaults.AuthenticationScheme, "refresh_token") + "<br>");
+                await context.Response.WriteAsync("Token Type: " + await AuthenticationToken.GetTokenAsync(context, CookieAuthenticationDefaults.AuthenticationScheme, "token_type") + "<br>");
+                await context.Response.WriteAsync("expires_at: " + await AuthenticationToken.GetTokenAsync(context, CookieAuthenticationDefaults.AuthenticationScheme, "expires_at") + "<br>");
```
Tratcher: And the flow if you need to update any tokens is as follows?
AuthenticateAsync(SignInScheme) to get the ClaimsPrincipal and AuthProperties,
Read the tokens from AuthProperties
Refresh the tokens
Store the tokens back in AuthProperties (or create a new instance?)
Call SignIn(SignInScheme) to persist the changes?

Tratcher: How about `context.Authentication.GetTokenAsync("access_token")` that used Automatic by default? That would avoid the weirdness of specifying signin-scheme.

HaoK: Updating the tokens that way seems pretty awkward.  But I guess so at this layer.

Sure I can add overloads with the automatic scheme, that does make things a lot nicer.

HaoK:
Updated with new extension methods, and AutomaticScheme overload which simplifies the calling code a bit

src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+            return tokens;
+        }
+
+        public static Task<string> GetTokenAsync(this AuthenticationManager manager, string tokenName)
+        {
+            return manager.GetTokenAsync(AuthenticationManager.AutomaticScheme, tokenName);
+        }
+
+
+        public static async Task<string> GetTokenAsync(this AuthenticationManager manager, string signedInScheme, string tokenName)
```
Tratcher: It's just called `signInScheme`, not `signedInScheme`.

Tratcher:
:watch: Curious what others think, and what it looks like in Identity.

leastprivilege:
This seems like a basic enough abstraction.

PinpointTownes:
I like this approach, though it requires too much plumbing to my taste :smile:

IMHO, updating `AuthenticationManager.AuthenticateAsync` to simply return an `AuthenticationTicket` would be much simpler (the current signature is ABSOLUTELY insane anyway).

```csharp
var ticket = context.Authentication.AuthenticateAsync("Cookies");
var token = ticket?.Properties?.Items?["access_token"];
```

We could also provide a `GetProperty` helper to make the properties easier to extract. Here's the one used by ASOS:

```csharp
/// <summary>
/// Gets a given property from the authentication ticket.
/// </summary>
/// <param name="ticket">The authentication ticket.</param>
/// <param name="property">The specific property to look for.</param>
/// <returns>The value corresponding to the property, or <c>null</c> if the property cannot be found.</returns>
public static string GetProperty(this AuthenticationTicket ticket, string property) {
    if (ticket == null) {
        throw new ArgumentNullException(nameof(ticket));
    }

    string value;
    if (!ticket.Properties.Items.TryGetValue(property, out value)) {
        return null;
    }

    return value;
}

```

```csharp
var ticket = context.Authentication.AuthenticateAsync("Cookies");
var token = ticket?.GetProperty("access_token");
```

test/Microsoft.AspNetCore.Authentication.Test/Google/GoogleMiddlewareTests.cs
```diff
@@ -781,6 +789,13 @@ private static TestServer CreateServer(GoogleOptions options, Func<HttpContext,
                         {
                             await context.Authentication.ChallengeAsync("Google");
                         }
+                        else if (req.Path == new PathString("/tokens"))
+                        {
+                            var authContext = new AuthenticateContext(TestExtensions.CookieAuthenticationScheme);
```
PinpointTownes: :gun:

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -85,21 +85,19 @@ public OAuthHandler(HttpClient backchannel)
 
             var identity = new ClaimsIdentity(Options.ClaimsIssuer);
 
-            if (Options.SaveTokensAsClaims)
+            if (Options.SaveTokens)
             {
-                identity.AddClaim(new Claim("access_token", tokens.AccessToken,
-                                            ClaimValueTypes.String, Options.ClaimsIssuer));
+                var toks = new List<AuthenticationToken>();
```
PinpointTownes: `toks`? :trollface:

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -107,13 +105,18 @@ public OAuthHandler(HttpClient backchannel)
                     int value;
                     if (int.TryParse(tokens.ExpiresIn, NumberStyles.Integer, CultureInfo.InvariantCulture, out value))
                     {
-                        var expiresAt = Options.SystemClock.UtcNow + TimeSpan.FromSeconds(value);
                         // https://www.w3.org/TR/xmlschema-2/#dateTime
                         // https://msdn.microsoft.com/en-us/library/az4se3k1(v=vs.110).aspx
-                        identity.AddClaim(new Claim("expires_at", expiresAt.ToString("o", CultureInfo.InvariantCulture),
-                                                    ClaimValueTypes.DateTime, Options.ClaimsIssuer));
+                        var expiresAt = Options.SystemClock.UtcNow + TimeSpan.FromSeconds(value);
+                        toks.Add(new AuthenticationToken
```
PinpointTownes: nit: semantically, `expires_at` is not really a token.

src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+            var tokenNames = new List<string>();
+            foreach (var token in tokens)
+            {
+                // REVIEW: should probably check that there are no ; in the token name and throw or encode
+                tokenNames.Add(token.Name);
+                properties.Items[token.Name] = token.Value;
+            }
+            if (tokenNames.Count > 0)
+            {
+                properties.Items[TokenNamesKey] = string.Join(";", tokenNames.ToArray());
```
PinpointTownes: This approach seems a bit unusual (and probably suboptimal). Instead, why not simply using a prefix key?

src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+
+        public static async Task<string> GetTokenAsync(this AuthenticationManager manager, string signedInScheme, string tokenName)
+        {
+            if (manager == null)
+            {
+                throw new ArgumentNullException(nameof(manager));
+            }
+            if (signedInScheme == null)
```
PinpointTownes: Are we fine with empty schemes?

src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+            {
+                throw new ArgumentNullException(nameof(signedInScheme));
+            }
+            if (tokenName == null)
```
PinpointTownes: ... and with empty keys?

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
Tratcher: *shrug* we want to keep it grouped together.

src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
Tratcher: It's a little weird, but I'd prefer this to key prefixing. With this they individual keys remain directly accessible. You only need this for enumeration.

HaoK: It's a little weird, but I'd prefer this to key prefixing. With this they individual keys remain directly accessible. You only need this for enumeration.

Yeah that said. I should prefix the token keys a bit since it's very easy to stomp on our reserved properties without some kind of .t prefix.

src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
HaoK: This is just a sugar/convention for callers to use, I don't think we should be opinionated on whether empty schemes or keys are valid or not, shrug.  My inclination is to leave them be as opposed to throwing.

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
HaoK: tokens is already a local variable :)

HaoK:
> updating AuthenticationManager.AuthenticateAsync to simply return an AuthenticationTicket would be much simpler (the current signature is ABSOLUTELY insane anyway).

I don't think AuthenticationProperties is really used much except for shuttling data around (identity uses it a bit, and now we stick tokens there), but in general it should be hidden.  That said, go ahead and file a new issue to discuss this if you want, but I don't want rip open any lower layers as part of this change if I don't have to.

test/Microsoft.AspNetCore.Authentication.Test/Google/GoogleMiddlewareTests.cs
HaoK: What's the gun for?  I can't nuke the scheme because its a required argument, I could switch it to automatic scheme but that's just less clear I think

PinpointTownes: The gun is for me, this syntax is literally killing me (though it's absolutely not specific to this PR).

HaoK: Ah okay...  Yeah its not clunky yes, this is the raw no sugar api :)

HaoK: Internally we have a fairly widespread pattern of creating a context, and passing it to a method, I don't think its THAT bad.  But its not the best public API no :)

PinpointTownes: Well, I'm totally fine using this pattern with protected members you override (it's not really different from calling a method on event handler args) but I think it's a terrible pattern for public APIs, as it's counterintuitive: instantiating a context object and expect the method you call to alter it is rather unusual.

HaoK: Yeah I agree, basically its the internal guts exposed, which is why it feels a bit gross...

HaoK: But this actually IS the public API between the lower HttpAbstractions layer and Auth middleware :)  Its just usually sugar'd over by the AuthenticationManager.  I think the reason why we felt this API didn't need sugar is because the developers who are using AuthenticationProperties are typically the ones who would be familiar enough with this pattern.

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
Tratcher: tokenList?

Tratcher:
@HaoK sounds like there are no major objections so far. Want to try filling in the other providers, and give Jwt a try while you're at it?

HaoK:
-Updated the rest of the middleware
-store JwtBearer token as Bearer
-StoreTokens deletes previous tokens, so last store wins
-Added tests

src/Microsoft.AspNetCore.Authentication/AuthenticationOptions.cs
```diff
@@ -47,5 +47,13 @@ public string AuthenticationScheme
...
         /// </summary>
         public AuthenticationDescription Description { get; set; } = new AuthenticationDescription();
+
+        /// <summary>
+        /// Defines whether access and refresh tokens should be stored in the
+        /// <see cref="AuthenticationProperties"/> after a successful authorization.
+        /// This property is set to <c>false</c> by default to reduce
+        /// the size of the final authentication cookie.
+        /// </summary>
+        public bool SaveTokens { get; set; }
```
PinpointTownes: Hum, I don't think this property is generic enough to be in the `AuthenticationOptions` base class.

blowdart: Agree

Tratcher: This could be true by default for Bearer as it wouldn't have the cookie size concerns.

HaoK: So the only one that doesn't use this now is Cookies...  And there's nothing preventing that from saving tokens either...

Its basically a flag that tells the auth middleware to save tokens into AuthenticationProperties if it happens to have any tokens.  If JwtBearer can use this flag, I don't see why its not generically applicable.  AuthenticationMiddleware who don't have any tokens can just ignore this option...

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
HaoK: went with authTokens

src/Microsoft.AspNetCore.Authentication/AuthenticationOptions.cs
Tratcher: RemoteAuth + Bearer, right?

Not Cookies, not Basic, not cert auth, not Windows, etc..

HaoK: Made it true by default for JwtBearer,

HaoK: Okay, moved SaveTokens back to RemoteAuth and added a new SaveToken which defaults to True for JwtBearer

Tratcher:
:+1:
How's it looking in Identity?

HaoK:
Identity PR here: https://github.com/aspnet/Identity/pull/748

HaoK:
@Tratcher want to take a final look over this before I merge?

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -107,13 +105,18 @@ public OAuthHandler(HttpClient backchannel)
...
                     {
-                        var expiresAt = Options.SystemClock.UtcNow + TimeSpan.FromSeconds(value);
                         // https://www.w3.org/TR/xmlschema-2/#dateTime
                         // https://msdn.microsoft.com/en-us/library/az4se3k1(v=vs.110).aspx
-                        identity.AddClaim(new Claim("expires_at", expiresAt.ToString("o", CultureInfo.InvariantCulture),
-                                                    ClaimValueTypes.DateTime, Options.ClaimsIssuer));
+                        var expiresAt = Options.SystemClock.UtcNow + TimeSpan.FromSeconds(value);
+                        authTokens.Add(new AuthenticationToken
+                        {
+                            Name = "expires_at",
```
PinpointTownes: So, no plan to make it a property on `AuthenticationToken`?

PinpointTownes:
@HaoK don't forget to update HandleSignOutAsync: https://github.com/aspnet/Security/blob/haok/2-16tok/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L108-L109.

HaoK:
Thanks @PinpointTownes oh how I love lack of test coverage :)

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
HaoK: Doesn't sound like you've convinced @Tratcher and it definitely makes things more complicated, so my default is no for now as well, you can file an issue to consider adding it later

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -153,6 +153,14 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
                             return AuthenticateResult.Skip();
                         }
 
+                        if (Options.SaveToken)
+                        {
+                            ticket.Properties.StoreTokens(new[]
+                            {
+                                new AuthenticationToken { Name = "Bearer", Value = token }
```
PinpointTownes: Should we name it `access_token` for consistency?

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
PinpointTownes: Okay, will do.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -105,9 +105,8 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
                     message.PostLogoutRedirectUri = logoutRedirectUri;
                 }
 
-                var principal = await Context.Authentication.AuthenticateAsync(Options.SignInScheme);
-                message.IdTokenHint = principal?.FindFirst(OpenIdConnectParameterNames.IdToken)?.Value;
-
+                // REVIEW: is the id token hint required?
+                message.IdTokenHint = await Context.Authentication.GetTokenAsync(OpenIdConnectParameterNames.IdToken);
```
PinpointTownes: Nope, it's not required.

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
HaoK: I don't have enough context to say, I would say we should name it 'correctly' in the context of JwtBearer as opposed to consistently.  I think it's better for it to be as clear as possible what this token is based on the name.  That said, if access_token is correct-ish, consistency is great :)

PinpointTownes: Yep, it's correct (it's the official name used by the specs: https://tools.ietf.org/html/rfc6750#section-2.3)

Tratcher:
:shipit:

HaoK:
ace166fa31af61a6c9924873e6bdcb45e7785d5d

PR #719, opened 2016-02-27
Remove usage of IOptions

HaoK:
https://github.com/aspnet/Options/issues/122

New behavior now with the UseAuth() that don't take an options instance, they will fail now unless someone registered AuthOptions before using that method unlike before...


PinpointTownes:
> New behavior now with the UseAuth() that don't take an options instance, they will fail now unless someone registered AuthOptions before using that method unlike before...

That was my biggest fear concerning this change... :sweat:

HaoK:
That was my biggest fear concerning this change... [:sweat:]


Well we can always nuke that behavior and go back to the empty overload using a new instance of options. And making it up to the developer to get the options out of DI to pass it to the middleware.

Otherwise yes you have to register the options manually to use the overloads since today there is no AddAuth unlike frameworks.

Issue #718, opened 2016-02-26
How can you inject a service into an implementation of IClaimsTransformer?


csaloio:
I need to be able to query a DB in my implementation.

My understanding is that you must add the transform like this in Startup.Configure:
```
app.UseClaimsTransformation(o => o.Transformer = new MyClaimsTransformer());
```

But how can it get access to a scoped service from within `MyClaimsTransformer` ?

blowdart:
As of right now you can't. @HaoK we did talk about giving claims transformation more love ...

csaloio:
@blowdart @HaoK I'm not sure how meaningful a transformation you can make without access to your data :(

HaoK:
Since this is inside of Configure, you can service locate for it to work

```C#
var myTransformer = app.Services.GetRequiredService<MyTransformer>();
app.UseClaimsTransformation(o => o.Transformer = myTransformer);
```

blowdart:
Kinda ugly though :)

HaoK:
Also with the new options changes we can just make this work now that options will be a normal service.

csaloio:
@HaoK I don't understand your example. `IApplicationBuilder` doesn't seem to have a `Services` property (at least in RC1). It does have 'ApplicationServices' but my understanding was that  his is only appropriate for services that have an application scope. I think it is important that any DB service injected into the transformer have request scope and not be scoped to the application.

DovydasNavickas:
@HaoK, @blowdart, this really does not make sense, if the request is not exposed to the transformer. Any reasons why you cannot expose that?

brockallen:
Seems like the `Options.Transformer` really should be pulled from DI.

DovydasNavickas:
And within current request scope :clap:

lasrol:
@csaloio getting db or usermanager from the ApplicationServices gave me a lot of open data reader errors

csaloio:
@lasrol I don't suggest using ApplicationServices. What I think could be done (though it is **not ideal**), is to just new up your context inside of the `TransformAsync` method in the transformer making sure to wrap it in a using block so it is disposed of. Then use it read whatever data you need to transform the `ClaimsPrincipal`.

But as I said this is far from ideal. Everything about the .Net core design encourages the use of DI and but `IClaimsTransformer` appears to be a design oversight.

Eilon:
This would indeed be nice but we are not taking many new features for this release.

csaloio:
[...]

PR #732, opened 2016-03-05
#718 IClaimsTransofrmer updated to take ClaimsTransformationContext

DovydasNavickas:
IClaimsTransofrmer updated to take ClaimsTransformationContext instead of ClaimsPrincipal in TransformAsync method.

#718

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication/ClaimsTransformationContext.cs
+++ b/src/Microsoft.AspNetCore.Authentication/ClaimsTransformationContext.cs
@@ -1,0 +1,15 @@
+﻿using Microsoft.AspNetCore.Http;
+using System.Security.Claims;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public class ClaimsTransformationContext
+    {
+        public ClaimsTransformationContext(HttpContext context)
+        {
+            Context = context;
+        }
+        public HttpContext Context { get; }
+        public ClaimsPrincipal Principal { get; set; }
+    }
+}
--- a/src/Microsoft.AspNetCore.Authentication/ClaimsTransformationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication/ClaimsTransformationHandler.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features.Authentication;
 
 namespace Microsoft.AspNetCore.Authentication
@@ -12,10 +13,12 @@ namespace Microsoft.AspNetCore.Authentication
     public class ClaimsTransformationHandler : IAuthenticationHandler
     {
         private readonly IClaimsTransformer _transform;
+        private readonly HttpContext _httpContext;
 
-        public ClaimsTransformationHandler(IClaimsTransformer transform)
+        public ClaimsTransformationHandler(IClaimsTransformer transform, HttpContext httpContext)
         {
             _transform = transform;
+            _httpContext = httpContext;
         }
 
         public IAuthenticationHandler PriorHandler { get; set; }
@@ -27,8 +30,12 @@ namespace Microsoft.AspNetCore.Authentication
                 await PriorHandler.AuthenticateAsync(context);
                 if (_transform != null && context?.Principal != null)
                 {
+                    var transformationContext = new ClaimsTransformationContext(_httpContext)
+                    {
+                        Principal = context.Principal
+                    };
                     context.Authenticated(
-                        await _transform.TransformAsync(context.Principal),
+                        await _transform.TransformAsync(transformationContext),
                         context.Properties,
                         context.Description);
                 }
--- a/src/Microsoft.AspNetCore.Authentication/IClaimsTransformer.cs
+++ b/src/Microsoft.AspNetCore.Authentication/IClaimsTransformer.cs
@@ -14,8 +14,8 @@ namespace Microsoft.AspNetCore.Authentication
         /// <summary>
         /// Provides a central transformation point to change the specified principal.
         /// </summary>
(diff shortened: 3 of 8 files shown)


src/Microsoft.AspNetCore.Authentication/ClaimsTransformationHandler.cs
```diff
@@ -27,8 +30,13 @@ public ClaimsTransformationHandler(IClaimsTransformer transform)
                 await PriorHandler.AuthenticateAsync(context);
                 if (_transform != null && context?.Principal != null)
                 {
+                    var transformationContext = new ClaimsTransformationContext
+                    {
+                        Context = _httpContext,
+                        Principal = context.Principal
+                    };
```
DovydasNavickas: Is it ok to create `transformationContext` object separately or should it be inlined into `TransformAsync` call?

src/Microsoft.AspNetCore.Authentication/ClaimsTransformationHandler.cs
```diff
@@ -12,10 +13,12 @@ namespace Microsoft.AspNetCore.Authentication
     public class ClaimsTransformationHandler : IAuthenticationHandler
     {
         private readonly IClaimsTransformer _transform;
+        private readonly HttpContext _httpContext;
 
-        public ClaimsTransformationHandler(IClaimsTransformer transform)
+        public ClaimsTransformationHandler(IClaimsTransformer transform, HttpContext httpContext)
```
DovydasNavickas: I've added `HttpContext` instead of actual `ClaimsTransformationContext`, because `AuthenticateAsync` method below receives `AuthenticateContext` param with current `Principal` in it.
Therefore I left construction of `ClaimsTransformationContext` just before `TransformAsync` call. This way it gets most recent `Principal` from `AuthenticateContext`.

src/Microsoft.AspNetCore.Authentication/IClaimsTransformer.cs
```diff
@@ -14,8 +15,8 @@ public interface IClaimsTransformer
         /// <summary>
         /// Provides a central transformation point to change the specified principal.
         /// </summary>
-        /// <param name="principal">The principal to transform.</param>
+        /// <param name="context"><see cref="ClaimsTransformationContext"/> containing principal to transform and current <see cref="HttpContext"/>.</param>
```
DovydasNavickas: Not sure if `<see cref="HttpContext"/>` should be here. For VS to find it here adding `using Microsoft.AspNetCore.Http;` is needed and when getting IntelliSense it adds `Http.HttpContext` anyways:

image


Maybe it should be written in plain text?

test/Microsoft.AspNetCore.Authentication.Test/Google/GoogleMiddlewareTests.cs
```diff
@@ -273,7 +273,7 @@ public class GoogleMiddlewareTests
             {
                 var transaction = await sendTask;
                 Assert.Equal(HttpStatusCode.Redirect, transaction.Response.StatusCode);
-                Assert.Equal("/error?FailureMessage=OMG"+UrlEncoder.Default.Encode(";Description=SoBad;Uri=foobar"), transaction.Response.Headers.GetValues("Location").First());
+                Assert.Equal("/error?FailureMessage=OMG" + UrlEncoder.Default.Encode(";Description=SoBad;Uri=foobar"), transaction.Response.Headers.GetValues("Location").First());
```
DovydasNavickas: Formatting code made this change. Should I exclude it or leave it as it is?

src/Microsoft.AspNetCore.Authentication/ClaimsTransformationContext.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Microsoft.AspNetCore.Http;
+using System.Security.Claims;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public class ClaimsTransformationContext
+    {
+        public HttpContext Context { get; set; }
+        public ClaimsPrincipal Principal { get; set; }
```
DovydasNavickas: @HaoK, you offered to use only getters. Should I leave only getters and take both params in a constructor or setters are ok?

HaoK: Setter seems fine for Principal in case of wanting to replace it.  I think its less reasonable for replacing the HttpContext, so make that readonly/ctor param

src/Microsoft.AspNetCore.Authentication/IClaimsTransformer.cs
HaoK: Plain text is fine then

test/Microsoft.AspNetCore.Authentication.Test/Google/GoogleMiddlewareTests.cs
HaoK: Its a good format change

src/Microsoft.AspNetCore.Authentication/project.json
```diff
@@ -19,7 +19,8 @@
       "version": "1.0.0-*"
     },
     "Microsoft.Extensions.Options": "1.0.0-*",
-    "Microsoft.Extensions.WebEncoders": "1.0.0-*"
+    "Microsoft.Extensions.WebEncoders": "1.0.0-*",
+    "Microsoft.AspNetCore.Http.Abstractions": "1.0.0-*"
```
Tratcher: Microsoft.AspNetCore.Http already pulls this in.

DovydasNavickas:
I've updated code according to comments and all checks have passed. Should I squash commits or are there any more comments?

HaoK:
Sure go ahead and squash and I'll merge, thanks!

HaoK:
Merged, thanks again!  f5c1c45db20fd4592a3f49d4dc1d39dc5f62108c

PR #745, opened 2016-03-15
OIDC & JWT event refactoring and cleanup

Tratcher:
#690 @brentschmaltz @muratg @Eilon @PinpointTownes @HaoK 
- Feedback from @vibronet around event names and placements.
- Consolidated some events that were close together or close to SecurityTokenValidator.ValidateToken.
- Enabling some of the AutomaticChallenge and AutomaticAuthenticate settings by default.
- Fixed up the samples do to runtime changes.


HaoK:
:shipit:

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
```diff
@@ -38,17 +28,22 @@ public interface IOpenIdConnectEvents : IRemoteAuthenticationEvents
         /// <summary>
         /// Invoked before redirecting to the identity provider to authenticate.
         /// </summary>
-        Task RedirectToAuthenticationEndpoint(RedirectContext context);
+        Task RedirectToIdentityProvider(RedirectContext context);
```
PinpointTownes: Not sure reintroducing the old name makes this event really clearer.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
```diff
@@ -38,17 +28,22 @@ public interface IOpenIdConnectEvents : IRemoteAuthenticationEvents
...
 
         /// <summary>
         /// Invoked before redirecting to the identity provider to sign out.
         /// </summary>
-        Task RedirectToEndSessionEndpoint(RedirectContext context);
+        Task RedirectToIdentityProviderForSignOut(RedirectContext context);
```
PinpointTownes: Idem.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
```diff
@@ -38,17 +28,22 @@ public interface IOpenIdConnectEvents : IRemoteAuthenticationEvents
...
 
         /// <summary>
         /// Invoked after "authorization code" is redeemed for tokens at the token endpoint.
         /// </summary>
-        Task TokenResponseReceived(TokenResponseReceivedContext context);
+        Task TokenResponseMessageReceived(TokenResponseMessageReceivedContext context);
```
PinpointTownes: What adding `Message` in this event? It sounds redundant, since a response is semantically a message.

HaoK: I think there must be some rule somewhere against having 5 words in a class name... :)

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/OpenIdConnectEvents.cs
```diff
@@ -39,17 +29,22 @@ public class OpenIdConnectEvents : RemoteAuthenticationEvents, IOpenIdConnectEve
...
         /// <summary>
         /// Invoked after "authorization code" is redeemed for tokens at the token endpoint.
         /// </summary>
-        public Func<TokenResponseReceivedContext, Task> OnTokenResponseReceived { get; set; } = context => Task.FromResult(0);
+        public Func<TokenResponseMessageReceivedContext, Task> OnTokenResponseMessageReceived { get; set; } = context => Task.FromResult(0);
+
+        /// <summary>
+        /// Invoked when an IdToken has been validated and produced an AuthenticationTicket.
+        /// </summary>
+        public Func<TokenValidatedContext, Task> OnTokenValidated { get; set; } = context => Task.FromResult(0);
```
PinpointTownes: FWIW, I'm not a huge fan of the new name, which will sound too restrictive for most people: the validation routine doesn't just validate the `id_token`, but also the access token, the code and the nonce.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/TokenValidatedContext.cs
```diff
@@ -8,16 +10,22 @@
...
+        public TokenValidatedContext(HttpContext context, OpenIdConnectOptions options)
             : base(context, options)
         {
-            Properties = properties;
         }
 
-        public AuthenticationProperties Properties { get; }
+        public AuthenticationProperties Properties { get; set; }
+        
+        public JwtSecurityToken JwtSecurityToken { get; set; }
```
PinpointTownes: I'm sure I've already mentioned it, but `IdToken` or `IdentityToken` would be clearer.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -350,27 +349,38 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
 
             try
             {
-                var messageReceivedContext = await RunMessageReceivedEventAsync(authorizationResponse);
-                if (CheckEventResult(messageReceivedContext, out result))
+                AuthenticationProperties properties = null;
+                if (!string.IsNullOrEmpty(authorizationResponse.State))
+                {
+                    properties = Options.StateDataFormat.Unprotect(Uri.UnescapeDataString(authorizationResponse.State));
```
PinpointTownes: Is `Uri.UnescapeDataString` really mandatory? I'm pretty sure the OAuth2 MW doesn't use it.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/TokenValidatedContext.cs
HaoK: At the very least, naming it something different from the type? Even just calling it Token I think would be an improvement...

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -350,27 +349,38 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
+                }
+
+                var messageReceivedContext = await RunMessageReceivedEventAsync(authorizationResponse, properties);
```
PinpointTownes: Curious, what's the concrete use case behind this change? Why do we need to restore the properties bag before calling the event?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -487,6 +498,18 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
                         {
                             nonce = ReadNonceCookie(nonce);
                         }
+
+                        var tokenValidatedContext = await RunTokenValidatedEventAsync(authorizationResponse, tokenEndpointResponse, properties, ticket, jwt, nonce);
+                        if (tokenValidatedContext.CheckEventResult(out result))
+                        {
+                            return result;
+                        }
+                        authorizationResponse = tokenValidatedContext.ProtocolMessage;
```
PinpointTownes: It really became noisy, IMHO.
But I guess that's the price to pay if you want to keep all these properties publicly settable.
(I've never understood why, BTW, as the types themselves are already fully mutable)

test/Microsoft.AspNetCore.Authentication.Test/JwtBearer/JwtBearerMiddlewareTests.cs
```diff
@@ -215,67 +183,37 @@ public class JwtBearerMiddlewareTests
...
                 AutomaticAuthenticate = true,
                 Events = new JwtBearerEvents()
                 {
-                    OnReceivingToken = context =>
+                    OnMessageReceived = context =>
                     {
                         context.Token = "CustomToken";
                         return Task.FromResult<object>(null);
-                    },
-                    OnReceivedToken = context =>
```
PinpointTownes: Alright, I guess this change is the last step of the "completely revamp the bearer middleware to get rid of simple scenarios" plan initiated by @brentschmaltz last year? :smile: 

Without this event, it will be hard for people to simply associate an `AuthenticationTicket` with a token without relying on the default `ISecurityTokenValidator` logic.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
PinpointTownes: Yeah. Specially here, where it's absolutely not needed.

muratg:
cc @vibronet

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: Hmm, internally StateDataFormat uses Base64UrlTextEncoder. UnescapeDataString may be redundant.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: AuthProperties were the only significant contribution of AuthorizationResponseReceived, so by exposing them earlier we do not need the other event. Unless there's another scenario for AuthorizationResponseReceived that can't be managed with MessageReceived?

test/Microsoft.AspNetCore.Authentication.Test/JwtBearer/JwtBearerMiddlewareTests.cs
Tratcher: Or replacing the ISecurityTokenValidator.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
brentschmaltz: There is no AuthenticationEndpoint. There is an AuthorizationEndpoint. The thought is that this name spans different protocols: WsFed, OIDC, OAuth.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
brentschmaltz: Abstracting away protocol.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
brentschmaltz: If 'properties' is null, then the user will need to examine authorizationResponse.State to determine if 'State' was missing OR Unprotect failed. A delegate could be passed, so the users could get the properties. May be a bit more obvious what is going on.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
PinpointTownes: lol, it's really nitpicking... though it's not the official name, "authentication endpoint" is not wrong since an OIDC request is both an authorization and an authentication request.

`RedirectToIdentityProvider` was renamed a few months because it was not clear enough. I'm not sure why you mentioned WS-Fed, but I guess you'd like to have a similar name for both protocols/middleware? Sounds like a bad idea, IMHO.

`RedirectToAuthorizationEndpoint` would be much better.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
PinpointTownes: It's an OpenID Connect-specific middleware, what's the point of trying to abstract the protocol?

test/Microsoft.AspNetCore.Authentication.Test/JwtBearer/JwtBearerMiddlewareTests.cs
PinpointTownes: I'm fine. This middleware was totally inappropriate for simple token validation since @brentschmaltz's last year changes anyway.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
Tratcher: Hmm, we do call it RedirectToAuthorizationEndpoint for all of our other providers.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/TokenValidatedContext.cs
Tratcher: We use `Token` to refer to the raw `string` elsewhere. How about `ValidatedToken`?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
vibronet: @PinpointTownes Here there's some more background. Although people who are familiar with the protocol details should be able to consume the object model, familiarity and full understanding of the protocol stack should not be a strict requirement. Mentioning "authorization" would be misleading given that that's not the main reason for which someone would use the middleware (signing in in itself does not entail authorization) and only people who know about the authorization endpoint and its role in those grants would understand what this is about. "AuthenticationEndpoint" would be also confusing, given that authentication is not the only reason for which this roundtrip is made (consent alone is a possibility) and the assonance with "authorization" would be a source of further confusion. Sticking with the old name has the advantage of describing a phase of the exchange without tying to the details  of this specific protocol - a developer who needs to augment the message back to the authority will have a coherent place to work in across all redirect based protocols.
@Tratcher re:other middlewares, that's the usual discussion. The other OAuth2 based,. provider specific middlewares are actually ALWAYS redirecting for authorization purposes - you are piggybacking on that for "simulating" sign in but the semantic of the operation is NEVER sing in (sign in with the provider happens as a side effect and it is performed transparently from your middleware, your entry and exit points are of an authorization request regardless of what happens in the middle).

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
vibronet: @PinpointTownes it's OpenID Connect-specific, but it is not meant to be usable only by people who are familiar with OpenID Connect. Basic extensibility should be accessible to people who understand the high level pattern but aren't familiar with the trappings of the specific protocol. As the customizations get more granular, I agree that requiring familiarity is inevitable: but every developer understand what sign out is, without the need to know what the session endpoint is or how it works.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
vibronet: @HaoK are you kidding or are you serious? If the latter, could you provide a reference to the rule? :)
@PinpointTownes that's for being coherent with messagereceived - that should help the developer to understand what kind (as in class) of event they are receiving.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/OpenIdConnectEvents.cs
vibronet: Are you referring to OnTokenValidated? That is supposed to trigger only after the id_token has been validated. The fact that other validations take place at other times should not change the fact that the id_token was validated here. Also: what do you mean by validating the access token?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
HaoK: I'm obviously kidding about the rule, but I agree with pinpoint that these names feel long without adding any clarity.  Is TokenReceived not sufficient?  It aligns with TokenValidated

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/TokenValidatedContext.cs
HaoK: Maybe just drop the Jwt?  `SecurityToken` or `IdentityToken` like @PinpointTownes suggested?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: Removed.

Tratcher:
Updated with minor review feedback.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
vibronet: @HaoK see, I think that your comment makes exactly my point. This event is not signaling that we received a token - it is signaling that we received a message from the token endpoint. The kind of extensions a developer can inject here might have nothing to do with the token itself. Initially I suggested that here we should simply raise another MessageReceived, and have something in the event parameters to signal the nature of the message - but @Tratcher convinced me otherwise. This kind of message is specific enough to the OIDC flows that, like the authorizationcodereceived, it deserves a protocol specific name (whereas MessageReceived, meant as a frontchannel auth message, is an artifact that is common to all redirect based protocols).
Note, I am not saying that names length isn't important - but unless we end up in the realm of the absurd (exhibit A: usewindowsactivedirectorybearerauthentication) I would consider clarity the highest order bit. Perhaps we have different threshold for absurd, tho :) do you still feel strongly against adding "message" after having read the above?

HaoK: >do you still feel strongly against adding "message" >after having read the above?

Ok but then either message or response should be sufficient to convey that distinction as opposed to both. TokenResponseRecieved or TokenMessageRecieved as opposed to both.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
PinpointTownes: Like @HaoK, I really think the events model is for advanced scenarios/users (TBH, it's not the easiest thing to use, though it makes total sense when you understand how it works).

Since it's advanced stuff, I think we can safely assume that the developers using them are familiar enough with OIDC to directly alter the authorization message (if you don't know anything about OIDC, playing with `OpenIdConnectMessage` seems a bit risky anyway).

In this case, I don't see why this event should have a super user-friendly name, since it's not meant for average users.

If you have particular scenarios that require using this event in mind, maybe you should list them here to determine if we can't introduce new first-class properties. Personally, I'm not opposed to having more options, even if they are mainly AAD concepts (as long as they are general enough to be reused by other providers, e.g the `Resource` property that determines the audience of the access tokens returned by AAD)

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
PinpointTownes: Same remark here, I don't think the events are made for "basic extensibility": they expose low-level properties (mainly `OpenIdConnectMessage`) that requires a minimum understanding of the OIDC logout parameters.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/OpenIdConnectEvents.cs
PinpointTownes: > Are you referring to OnTokenValidated?

Yes.

> Also: what do you mean by validating the access token?

When receiving both an identity and an access token, Wilson validates them both in the same call (by computing the access token hash and comparing it to the `at_hash` claim extracted from the identity token). That's why I think `OnTokenValidated` seems too restrictive, as it doesn't simply validate the `id_token` but covers more things.

vibronet: This event has been historically used for performing any work that requires knowledge of the incoming user. The goal is signaling the fact that the token carrying identity information has been validated and its content is now available for processing. This should be consistent with all the other middlewares which validate tokens. The fact that the middleware as a whole *might* be doing more work in the OIDC case is a no-op for the implementer - if they want to do something with the access token (if present) they can do it later on.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
vibronet: We actually have abundant data from the WIF days, showing that most developers used the corresponding event without any deep knowledge of WS-Federation. The ability to modify the outgoing messages is a generic extensibility point, which has nothing to do with Azure AD - again, this event was in the WIF OM well before Azure AD even existed. _Modifying _the message does require more knowledge of the protocol, however _augmenting _the message  does not - it's common for people to add extra stuff to communicate to their own provider. And even modifying the message does not automatically entail _perfect_ knowledge of the protocol. A developer can have some familiarity with protocol artifacts that touch directly their scenarios (for example, scopes required by the API they talk to) but not with the protocol endpoints or flows.

PinpointTownes: Heh, I've never said the events model was specific to AAD :open_mouth: 

I just said that `OpenIdConnectOptions` already has AAD-specific options like `Resource` and that I'm much more comfortable with adding new first-class options than forcing folks to use the events approach, which requires more work and is more complicated for beginners.

Anyway, it doesn't change anything to the fact I believe `RedirectToIdentityProvider` is a poor name.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
Tratcher: Agreed, the length is excessive without strong benefit. Reverting.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
vibronet: I agree it does not sound awesome, but it has the advantage of being a clear correspective of the events that have been used for handling the same phase in the last 2 generations of auth software. Unless we have a new name that has excellent explicative power and does not require deep protocol knowledge, this looks the alternative with the least impact.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
vibronet: I realized I didn't comment on this one. The comment is the same as above: events in WIF and Katana have been, and are being used for customizing the sw behavior by people who do not have perfect knowledge of the protocol, for legit reasons. Details in the note above.

PR #752, opened 2016-03-23
Ensure an `ObjectPoolProvider` is registered

dougbu:
- take advantage of aspnet/HttpAbstractions#561 fix


dougbu:
Note aspnet/HttpAbstractions#594 is _not_ a breaking change. This PR just ensures the authentication middleware takes advantage of that perf improvement.

dougbu:
Will do this in Hosting instead.

PR #756, opened 2016-03-28
Add comments to AddAuthentication extension methods

javiercn:


Eilon:
Don't care much either way because we often use either phrasing, but :shipit:

PR #759, opened 2016-03-31
Add fallback for username logging in AuthZ

HaoK:
cc @Tratcher @blowdart

### DIFF
--- a/src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationService.cs
+++ b/src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationService.cs
@@ -74,21 +74,14 @@ namespace Microsoft.AspNetCore.Authorization
                 }
                 return GetClaimValue(identity, "sub")
                     ?? GetClaimValue(identity, ClaimTypes.Name)
-                    ?? GetClaimValue(identity, ClaimTypes.NameIdentifier)
-                    ?? GetClaimValue(identity, ClaimTypes.WindowsAccountName)
-                    ?? GetClaimValue(identity, ClaimTypes.X500DistinguishedName)
-                    ?? GetClaimValue(identity, ClaimTypes.PrimarySid)
-                    ?? GetClaimValue(identity, ClaimTypes.Sid)
-                    ?? GetClaimValue(identity, ClaimTypes.Upn);
+                    ?? GetClaimValue(identity, ClaimTypes.NameIdentifier);
             }
             return null;
         }
 
         private static string GetClaimValue(IIdentity identity, string claimsType)
         {
-            return (identity as ClaimsIdentity)
-                ?.FindFirst(c => string.Equals(c.Type, claimsType, StringComparison.OrdinalIgnoreCase))
-                ?.Value;
+            return (identity as ClaimsIdentity)?.FindFirst(claimsType)?.Value;
         }
 
         public async Task<bool> AuthorizeAsync(ClaimsPrincipal user, object resource, string policyName)
--- a/src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationService.cs
+++ b/src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationService.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Security.Claims;
+using System.Security.Principal;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationService.cs
```diff
@@ -51,16 +52,45 @@ public DefaultAuthorizationService(IAuthorizationPolicyProvider policyProvider,
...
+            {
+                var name = identity.Name;
+                if (name != null)
+                {
+                    return name;
+                }
+                return GetClaimValue(identity, "sub")
+                    ?? GetClaimValue(identity, ClaimTypes.Name)
+                    ?? GetClaimValue(identity, ClaimTypes.NameIdentifier)
+                    ?? GetClaimValue(identity, ClaimTypes.WindowsAccountName)
```
blowdart: Vittorio wanted "preferred_username" as well. So once that's in :shipit:

HaoK: You do realize this is only for the log message right?  This already seems like insane overkill, my vote is we pick the 3 most common ones as opposed to this crazy list...

PinpointTownes: I totally agree with @HaoK, this list is insane and not every claim is accurate for a "username" (e.g the JWT-specific `sub` claim shouldn't be used as a display name, unless you want to end up with random strings) :+1: 

IMHO, we should only use `IIdentity.Name` here and let the identity implementation decide what's appropriate for a username (`ClaimsIdentity` checks for a claim corresponding to `NameClaimType`).

blowdart: @PinpointTownes Unfortunately not every provider sets it. Because, meh. Fine, name, sub, nameidentifier,

PinpointTownes: > Unfortunately not every provider sets it

In this case, I'm convinced we shouldn't try to guess what a valid username could be: if the provider didn't add a name claim, it probably means the (user)name was not available.

src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationService.cs
```diff
@@ -51,16 +52,45 @@ public DefaultAuthorizationService(IAuthorizationPolicyProvider policyProvider,
...
+                    ?? GetClaimValue(identity, ClaimTypes.Sid)
+                    ?? GetClaimValue(identity, ClaimTypes.Upn);
+            }
+            return null;
+        }
+
+        private static string GetClaimValue(IIdentity identity, string claimsType)
+        {
+            return (identity as ClaimsIdentity)
+                ?.FindFirst(c => string.Equals(c.Type, claimsType, StringComparison.OrdinalIgnoreCase))
```
PinpointTownes: Curious: why not using the simpler overload that takes a claim type?

src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationService.cs
HaoK: Cool, so checking .Name, then sub, Name, NameIdentifier.

src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationService.cs
HaoK: Good point, fixed thanks!

src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationService.cs
```diff
@@ -51,16 +52,38 @@ public DefaultAuthorizationService(IAuthorizationPolicyProvider policyProvider,
...
+        {
+            var identity = user?.Identity;
+            if (identity != null)
+            {
+                var name = identity.Name;
+                if (name != null)
+                {
+                    return name;
+                }
+                return GetClaimValue(identity, "sub")
```
PinpointTownes: If you really want to keep `sub` and `NameIdentifier`, please consider changing the order of these claims to `Name`, `NameIdentifier` and `sub`.

HaoK: Well we are checking Identity.Name first, but the order comes from @blowdart, I'll let you guys hash out the appropriate priority :)

src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationService.cs
PinpointTownes: If you want to use fallback claims, I'm fine, but note that it might be a bit confusing: the authorization block would log a username the developer couldn't retrieve in his own code by just using `identity.Name`.

src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationService.cs
blowdart: I'd hope people setting name would have set the name claim type :)

src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationService.cs
HaoK: Sure but current behavior is worse as you get user: null which makes it impossible to correlated, at least you have a chance to find the identifier matching a user this way...that's the intent anyways...

PinpointTownes: As long as it's just for logging (and is not extended to the authentication middleware themselves), I'm fine.

HaoK:
4d6ad51f8a9354d8e62cf8d66db97bd991b7d93c

Issue #786, opened 2016-04-18
Better abstraction for AuthorizationPolicyBuilder


MonkeyJamboree:
I suggest moving most of the `Require*` methods on AuthorizationPolicyBuilder to extensions that rely on either the `AddRequirements` method or directly adds the requirements to the `Requirements` property.

The intent of this change would be to keep the builder's API simple and abstracted away from any given `IAuthorizationRequirement` implementation.

Eilon:
At this time we are not looking to make such changes to the system.

Issue #785, opened 2016-04-18
Authorization Block comments and change to IAuthorizationService


MonkeyJamboree:
Hey all,

After some discussion with @PinpointTownes about the JWT middleware and looking at the Authorization block (at https://github.com/aspnet/Security/tree/dev/src/Microsoft.AspNetCore.Authorization), I have some comments and suggestions for improvement. 

First off, Awesome job at abstracting out the interfaces, particularly the `IAuthorizationRequirement` abstraction and `AuthorizationHandler` class. I think it was implemented well and allows the stack to handle custom implementations and custom authorization requirements as needed by consumers. I also like that the different `IAuthorizationRequirement` are completely separated out.

I'll post the improvements as separate issues so they can be tracked and resolved independently. This is the first:

The `IAuthorizationService` exposes `AuthorizeAsync` that requires a `ClaimsPrincipal`. Several use cases come to mind wherein a consumer of the Authorization Block might want to make authorization decisions based on something in the `HttpContext` itself rather than anything available in a `ClaimsPrincipal`:

- IP Address or host name based authorization for enterprise-level server trust authorization
- Using `AuthenticationProperties` in an `AuthenticationTicket` that shouldn't really be exposed as claims in a principal (example: OAuth/OIDC scopes, as discussed at https://github.com/aspnet-contrib/AspNet.Security.OAuth.Extensions/issues/8)
- Using tokens or other authentication mechanisms that are not explicitly tied to a principal, that may not conform to a standard or be associated with a principal (such as custom time- or use quantity-limited tokens, or server- or machine-level tokens)
- Custom `HttpRequest`/`HttpResponse` header or query string decisions (handling something similar to Google's "API key" seen at https://developers.google.com/+/web/api/rest/oauth#acquiring-and-using-an-api-key or writing custom WWW-Authenticate headers for simplistic authentication/authorization schemes)
- Request or server date/time scenarios where services are only available between specific hours
- Probably many others including supporting authorization schemes from legacy systems while migration to new approaches happens

The main reason behind requesting this change is that a `ClaimsPrincipal` is not really all that appropriate for all authorization decisions without a lot of hacky workarounds (such as creating specific custom claims for standard things already available in the `HttpContext` class).

MonkeyJamboree:
What, exactly, is the reasoning for using ClaimsPrincipal instead of something more flexible or all-encompassing? Also, could generics be used on the handler/requirements so that the stack can be used without any explicit dependencies?

[...]

PR #791, opened 2016-04-22
Authorization abstraction and dependency improvements

MonkeyJamboree:
I created this branch to improve the Authorization block such that it would have no direct requirement on `ClaimsPrincipal` and instead take a more generic approach. I attempted to this so in a backward-compatible way for code that already uses it, however there are some relatively minor breaking changes, particularly with `AuthorizationContext`.

Pros to this approach: With these changes, any `object` could be used to make authorization decisions (be it a `ClaimsPrincipal`, `AuthenticationTicket`, `HttpContext`, list of roles or scopes, and so on) and custom `IAuthorizationRequirement`/`AuthorizationHandler`s could be made to use anything for authorization. Also, the `AuthorizationPolicyBuilder` has been simplified and extensions made to support the existing builder functionality that sets up required claims, roles, etc. 

Drawbacks to this approach: `object` needs to be cast to the required type or the type inferred once within an `AuthorizationHandler`, and consumers may not know what type to use when creating an `AuthenticationContext` manually. I attempted an approach using generics but felt that it would disrupt the existing codebase too much and indeed be much harder for consumers to use.

@PinpointTownes @DovydasNavickas @blowdart @Eilon and whoever else would like to comment... I welcome feedback on this change. I felt these changes necessary and decided to take on the task myself since the issues requesting these changes was closed as `wontfix`.


Eilon:
@MonkeyJamboree it would be helpful to understand exactly which specific concrete scenarios are not possible or are too difficult to achieve with the current implementation and why. In https://github.com/aspnet/Security/issues/786 and https://github.com/aspnet/Security/issues/785 there are some relatively abstract concepts and design topics, but it isn't clear to me the specific scenarios that are being addressed. Though there are some specific scenarios mentioned, it isn't clear why they can't be implemented with the current implementation.

blowdart:
The vast majority of our use cases are built on identities. However MVC goes further, deriving it's own AuthorizationContext, which does include the request, IP address and everything else. That would be the "ideal" way to do this, because those details framework specific. We have tried to make authorization not base itself on assuming MVC etc.

Relying on people to make assumptions, and cast objects to something that might be there is a big overhead, both in terms of conceptual understanding, ensuring correctness and speed. The drawback you acknowledge is much larger than any advantage.

MonkeyJamboree:
@Eilon There are a plenty of use cases that don't seem to really fit with using a `ClaimsPrincipal` as the current implementation does. That doesn't mean it's not possible or necessarily hard to do so. It's actually more of a conceptualization problem and misuse of a data structure, in my eyes. In cases that don't involve identities, forcing things to use them simply doesn't make sense.

Use Case 1: Authorization based on something peripheral or ancillary to a `ClaimsPrincipal`.
In this case, we have a `ClaimsPrincipal`, but also something extraneous - Examples include the additional properties required for the MVC framework (`HttpContext`, `RouteData`, etc as seen on an `ActionContext`) and `AuthenticationTicket` (as it also has `AuthenticationProperties`). Shoving the additional properties from the peripheral data into the principal or identity isn't the best option as you'd then essentially be using a `ClaimsPrincipal` as a fancy `Dictionary` or collection of properties.

Use Case 2: Authorization based on something that won't easily go into a `Claim` or set of `Claim`s.
In some cases, you might want to do authorization based on a large, complex hierarchical data model. Having to serialize a complex model or set of models into strings to be placed into Claims makes little to no sense. Example:
```
public class UserDataModel
{
    public Guid Id { get; set; }
    public string Name { get; set; }
    public IReadOnlyCollection<SubModel<IData>> Props { get; set; }
    public ConcreteDataModelA { get; set; }
    public ConcreteDataModelB { get; set; }
    public IEnumerable<string> Roles { get; set; }
    public UserAddress Address { get; set; }
    public string XmlModelRepresentation { get; set; }
}

public interface IData { }

public class StringData : String, IData { }

public class EnumData : Enum, IData { }

public class HierarchicalData : IData
{
    ICollection<IData> SubModelData { get; set; }
}

public class SubModel<T>
{
    public string Name { get; set; }
    public T Data { get; set; }
    public DateTime DateAdded { get; set; }
}

public class ConcreteDataModelA { /* properties */ }

public class ConcreteDataModelB { /* properties */ }

public class UserAddress
{
    public string Street { get; set; }
    public IRegion State { get; set; }
    // and so on
}

public interface IRegion { }

public class State IRegion
{
    public string FullName { get; set; }
    public char[] Abbreviation { get; set; }
}
```

This is similar to some legacy data models I have seen around. Many of the model properties could be transformed into `Claim`s in some way, however how to do so might be problematic and complex. Authorization based on this model would be complex in and of itself, let alone attempting to transform the model before doing the authorization. Instead of requiring the transformation (or doing some sort of serialization/deserialization to a `string`), allowing consumers to simply use and make authorization decisions based on a given model as-is would be beneficial.

Use Case 3: No authentication has or will have happened.
While extremely uncommon and not recommended in most cases, it is possible one would want to make authorization decisions based on information when authentication can't or won't be performed. Service- or machine-level communication could very well require authorization without authentication, such as two servers who have a predefined trust relationship communicating on specific ports. Authorizing connections based on an IP address, Host name, etc and applying a whitelist or blacklist to that is a common theme in these regards. Authorization decisions in this regard could additionally be applied alongside a `ClaimsPrincipal` to ensure users are only allowed to access certain applications or services from specific machines, but the user themselves may migrate in and out of allowed zones. Forcing all of this into a principal/identity is certainly possible, but does not seem appropriate.

@blowdart You say that MVC derives its own AuthorizationContext, but from my searching, the MVC stack doesn't use the `AuthorizationContext` that's in the Authorization block at all. Furthermore, it seems to not use a lot of what's built in, and as far as I can tell only hooks into the block in AuthorizeFilter, and then creates a default (anonymous) principal even when authentication fails (which seems odd to me). This alone tells me that there's a problem inherent to the block. Since the framework doesn't really even use it, what's the reason for requiring the usage of a `ClaimsPrincipal` instead of something more generic/agnostic?

Making authorization not base itself on MVC is good, however there's still an assumption at requiring identities or principals when that may not be appropriate. Yes, that's the most common case and more than the majority of scenarios, but for anything that doesn't quite fit, custom code and workarounds need to be implemented. I have a feeling this custom code will need to happen more often than expected because of the limitations.

Regarding the issues with making assumptions and casting, that's only the initial approach I came up with so far. There may well be a better approach or a way to use generics that hasn't been considered that would work much better. Besides, `AuthorizationContext`'s resource constructor parameter is an `object` as well that would need a cast or making assumptions, so that's no different if the resource needs to be used. If anyone has some constructive criticism on how to improve this design, or perhaps a better direction to attempt, I welcome it. I may take another stab at it with generics, interfaces, or both, but comments are appreciated.

blowdart:
There's one other thing; authorization is generally accepted to be checking a user's credentials and their ability to perform the action requested. A user is not a machine (IP address), route (HttpContext) or anything else. Think of your drivers license, or your atm card, a passport or even a OTP. Those are all user attributes. The presence of an ATM card isn't enough to withdraw money, the user must authenticate before authorization happens.

IP address access is usually by filtering, before authentication/authorization happens. Thus, really, we're not going to widen authorization to be anything other than dealing with a user's credentials. You could, of course, write authentication middleware to authenticate based on IP address, construct user credentials from it, and then act upon it within authorization.

Whilst we appreciate the feedback, I feel it's going to be more confusing for the vast majority of developers, and there are other ways to achieve what you want.

Eilon:
Given @blowdart 's comments, which I agree with, I am closing this PR. We appreciate the feedback and design thoughts, but unfortunately this change does not fit with the model we are developing.

PinpointTownes:
> There's one other thing; authorization is generally accepted to be checking a user's credentials and their ability to perform the action requested.

TBH, I can't understand why your definition of "authorization" is that limited (though it explains why the authorization block is itself limited to basically only user claims-based scenarios).

IMHO, authorization is so much more than just "checking a user's credentials and their ability to perform the action requested".

To reuse your ATM card sample, your bank will probably want to ensure that the ATM is a trusted machine and still has money before authorizing the transaction. These two variables are niehter user attributes/credentials (since they are directly bound to the ATM, not the user) nor resources (the "resource" here is the transaction itself) and yet may be fundamental parts of your authorization process.

Note that this decision also impacts simple scenarios like ensuring an OAuth2/OIDC client app' has the appropriate scopes before authorizing a request (https://github.com/aspnet-contrib/AspNet.Security.OAuth.Extensions/issues/8#issuecomment-210796367).

blowdart:
But those are in additional to the authentication the person performs (the identity of the machine is an identity), and are still authenticating part of the process. Authorization then builds on that identity.

PinpointTownes:
It is. Unfortunately, the authorization block doesn't offer anything for delegation chains (`ClaimsIdentity.Actor`) so having two distinct identities (one for the user, one for the caller) is really hard in practice. Merging the two in a single one is of course very bad from a semantic perspective, since the user and the caller are two different entities.

blowdart:
It's not hard at all. The entire stack supports multiple identities, that's what claims principal *does*

Anyway, regardless, this change is not being made.

PinpointTownes:
> It's not hard at all. The entire stack supports multiple identities, that's what claims principal does

I'm not talking about the entire stack, I'm talking about authorization. You wouldn't merge the two identities in the same principal, would you? (representing distinct entities with the same principal doesn't make sense, sorry).

BTW, do you have an example of how to do scopes authorization? Definitely worth a page in the doc' IMHO.

PinpointTownes:
> Anyway, regardless, this change is not being made.

Understood. Just wanted to make sure you were not missing obvious scenarios. You know, real world apps... :smile:

MonkeyJamboree:
> Given @blowdart 's comments, which I agree with, I am closing this PR. We appreciate the feedback and design thoughts, but unfortunately this change does not fit with the model we are developing.

Perhaps then it should be renamed to "ClaimsAuthorization" instead, or something that indicates it's tied to the model you're developing.

> There's one other thing; authorization is generally accepted to be checking a user's credentials and their ability to perform the action requested. A user is not a machine (IP address), route (HttpContext) or anything else.

Generally, perhaps, but according to https://technet.microsoft.com/en-us/library/ff687657(v=ws.10).aspx (and quite a few other resources out there as google searches yield), Authentication is essentially "who are you?" and Authorization is "given who you are, what can you do?" While related, and while Authorization is often reliant upon Authentication, they're two separate concerns.

> You could, of course, write authentication middleware to authenticate based on IP address, construct user credentials from it, and then act upon it within authorization.

> But those are in additional to the authentication the person performs (the identity of the machine is an identity), and are still authenticating part of the process. Authorization then builds on that identity.

I hadn't actually thought about the machine as having an "identity" so to speak, but I suppose that could be the case. I worry that conceptually things are being forced into boxes they don't belong in. I was just trying to provide something that's a better conceptual fit, though I suppose I failed in that.

I suppose next time I think about contributing code to help, I'll state my points on the issue(s) first, so none of us wastes time or effort in debating a PR.

PR #796, opened 2016-04-23
Cleanup social sample

benaadams:
Move Authentication from startup to extensions; reduce the using size

### DIFF
--- a/samples/SocialSample/AuthenticationExtensions.cs
+++ b/samples/SocialSample/AuthenticationExtensions.cs
@@ -1,0 +1,257 @@
+﻿using System.Text.Encodings.Web;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authentication.Cookies;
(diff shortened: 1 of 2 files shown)


Tratcher:
I don't think we should change it like this.
A) This isn't a pattern we've used or recommended anywhere.
B) This is a sample to show how to set up these components, but this change obscures that. Extension methods are not easy to follow when just reading sources.

If Configure is too complicated, one pattern we have used is to break it down into submethods like: 
`void ConfigureCookies(IApplicationBuilder builder, IConfiguration config)`
`void ConfigureFacebook(IApplicationBuilder builder, IConfiguration config)`
etc..

benaadams:
Its more that the file starts with 27 using lines

benaadams:
Still ends up with 13 in each file, but its less of a shock :)

Happy for it to be closes, or to close it

PR #799, opened 2016-04-25
Update web.config and add publish tool

BrennanConroy:
@Tratcher @JunTaoLuo @Eilon

### DIFF
--- a/samples/CookieSample/project.json
+++ b/samples/CookieSample/project.json
@@ -26,5 +26,14 @@
   },
   "content": [
     "web.config"
-  ]
+  ],
+  "tools": {
+    "Microsoft.AspNetCore.Server.IISIntegration.Tools": {
+      "version": "1.0.0-*",
+      "imports": "portable-net45+wp80+win8+wpa81+dnxcore50"
+    }
+  },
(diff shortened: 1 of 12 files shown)


Tratcher:
:shipit:

PinpointTownes:
The approach used for this PR doesn't seem to be consistent with the one used for the cli-samples repo, that doesn't have any `web.config` (https://github.com/aspnet/cli-samples/pull/46).

Is including `web.config` (even if it doesn't have custom sections) the recommended approach?

/cc @moozzyk @Guardrex

moozzyk:
If you open a project with VS it will add a web.config for you if one is not there. I did not know that when I removed web.config files from cli-samples. The net effect for default settings will be the same though because publish-iis will create a web.config for you in the publish location if one does not exist - so you don't have to have a web.config file if you don't want one. A side effect of not having/wanting to have a web.config file and using VS and git is that you may need to add web.config to .gitignore or it will show up as a new file each time you open VS.

PinpointTownes:
Woot, thanks for the precious details, truly helpful :clap:

PR #817, opened 2016-05-13
Use a shared CookieSecurePolicy

Tratcher:
See https://github.com/aspnet/HttpAbstractions/issues/612
Depends on: https://github.com/aspnet/HttpAbstractions/pull/631
@blowdart @HaoK @muratg


muratg:
:shipit:

PR #822, opened 2016-05-17
Clean up samples

Tratcher:
Lots of the samples don't work after all of the hosting changes. This updates them to work, and makes them conform more to the templates.

@JunTaoLuo


samples/CookieSample/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
+
+namespace CookieSample
+{
+    public static class Program
+    {
+        public static void Main(string[] args)
+        {
+            var config = new ConfigurationBuilder().AddEnvironmentVariables("ASPNETCORE_").Build();
```
Tratcher: @glennc this is similar to our Docker discussion. Compare with the launchSettings.json.

JunTaoLuo:
:shipit:

Issue #801, opened 2016-04-27
Authorize GitHub causes infinite redirects or Correlation failed


Tratcher:
Trying to directly authorize one of the redirecting middleware causes infinite loops. This only works for middleware that directly contain the identity like Cookies or Bearer. We should throw NotSupportedException if you try to call AuthorizeAsync("GitHub").

`[Authorize(ActiveAuthenticationSchemes = "GitHub-AccessToken")]`

Chrome gets infinite redirects. IE does as well, but sometimes gets:
```
System.Exception: Failed to retrieve access token.
 AggregateException: Unhandled remote failure. (Failed to retrieve access token.)
Microsoft.AspNetCore.Authentication.RemoteAuthenticationHandler`1.<HandleRemoteCallbackAsync>d__5.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
Microsoft.AspNetCore.Authentication.RemoteAuthenticationHandler`1.<HandleRequestAsync>d__4.MoveNext()
```
Which appears to be caused by rate limiting.

Edge gets a different error:
```
•	System.Exception: Correlation failed.
•	AggregateException: Unhandled remote failure. (Correlation failed.)
o	Microsoft.AspNetCore.Authentication.RemoteAuthenticationHandler`1.<HandleRemoteCallbackAsync>d__5.MoveNext()
o	--- End of stack trace from previous location where exception was thrown ---
o	System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
o	System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
o	Microsoft.AspNetCore.Authentication.RemoteAuthenticationHandler`1.<HandleRequestAsync>d__4.MoveNext()
o	--- End of stack trace from previous location where exception was thrown ---
o	System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
o	System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
o	Microsoft.AspNetCore.Authentication.AuthenticationMiddleware`1.<Invoke>d__18.MoveNext()
o	--- End of stack trace from previous location where exception was thrown ---
[...]

Issue #667, opened 2016-01-21
Flow for authenticated but unauthorized users with OIDC is broken (infinite redirect)


DamianEdwards:
The [live.asp.net](https://github.com/aspnet/live.asp.net) web site is configured to authenticate with OIDC (Azure AD) **only** (as emitted by the template during File -> New). The `AdminController` is configured to require authorization with a policy "Admin", which basically just [asserts that the user is signed in with one of three usernames](https://github.com/aspnet/live.asp.net/blob/dev/src/live.asp.net/Startup.cs#L54).

The issue is that anybody in the Azure AD org (microsoft.com) can sign-in to the site (authenticate), but when they try to access the protected resource (/admin) the policy fails and they're redirected back to Azure to sign-in. As they're already signed in, Azure redirects back to the originally requested resource, which fails the policy, and the loop continues forever.

The desired flow would appear to be that in the case a user is authenticated but not authorized for a resource when using OIDC, a "Forbidden" page is shown (as configured by the app, e.g. /account/forbidden) with a suitable message and a button that enables the user to sign-in as another user. The button would basically sign them out, then redirect them back to the originally requested resource, which would start the proper authentication flow again.

@Tratcher @blowdart @vibronet

blowdart:
Ah, forbidden only kicks in when using cookies. @HaoK

Tratcher:
I think OIDC Challenge needs to check Authenticated(SignInScheme), and if true, call Forbid(SignInScheme).

Tratcher:
Note all the redirecting auth providers could have this issue if used stand-alone. They could all use the same fix.

Eilon:
Backlog for now.

LeandroPT:
Is this still for backlog?

Any workarounds?

Mendallas:
Sorry, I'm confused, 
What do we need to do to resolve this issue ? 
Could you give me an example ? 

Regards

Charles

Tratcher:
@Mendallas we made internal changes for 1.0 that should address this (https://github.com/aspnet/Security/pull/823). You shouldn't have to make any changes yourself. You can try the pre-release packages on myget.org or wait for 1.0 to release.

brockallen:
For now if you disable AutomaticChallenge on the OIDC MW, then it will fall thru to the cookie middleware and get the desired AccessDeniedPath rendering.

PR #823, opened 2016-05-17
Handle authorize and forbid for redirecting handlers.

Tratcher:
#667 - Authenticated but not authorized infinite loop
#801 - Authenticate remote provider by name

When a user lacks a specific claim the authorization service issues a Challenge. For Bearer or Cookies they accept this challenge, verify that they produced the identity, and switch to Forbid. This isn't happening for the remote providers because they do not maintain the identity, they delegate that to Cookies. As such they end up in an infinite challenge loop. See https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs#L310-L332 for reference on Challenge behaviors.

Similarly, `[Authorize(ActiveAuthenticationSchemes = OpenIdConnectDefaults.AuthenticationScheme)]` doesn't work because the remote auth handler does not store the identity internally so `HttpContext.Authentication.AuthenticateAsync(scheme)` always returns null. `[Authorize]` works because Cookies has AutomaticAuthenticate and OIDC has AutomaticChallenge.

Fix: When a remote handler is 'Authenticating', have it delegate to the handler specified by SignInScheme.

@HaoK @PinpointTownes Thoughts?

@vibronet I think this will break your background refresh sample where you challenge for the current auth type to refresh the identity. I think it can still work if we make an easy way to avoid ChallengeBehavior.Automatic.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
@@ -89,9 +89,32 @@ namespace Microsoft.AspNetCore.Authentication
 
         protected abstract Task<AuthenticateResult> HandleRemoteAuthenticateAsync();
 
-        protected override Task<AuthenticateResult> HandleAuthenticateAsync()
+        protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
         {
-            return Task.FromResult(AuthenticateResult.Fail("Remote authentication does not support authenticate"));
+            // Most RemoteAuthenticationHandlers will have a PriorHandler, but it might not be set up during unit tests.
+            if (PriorHandler != null)
+            {
+                var authenticateContext = new AuthenticateContext(Options.SignInScheme);
+                await PriorHandler.AuthenticateAsync(authenticateContext);
+                if (authenticateContext.Accepted)
+                {
+                    if (authenticateContext.Error != null)
+                    {
+                        return AuthenticateResult.Fail(authenticateContext.Error);
+                    }
+
+                    if (authenticateContext.Principal != null)
+                    {
+                        return AuthenticateResult.Success(new AuthenticationTicket(authenticateContext.Principal,
+                            new AuthenticationProperties(authenticateContext.Properties), Options.AuthenticationScheme));
+                    }
+
+                    return AuthenticateResult.Fail("Not authenticated");
+                }
+
+            }
+
+            return AuthenticateResult.Fail("Remote authentication does not support authenticate");
         }
 
         protected override Task HandleSignOutAsync(SignOutContext context)
@@ -104,9 +127,11 @@ namespace Microsoft.AspNetCore.Authentication
             throw new NotSupportedException();
         }
 
-        protected override Task<bool> HandleForbiddenAsync(ChallengeContext context)
+        protected override async Task<bool> HandleForbiddenAsync(ChallengeContext context)
         {
-            throw new NotSupportedException();
+            var challengeContext = new ChallengeContext(Options.SignInScheme, context.Properties, ChallengeBehavior.Forbidden);
+            await PriorHandler.ChallengeAsync(challengeContext);
+            return challengeContext.Accepted;
         }
 
         protected virtual void GenerateCorrelationId(AuthenticationProperties properties)
--- a/samples/OpenIdConnectSample/Startup.cs
+++ b/samples/OpenIdConnectSample/Startup.cs
@@ -6,6 +6,7 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Authentication;
+using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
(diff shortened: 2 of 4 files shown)


HaoK:
Hrm this feels a bit weird, but I guess that's what this handler boils down to logically, we want the challenge to go remote, but Forbid/Authenticate really are going against the SignInScheme.

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -89,9 +89,27 @@ public abstract class RemoteAuthenticationHandler<TOptions> : AuthenticationHand
...
+            var authenticateContext = new AuthenticateContext(Options.SignInScheme);
+            await PriorHandler.AuthenticateAsync(authenticateContext);
+            if (authenticateContext.Accepted)
+            {
+                if (authenticateContext.Error != null)
+                {
+                    return AuthenticateResult.Fail(authenticateContext.Error);
+                }
+                // Compare the ClaimsIssuer to make sure the identity was produced by this provider and not by some other one.
+                if (string.Equals(authenticateContext.Principal?.Identity?.AuthenticationType, Options.ClaimsIssuer, StringComparison.Ordinal))
```
HaoK: Is this something we explicitly need to guard against?  If so, is ClaimsIssuer really a strong enough guard to ensure that the identity was produced by this provider?

Tratcher: I think it would be very strange for `[Authorize("Github")]` to return a Facebook identity. I'm imagining a scenario where you need to sign in with a specific provider to access specific functionality like a github API call.  I'll update the Social Sample. Of course this approach doesn't work when you're using Identity as it has its own auth types.

ClaimsIssuer is used to set the ClaimsIdentity AuthType in all the handlers, and it's set to the middleware AuthType by default.

HaoK: I guess my point was more that the remote handler shouldn't care what is in the sign in identity. And if something like claims transformation changes the issuer would now block this behavior.

PinpointTownes:
> When a user lacks a specific claim the authorization service issues a Challenge. For Bearer or Cookies they accept this challenge, verify that they produced the identity, and switch to Forbid.

This is the real problem. Get rid of this terrible voodoo magic and all these weird problems will suddenly disappear, without having to introduce odd work arounds.

> @HaoK @PinpointTownes Thoughts?

IMHO, this fix seems fragile as it relies on the informational `ClaimsIssuer` property. If you really like this approach, I think it would be preferable to bind `ClaimsIdentity.AuthenticationType` with `AuthenticationScheme`.

HaoK:
I think its too late at this point to revisit the Challenge behavior again.  But that magic is specific to  MVC's authorize filter implementation, in general you can opt out of that challenge magic anyways by calling the Challenge passing in the specific ChallengeBehavior you want (Unauthorized/Forbidden)

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
HaoK: Also your example of an authorization policy is a bit suspect.  In general you would have a policy asking for certain claims, and then specify which auth schemes will be queried for the identities/claims.  So it's unlikely you would have a "Github" policy, you might have a "SocialAuth" policy which listed "Github" as one of the authentication schemes for the policy.  There shouldn't be anything wrong with the policy working against a claim requirement if it was satisfied from a Facebook identity or a github identity, that's sort of the point.  Its supposed to allow authorization against a set of claims, rather than worrying about where they came from.

Tratcher:
Updated to be less strict about the identity check. We'll see if it causes any confusion.

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerOptions.cs
```diff
@@ -113,6 +113,18 @@ public JwtBearerOptions() : base()
         /// <exception cref="ArgumentNullException">if 'value' is null.</exception>
         public TokenValidationParameters TokenValidationParameters { get; set; } = new TokenValidationParameters();
 
+        public override string ClaimsIssuer
```
Tratcher: I'll revert this part

HaoK:
:shipit:

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -89,9 +89,32 @@ public abstract class RemoteAuthenticationHandler<TOptions> : AuthenticationHand
...
-        protected override Task<AuthenticateResult> HandleAuthenticateAsync()
+        protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
         {
-            return Task.FromResult(AuthenticateResult.Fail("Remote authentication does not support authenticate"));
+            // Most RemoteAuthenticationHandlers will have a PriorHandler, but it might not be set up during unit tests.
+            if (PriorHandler != null)
+            {
+                var authenticateContext = new AuthenticateContext(Options.SignInScheme);
+                await PriorHandler.AuthenticateAsync(authenticateContext);
+                if (authenticateContext.Accepted)
```
PinpointTownes: So asking for `Facebook` might return a `Google` ticket if the two middleware share the same cookies middleware for identity persistence? :gun:

HaoK: Well given that previously asking for Facebook would never give you any identity anyways, yes... now remote schemes will give you the 'sign in scheme' identity when you ask.  If you are sharing sign in schemes, that may give you someone else's identity...  Kind of weird but I think it makes sense.

HaoK: Note: I am talking mostly about the 'external' public API call of Authenticate("Facebook") making sense, it is true what you say that if direct calls to HandleAuthenticateAsync will show the much weirder behavior of getting a "Google" ticket back potentially.

PinpointTownes: > Well given that previously asking for Facebook would never give you any identity anyways, yes... 

IMHO, that was a much more reasonable behavior. Now, `HandleAuthenticateAsync` will just lie hard.

Tratcher: @PinpointTownes Did you see my prior version that tried to verify the provider?

PinpointTownes: Yep, here was my remark:

> IMHO, this fix seems fragile as it relies on the informational ClaimsIssuer property. If you really like this approach, I think it would be preferable to bind ClaimsIdentity.AuthenticationType with AuthenticationScheme.

That said, it's probably better than no check at all... :fearful:

Issue #814, opened 2016-05-06
CookiePolicy middleware can't affect CookieAuthentication middleware


anurse:
Because the CookieAuthentication middleware does it's own cookie-writing manually (see https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.Cookies/ChunkingCookieManager.cs), the CookiePolicy middleware can't affect it. This is easily seen by doing something like this in your Startup.cs

```csharp
app.UseCookiePolicy(new CookiePolicyOptions()
{
    HttpOnly = HttpOnlyPolicy.Always
});

app.UseCookieAuthentication(new CookieAuthenticationOptions()
{
    AuthenticationScheme = "MagicSignin",
    CookieHttpOnly = false,
    LoginPath = new PathString("/Account/Login"),
    AccessDeniedPath = new PathString("/Account/AccessDenied"),
    AutomaticAuthenticate = true,
    AutomaticChallenge = true
});
```

CookiePolicy works by replacing the `IResponseCookiesFeature`, but CookieAuth sets the header manually so it ignores the policy.

brockallen:
Does this mean there are 2 different ways to emit cookies in ASP.NET Core then?

Tratcher:
Sortof. HttpContext.Response.Cookies.Append and HttpContext.Response.Headers.Append can both be used to set a set-cookie header. Note you could also do this in Asp.Net 4. The difference between 4 and Core is that in Core the headers are the source of truth, where in 4 there was a cookies collection that would overwrite the headers just before sending the response.

davidfowl:
This feels like something we should fix post RC2.

anurse:
@davidfowl Agree entirely. It's a functional change, yes, but a fairly edge-case scenario and I agree it need not be a priority for RC2.

PR #843, opened 2016-05-24
Rework CookieAuth for compat with CookiePolicy.

Tratcher:
#814 @anurse @HaoK @muratg 
Change ChunkingCookieManager from appending raw headers to appending via the response cookies API. Here are a few side effects:
- Removed the key & value encoding. The lower level will handle this and most of the time no encoding is required. Worst case we loose a little bit of length precision and devs need to lower ChunkSize accordingly.
- Removed the quote handling. Cookie auth never uses quotes.
- Change the format of the first cookie from `key=chunks:2` to `key=chunks-2`. `:` would be escaped by the response cookies API.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationMiddleware.cs
@@ -42,7 +42,7 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
             }
             if (Options.CookieManager == null)
             {
-                Options.CookieManager = new ChunkingCookieManager(urlEncoder);
+                Options.CookieManager = new ChunkingCookieManager();
             }
             if (!Options.LoginPath.HasValue)
             {
--- a/src/Microsoft.Owin.Security.Interop/ChunkingCookieManager.cs
+++ b/src/Microsoft.Owin.Security.Interop/ChunkingCookieManager.cs
@@ -5,12 +5,11 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
-using Microsoft.AspNetCore.Http;
-using Microsoft.Extensions.Primitives;
-using Microsoft.Net.Http.Headers;
+using Microsoft.Owin.Infrastructure;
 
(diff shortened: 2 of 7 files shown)


HaoK:
:shipit:

anurse:
:shipit:

PinpointTownes:
I wonder why the cookie manager is not part of a more general repository like HttpAbstractions. It's not really specific to authentication cookies, actually.

Tratcher:
That's what we thought in Katana, but then nobody else ever used it.

Tratcher:
Hmm, this format change will break interop with Katana. However the ICookieManager could be replaced there to read the modified format. @HaoK is this a problem for the interop work you did?

HaoK:
Did you change the TicketSerializer?

Tratcher:
No, I changed the format of the first cookie from `key=chunks:2` to `key=chunks-2` because the `:` would be escaped by the response cookies API.

HaoK:
Well I guess if katana is broken by definition the cookie interop will be as well regardless.

Tratcher:
Should I add a compatible cookie manager to the interop package?

PinpointTownes:
> Should I add a compatible cookie manager to the interop package?

If you want to bring back a dependency on the OWIN cookies package, that would be great to do that in a separate package (something like `Microsoft.Owin.Security.Cookies.Interop`).

HaoK:
Is cookie chunking on by default?

Tratcher:
Yes, chunking is on by default, but it only takes effect if the cookie is over 4kb. This happens pretty often for OIDC.

HaoK:
A cookie interop package is good since it would be a better home for the extensions methods needed for cookie data protection sharing than the Identity compat package.

Tratcher:
Updated. I was able to add a compatible ChunkingCookieManager to the interop package without adding any dependencies because the old interface was in the Microsoft.Owin package.

PinpointTownes:
> I was able to add a compatible ChunkingCookieManager to the interop package without adding any dependencies because the old interface was in the Microsoft.Owin package.

Nice :+1:

HaoK:
Nice :shipit:

Issue #776, opened 2016-04-12
JwtBearer does not return any useful info when failing to validate/accept a token


vibronet:
It is currently very challenging to troubleshoot validation failures originated in the JwtBearer middleware. The client doesn't get back the reasons for which the token failed the validation, which makes it very tough to identify common error situations (such as audience mismatch, untrusted issuers, and so on).
I suggest that the JwtBearer middleware should return some indication about the failure in the response, and specifically in the WWW-Authenticate header.
It goes without saying that the kind of information returned should undergo rigorous security reviews to ensure that the middleware does not leak information that could be used for refining an attack against the protected service.

Tratcher:
@blowdart

Tratcher:
The first problem is getting meaningful errors out of ValidateToken. See https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/312.

blowdart:
If we can get the information then we should return it somehow.

Eilon:
Need to get a grasp on what the top hard-to-debug/hard-to-diagnose issues are.

Some problem areas:
* Audience mismatch
* Issuers that the app doesn't like
* Tokens outside of valid window (e.g. expired, not yet valid)
* Signature failures (e.g. key not found)

Some of these already have specific exceptions associated with them.

Eilon:
Should look into what's valid in the WWW-Authenticate header to transmit the diagnostic message. http://www.ietf.org/rfc/rfc2617.txt

PinpointTownes:
Definitely a good idea (FWIW, we'll implement a similar thing in the aspnet-contrib validation and introspection middleware in the next version).

That said, it might be interesting to provide an option to turn off these details, for people who don't want to disclose them (we had a similar discussion with @MonkeyJamboree about ASOS' introspection endpoint, and whether we should tell the caller why the token was invalid: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/issues/157)

Eilon:
@PinpointTownes good point!

Eilon:
@brentschmaltz can you add a comment here with which exceptions are thrown for the cases we are concerned about?

Eilon:
Ping @brentschmaltz

Tratcher:
Problem | Exception
----------- | --------------
Audience mismatch | SecurityTokenInvalidAudienceException
Issuers that the app doesn't like | SecurityTokenInvalidIssuerException
[...]

PR #845, opened 2016-05-25
Show some JwtBearer errors in response headers

Tratcher:
#776 @vibronet @blowdart @brentschmaltz @HaoK @PinpointTownes @Eilon @muratg 

For a specific set of exceptions, display those exception messages in the response headers to help the consumer report and debug configuration issues. The list of exceptions to report is configurable. Note we're using headers rather than the body because the auth middleware is never in control of the response body.

Example:
```
Content-Length:0
Date:Wed, 25 May 2016 20:54:58 GMT
Server:Kestrel
WWW-Authenticate:Bearer
X-Authentication-Failure:IDX10214: Audience validation failed. Audiences: '63a87a83-64b9-4ac1-b2c5-092126f8474f'. Did not match:  validationParameters.ValidAudience: '63a87a83-64b9-4ac1-b2c5-092126f8474f-wrongaudiance' or validationParameters.ValidAudiences: 'null'
```

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Security.Claims;
+using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
@@ -103,7 +104,8 @@ namespace Microsoft.AspNetCore.Authentication.JwtBearer
                             Logger.TokenValidationFailed(token, ex);
 
                             // Refresh the configuration for exceptions that may be caused by key rollovers. The user can also request a refresh in the event.
-                            if (Options.RefreshOnIssuerKeyNotFound && ex.GetType().Equals(typeof(SecurityTokenSignatureKeyNotFoundException)))
+                            if (Options.RefreshOnIssuerKeyNotFound && Options.ConfigurationManager != null
+                                && ex is SecurityTokenSignatureKeyNotFoundException)
                             {
                                 Options.ConfigurationManager.RequestRefresh();
                             }
@@ -183,7 +185,12 @@ namespace Microsoft.AspNetCore.Authentication.JwtBearer
 
         protected override async Task<bool> HandleUnauthorizedAsync(ChallengeContext context)
         {
-            var eventContext = new JwtBearerChallengeContext(Context, Options, new AuthenticationProperties(context.Properties));
+            var authResult = await HandleAuthenticateOnceAsync();
+
+            var eventContext = new JwtBearerChallengeContext(Context, Options, new AuthenticationProperties(context.Properties))
+            {
+                AuthenticateFailure = authResult?.Failure,
+            };
             await Options.Events.Challenge(eventContext);
             if (eventContext.HandledResponse)
             {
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -103,7 +106,8 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
                             Logger.TokenValidationFailed(token, ex);
 
                             // Refresh the configuration for exceptions that may be caused by key rollovers. The user can also request a refresh in the event.
-                            if (Options.RefreshOnIssuerKeyNotFound && ex.GetType().Equals(typeof(SecurityTokenSignatureKeyNotFoundException)))
+                            if (Options.RefreshOnIssuerKeyNotFound && Options.ConfigurationManager != null
+                                && ex.GetType().Equals(typeof(SecurityTokenSignatureKeyNotFoundException)))
```
brentschmaltz: this will skip derived types of SecurityTokenSignatureKeyNotFoundException. if it matters to you, you could do 'is' or 'as'.

PinpointTownes:
Holy sh.. what daf... is this? :trollface: 

I **hate** this approach for at least 2 reasons:
  - It directly exposes the exception messages, which might disclose confidential/private data. Even if the exceptions are white-listed, IdentityModel could be updated to return more data than what you expect. Don't do that, really.

  - It's absolutely not standard. Why reinventing the wheel? You can use the `WWW-Authenticate` header and one of the listened OAuth2 error codes: https://tools.ietf.org/html/rfc6750#section-3.1

Tratcher:
Ah, error and error_description are good ideas. But to your first point, what would you say in them that would actually be helpful?

PinpointTownes:
> But to your first point, what would you say in them that would actually be helpful?

Helpful for a bad guy? For instance, IdentityModel could reveal the expected signature of a security token, which would be *terrible* (AFAIK, it doesn't do that ATM, but we never know).

Or helpful for the consumer of the API? Personally, I'd associate a "manual" message with the exceptions you've listed to ensure we don't share too much details.

E.g for an invalid audience:
  - error: invalid_token
  - error_description: the token was rejected because its audience(s) was/were not acceptable for this resource server.

brockallen:
https://tools.ietf.org/html/rfc6750#section-3.1

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerOptions.cs
```diff
@@ -118,5 +118,21 @@ public JwtBearerOptions() : base()
         /// <see cref="Http.Authentication.AuthenticationProperties"/> after a successful authorization.
         /// </summary>
         public bool SaveToken { get; set; } = true;
+
+        /// <summary>
+        /// Authentication failures with one of these exception types, or types derived from these, will have their messages
+        /// included in a 'X-Authentication-Failure' response header.
+        /// </summary>
+        public ICollection<Type> ReportableFailures { get; set; } = new List<Type>()
```
HaoK: Not a huge fan of a generic white list feature like this, I think for something like this, we should just hardcode this list and turn it into a custom error.  Or if we wanted to make it somewhat extensible, then I would rather something like an `Func<string, Exception>` transform, where the default instance would take these exception types and produce the appropriate error for them.

PinpointTownes:
@brockallen it's precisely the link I posted ~45 minutes ago :smile: 

> Not a huge fan of a generic white list feature like this, I think for something like this, we should just hardcode this list and turn it into a custom error.

@HaoK yep, exactly what I had in mind.

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerOptions.cs
Tratcher: You mean `Func<Exception, string>`? And returning `null` or `string.Empty` would say don't return anything?

brockallen:
> @brockallen it's precisely the link I posted ~45 minutes ago 

You're much younger than me, so I have a hard time keeping up.

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerOptions.cs
HaoK: Yeah exactly

PinpointTownes: Is it really worth providing an option for that? One could use the `Challenge` event to replace the `error`/`error_description` in the `WWW-Authenticate` header.

PinpointTownes: On the other hand, we really need an option to disable this feature when needed.

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
Eilon: Though if you really want equality, use `==`. Not sure if you want derived types or not.

Tratcher:
Notes: Have our own messages mapped from the different exception types. Consider having the exceptions add a few properties with relevant data like the invalid issuer.

Tratcher:
Updated.
- We now map the exception types to our own messages. These messages go into the error_description field.
- The exception is made available in the Challenge event where you can handle it, clear it, etc..

I'm filing a bug against IdentityModel to details to the exception properties like the issuer that failed so we can provide more detailed messages.

Tratcher:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/422

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -195,11 +202,103 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
             }
 
             Response.StatusCode = 401;
-            Response.Headers.Append(HeaderNames.WWWAuthenticate, Options.Challenge);
+
+            var errorDescription = CreateErrorDescription(eventContext.AuthenticateFailure);
```
PinpointTownes: Have you considered adding new `Error`/`ErrorDescription` properties to the challenge context and computing the error description before invoking the `Challenge` event? This way, one could easily replace the error code/description without having to manually create the `WWW-Authenticate` header.

PinpointTownes:
It looks much better than the original proposal :smile: 

I know you're not a huge fan of that, but a "write the error to the response body" option would be cool too (not necessarily for this PR or even for RTM). Facebook's API are particularly appreciated for this reason.

test/Microsoft.AspNetCore.Authentication.Test/JwtBearer/JwtBearerMiddlewareTests.cs
```diff
@@ -140,6 +142,69 @@ public class JwtBearerMiddlewareTests
...
+        [InlineData(typeof(SecurityTokenInvalidLifetimeException), "The token lifetime is invalid")]
+        [InlineData(typeof(SecurityTokenNotYetValidException), "The token is not valid yet")]
+        [InlineData(typeof(SecurityTokenExpiredException), "The token is expired")]
+        [InlineData(typeof(SecurityTokenInvalidSignatureException), "The signature is invalid")]
+        [InlineData(typeof(SecurityTokenSignatureKeyNotFoundException), "The signature key was not found")]
+        public async Task ExceptionReportedInHeaderForAuthenticationFailures(Type errorType, string message)
+        {
+            var options = new JwtBearerOptions
+            {
+                AutomaticAuthenticate = true
```
PinpointTownes: nit: you could remove that, since it's the default value.

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
Tratcher: The www-Authenticate header is pretty easy in this scenario, I don't expect people to have trouble creating it.

Tratcher:
@HaoK please review.

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -195,11 +202,103 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
...
+                else if (ex is SecurityTokenInvalidSignatureException)
+                {
+                    Append("The signature is invalid", errorDescription);
+                }
+            }
+
+            return errorDescription.ToString();
+        }
+
+        private static void Append(string message, StringBuilder errorDescription)
```
HaoK: Nit: could consider just appending to a `List<string> errors` and then call String.Join(errors, ';') at the end, might be a little bit more compact without the Append helper.

HaoK:
:shipit:

PR #846, opened 2016-05-25
Support OIDC distributed sign-out.

Tratcher:
#423 @vibronet @brentschmaltz @blowdart @HaoK @Eilon @muratg @PinpointTownes 

Requests to `/signout-oidc` will be intercepted by the middleware and it will sign out the auth cookie.

AAD setup instructions:
manage.windowsazure.com --> active directory --> [tenant] --> applications --> [app] --> configure --> download manifest --> change logoutUrl --> upload manifest

Most of the sample changes are to enable https on the site. Otherwise the browser blocked the request as mixed content.

Open questions: 
- Should there be an event here?


Tratcher:
Note http://openid.net/specs/openid-connect-logout-1_0.html  says there may be a sid in the query to cross check, but AAD doesn't send that.

PinpointTownes:
> Note http://openid.net/specs/openid-connect-logout-1_0.html says there may be a sid in the query to cross check, but AAD doesn't send that.

The real question is: are we fine with the fact it would leave an OIDC client app vulnerable to unwanted signout?

Eilon:
@PinpointTownes yeah we discussed the unwanted sign-out thing and though we're not super happy with it, it's rather common amongst auth systems (unfortunately). The worst case is you can be griefed (a very mild form of DoS), there's no information disclosure, elevation of privileges, etc.

brockallen:
Please ensure there's an event so custom code can do the `sid` param/claims check when you are using a more spec compliant OIDC server. I need to be able to prevent this as default behavior in this event.

PinpointTownes:
Here's the compromise I suggest: if the `sid` claim can be found in the identity token returned by the OP as part of the initial authorization/token request, require a `sid` parameter and reject the frontchannel logout request if it's missing or doesn't correspond to the `sid` claim. 

I guess it would make everyone happy: the AzureAD guys because it wouldn't prevent AAD from working correctly and @brockallen and me because unwanted signout would be avoided with servers supporting `sid`.

brockallen:
Sure, sounds like a decent compromise.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -62,6 +62,17 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
             HtmlEncoder = htmlEncoder;
         }
 
+        public override async Task<bool> HandleRequestAsync()
+        {
+            if (Options.SignOutPath.HasValue && Options.SignOutPath == Request.Path)
```
PinpointTownes: You should consider adding a tiny HTTP method constraint somewhere. GET is a must, but I guess POST could be also supported (that would be triggered from an auto-post form in an iframe).

PinpointTownes: Also, is it worth adding the no-cache directives recommended by the spec?

samples/OpenIdConnectSample/Program.cs
```diff
@@ -1,24 +1,46 @@
...
                 .UseStartup<Startup>()
                 .Build();
 
             host.Run();
         }
+
+        private static X509Certificate2 LoadCertificate()
+        {
+            var assembly = typeof(Startup).GetTypeInfo().Assembly;
+            var embeddedFileProvider = new EmbeddedFileProvider(assembly, "OpenIdConnectSample");
```
PinpointTownes: nit: why using `EFP` instead of directly loading the certificate from the assembly?

PinpointTownes:
I agree with @brockallen, we need an event to have more control over what happens in this new frontchannel logout endpoint.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: Cookie auth sign-out adds the no-cache directives.

samples/OpenIdConnectSample/Program.cs
Tratcher: It simplifies publishing to embed the resource.

Tratcher:
Rebased. Updated to add an event.

PinpointTownes:
So no plan to add an optional sid check as I suggested? :sweat_smile:

Tratcher:
@PinpointTownes @brockallen I'll check this in today and you can try it with your servers. If you want to send a PR for the sid check by Tuesday morning we'll consider it. We're locking down for RTM after that.

@HaoK please review.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -62,6 +62,33 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
+                if (remoteSignOutContext.Skipped)
+                {
+                    Logger.RemoteSignOutSkipped();
+                    return false;
+                }
+
+                Logger.RemoteSignOut();
+
+                // We've received a remote sign-out request
+                await Context.Authentication.SignOutAsync(Options.SignOutScheme ?? Options.SignInScheme);
```
HaoK: This looks a bit weird, there's a SignOutScheme and a SignInScheme?

HaoK:
:shipit:

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: This is so it can still work with Identity where there are multiple cookies.

PinpointTownes:
> If you want to send a PR for the sid check by Tuesday morning we'll consider it. We're locking down for RTM after that.

This is a very short window :fearful: 

@brockallen any chance you could do that?

brockallen:
I don't have time to try any nightlys. Maybe @leastprivilege  does.

leastprivilege:
No he doesn't ;)

I didn't even follow this discussion - what's needed is an extensibility point that gets called when the signout request is received. beyond that it is too implementation specific (e.g. which claim represents the session id etc.)

Maybe you already came to that conclusion ;)

Tratcher:
@PinpointTownes Sorry, I know, that's the same reason we needed to get the base feature in. We can continue to work on this, it just won't make it in the 1.0.0 release.

PinpointTownes:
> beyond that it is too implementation specific (e.g. which claim represents the session id etc.)

From the client perspective, the claim is always the same as it's defined by the logout spec: `sid`.

Do you have an IdentityServer sample I could use to test my PR? It works fine with my ASOS demo, but I'd like to test with another server.

leastprivilege:
We always emit the sid claim.

You can use https://demo.identityserver.io

IIRC - the client id is e.g. `implicit` or https://github.com/leastprivilege/AspNetCoreSecuritySamples/tree/master/OIDC

PinpointTownes:
@leastprivilege FYI, I haven't been able to make your demo work. I presume the `implicit` client doesn't have any `frontchannel_logout_uri` associated, since it's not included in the `iFrameUrls` list:

```js
<script id='modelJson' type='application/json'>{&quot;iFrameUrls&quot;:[&quot;https://demo.identityserver.io/connect/endsessioncallback?sid=2f3d4eb209a688de67012c083a02e3e7&quot;],&quot;clientName&quot;:&quot;Hybrid Flow Client&quot;,&quot;redirectUrl&quot;:&quot;https://localhost:3308/signedout&quot;,&quot;autoRedirect&quot;:false,&quot;autoRedirectDelay&quot;:0,&quot;siteUrl&quot;:&quot;https://demo.identityserver.io/&quot;,&quot;siteName&quot;:&quot;IdentityServer3 Demo&quot;,&quot;currentUser&quot;:null,&quot;logoutUrl&quot;:null,&quot;custom&quot;:null}</script>
```

leastprivilege:
yes - sorry - forgot about that.

not sure how to fix that for you quickly - unless you want to use one of our demo hosts from the samples repo.

PinpointTownes:
> unless you want to use one of our demo hosts from the samples repo.

Thanks, it worked like a charm.

PR #848, opened 2016-05-26
AuthZ: Combine needs to use the PolicyProvider

HaoK:
cc @blowdart @divega


HaoK:
c257c9528f16db78a33eed810b9a66f796146b3c

Issue #780, opened 2016-04-15
CookieAuthenticationHandler IsPersistent with UTC dates


nickcam:
Hi Guys,

Just stumbled across a small bug I think.
Line 243 in CookieAuthenticationHandler.
```c#
if (signInContext.Properties.IsPersistent)
{
      var expiresUtc = signInContext.Properties.ExpiresUtc ?? issuedUtc.Add(Options.ExpireTimeSpan);
      signInContext.CookieOptions.Expires = expiresUtc.ToUniversalTime().DateTime;
}
```
The date signInContext.CookieOptions.Expires has the correct value, but it's kind is Unspecified, so when added to the cookie it gets converted to UTC again.

I've got a local copy of this class and just changed that line to
`signInContext.CookieOptions.Expires = DateTime.SpecifyKind(expiresUtc.ToUniversalTime().DateTime, DateTimeKind.Utc);`

and it's all good.
Thanks.

Eilon:
Can we look into why we aren't using `DateTimeOffset`? That might help resolve some of this.

PR #851, opened 2016-05-27
Remove obsolete DateTime conversions

Tratcher:
#780 @nickcam @HaoK @muratg @Eilon 

Katana used DateTime for all of its fields/properties. We've since converted everything to use DateTimeOffset as it more accurately preserves things like time zone. However, we forgot to remove a few conversions like the ones in this PR. It wasn't a compiler break because of the implicit conversion from DateTime to DateTimeOffset, but we were loosing information in the downgrade.


HaoK:
:shipit:

Issue #772, opened 2016-04-11
Add OIDC test coverage


Tratcher:
OIDC does not have sufficient test coverage. Many tests were removed or disabled due to breaking changes or redesigns. Check our coverage and fill in the gaps.

Note there are some pseudo-functional tests in MusicStore that could be moved here.

Related: https://github.com/aspnet/Security/issues/349 - Move OIDC tests to their own project.

Eilon:
@BrennanConroy please discuss with @Tratcher .

troydai:
Coverage list:

_In progress_

#### Configuration
 - [ ] Middleware is added correctly.
 - [ ] Options are set correctly.
 - [ ] Missing required settings.

#### Sign in work flow
 - [ ] 

#### Sign out work flow
 - [ ] 

#### Error handling
 - [ ]

Tratcher:
- Events and their parameters
- Event flow control (Handled, skipped, throws, etc.)
https://github.com/aspnet/Security/issues/884

PinpointTownes:
... and also userinfo :smile:

Eilon:
Let's spend 2-3 days adding a basic Hello World type of test that goes through a full successful login flow.

Also tagging @javiercn because of the work he did with Web Application Factory, and that could help in doing this.

PR #852, opened 2016-05-31
Add a 'sid' check to the OIDC middleware to prevent unsolicited logout when possible

PinpointTownes:
Partially addresses the security concern discussed [here](https://github.com/aspnet/Security/pull/846#issuecomment-221844918).

@Eilon @Tratcher @HaoK @blowdart please review when you have a minute :smile:

### DIFF
--- a/samples/OpenIdConnectSample/Startup.cs
+++ b/samples/OpenIdConnectSample/Startup.cs
@@ -98,6 +98,7 @@ namespace OpenIdConnectSample
                 if (context.Request.Path.Equals("/signout-remote"))
                 {
                     // Redirects
+                    await context.Authentication.SignOutAsync(CookieAuthenticationDefaults.AuthenticationScheme);
                     await context.Authentication.SignOutAsync(OpenIdConnectDefaults.AuthenticationScheme, new AuthenticationProperties()
                     {
                         RedirectUri = "/signedout"
--- a/samples/OpenIdConnectSample/Startup.cs
+++ b/samples/OpenIdConnectSample/Startup.cs
@@ -6,10 +6,8 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Authentication;
-using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Logging;
 using Microsoft.IdentityModel.Protocols.OpenIdConnect;
 
 namespace OpenIdConnectSample
@@ -39,10 +37,8 @@ namespace OpenIdConnectSample
                 sharedOptions.SignInScheme = CookieAuthenticationDefaults.AuthenticationScheme);
         }
 
-        public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
+        public void Configure(IApplicationBuilder app)
         {
-            loggerfactory.AddConsole(LogLevel.Information);
-
             // Simple error page
             app.Use(async (context, next) =>
             {
@@ -102,7 +98,6 @@ namespace OpenIdConnectSample
                     {
                         RedirectUri = "/signedout"
                     });
-                    await context.Authentication.SignOutAsync(CookieAuthenticationDefaults.AuthenticationScheme);
                     return;
                 }
                 if (context.Request.Path.Equals("/Account/AccessDenied"))
--- a/samples/OpenIdConnectSample/Startup.cs
+++ b/samples/OpenIdConnectSample/Startup.cs
@@ -8,6 +8,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Authentication;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 using Microsoft.IdentityModel.Protocols.OpenIdConnect;
 
 namespace OpenIdConnectSample
@@ -37,8 +38,11 @@ namespace OpenIdConnectSample
                 sharedOptions.SignInScheme = CookieAuthenticationDefaults.AuthenticationScheme);
         }
 
-        public void Configure(IApplicationBuilder app)
+        public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
         {
+            loggerfactory.AddConsole(LogLevel.Information);
+            loggerfactory.AddDebug(LogLevel.Information);
+
             // Simple error page
             app.Use(async (context, next) =>
             {
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/RemoteSignoutContext.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/RemoteSignoutContext.cs
@@ -3,14 +3,24 @@
 
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
+using Microsoft.IdentityModel.Protocols.OpenIdConnect;
 
 namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
 {
     public class RemoteSignOutContext : BaseOpenIdConnectContext
     {
-        public RemoteSignOutContext(HttpContext context, OpenIdConnectOptions options)
+        public RemoteSignOutContext(
+            HttpContext context,
+            OpenIdConnectOptions options,
+            OpenIdConnectMessage message)
             : base(context, options)
         {
+            Message = message;
         }
+
+        /// <summary>
+        /// Gets the remote signout message.
+        /// </summary>
+        public OpenIdConnectMessage Message { get; }
     }
 }
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/RemoteSignoutContext.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/RemoteSignoutContext.cs
@@ -15,12 +15,7 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
             OpenIdConnectMessage message)
             : base(context, options)
         {
-            Message = message;
+            ProtocolMessage = message;
         }
-
-        /// <summary>
-        /// Gets the remote signout message.
-        /// </summary>
-        public OpenIdConnectMessage Message { get; }
     }
 }
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
@@ -112,9 +112,7 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
             // If the identifier cannot be found, bypass the session identifier checks: this may indicate that the
             // authentication cookie was already cleared, that the session identifier was lost because of a lossy
             // external/application cookie conversion or that the identity provider doesn't support sessions.
-            var sid = (await Context.Authentication.AuthenticateAsync(Options.SignOutScheme))?.FindFirst("sid")?.Value
-                   ?? (await Context.Authentication.AuthenticateAsync(Options.SignInScheme))?.FindFirst("sid")?.Value
-                   ?? Context.User?.FindFirst("sid")?.Value;
+            var sid = (await Context.Authentication.AuthenticateAsync(Options.SignOutScheme))?.FindFirst("sid")?.Value;
             if (!string.IsNullOrEmpty(sid))
             {
                 // Ensure a 'sid' parameter was sent by the identity provider.
@@ -182,9 +180,7 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
                 }
 
                 // Attach the identity token to the logout request when possible.
-                message.IdTokenHint = await Context.Authentication.GetTokenAsync(Options.SignOutScheme, OpenIdConnectParameterNames.IdToken)
-                                   ?? await Context.Authentication.GetTokenAsync(Options.SignInScheme, OpenIdConnectParameterNames.IdToken)
-                                   ?? await Context.Authentication.GetTokenAsync(OpenIdConnectParameterNames.IdToken);
+                message.IdTokenHint = await Context.Authentication.GetTokenAsync(Options.SignOutScheme, OpenIdConnectParameterNames.IdToken);
 
                 var redirectContext = new RedirectContext(Context, Options, properties)
                 {
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
@@ -89,11 +89,6 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
                 message = new OpenIdConnectMessage(form.Select(pair => new KeyValuePair<string, string[]>(pair.Key, pair.Value)));
             }
 
-            if (message == null)
-            {
-                return false;
-            }
-
             var remoteSignOutContext = new RemoteSignOutContext(Context, Options, message);
             await Options.Events.RemoteSignOut(remoteSignOutContext);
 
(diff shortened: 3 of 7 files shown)


PinpointTownes:
AFAICT, there's no dedicated test for this remote signout feature. Given the short window I had to submit this PR, adding new tests was not possible.

I'd recommend adding this task to https://github.com/aspnet/Security/issues/772.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -66,27 +66,82 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
+            {
+                Logger.RemoteSignOutSkipped();
+                return false;
+            }
+
+            // Try to extract the session identifier from the authentication ticket persisted by the sign-in handler.
+            // If the identifier cannot be found, bypass the session identifier checks: this may indicate that the
+            // authentication cookie was already cleared, that the session identifier was lost because of a lossy
+            // external/application cookie conversion or that the identity provider doesn't support sessions.
+            var sid = await Context.Authentication.GetTokenAsync(Options.SignInScheme, "sid");
```
PinpointTownes: @brentschmaltz any chance we could get a `Sid` entry in `OpenIdConnectParameterNames`/`JwtRegisteredClaimNames`?

samples/OpenIdConnectSample/Startup.cs
```diff
@@ -39,10 +37,8 @@ public void ConfigureServices(IServiceCollection services)
                 sharedOptions.SignInScheme = CookieAuthenticationDefaults.AuthenticationScheme);
         }
 
-        public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
```
Tratcher: We try to keep these in sync with the templates for things unrelated to the current feature:
https://github.com/aspnet/Templates/blob/dev/src/BaseTemplates/StarterWeb/Startup.cs

samples/OpenIdConnectSample/Startup.cs
```diff
@@ -102,7 +98,6 @@ await context.Authentication.SignOutAsync(OpenIdConnectDefaults.AuthenticationSc
                     {
                         RedirectUri = "/signedout"
                     });
-                    await context.Authentication.SignOutAsync(CookieAuthenticationDefaults.AuthenticationScheme);
```
Tratcher: Keep this. We want the local sign-out too.

leastprivilege: But this hard-codes the name. Rather use the SignInScheme.

samples/OpenIdConnectSample/Startup.cs
PinpointTownes: The templates suck, configuring logging from `Main()` feels much cleaner.
But okay, I'll revert this change :smile:

samples/OpenIdConnectSample/Startup.cs
Tratcher: In user code. SignInScheme isn't very accessible here. Also keeping in sync with templates: 
https://github.com/aspnet/Templates/blob/dev/src/Rules/StarterWeb/OrganizationalAuth/Single/Common/Controllers/AccountController.cs#L24

samples/OpenIdConnectSample/Startup.cs
Tratcher: Then fix the templates :-)

samples/OpenIdConnectSample/Startup.cs
PinpointTownes: I'm fine reverting this change, since it's what any client app should do. But note that removing the local cookie before redirecting the user agent to the identity provider doesn't allow debugging the `sid` check.

Since this sample is more a sandbox than a real sample, I thought it was better to remove it.

samples/OpenIdConnectSample/Startup.cs
PinpointTownes: I opened a ticket to track that: https://github.com/aspnet/Templates/issues/586.

In the meantime, I'll revert this change.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -66,27 +66,83 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
+                message = new OpenIdConnectMessage(Request.Query.Select(pair => new KeyValuePair<string, string[]>(pair.Key, pair.Value)));
+            }
+            // assumption: if the ContentType is "application/x-www-form-urlencoded" it should be safe to read as it is small.
+            else if (string.Equals(Request.Method, "POST", StringComparison.OrdinalIgnoreCase)
+              && !string.IsNullOrEmpty(Request.ContentType)
+              // May have media/type; charset=utf-8, allow partial match.
+              && Request.ContentType.StartsWith("application/x-www-form-urlencoded", StringComparison.OrdinalIgnoreCase)
+              && Request.Body.CanRead) {
+                var form = await Request.ReadFormAsync();
+                message = new OpenIdConnectMessage(form.Select(pair => new KeyValuePair<string, string[]>(pair.Key, pair.Value)));
```
Tratcher: Wrapping the query or form in an OIDC message just gets you named properties for the keys, but what keys are you expecting? Only `sid` right? Which isn't even on the OIDCMessage. Are you expecting any other OIDC fields? If not this seems like the wrong abstraction for this endpoint.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -66,27 +66,83 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
+            {
+                Logger.RemoteSignOutSkipped();
+                return false;
+            }
+
+            // Try to extract the session identifier from the authentication ticket persisted by the sign-in handler.
+            // If the identifier cannot be found, bypass the session identifier checks: this may indicate that the
+            // authentication cookie was already cleared, that the session identifier was lost because of a lossy
+            // external/application cookie conversion or that the identity provider doesn't support sessions.
+            var sid = await Context.Authentication.GetTokenAsync(Options.SignInScheme, "sid")
```
Tratcher: Try the SignOutScheme if control has been transferred to a second cookie auth instance?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
PinpointTownes: > Only sid right?

Yep. That said, nothing prevents an identity provider from adding more keys, and an application might be interested in retrieving them. Using an `OpenIdConnectMessage` allows flowing it through the event in a typed way.

> Which isn't even on the OIDCMessage.

Not yet :trollface: 

If you think if it's not worth using `OpenIdConnectMessage` here, I'm fine killing it.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -757,6 +825,12 @@ private void SaveTokens(AuthenticationProperties properties, OpenIdConnectMessag
                 }
             }
 
+            var sid = principal.FindFirst("sid")?.Value;
+            if (!string.IsNullOrEmpty(sid))
+            {
+                tokens.Add(new AuthenticationToken { Name = "sid", Value = sid });
```
Tratcher: Why treat it like a token at all when it comes from a claim?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: Its harmless for now, I just wanted to make sure you weren't expecting something else.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
PinpointTownes: No particular reason. It was mainly for consistency with how we extract `id_token_hint`, but I'll revert it if you prefer.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
PinpointTownes: This `SignInScheme`/`SignOutScheme` seems a bit strange, but why not :+1:

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: lets keep it as a claim

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/RemoteSignoutContext.cs
```diff
@@ -3,14 +3,24 @@
...
+            OpenIdConnectMessage message)
             : base(context, options)
         {
+            Message = message;
         }
+
+        /// <summary>
+        /// Gets the remote signout message.
+        /// </summary>
+        public OpenIdConnectMessage Message { get; }
```
Tratcher: Just use ProtocolMessage from the base class.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -66,27 +66,83 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
+            else if (string.Equals(Request.Method, "POST", StringComparison.OrdinalIgnoreCase)
+              && !string.IsNullOrEmpty(Request.ContentType)
+              // May have media/type; charset=utf-8, allow partial match.
+              && Request.ContentType.StartsWith("application/x-www-form-urlencoded", StringComparison.OrdinalIgnoreCase)
+              && Request.Body.CanRead) {
+                var form = await Request.ReadFormAsync();
+                message = new OpenIdConnectMessage(form.Select(pair => new KeyValuePair<string, string[]>(pair.Key, pair.Value)));
+            }
+
+            if (message == null)
```
Tratcher: Put this after the event? It allows more control.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
PinpointTownes: Should we use both `SignOutScheme` + `SignInScheme` or just `SignOutScheme` here?

Tratcher: Well, you're getting rid of all the Tokens indirection, but for AuthenticateAsync I'd go with `SignOutScheme ?? SignInScheme` like SignOut uses.

samples/OpenIdConnectSample/Startup.cs
PinpointTownes: Reverted, but moved before OIDC, as trying to update the response headers after writing to the response stream (e.g when `AuthenticationMethod = OpenIdConnectRedirectBehavior.FormPost`) is not safe and will likely throw an exception.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -66,27 +66,76 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
+            {
+                return false;
+            }
+
+            // Try to extract the session identifier from the authentication ticket persisted by the sign-in handler.
+            // If the identifier cannot be found, bypass the session identifier checks: this may indicate that the
+            // authentication cookie was already cleared, that the session identifier was lost because of a lossy
+            // external/application cookie conversion or that the identity provider doesn't support sessions.
+            var sid = (await Context.Authentication.AuthenticateAsync(Options.SignInScheme))?.FindFirst("sid")?.Value
```
PinpointTownes: @Tratcher I think this order is the best compromise:

  - SignInScheme = external cookie
  - SignOutScheme = application cookie
  - Automatic scheme

No objection? :smile:

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectMiddleware.cs
```diff
@@ -100,6 +100,10 @@ public class OpenIdConnectMiddleware : AuthenticationMiddleware<OpenIdConnectOpt
             {
                 throw new ArgumentException("Options.SignInScheme is required.");
             }
+            if (string.IsNullOrEmpty(Options.SignOutScheme))
```
PinpointTownes: Setting the `SignOutScheme` makes the handler code a bit simpler.

PinpointTownes:
Updated.

PinpointTownes:
FYI, I sent a PR to add the `sid` constant we need in Wilson: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/421.

That said, it's not blocking for this PR, and the magic strings could be replaced by the constant as part of a future code cleanup.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: SignOutScheme is highest priority, if set. Then SignInScheme, then Automatic.

Tratcher: And since SignOutScheme is already set, why bother with SignInScheme?

PinpointTownes: Because they point to different cookies in practice?
There are far more chances to find the `sid` in the external cookie than in the application cookie, specially since Identity doesn't preserve the claims between the two cookies.

Or maybe I don't get the point of having two separate schemes :open_mouth:

PinpointTownes: Order changed.

Tratcher:
@brentschmaltz Any thoughts? This should be a no-op for AAD as it never sends a SID in the claims or signout request.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -66,27 +66,76 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
+            if (message == null)
+            {
+                return false;
+            }
+
+            // Try to extract the session identifier from the authentication ticket persisted by the sign-in handler.
+            // If the identifier cannot be found, bypass the session identifier checks: this may indicate that the
+            // authentication cookie was already cleared, that the session identifier was lost because of a lossy
+            // external/application cookie conversion or that the identity provider doesn't support sessions.
+            var sid = (await Context.Authentication.AuthenticateAsync(Options.SignOutScheme))?.FindFirst("sid")?.Value
```
Tratcher: Thinking about this more, why look in all of these places for the sid if we'll only call SingOutAsync for SignOutScheme?
Note there's rarely an external cookie and an app cookie at the same time, you switch from one to the other.

PinpointTownes: > why look in all of these places for the sid if we'll only call SingOutAsync for SignOutScheme?

To improve our chances to find the precious `sid` somewhere :smile: 

If you think it's not worth it, I'm fine reducing the search to just `SignOutScheme`.

Tratcher: Stick to SignOutScheme. It would be inaccurate to search for a sid in an auth session/cookie we weren't going to sign-out.

PinpointTownes: Alright, I will change that.

PinpointTownes:
@brentschmaltz no opinion on that?

Tratcher:
Rebased, squashed, and merged. Thanks

PinpointTownes:
:clap:

PR #854, opened 2016-05-31
Removing abstract method in AuthorizationHandler

### LINKED ISSUE #849: AuthorizationHandler<TRequirement> design questions
Link to the class: https://github.com/aspnet/Security/blob/2634fe318d8b81a3fc6e7e67a2dca980fb9bcc69/src/Microsoft.AspNetCore.Authorization/AuthorizationHandler.cs

#### Assert
This class contains 
- Handle(c)
- HandleAsync(c)
- Handle(c,r)
- HandleAsync(c,r)

It implements `IAuthorizationHandler` and hence its `HandlerAsync(context)` method is called for every requirement. 

The goal of this class is to let developers implement handlers that only get called for a specific requirement (of the generic type).

#### Arrange
1- Because of the goal of this class I think it should only expose `Handle(context, requirement)` and `HandleAsync(context, requirement)`. If a handler doesn't use the requirement, then it can implement the interface directly.
2- `Handle(context)` is currently dead code (not called, and can't be overrided)
3- `HandleAsync(context, req)` should not be `virtual`. IMO it just adds confusion as inheriting from the class show this member to be overridable, but this would break the inner logic of this class. Any reason to re-implement this method should actually be a custom implementation of the interface itself.
4- `Handle(context, req)` should not be `abstract`, as it makes it mandatory in any sub-class, even if we just want to implement `HandleAsync(context, req)`

#### Act
Here is my proposed implementation

```csharp
public abstract class AuthorizationHandler<TRequirement> : IAuthorizationHandler
        where TRequirement : IAuthorizationRequirement
{
    async Task IAuthorizationHandler.HandleAsync(AuthorizationHandlerContext context)
    {
        foreach (var req in context.Requirements.OfType<TRequirement>())
        {
            await HandleAsync(context, req);
        }
    }

    protected virtual void Handle(AuthorizationHandlerContext context, TRequirement requirement)
    {
        return;
    }

    protected virtual Task HandleAsync(AuthorizationHandlerContext context, TRequirement requirement)
    {
        Handle(context, requirement);
        return Task.FromResult(0);
    }
}
```

Any sub-class can then implement either `Handle(ctx, req)` or `HandleAsync(ctx, req)`.

The only drawback is that hiding `HandleAsync(ctx)` is a breaking change. But as mentioned, it would not make sense to override it as it the inner logic of the class, and all samples and the fact that `Handle(ctx, req)` is currently abstract already had you implement `Handle(ctx, req)` so the probability that anyone is doing that is low.

sebastienros:
Fixes #849

### DIFF
--- a/src/Microsoft.AspNetCore.Authorization/AuthorizationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authorization/AuthorizationHandler.cs
@@ -6,18 +6,14 @@ using System.Threading.Tasks;
 
 namespace Microsoft.AspNetCore.Authorization
 {
+    /// <summary>
+    /// Base class for authorization handlers that need to be called for a specific requirement type.
+    /// </summary>
+    /// <typeparam name="TRequirement">The type of the requirement to handle.</typeparam>
     public abstract class AuthorizationHandler<TRequirement> : IAuthorizationHandler
-        where TRequirement : IAuthorizationRequirement
+            where TRequirement : IAuthorizationRequirement
     {
-        public void Handle(AuthorizationHandlerContext context)
-        {
-            foreach (var req in context.Requirements.OfType<TRequirement>())
-            {
-                Handle(context, req);
-            }
-        }
-
-        public virtual async Task HandleAsync(AuthorizationHandlerContext context)
+        async Task IAuthorizationHandler.HandleAsync(AuthorizationHandlerContext context)
         {
             foreach (var req in context.Requirements.OfType<TRequirement>())
             {
@@ -25,8 +21,21 @@ namespace Microsoft.AspNetCore.Authorization
             }
         }
 
-        protected abstract void Handle(AuthorizationHandlerContext context, TRequirement requirement);
+        /// <summary>
+        /// Makes a decision if authorization is allowed based on a specific requirement.
+        /// </summary>
+        /// <param name="context">The authorization information.</param>
+        /// <param name="requirement">The requirement to evaluate.</param>
+        protected virtual void Handle(AuthorizationHandlerContext context, TRequirement requirement)
+        {
+            return;
+        }
 
+        /// <summary>
+        /// Makes a decision if authorization is allowed based on a specific requirement.
+        /// </summary>
+        /// <param name="context">The authorization information.</param>
+        /// <param name="requirement">The requirement to evaluate.</param>
         protected virtual Task HandleAsync(AuthorizationHandlerContext context, TRequirement requirement)
         {
             Handle(context, requirement);
@@ -34,10 +43,16 @@ namespace Microsoft.AspNetCore.Authorization
         }
     }
 
+    /// <summary>
+    /// Base class for authorization handlers that need to be called for specific requirement and
+    /// resource types.
+    /// </summary>
+    /// <typeparam name="TRequirement">The type of the requirement to evaluate.</typeparam>
+    /// <typeparam name="TResource">The type of the resource to evaluate.</typeparam>
     public abstract class AuthorizationHandler<TRequirement, TResource> : IAuthorizationHandler
         where TRequirement : IAuthorizationRequirement
     {
-        public virtual async Task HandleAsync(AuthorizationHandlerContext context)
+        async Task IAuthorizationHandler.HandleAsync(AuthorizationHandlerContext context)
         {
             if (context.Resource is TResource)
             {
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.Authorization/AuthorizationHandler.cs
```diff
@@ -6,38 +6,53 @@
...
+    /// <summary>
+    /// Base class for authorization handlers that need to be called for a specific requirement type.
+    /// </summary>
+    /// <typeparam name="TRequirement">The type of the requirement to handle.</typeparam>
     public abstract class AuthorizationHandler<TRequirement> : IAuthorizationHandler
-        where TRequirement : IAuthorizationRequirement
+            where TRequirement : IAuthorizationRequirement
     {
-        public void Handle(AuthorizationHandlerContext context)
+        async Task IAuthorizationHandler.HandleAsync(AuthorizationHandlerContext context)
```
HaoK: These should be public virtual, to enable extending and calling base.

sebastienros: My view on this is that this class is a helper class to provide a behavior (filtering on requirements) it doesn't make sense to rewrite it, or then the user would just have to implement the `IAuthorizationHandler` interface directly. On the opposite, making it public would let developers think then can override it instead of `Handle` or `HandleAsync`.

HaoK: The idea is for this class to be as flexible as possible. There is nothing wrong with overriding the public handle to add behavior.

blowdart: I'm on Sebastien's side on this. Damnit. People will end up overriding async, and that's not what we want. If you want to do that, then use the interface.

HaoK: That's not the point of it being virtual. What if they want to log every requirement or change the filtering slightly. I'm not sure why this one is special where we need to put a baby guard around it. All this accomplishes is waiting around until someone asks for this to be virtual again for any number of legitimate scenarios and we will add it back in 1.0.1 just like we did in for identity :)

blowdart: Then the non-async needs to go, and we only call the async one. Having both is nonsensical and ends up with people overriding the wrong one and wondering why it's never called.

HaoK: I think you are confused barry :)  There's two protected virtual (async and sync)...  The default protected Async behavior is to call the sync, so if you override that, the sync is never called.  That's different from the public virtual HandleAsync method which is doing the filtering, which is what I'm talking about.  But regardless, I want all 3 to be virtual :)

divega: > My view on this is that this class is a helper class to provide a behavior (filtering on requirements) it doesn't make sense to rewrite it, or then the user would just have to implement the `IAuthorizationHandler` interface directly. 

Reimplementing the interface member is much uglier an uncommon than just overriding. And I don't think we should be in the business of anticipating what valid scenarios customers are going to come up with. That is why we don't seal stuff that much anymore.

> On the opposite, making it public would let developers think then can override it instead of `Handle` or `HandleAsync`.

Sounds like something good doc comments can address.

src/Microsoft.AspNetCore.Authorization/AuthorizationHandler.cs
```diff
@@ -6,38 +6,53 @@
...
             }
         }
 
-        public virtual async Task HandleAsync(AuthorizationHandlerContext context)
+        /// <summary>
+        /// Makes a decision if authorization is allowed based on a specific requirement.
+        /// </summary>
+        /// <param name="context">The authorization information.</param>
+        /// <param name="requirement">The requirement to evaluate.</param>
+        protected virtual void Handle(AuthorizationHandlerContext context, TRequirement requirement)
```
divega: Why not just make `HandleAsync(context, requirement)` abstract and get rid of the sync version? Customers implementing custom `AuthorizationHandlers` that don't need to do async work should be able to figure out `Task.CompletedTask`.

sebastienros: I don't disagree but this is a much bigger breaking change as as of today this was the one that you HAD to implement. I also agree as the interface only expose the async one. But from @HaoK this class is also a way to provide a sugar around the sync version of Handle.

I really think that if the purpose of the class was clear, there wouldn't be any question about which methods to expose:
1- It's from filtering `TRequirement` then we should only expose `Handle(context, requirement)`
2- If it's to expose a synchronous method, then we need `Handle(context, requirement)`
3- Can be both

HaoK: Lets go with the more drastic but simpler for long term options.  Lets get rid of sync entirely and have only two methods, and rename the Handle to not have the same name.

Thoughts?
```
   public abstract AuthenticationHandler<TReq> : IAuthenticationHandler {
      public virtual Task HandleAsync(AuthenticationHandlerContext context);
      protected abstract Task HandleRequirementAsync(AuthenticationHandlerContext context, TReq requirement)
   }
```

sebastienros: Fine. But this is a significant breaking change. We also need to update the docs.

And if you decide to change your mind and not make `HandleAsync` virtual, feel free.

HaoK: Cool, I can make this change since its a bit bigger if you want.

@blowdart heads up.

sebastienros: You can work on the same PR to keep the comments.

HaoK:
Updated, this will pretty much break all existing users of the AuthenticationHandler but I think disambiguating the Handle methods is worth it.  The removal of the sync is a bit more questionable but I can go either way I guess...

@blowdart Thoughts?

HaoK:
35d0592701c4aa2edb6da919b629f87b8701807a

PR #856, opened 2016-05-31
Update sample ADAL dependency

Tratcher:
@rustd

### DIFF
--- a/samples/OpenIdConnect.AzureAdSample/Startup.cs
+++ b/samples/OpenIdConnect.AzureAdSample/Startup.cs
@@ -45,7 +45,7 @@ namespace OpenIdConnect.AzureAdSample
 
         public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
         {
-            loggerfactory.AddConsole(LogLevel.Information);
+            loggerfactory.AddConsole(Microsoft.Extensions.Logging.LogLevel.Information);
 
             // Simple error page
             app.Use(async (context, next) =>
--- a/samples/OpenIdConnect.AzureAdSample/project.json
+++ b/samples/OpenIdConnect.AzureAdSample/project.json
@@ -8,7 +8,7 @@
     "Microsoft.Extensions.Configuration.EnvironmentVariables": "1.0.0-*",
     "Microsoft.Extensions.Configuration.UserSecrets": "1.0.0-*",
     "Microsoft.Extensions.Logging.Console": "1.0.0-*",
-    "Microsoft.IdentityModel.Clients.ActiveDirectory": "3.9.302261508-alpha"
+    "Microsoft.IdentityModel.Clients.ActiveDirectory": "3.11.305310302-alpha"
   },
   "frameworks": {
     "net451": { },


samples/OpenIdConnect.AzureAdSample/Startup.cs
```diff
@@ -45,7 +45,7 @@ public void ConfigureServices(IServiceCollection services)
 
         public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
         {
-            loggerfactory.AddConsole(LogLevel.Information);
+            loggerfactory.AddConsole(Microsoft.Extensions.Logging.LogLevel.Information);
```
Tratcher: Conflict with ActiveDirectory.LogLevel

samples/OpenIdConnect.AzureAdSample/project.json
```diff
@@ -8,7 +8,7 @@
     "Microsoft.Extensions.Configuration.EnvironmentVariables": "1.0.0-*",
     "Microsoft.Extensions.Configuration.UserSecrets": "1.0.0-*",
     "Microsoft.Extensions.Logging.Console": "1.0.0-*",
-    "Microsoft.IdentityModel.Clients.ActiveDirectory": "3.9.302261508-alpha"
+    "Microsoft.IdentityModel.Clients.ActiveDirectory": "3.11.305310302-alpha"
   },
   "frameworks": {
     "net451": { },
```
rustd: We should remove the net451 target and only target .NET Core

Tratcher: No, all of these samples have this dual targeting. We use it for debugging.

rustd: Would this catch the current issue where ADAL was not working with .NET Core?

Tratcher: 3.9.302261508-alpha already worked for me, we never did find out why you had so much trouble with it.

rustd:
It did not work with the version of .NET Core and packages used in the templates in the stable builds for RC2. As long as some tests that scenario we should be good.

samples/OpenIdConnect.AzureAdSample/Startup.cs
davidfowl: That sucks... Will everyone run into this?

samples/OpenIdConnect.AzureAdSample/project.json
PinpointTownes: Hum, isn't `3.11.305310302-alpha` considered as a lower version by SemVer 1/NuGet?

Tratcher: no, the major, minor, and patch are all ints, it only gets into strict alphabetical in the prerelease tag.

samples/OpenIdConnect.AzureAdSample/Startup.cs
Tratcher: Using both ADAL and LogLevel shouldn't be too common. It only happens in one of our templates.

samples/OpenIdConnect.AzureAdSample/project.json
PinpointTownes: Ah yeah, I need glasses.

PR #862, opened 2016-06-02
Add a new IncludeErrorDetails option to prevent the JWT middleware from returning error/error_description

PinpointTownes:
We had agreed on the fact an option allowing to remove the error details from the WWW-Authenticate header was important, but it looks like it was not added by @Tratcher in his PR.

https://github.com/aspnet/Security/issues/776#issuecomment-210263382

/cc @Eilon @HaoK @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
@@ -191,6 +191,14 @@ namespace Microsoft.AspNetCore.Authentication.JwtBearer
             {
                 AuthenticateFailure = authResult?.Failure,
             };
+
+            // Avoid returning error=invalid_token if the error is not caused by an authentication failure (e.g missing token).
+            if (Options.IncludeErrorDetails && eventContext.AuthenticateFailure != null)
+            {
+                eventContext.Error = "invalid_token";
+                eventContext.ErrorDescription = CreateErrorDescription(eventContext.AuthenticateFailure);
+            }
+
             await Options.Events.Challenge(eventContext);
             if (eventContext.HandledResponse)
             {
@@ -203,9 +211,9 @@ namespace Microsoft.AspNetCore.Authentication.JwtBearer
 
             Response.StatusCode = 401;
 
-            var errorDescription = CreateErrorDescription(eventContext.AuthenticateFailure);
-
-            if (errorDescription.Length == 0)
+            if (string.IsNullOrEmpty(eventContext.Error) &&
+                string.IsNullOrEmpty(eventContext.ErrorDescription) &&
+                string.IsNullOrEmpty(eventContext.ErrorUri))
             {
                 Response.Headers.Append(HeaderNames.WWWAuthenticate, Options.Challenge);
             }
@@ -219,9 +227,35 @@ namespace Microsoft.AspNetCore.Authentication.JwtBearer
                     // Only add a comma after the first param, if any
                     builder.Append(',');
                 }
-                builder.Append(" error=\"invalid_token\", error_description=\"");
-                builder.Append(errorDescription);
-                builder.Append('\"');
+                if (!string.IsNullOrEmpty(eventContext.Error))
+                {
+                    builder.Append(" error=\"");
+                    builder.Append(eventContext.Error);
+                    builder.Append("\"");
+                }
+                if (!string.IsNullOrEmpty(eventContext.ErrorDescription))
+                {
+                    if (!string.IsNullOrEmpty(eventContext.Error))
+                    {
+                        builder.Append(",");
+                    }
+
+                    builder.Append(" error_description=\"");
+                    builder.Append(eventContext.ErrorDescription);
+                    builder.Append('\"');
+                }
+                if (!string.IsNullOrEmpty(eventContext.ErrorUri))
+                {
+                    if (!string.IsNullOrEmpty(eventContext.Error) ||
+                        !string.IsNullOrEmpty(eventContext.ErrorDescription))
+                    {
+                        builder.Append(",");
+                    }
+
+                    builder.Append(" error_uri=\"");
+                    builder.Append(eventContext.ErrorUri);
+                    builder.Append('\"');
+                }
 
                 Response.Headers.Append(HeaderNames.WWWAuthenticate, builder.ToString());
             }
@@ -231,11 +265,6 @@ namespace Microsoft.AspNetCore.Authentication.JwtBearer
 
         private static string CreateErrorDescription(Exception authFailure)
         {
-            if (authFailure == null)
-            {
-                return string.Empty;
-            }
-
             IEnumerable<Exception> exceptions;
             if (authFailure is AggregateException)
             {
--- a/test/Microsoft.AspNetCore.Authentication.Test/JwtBearer/JwtBearerMiddlewareTests.cs
+++ b/test/Microsoft.AspNetCore.Authentication.Test/JwtBearer/JwtBearerMiddlewareTests.cs
@@ -7,6 +7,7 @@ using System.Net;
 using System.Net.Http;
 using System.Reflection;
 using System.Security.Claims;
+using System.Text;
 using System.Threading.Tasks;
 using System.Xml.Linq;
 using Microsoft.AspNetCore.Builder;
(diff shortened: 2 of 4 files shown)


Tratcher:
You can clear JwtBearerChallengeContext.AuthenticateFailure to prevent any errors from being displayed. Looks like you want an Option instead.

PinpointTownes:
I also pushed a new commit adding new properties in `JwtBearerChallengeContext`: `Error` and `ErrorDescription`. They will allow one to override the default error/description without having to rewrite the entire `WWW-Authenticate` header, as the syntax is particularly error-prone.

PinpointTownes:
> Looks like you want an Option instead.

Yes, I do. Events are for advanced (flow) control. Such a simple thing deserves an easily "discoverable" option.

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -219,14 +224,14 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
...
-                builder.Append(errorDescription);
+                builder.Append($" error=\"{eventContext.Error ?? "invalid_token"}\", error_description=\"");
+                builder.Append(eventContext.ErrorDescription ?? "An error occurred when validating the access token.");
                 builder.Append('\"');
 
                 Response.Headers.Append(HeaderNames.WWWAuthenticate, builder.ToString());
             }
 
-            return false;
+            return true;
```
PinpointTownes: @Tratcher I believe returning `true` here was incorrect, since we're always handling the challenge response at this stage.

PinpointTownes:
Updated and rebased.

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
Tratcher: This should stay `false`. It doesn't just mean "did I handle this?", but rather "should I exclude others from handling this?" See https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs#L334
The redirecting providers are all exclusive and would return true, but providers like Basic, Bearer, etc. can all contribute challenges to the same response in an additive fashion.

I'll take care of this and get this merged today.

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -219,14 +225,14 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
                     // Only add a comma after the first param, if any
                     builder.Append(',');
                 }
-                builder.Append(" error=\"invalid_token\", error_description=\"");
-                builder.Append(errorDescription);
+                builder.Append($" error=\"{eventContext.Error}\", error_description=\"");
```
Tratcher: This defeats the point of using a StringBuilder. I'll fix it.

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
PinpointTownes: > The redirecting providers are all exclusive and would return true, but providers like Basic, Bearer, etc. can all contribute challenges to the same response in an additive fashion.

Why would non-interactive providers deserve a different treatment? There's fundamentally no difference between "redirecting providers" and the ones you mentioned.

Here, we could potentially end up with multiple `WWW-Authenticate` headers. While it's not forbidden by the HTTP spec (at least, AFAIK), it's not something most "REST clients" expect.

But if you think it's better, I'll revert this change.

PinpointTownes: Reverted.

Tratcher: Good point. Multiple WWW-Authenticate headers is legal, but browsers don't handle it well. We should be appending to a single header value.

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
PinpointTownes: Addressed :+1:

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
PinpointTownes: Pretty sure having multiple values in the same `WWW-Authenticate` header is valid, but in practice, I'm not sure many clients will correctly parse it when trying to extract the `error`/`error_description`.

Tratcher:
Thanks. Related: https://github.com/aspnet/Security/issues/864

PinpointTownes:
If you can wait 2 or 3 minutes, I'd like to try a variation of this PR (being able to specify error and/or error_description independently)

Tratcher:
Sure, go for it.

PinpointTownes:
Well, it needed a bit more than 2/3 minutes :smile: 

2 variations:

  - One allowing to specify error and error_description independently (you can return both, one of them or none).
  - One adding error_uri support and allowing error/error_description/error_uri to be set independently.

PinpointTownes:
I also updated the `if (Options.IncludeErrorDetails)` check to avoid returning `error`/`error_description` when the token was not included in the request.

```csharp
if (Options.IncludeErrorDetails && (authResult != null && !authResult.Skipped))
{
    eventContext.Error = "invalid_token";
    eventContext.ErrorDescription = CreateErrorDescription(eventContext.AuthenticateFailure);
}
```

test/Microsoft.AspNetCore.Authentication.Test/JwtBearer/JwtBearerMiddlewareTests.cs
```diff
@@ -179,11 +180,103 @@ public class JwtBearerMiddlewareTests
...
+            Assert.Equal("Bearer", response.Response.Headers.WwwAuthenticate.First().ToString());
+            Assert.Equal("", response.ResponseText);
+        }
+
+        [Fact]
+        public async Task ExceptionNotReportedInHeaderWhenTokenWasMissing()
+        {
+            var server = CreateServer(new JwtBearerOptions
+            {
+                IncludeErrorDetails = false
```
Tratcher: This should be true, and when you set it I think this test will fail.

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -191,6 +191,13 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
             {
                 AuthenticateFailure = authResult?.Failure,
             };
+
+            if (Options.IncludeErrorDetails && (authResult != null && !authResult.Skipped))
+            {
+                eventContext.Error = "invalid_token";
```
Tratcher: This sets `error` even when there was no Failure. Auth may have been successful and the app could choose to return 401 anyways. Check for Failure rather than Skipped.

test/Microsoft.AspNetCore.Authentication.Test/JwtBearer/JwtBearerMiddlewareTests.cs
Tratcher: Ah, it would still pass because you're checking for Skipped...

PinpointTownes: Fixed, and yes, it passes :sweat_smile:

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
PinpointTownes: Fixed :+1:

Tratcher:
Great, that should do it. Rebase and squash please.

PinpointTownes:
Rebased and squashed.

PinpointTownes:
Thanks for merging it :+1:

PR #867, opened 2016-06-04
Update Json.NET to 9.0.1-beta1

javiercn:


Tratcher:
JwtBearer and OpenIdConnect both depend on this too, by I can't see where the dependency is declared.

PinpointTownes:
@Tratcher https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/src/Microsoft.IdentityModel.Tokens/project.json#L6

javiercn:
Good catch, we probably want to add the reference to the project.json directly to force it to the right version? @pranavkm

dougbu:
@javiercn we _like_ transitive references

Tratcher:
@javiercn file a bug in AzureAd for them to upgrade to the same version. I assume there's a plan for RTM?

javiercn:
@Tratcher I would think so, but will check with @Eilon

Eilon:
Fix for now here, but also log a bug in AzureAD.

javiercn:
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/431

PR #876, opened 2016-06-09
Add a middleware for authenticating users using Vkontakte social network

serber:
An ASP.NET Core middleware for authenticating users using Vkontakte (https://vk.com)


blowdart:
Whilst your PR is appreciated we're only supporting the major providers, so we're unable to take this.

There's a community project for providers at https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers who would welcome your contribution.

Issue #816, opened 2016-05-12
Add docs to cover: AuthorizeAsync should not require a resource to be provided when passing a requirement


Eilon:
Given that we're not doing https://github.com/aspnet/Security/issues/753 for 1.0.0 RTM, need to add some docs to that effect.

Eilon:
@blowdart @HaoK anything we plan to do here for 1.0.0? If we can get something done ASAP we can try to sneak it in. Otherwise it'll have to wait.

HaoK:
I believe this was a doc doc issue, not xml doc

HaoK:
Or maybe @blowdart already updated the doc comments to address this

blowdart:
Nope. We probably should though.

How about

        /// <summary>
        /// Checks if a user meets a specific set of requirements for the specified resource
        /// </summary>
        /// <param name="user">The user to evaluate the requirements against.</param>
        /// <param name="resource">
        /// An optional resource the policy should be checked with.
        /// If a resource is not required for policy evaluation you may pass null as the value.
        /// </param>
        /// <param name="requirements">The requirements to evaluate.</param>
        /// <returns>
        /// A flag indicating whether authorization has succeeded.
        /// This value is <value>true</value> when the user fulfills the policy; otherwise <value>false</value>.
        /// </returns>
        /// <remarks>
        /// Resource is an optional parameter and may be null. Please ensure that you check it is not 
        /// null before acting upon it.
        /// </remarks>
        Task<bool> AuthorizeAsync(ClaimsPrincipal user, object resource, IEnumerable<IAuthorizationRequirement> requirements);

        /// <summary>
        /// Checks if a user meets a specific authorization policy
        /// </summary>
        /// <param name="user">The user to check the policy against.</param>
        /// <param name="resource">
        /// An optional resource the policy should be checked with.
        /// If a resource is not required for policy evaluation you may pass null as the value.
        /// </param>
        /// <param name="policyName">The name of the policy to check against a specific context.</param>
        /// <returns>
        /// A flag indicating whether authorization has succeeded.
        /// Returns a flag indicating whether the user, and optional resource has fulfilled the policy.    
        /// <value>true</value> when the the policy has been fulfilled; otherwise <value>false</value>.
        /// </returns>
        /// <remarks>
        /// Resource is an optional parameter and may be null. Please ensure that you check it is not 
        /// null before acting upon it.
        /// </remarks>
[...]

PR #881, opened 2016-06-13
Doc comment update

HaoK:
Fixes https://github.com/aspnet/Security/issues/816

cc @Eilon @blowdart


Eilon:
:shipit:

anurse:
Should this comment also be applied to `DefaultAuthorizationService`? It looks like it is otherwise identical.

HaoK:
Where is inheritdoc when you need it

PR #893, opened 2016-07-08
Remote auth expiration fix

### LINKED ISSUE #855: Cookie ExpireTimeSpan not honoured using Auzure AD OpenIDConnect authentication
I am currently struggling with setting the timeout on the cookie/auth token when authenticating my .NET Core App using Azure AD via the OpenIdConnect authentication model.

The sign-in scheme is being set in the ConfigureServices method via the following:

    services.AddAuthentication(options => options.SignInScheme = CookieAuthenticationDefaults.AuthenticationScheme);

I am then setting up my configuration as follows:

    app.UseCookieAuthentication(new CookieAuthenticationOptions()
    {
        CookieName = "MyCookie",
        ExpireTimeSpan = TimeSpan.FromHours(2)
    });

    app.UseOpenIdConnectAuthentication(new OpenIdConnectOptions()
    {
        Authority = authorityUri.AbsoluteUri,
        ClientId = azureOptions.ClientId,
        ClientSecret = azureOptions.ClientSecret,
        ResponseType = OpenIdConnectResponseTypes.CodeIdToken,
        Events = new OpenIdConnectEvents()
        {
            OnAuthorizationCodeReceived = async context =>
            {
                await aAuthenticateMiddleware.OnAuthenticate(context, logger);
            }
        }
    });

    app.UseMiddleware<aAuthenticateMiddleware>();

Note, that I am not using the built in Identity (as its not practical for our purposes) but rather using a custom middleware.

Within the middleware layer I am checking whether the user is authenticated and if not a challenge is issued:

    var authenticationProperties = new AuthenticationProperties() { RedirectUri = context.Request.Path.Value ?? "/" };
    authenticationProperties.AllowRefresh = false;
    authenticationProperties.IssuedUtc = DateTime.Now;
    authenticationProperties.ExpiresUtc = DateTime.Now.AddHours(2);
    await context.Authentication.ChallengeAsync(
        authenticationManager.IdentityProvider.AuthenticationScheme,
        authenticationProperties,
        ChallengeBehavior.Automatic
    );

This is all works fine and authenticates the user correctly etc however this is issuing the auth token (and cookie) with a 15 minute expiry and ignoring my 2 hour expiry that I have tried setting.

I have been referring to the latest source examples from GitHub from the aspnet/security repository for examples.... however none of these mention anything about overriding the default expiry issued.

https://github.com/aspnet/Security/tree/dev/samples/OpenIdConnect.AzureAdSample

Most examples I have found are still referencing the old AspNet libraries rather than the AspNetCore libraries.

Some articles suggest that using the SignInAsync with persistent set to True allows the ExpireTimeSpan to be honored, however this throws a "Not Supported Exception" when calling it. Perhaps SignInAsync is not supported via Azure AD?

Not sure whether this is a bug???

tuespetre:
Fixes #855, which documents a bug where the `RemoteAuthenticationTimeout` (intended to be used for a correlation cookie) also ends up getting used for the `ExpiresUtc` value of the `AuthenticationProperties` of the actual authentication ticket that gets issued. 

I did not see a generic set of tests for `RemoteAuthenticationHandler` so I refrained from adding a test for now.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
@@ -149,7 +149,7 @@ namespace Microsoft.AspNetCore.Authentication
             {
                 HttpOnly = true,
                 Secure = Request.IsHttps,
-                Expires = properties.ExpiresUtc
+                Expires = Options.SystemClock.UtcNow.Add(Options.RemoteAuthenticationTimeout),
             };
 
             properties.Items[CorrelationProperty] = correlationId;
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
@@ -139,12 +139,6 @@ namespace Microsoft.AspNetCore.Builder
         public ISecureDataFormat<AuthenticationTicket> TicketDataFormat { get; set; }
 
         /// <summary>
-        /// For testing purposes only.
-        /// </summary>
-        [EditorBrowsable(EditorBrowsableState.Never)]
-        public ISystemClock SystemClock { get; set; }
-
-        /// <summary>
         /// The component used to get cookies from the request or set them on the response.
         ///
         /// ChunkingCookieManager will be used by default.
--- a/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerOptions.cs
@@ -96,12 +96,6 @@ namespace Microsoft.AspNetCore.Builder
         public bool RefreshOnIssuerKeyNotFound { get; set; } = true;
 
         /// <summary>
-        /// For testing purposes only.
-        /// </summary>
-        [EditorBrowsable(EditorBrowsableState.Never)]
-        public ISystemClock SystemClock { get; set; } = new SystemClock();
-
-        /// <summary>
         /// Gets the ordered list of <see cref="ISecurityTokenValidator"/> used to validate access tokens.
         /// </summary>
         public IList<ISecurityTokenValidator> SecurityTokenValidators { get; } = new List<ISecurityTokenValidator> { new JwtSecurityTokenHandler() };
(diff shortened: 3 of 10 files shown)


Tratcher:
Notes: Moving ISystemClock to AuthenticationOptions should be OK, it was already on every child class in this repo.

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -149,7 +149,7 @@ protected virtual void GenerateCorrelationId(AuthenticationProperties properties
             {
                 HttpOnly = true,
                 Secure = Request.IsHttps,
-                Expires = properties.ExpiresUtc
+                Expires = Options.SystemClock.UtcNow.Add(Options.RemoteAuthenticationTimeout),
```
Tratcher: Please add this to TwitterHandler.HandleUnauthorizedAsync as well.

Tratcher:
:shipit: pending the update to Twitter.

tuespetre:
:up::date:

HaoK:
Should we consider putting the system clock in DI as a service instead of hanging them off the options?  Its not like we would ever expect more than one system clock across all the different middlewares, etc right?

HaoK:
Basically what I'm suggesting instead of putting the clock in AuthenticationOptions, instead add it as a property to AuthenticationMiddleware/Handler that gets injected optionally (and defaults to a new instance of SystemClock same as today)

```
   // On AuthenticationMiddleware and AuthenticationHandler
   public virtual ISystemClock SystemClock { get; set; } = new SystemClock()

  // Invoke needs to set the system clock on AuthenticationHandler after its created (patching intialize basically)
        {
            var handler = CreateHandler();  
            handler.SystemClock = SystemClock;
            ...

  // Update all of the various AuthMiddleware ctors 
        public CookieAuthenticationMiddleware(
            RequestDelegate next,
            IDataProtectionProvider dataProtectionProvider,
            ILoggerFactory loggerFactory,
            UrlEncoder urlEncoder,
            IOptions<CookieAuthenticationOptions> options,
            ISystemClock clock = null)  
        if (clock != null) {
            SystemClock = clock;
        }
        ...
  
  
```

Tratcher:
We've considered moving it to a service before, but it's only used for unit tests and its easier to set in options.

PR #897, opened 2016-07-12
Two minor updates

troydai:


src/Microsoft.AspNetCore.Authentication/Win32.cs
```diff
@@ -3,7 +3,6 @@
 
 
 using System;
-using System.ComponentModel;
 using System.Runtime.InteropServices;
 
 namespace Microsoft.Win32
```
PinpointTownes: BTW, is this file still used? IIRC, it was used by the Twitter provider for cert pinning, but I'm pretty sure it was removed last year. /cc @Tratcher

.vscode/settings.json
```diff
@@ -0,0 +1,14 @@
+// Place your settings in this file to overwrite default and user settings.
```
Tratcher: Do we have this checked into any of our other repos?

troydai: No. But it is for the convenience of the VS code user.

Tratcher: @Eilon @muratg, do you want this in all of our repos? We shouldn't put it in just one, there's nothing special about this repo.

troydai:
Remove the vs code settings. We can think about it next time.

src/Microsoft.AspNetCore.Authentication/Win32.cs
Tratcher: Remove?

Tratcher:
Can you remove Win32.cs?

troydai:
sure.

troydai:
Updated

troydai:
@Tratcher looking good?

Tratcher:
:shipit:

Issue #53, opened 2014-08-27
Google: Need better way to discover when google+ api not enabled


HaoK:
It would be helpful for the google middleware to handle the situation where just the google+ me request fails.  Currently today this results in a 403, which is caught and swallowed.  OnAuthenticated isn't called, and I had to debug into sources to see that it was the google plus me request failing here:

                // Get the Google user
                HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, UserInfoEndpoint);
                request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", accessToken);
                HttpResponseMessage graphResponse = await _httpClient.SendAsync(request, Context.RequestAborted);
                graphResponse.EnsureSuccessStatusCode();

Do we need some kind of error callback?  Instead of just getting called back with a null Identity in the authentication result?

HaoK:
@Tratcher So I tried this again with our new changes, the main problem is we only get a 403 back with no other error when things are misconfigured, so we currently just blow up when we try to EnsureSuccessCode.  The best we can do is special case 403 to say "failed, Forbidden by OAuth provider"?  Any other ideas?

muratg:
POR: do not throw, give the status code 403.  We should also consider documenting why this may be happening.

HaoK:
238fdf24e8b0e3c5e84cd8dc7f7324502a439429

anurse:
This came up during my testing and it was hard to diagnose. Can we improve the error message? Right now it just says:

```
An unhandled exception occurred while processing the request.

HttpRequestException: Response status code does not indicate success: 403 ().
System.Net.Http.HttpResponseMessage.EnsureSuccessStatusCode()
```

Eilon:
Yes, but not in RC2 😄

troydai:
Fixed in https://github.com/aspnet/Security/pull/905/commits/5de4816f59a87d00436bf386b23d9ff10e9f6bf6

PR #905, opened 2016-07-19
Handle  the remote authentication failure gracefully

troydai:
Address issue #53 

Error message now:
image

The `GoogleHandler` throws when the response from back channel is out of 2XX range. The exception neither provides sufficient information for user to correct the error or is handled gracefully in `OAuthHandler`

The PR contains two commit. The first one is purely for refactoring and readability improvement. For the actual logic change you can focus on the [second commit](https://github.com/aspnet/Security/commit/967c7cdeb6838368ac775a526b84abd164dd6fae)

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Globalization;
 using System.Net.Http;
 using System.Security.Claims;
@@ -35,7 +36,11 @@ namespace Microsoft.AspNetCore.Authentication.Facebook
             }
 
             var response = await Backchannel.GetAsync(endpoint, Context.RequestAborted);
-            response.EnsureSuccessStatusCode();
+            if (!response.IsSuccessStatusCode)
+            {
+                var errorMessage = $"Failed to retrived Facebook user information ({response.StatusCode}) Please check if the authentication information is correct and the corresponding Google API is enabled.";
+                throw new InvalidOperationException(errorMessage);
+            }
 
             var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
 
@@ -119,7 +124,7 @@ namespace Microsoft.AspNetCore.Authentication.Facebook
             {
                 identity.AddClaim(new Claim(ClaimTypes.Name, name, ClaimValueTypes.String, Options.ClaimsIssuer));
             }
- 
+
             var timeZone = FacebookHelper.GetTimeZone(payload);
             if (!string.IsNullOrEmpty(timeZone))
             {
--- a/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
@@ -38,8 +38,7 @@ namespace Microsoft.AspNetCore.Authentication.Facebook
             var response = await Backchannel.GetAsync(endpoint, Context.RequestAborted);
             if (!response.IsSuccessStatusCode)
             {
-                var errorMessage = $"Failed to retrived Facebook user information ({response.StatusCode}) Please check if the authentication information is correct and the corresponding Google API is enabled.";
-                throw new InvalidOperationException(errorMessage);
+                throw new HttpRequestException($"Failed to retrived Facebook user information ({response.StatusCode}) Please check if the authentication information is correct and the corresponding Facebook API is enabled.");
             }
 
             var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
--- a/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
@@ -38,7 +38,7 @@ namespace Microsoft.AspNetCore.Authentication.Facebook
             var response = await Backchannel.GetAsync(endpoint, Context.RequestAborted);
             if (!response.IsSuccessStatusCode)
             {
-                throw new HttpRequestException($"Failed to retrived Facebook user information ({response.StatusCode}) Please check if the authentication information is correct and the corresponding Facebook API is enabled.");
+                throw new HttpRequestException($"Failed to retrived Facebook user information ({response.StatusCode}) Please check if the authentication information is correct and the corresponding Facebook Graph API is enabled.");
             }
 
             var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
--- a/src/Microsoft.AspNetCore.Authentication.Google/GoogleHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Google/GoogleHandler.cs
@@ -34,8 +34,7 @@ namespace Microsoft.AspNetCore.Authentication.Google
             var response = await Backchannel.SendAsync(request, Context.RequestAborted);
             if (!response.IsSuccessStatusCode)
             {
-                var errorMessage = $"Failed to retrived Google user information ({response.StatusCode}) Please check if the authentication information is correct and the corresponding Google API is enabled.";
-                throw new InvalidOperationException(errorMessage);
+                throw new HttpRequestException($"An error occurred when retrieving user information ({response.StatusCode}). Please check if the authentication information is correct and the corresponding Google API is enabled.");
             }
 
             var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
--- a/src/Microsoft.AspNetCore.Authentication.Google/GoogleHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Google/GoogleHandler.cs
@@ -34,7 +34,7 @@ namespace Microsoft.AspNetCore.Authentication.Google
             var response = await Backchannel.SendAsync(request, Context.RequestAborted);
             if (!response.IsSuccessStatusCode)
             {
-                throw new HttpRequestException($"An error occurred when retrieving user information ({response.StatusCode}). Please check if the authentication information is correct and the corresponding Google API is enabled.");
+                throw new HttpRequestException($"An error occurred when retrieving user information ({response.StatusCode}). Please check if the authentication information is correct and the corresponding Google+ API is enabled.");
             }
 
             var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
--- a/src/Microsoft.AspNetCore.Authentication.Google/GoogleHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Google/GoogleHandler.cs
@@ -32,7 +32,11 @@ namespace Microsoft.AspNetCore.Authentication.Google
             request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", tokens.AccessToken);
 
             var response = await Backchannel.SendAsync(request, Context.RequestAborted);
-            response.EnsureSuccessStatusCode();
+            if (!response.IsSuccessStatusCode)
+            {
+                var errorMessage = $"Failed to retrived Google user information ({response.StatusCode}) Please check if the authentication information is correct and the corresponding Google API is enabled.";
+                throw new InvalidOperationException(errorMessage);
+            }
 
             var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
 
--- a/src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
@@ -119,7 +119,15 @@ namespace Microsoft.AspNetCore.Authentication.OAuth
                 properties.StoreTokens(authTokens);
             }
 
-            return AuthenticateResult.Success(await CreateTicketAsync(identity, properties, tokens));
+            try
+            {
+                var ticket = await CreateTicketAsync(identity, properties, tokens);
+                return AuthenticateResult.Success(ticket);
+            }
+            catch (Exception ex)
+            {
+                return AuthenticateResult.Fail(ex);
+            }
         }
 
         protected virtual async Task<OAuthTokenResponse> ExchangeCodeAsync(string code, string redirectUri)
--- a/src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
@@ -8,6 +8,7 @@ using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Security.Claims;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http.Authentication;
@@ -15,7 +16,6 @@ using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Http.Features.Authentication;
 using Microsoft.Extensions.Primitives;
 using Newtonsoft.Json.Linq;
-using System.Threading;
 
 namespace Microsoft.AspNetCore.Authentication.OAuth
 {
--- a/src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
@@ -15,6 +15,7 @@ using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Http.Features.Authentication;
 using Microsoft.Extensions.Primitives;
 using Newtonsoft.Json.Linq;
+using System.Threading;
 
 namespace Microsoft.AspNetCore.Authentication.OAuth
 {
@@ -119,14 +120,14 @@ namespace Microsoft.AspNetCore.Authentication.OAuth
                 properties.StoreTokens(authTokens);
             }
 
-            try
+            var ticket = await CreateTicketAsync(identity, properties, tokens);
+            if (ticket != null)
             {
-                var ticket = await CreateTicketAsync(identity, properties, tokens);
                 return AuthenticateResult.Success(ticket);
             }
-            catch (Exception ex)
+            else
             {
-                return AuthenticateResult.Fail(ex);
+                return AuthenticateResult.Fail("Failed to retrieve user information from remote server.");
             }
         }
 
--- a/src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
@@ -8,7 +8,6 @@ using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Security.Claims;
 using System.Text;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http.Authentication;
--- a/src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
@@ -87,6 +87,12 @@ namespace Microsoft.AspNetCore.Authentication
             return true;
         }
 
+        /// <summary>
+        /// Authenticate the user identity with the identity provider. 
+        ///
+        /// This could be done through a back channel communication with the identity provider. Exception thrown during
+        /// the authenticating should be saved to the AuthenticateResult.
+        /// </summary>
         protected abstract Task<AuthenticateResult> HandleRemoteAuthenticateAsync();
 
         protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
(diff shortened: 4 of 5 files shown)


src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -129,7 +129,15 @@ public OAuthHandler(HttpClient backchannel)
                 properties.StoreTokens(authTokens);
             }
 
-            return AuthenticateResult.Success(await CreateTicketAsync(identity, properties, tokens));
+            try
+            {
+                var ticket = await CreateTicketAsync(identity, properties, tokens);
+                return AuthenticateResult.Success(ticket);
+            }
+            catch(Exception ex)
```
HaoK: This try catch/fail should probably go lower down, perhaps around the HandleRemoteAuthenticate in the base class instead 

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs#L34

src/Microsoft.AspNetCore.Authentication.Google/GoogleHandler.cs
```diff
@@ -29,7 +29,11 @@ public GoogleHandler(HttpClient httpClient)
             request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", tokens.AccessToken);
 
             var response = await Backchannel.SendAsync(request, Context.RequestAborted);
-            response.EnsureSuccessStatusCode();
+            if (!response.IsSuccessStatusCode)
+            {
+                var errorMessage = $"Failed to retrived Google user information. Please check if the authentication information is correct and the corresponding Google API is enabled.";
```
Tratcher: Include the actual response status code?

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -35,26 +47,24 @@ public OAuthHandler(HttpClient backchannel)
...
             {
-                var failureMessage = new StringBuilder();
-                failureMessage.Append(error);
+                var failureMessage = new StringBuilder(error);
+
                 var errorDescription = query["error_description"];
                 if (!StringValues.IsNullOrEmpty(errorDescription))
                 {
-                    failureMessage.Append(";Description=").Append(errorDescription);
+                    failureMessage.Append($";Description={errorDescription}");
```
Tratcher: Stick with Append for 'perf' reasons.

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
Tratcher: Actually I think this is a good spot. The other parts of this code can handle their own errors and return AuthenticateResult.Fail, but CreateTicketAsync can't.

Tratcher:
Please don't mix style updates and functional changes in the same PR.

src/Microsoft.AspNetCore.Authentication.Google/GoogleHandler.cs
Tratcher: What about the other providers? Can they be any more specific?

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
troydai: ok.

troydai:
I'll remove the refactoing

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
HaoK: I think it would be nicer to trap all exceptions that escape HandleRemoteAuthenticateAsync, and surface them to the RemoteFailure event, otherwise today the remote failure event would get by passed

troydai: I'm with @Tratcher. I don't think `RemoteRemoteAuthenticateAsync` should throw. It returns `AuthenticateResult` which has a exception property on it. That indicate the method should handle the exception itself. On the other hand `CreateTicketAsync` returns `AuthenticationTicket` which doesn't have a property to carry error message.

HaoK: I don't think its an obvious expectation for developers to know to try/catch always around `HandleRemoteAuthenticateAsync`, seems much more defensive to just prevent exceptions from leaking ourselves.  Basically, the choice is between putting some doc comments saying uncaught exceptions in `HandleRemoteAuthenticateAsync` bypass the remote error handler, or just guarding against that possibility ourselves.

troydai: If we wrap every invocation to `HandleRemoteAuthenticateAsync` with a try-catch, a conscious decision should be made to justify the two means of report errors (exception property vs. thrown exception).

This maybe helpful, but I'll say we need a case of unhandled exception in `HandleRemoteAuthenticateAsync` itself to study if such measure is helpful. Rather than get ahead of this issue itself.

I'll add comment to the signature of `HandleRemoteAuthenticateAsync` to state user should handle exception in it or face a leaked exception.

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -87,6 +87,12 @@ public abstract class RemoteAuthenticationHandler<TOptions> : AuthenticationHand
             return true;
         }
 
+        /// <summary>
+        /// Authenticate the user identity with the identity provider. 
+        ///
+        /// This could be done through a back channel communication with the identity provider. Exception thrown during
+        /// the authenticating should be saved to the AuthenticateResult.
```
HaoK: I still think the try/catch should be moved here, the whole point of this class being created was to streamline the error/exception handling of Auth Handlers.  With this comment, we are basically saying its pretty much a bug if they don't have a try/catch/Fail in their implementation of `HandleRemoteAuthenticateAsync`... which at that point we might as well do the Try/Catch/Fail when we call this method so its foolproof.  That is my main argument.

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
```diff
@@ -35,7 +36,11 @@ public FacebookHandler(HttpClient httpClient)
             }
 
             var response = await Backchannel.GetAsync(endpoint, Context.RequestAborted);
-            response.EnsureSuccessStatusCode();
+            if (!response.IsSuccessStatusCode)
+            {
+                var errorMessage = $"Failed to retrived Facebook user information ({response.StatusCode}) Please check if the authentication information is correct and the corresponding Google API is enabled.";
```
PinpointTownes: That's the downside of long lines, you always end up missing something (obvious) when copying/pasting them :smile: 

> Failed to retrived Facebook user information ({response.StatusCode}) Please check if the authentication information is correct and the corresponding Google API is enabled.

PinpointTownes: FWIW, here's what we do in all the aspnet-contrib OAuth2 providers:

```csharp
var response = await Backchannel.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, Context.RequestAborted);
if (!response.IsSuccessStatusCode) {
    Logger.LogError("An error occurred when retrieving the user profile: the remote server " +
                    "returned a {Status} response with the following payload: {Headers} {Body}.",
                    /* Status: */ response.StatusCode,
                    /* Headers: */ response.Headers.ToString(),
                    /* Body: */ await response.Content.ReadAsStringAsync());

    throw new HttpRequestException("An error occurred when retrieving the user profile.");
}
```

Ideally, we shouldn't have to throw an exception, but returning `null` is not legal as `OAuthHandler.HandleRemoteAuthenticateAsync` crashes if the ticket is null.

IMHO, you should fix `HandleRemoteAuthenticateAsync` to detect if the ticket returned by `CreateTicketAsync` is null and ultimately return `AuthenticateResult.Failed` instead of calling `AuthenticateResult.Success` blindly.

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/src/AspNet.Security.OAuth.GitHub/GitHubAuthenticationHandler.cs#L33-L39

PinpointTownes:
FWIW, here's what we do in all the aspnet-contrib OAuth2 providers:

```csharp
var response = await Backchannel.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, Context.RequestAborted);
if (!response.IsSuccessStatusCode) {
    Logger.LogError("An error occurred when retrieving the user profile: the remote server " +
                    "returned a {Status} response with the following payload: {Headers} {Body}.",
                    /* Status: */ response.StatusCode,
                    /* Headers: */ response.Headers.ToString(),
                    /* Body: */ await response.Content.ReadAsStringAsync());

    throw new HttpRequestException("An error occurred when retrieving the user profile.");
}
```

Ideally, we shouldn't have to throw an exception, but returning `null` is not legal as `OAuthHandler.HandleRemoteAuthenticateAsync` crashes if the ticket is null.

IMHO, you should fix `HandleRemoteAuthenticateAsync` to detect if the ticket returned by `CreateTicketAsync` is null and ultimately return `AuthenticateResult.Failed` instead of calling `AuthenticateResult.Success` blindly.

https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/dev/src/AspNet.Security.OAuth.GitHub/GitHubAuthenticationHandler.cs#L33-L39

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
troydai: Thanks! I'll update it.

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
troydai: Thanks for the suggestion. I'll rethink the overall error handling mechanism here.

HaoK:
FWIW @PinpointTownes I agree with you, the error handling via throwing exceptions for something like a bad status code is pretty gross.  Not sure if we can/should do breaking changes to address this (ironically its likely you would be the one who's broken the most)

HaoK:
The problem with a easy fix like if CreateTicket returns null => Fail with generic error message, is that defeats many of our attempts to improve the error message.  So what we have today via throwing an "better" exception, still seems better than a bland generic (and fairly useless) "Failed to create OAuth ticket" error.

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
Tratcher: The .NET Exception guidelines are a useful reference: https://msdn.microsoft.com/en-us/library/ms229005(v=vs.100).aspx
The relevant bit here is to avoid catching too broad a range of exception types. E.g. only catch exceptions when you know what causes them.

troydai:
Updated:

1. Exeption will be still thrown from CreateTicketAsync to avoid breaking
change.
2. OAuthHandler won't try to capture the exception.
3. OAuthHandler will check if the AuthenticateTicket is null.
4. RemoteAuthenticationHandler's error handling flow is improved to cover
the exception thrown from HandleRemoteAuthenticateAsync. The exeption
thrown from it will go throw the user error handling logic, too.

/cc @HaoK @Tratcher

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -31,26 +32,49 @@ public abstract class RemoteAuthenticationHandler<TOptions> : AuthenticationHand
...
             {
-                return false;
+                authResult = await HandleRemoteAuthenticateAsync();
+                if (authResult != null && authResult.Skipped == true)
+                {
+                    return false;
+                }
+                else if (authResult == null)
+                {
+                    exception = new InvalidOperationException("Invalide return state, unable to redirect.");
```
HaoK: Invalid

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -31,26 +32,49 @@ public abstract class RemoteAuthenticationHandler<TOptions> : AuthenticationHand
...
+                }
+                else if (!authResult.Succeeded)
+                {
+                    exception = authResult?.Failure ??
+                                new InvalidOperationException("Invalide return state, unable to redirect.");
```
HaoK: Invalid

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -31,26 +32,49 @@ public abstract class RemoteAuthenticationHandler<TOptions> : AuthenticationHand
...
+                Logger.RemoteAuthenticationError(exception.Message);
+                var errorContext = new FailureContext(Context, exception);
                 await Options.Events.RemoteFailure(errorContext);
+
                 if (errorContext.HandledResponse)
                 {
                     return true;
                 }
-                if (errorContext.Skipped)
+                else if (errorContext.Skipped)
```
HaoK: nit: why nest? easier to read without unnecessary else/indenting

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -31,26 +32,49 @@ public abstract class RemoteAuthenticationHandler<TOptions> : AuthenticationHand
...
                 {
                     return false;
                 }
-
-                throw new AggregateException("Unhandled remote failure.", errorContext.Failure);
+                else
```
HaoK: Same thing, easier to read if the bottom is a clear throw, without having to scroll up to figure out what the else is for

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -31,26 +32,49 @@ public abstract class RemoteAuthenticationHandler<TOptions> : AuthenticationHand
 
         protected virtual async Task<bool> HandleRemoteCallbackAsync()
         {
-            var authResult = await HandleRemoteAuthenticateAsync();
-            if (authResult != null && authResult.Skipped)
+            AuthenticateResult authResult = null;
```
HaoK: nit: I think you can declare the ticket here instead of the result, and leave the authResult inside of the try catch, and only set the ticket in a final else

HaoK:
Looks good overall, I would suggest adding a test or two to ensure that various exceptions (caught/uncaught) do the right thing now against the base class (both with the built in middlewares and a custom derived one)

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
troydai: will do.

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
troydai: thanks. will update.

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
troydai: will update.

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -15,6 +15,7 @@
 using Microsoft.AspNetCore.Http.Features.Authentication;
 using Microsoft.Extensions.Primitives;
 using Newtonsoft.Json.Linq;
+using System.Threading;
```
Tratcher: sort

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.AspNetCore.Http.Features.Authentication;
 using Microsoft.AspNetCore.Http.Authentication;
 using Microsoft.Extensions.Logging;
+using System.Net.Http;
```
Tratcher: sort

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -119,7 +120,15 @@ public OAuthHandler(HttpClient backchannel)
...
 
-            return AuthenticateResult.Success(await CreateTicketAsync(identity, properties, tokens));
+            var ticket = await CreateTicketAsync(identity, properties, tokens);
+            if (ticket != null)
+            {
+                return AuthenticateResult.Success(ticket);
+            }
+            else
+            {
+                return AuthenticateResult.Fail("Failed to retrieve user information from remote server.");
```
Tratcher: What's this for? You're not returning null anywhere.

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -87,6 +111,11 @@ public abstract class RemoteAuthenticationHandler<TOptions> : AuthenticationHand
             return true;
         }
 
+        /// <summary>
+        /// Authenticate the user identity with the identity provider. 
+        ///
+        /// This could be done through a back channel communication with the identity provider.
```
Tratcher: The backchannel isn't the important part here, it's that this processes requests on the CallbackPath endpoint.

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
troydai: The method could be overridden.

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
```diff
@@ -35,7 +36,10 @@ public FacebookHandler(HttpClient httpClient)
             }
 
             var response = await Backchannel.GetAsync(endpoint, Context.RequestAborted);
-            response.EnsureSuccessStatusCode();
+            if (!response.IsSuccessStatusCode)
+            {
+                throw new HttpRequestException($"Failed to retrived Facebook user information ({response.StatusCode}) Please check if the authentication information is correct and the corresponding Facebook API is enabled.");
```
PinpointTownes: I suppose logging a warning message + returning `null` is not an option?

troydai:
@Tratcher mind take another look?

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
PinpointTownes: :+1: for allowing `null` to be returned.

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -31,30 +31,54 @@ public abstract class RemoteAuthenticationHandler<TOptions> : AuthenticationHand
...
-            var authResult = await HandleRemoteAuthenticateAsync();
-            if (authResult != null && authResult.Skipped)
+            AuthenticationTicket ticket = null;
+            Exception exception = null;
+
+            try
             {
-                return false;
+                var authResult = await HandleRemoteAuthenticateAsync();
+                if (authResult != null && authResult.Skipped == true)
```
PinpointTownes: nit: can't `authResult.Skipped == true` be simplified to `authResult.Skipped`?

troydai: yup. this is personal habit. I'll update.

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
troydai: Throw exception here is not ideal, but without breaking change, there is no other means to return the information to outer logic. Returns `null` can indicate an error, but the message composed here is more readable.

I think there is room for us to improve the error handling but it could be out of the scope of this PR. 

I'll update the comments, this is actually `Facebook Graph API`.

src/Microsoft.AspNetCore.Authentication.Google/GoogleHandler.cs
```diff
@@ -32,7 +32,10 @@ public GoogleHandler(HttpClient httpClient)
             request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", tokens.AccessToken);
 
             var response = await Backchannel.SendAsync(request, Context.RequestAborted);
-            response.EnsureSuccessStatusCode();
+            if (!response.IsSuccessStatusCode)
+            {
+                throw new HttpRequestException($"An error occurred when retrieving user information ({response.StatusCode}). Please check if the authentication information is correct and the corresponding Google API is enabled.");
```
PinpointTownes: IMHO, this message is not precise enough. AFAIK, it's actually the "Google+ API" that must be enabled.

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -31,30 +31,54 @@ public abstract class RemoteAuthenticationHandler<TOptions> : AuthenticationHand
...
+                    return false;
+                }
+
+                if (authResult == null)
+                {
+                    exception = new InvalidOperationException("Invalid return state, unable to redirect.");
+                }
+                else if (!authResult.Succeeded)
+                {
+                    exception = authResult?.Failure ??
```
Tratcher: `?.` is not needed here

Tratcher:
:shipit: after minor comments.

troydai:
Updated, wait for CI. I'll merge afterwards.

PR #906, opened 2016-07-19
Remove OfType allocations

ryanbrandenburg:
`System.Linq.OfType<>` allocates in order to create an enumerable. Here we've modified the function so as not to do that.

### DIFF
--- a/src/Microsoft.AspNetCore.Authorization/AuthorizationPolicy.cs
+++ b/src/Microsoft.AspNetCore.Authorization/AuthorizationPolicy.cs
@@ -122,28 +122,28 @@ namespace Microsoft.AspNetCore.Authorization
 
             var policyBuilder = new AuthorizationPolicyBuilder();
             var any = false;
-            foreach (var authorizeAttribute in authorizeData.OfType<AuthorizeAttribute>())
+            foreach (var authorizeDatum in authorizeData)
             {
                 any = true;
                 var useDefaultPolicy = true;
-                if (!string.IsNullOrWhiteSpace(authorizeAttribute.Policy))
+                if (!string.IsNullOrWhiteSpace(authorizeDatum.Policy))
                 {
-                    var policy = await policyProvider.GetPolicyAsync(authorizeAttribute.Policy);
+                    var policy = await policyProvider.GetPolicyAsync(authorizeDatum.Policy);
                     if (policy == null)
                     {
-                        throw new InvalidOperationException(Resources.FormatException_AuthorizationPolicyNotFound(authorizeAttribute.Policy));
+                        throw new InvalidOperationException(Resources.FormatException_AuthorizationPolicyNotFound(authorizeDatum.Policy));
                     }
                     policyBuilder.Combine(policy);
                     useDefaultPolicy = false;
                 }
-                var rolesSplit = authorizeAttribute.Roles?.Split(',');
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Authorization/AuthorizationPolicy.cs
```diff
@@ -122,8 +122,14 @@ public static AuthorizationPolicy Combine(IEnumerable<AuthorizationPolicy> polic
 
             var policyBuilder = new AuthorizationPolicyBuilder();
             var any = false;
-            foreach (var authorizeAttribute in authorizeData.OfType<AuthorizeAttribute>())
+            foreach (var authorizeDatum in authorizeData)
```
ilmax: May be worth adding a comment to explain this is not using Linq for perf reason in order to avoid regression?

HaoK:
+1 to the comment saying this is for perf, otherwise :shipit:

HaoK:
Actually, I'm not sure we need the OfType at all, the interface has all of the properties we need, can you try removing the if check/continue entirely?  I believe its not needed at all...

ryanbrandenburg:
I was under the impression that we were using it to filter out things which weren't of type `AuthorizeAttribute`. It appears that we don't have any implementations of `IAuthorizeData` other than `AuthorizeAttribute`. Would we want a hypothetical future implementation of `IAuthorizeData` to be processed here? If yes then we should make that change. If no then that change seems like a bug in waiting. Or am I misunderstanding something?

HaoK:
Looks like a vestige carried from times before there was the interface.  Its actually a bug that it only looks at AuthorizeAttributes today instead of all `IAuthorizeData`.  The fact that there's only one implementation is why this bug isn't very noticeable today until I thought it :)

HaoK:
This code was originally based on what the old mvc auth filter was doing, which uses an OfType to get all AuthorizeAttributes out.  i.e.: the caller is already filtering

https://github.com/aspnet/Mvc/blob/c7a46e4caf0d1f25dd5f8bcd385af50b34b0f5dd/src/Microsoft.AspNetCore.Mvc.Core/Internal/AuthorizationApplicationModelProvider.cs#L39

While the policy actually shouldn't care what the implementation is.  So this perf bug actually uncovered a subtle functional bug :)

ryanbrandenburg:
Neat, will make the change then. Thanks!

troydai:
:shipit: 

We may need to sweep every `OfType` in this repo to see if they're necessary.

HaoK:
Looks good :shipit:

NKnusperer:
Wasted a day because of this while using the NuGet version and a custom attribute implementing `IAuthorizeData`. I'am integrating `Microsoft.AspNetCore.Authorization` into a custom project which has nothing to do with Asp.Net so I guess no one else had this problem yet.
As a workaround I now wrap the `IAuthorizeData` into the build-in `AuthorizeAttribute` like this before calling `AuthorizationPolicy.CombineAsync`:
```
private static IEnumerable<IAuthorizeData> WrapInAuthorizeAttribute(IEnumerable<IAuthorizeData> authorizationData)
{
	foreach (var authorizeData in authorizationData)
		yield return new AuthorizeAttribute
		{
			Policy = authorizeData.Policy,
			Roles = authorizeData.Roles,
			ActiveAuthenticationSchemes = authorizeData.ActiveAuthenticationSchemes
		};
}
```

PR #921, opened 2016-07-26
Improve code readability for Authentication.Google

troydai:


src/Microsoft.AspNetCore.Authentication.Google/GoogleDefaults.cs
```diff
@@ -3,9 +3,12 @@
 
 namespace Microsoft.AspNetCore.Authentication.Google
 {
+    /// <summary>
+    /// Default values for Google authentication
+    /// </summary>
     public static class GoogleDefaults
     {
-        public const string AuthenticationScheme = "Google";
+        public static readonly string AuthenticationScheme = "Google";
```
Tratcher: These need to be const so they can be used in the Authorize attribute.

src/Microsoft.AspNetCore.Authentication.Google/GoogleHandler.cs
```diff
@@ -100,8 +104,11 @@ protected override string BuildChallengeUrl(AuthenticationProperties properties,
             return authorizationEndpoint;
         }
 
-        private static void AddQueryString(IDictionary<string, string> queryStrings, AuthenticationProperties properties,
-            string name, string defaultValue = null)
+        private static void AddQueryString(
```
Tratcher: overdoing it

Tratcher:
:shipit:

src/Microsoft.AspNetCore.Authentication.Google/GoogleHandler.cs
troydai: I break the lines which are longer than 120 characters. 

By the way, in other repos we try to avoid default parameter. Do we do it here?

HaoK: Its only for public methods I believe

troydai: got it.

PR #924, opened 2016-07-27
Modify warning about policy behavior

ryanbrandenburg:
Make the warning about the behavior dependency MVC has on DefaultAuthorizationPolicyProvider a bit more explicit.

CC @Eilon, @rynowak, @HaoK, and @blowdart for review.

### DIFF
--- a/src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationPolicyProvider.cs
+++ b/src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationPolicyProvider.cs
@@ -45,7 +45,9 @@ namespace Microsoft.AspNetCore.Authorization
         /// <returns>The named <see cref="AuthorizationPolicy"/>.</returns>
         public virtual Task<AuthorizationPolicy> GetPolicyAsync(string policyName)
         {
-            // MVC relies on DefaultAuthorizationPolicyProvider providing the same policy for the same requests.
+            // MVC caches policies specifically for this class, so this method MUST return the same policy per
+            // policyName for every request or it could allow undesired access. It also must return synchronously.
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationPolicyProvider.cs
```diff
@@ -45,7 +45,8 @@ public Task<AuthorizationPolicy> GetDefaultPolicyAsync()
         /// <returns>The named <see cref="AuthorizationPolicy"/>.</returns>
         public virtual Task<AuthorizationPolicy> GetPolicyAsync(string policyName)
         {
-            // MVC relies on DefaultAuthorizationPolicyProvider providing the same policy for the same requests.
+            // MVC relies on this class providing the same policy for the same requests. If it were to provide
+            // different policies per request then the cached policy in MVC could cause undesired access.
```
HaoK: Nit: Maybe something more like?

MVC caches policies specifically for this class, so this method must return the same policy for every request or it could allow undesired access.

ryanbrandenburg: Yeah, that's a bit clearer, will go with that unless someone has a better one.

rynowak: I like @HaoK 's idea. I'd also mention that MVC also relies on it returning synchronously, and also that any changes to either of these behaviors will require us to ship a patch of MVC as well.

ryanbrandenburg:
:up::date:

Eilon:
:shipit: (I'm in a room with @blowdart @HaoK @muratg @Tratcher @troydai and we all :shipit: this)

Issue #743, opened 2016-03-14
OIDC signout should return to CallbackPath then locally redirect to AuthProperties.RedirectUri


Tratcher:
Feedback from template reviews:

Default PostLogoutRedirectUri to CallbackPath. Listen on that path for signout callbacks and then apply the local redirect specified in auth properties. This prevents needing to register a separate redirect uri address with the provider.

If we do this do we still need PostLogoutRedirectUri? How do we reconcile it with AuthProperties.RedirectUri if both are set?

@vibronet @phenning @rustd

PinpointTownes:
Interesting approach, I like it. That said, it might be tricky to implement, since there's no logout-specific parameter: http://openid.net/specs/openid-connect-session-1_0.html#RPLogout.

A way to implement that would be using the `state` parameter to store the callback type.

Tratcher:
When looking into this I found that:
- AAD does not restrict the value of post_logout_redirect_uri like it should
- AAD does not flow state to the post_logout_redirect_uri like it should
- Google doesn't implement end_session_endpoint at all

Putting this in backlog until AAD gets this sorted out on their end.

brockallen:
> AAD does not flow state to the post_logout_redirect_uri like it should

I was just testing with AAD today (via OIDC) and it was passing back the state param on post signout redirects.

Tratcher:
@brockallen Hmm, I just tested it today as well.

Here's what my initial signout redirect looks like:
```
Location: https://login.microsoftonline.com/cdc690f9-b6b8-4023-813a-bae7143d1f87/oauth2/logout?post_logout_redirect_uri=http%3A%2F%2Flocalhost%3A42023%2Fsignin-oidc&state=CfDJ8AaEPKTYgPVMjEyowKKgeJKyKJt3ENUSWXDuMtRWlpHb716h5OpgKR22nGFad0K5mdPpAJHZVCOI6sxfx0GYP3TPBgWXseU2C0ffXVFyMusoLj9UJ69uhyJ9ypTT7M15WuBZ_6NEPIKNIUTO-mYQ_zyA6wy27iGN0cQMwuH-PIEB
```
But all I get back is:
```
Location: http://localhost:42023/signin-oidc
```

How about you?

brockallen:
```
[...]

PR #925, opened 2016-07-28
Update OIDC signout flow

troydai:
Address issue #743, and react to recent AAD OIDC sign out flow change.

### DIFF
--- a/samples/OpenIdConnect.AzureAdSample/Startup.cs
+++ b/samples/OpenIdConnect.AzureAdSample/Startup.cs
@@ -20,8 +20,6 @@ namespace OpenIdConnect.AzureAdSample
 {
     public class Startup
     {
-        private const string GraphResourceID = "https://graph.windows.net";
-
         public Startup(IHostingEnvironment env)
         {
             var builder = new ConfigurationBuilder()
@@ -103,7 +101,20 @@ namespace OpenIdConnect.AzureAdSample
 
             app.Run(async context =>
             {
-                if (context.Request.Path.Equals("/signout"))
+                if (context.Request.Path.Equals("/signin"))
+                {
+                    if (context.User.Identities.Any(identity => identity.IsAuthenticated))
+                    {
+                        // User has already signed in
+                        context.Response.Redirect("/");
+                        return;
+                    }
+
+                    await context.Authentication.ChallengeAsync(
+                        OpenIdConnectDefaults.AuthenticationScheme,
+                        new AuthenticationProperties { RedirectUri = "/" });
+                }
+                else if (context.Request.Path.Equals("/signout"))
                 {
                     await context.Authentication.SignOutAsync(CookieAuthenticationDefaults.AuthenticationScheme);
                     await WriteHtmlAsync(context.Response,
--- a/samples/OpenIdConnect.AzureAdSample/Startup.cs
+++ b/samples/OpenIdConnect.AzureAdSample/Startup.cs
@@ -81,6 +81,7 @@ namespace OpenIdConnect.AzureAdSample
                 ClientSecret = clientSecret, // for code flow
                 Authority = authority,
                 ResponseType = OpenIdConnectResponseType.CodeIdToken,
+                PostLogoutRedirectUri = "/usersignout",
                 // GetClaimsFromUserInfoEndpoint = true,
                 Events = new OpenIdConnectEvents()
                 {
--- a/samples/OpenIdConnect.AzureAdSample/Startup.cs
+++ b/samples/OpenIdConnect.AzureAdSample/Startup.cs
@@ -81,7 +81,7 @@ namespace OpenIdConnect.AzureAdSample
                 ClientSecret = clientSecret, // for code flow
                 Authority = authority,
                 ResponseType = OpenIdConnectResponseType.CodeIdToken,
-                PostLogoutRedirectUri = "/usersignout",
+                PostLogoutRedirectUri = "/signed-out",
                 // GetClaimsFromUserInfoEndpoint = true,
                 Events = new OpenIdConnectEvents()
                 {
@@ -128,10 +128,16 @@ namespace OpenIdConnect.AzureAdSample
                 else if (context.Request.Path.Equals("/signout-remote"))
                 {
                     await context.Authentication.SignOutAsync(CookieAuthenticationDefaults.AuthenticationScheme);
-                    await context.Authentication.SignOutAsync(OpenIdConnectDefaults.AuthenticationScheme, new AuthenticationProperties
-                    {
-                        RedirectUri = "/remote-signedout"
-                    });
+                    await context.Authentication.SignOutAsync(OpenIdConnectDefaults.AuthenticationScheme);
+                }
+                else if (context.Request.Path.Equals("/signed-out"))
+                {
+                    await WriteHtmlAsync(context.Response, 
+                        async response =>
+                        {
+                            await response.WriteAsync($"<h1>You have been signed out.</h1>");
+                            await response.WriteAsync("<a class=\"btn btn-primary\" href=\"/signin\">Sign In</a>");
+                        });
                 }
                 else if (context.Request.Path.Equals("/remote-signedout"))
                 {
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
@@ -18,6 +18,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Authentication;
 using Microsoft.AspNetCore.Http.Features.Authentication;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Primitives;
 using Microsoft.IdentityModel.Protocols.OpenIdConnect;
 using Microsoft.IdentityModel.Tokens;
 using Microsoft.Net.Http.Headers;
@@ -68,7 +69,14 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
             {
                 return await HandleRemoteSignOutAsync();
             }
-            return await base.HandleRequestAsync();
+            else if (Options.SignedOutCallbackPath.HasValue && Options.SignedOutCallbackPath == Request.Path)
+            {
+                return await HandleSignOutCallbackAsync();
+            }
+            else
+            {
+                return await base.HandleRequestAsync();
+            }
         }
 
         protected virtual async Task<bool> HandleRemoteSignOutAsync()
@@ -79,12 +87,14 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
             {
                 message = new OpenIdConnectMessage(Request.Query.Select(pair => new KeyValuePair<string, string[]>(pair.Key, pair.Value)));
             }
+
             // assumption: if the ContentType is "application/x-www-form-urlencoded" it should be safe to read as it is small.
             else if (string.Equals(Request.Method, "POST", StringComparison.OrdinalIgnoreCase)
               && !string.IsNullOrEmpty(Request.ContentType)
               // May have media/type; charset=utf-8, allow partial match.
               && Request.ContentType.StartsWith("application/x-www-form-urlencoded", StringComparison.OrdinalIgnoreCase)
-              && Request.Body.CanRead) {
+              && Request.Body.CanRead)
+            {
                 var form = await Request.ReadFormAsync();
                 message = new OpenIdConnectMessage(form.Select(pair => new KeyValuePair<string, string[]>(pair.Key, pair.Value)));
             }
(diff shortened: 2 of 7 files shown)


samples/OpenIdConnect.AzureAdSample/Startup.cs
```diff
@@ -100,48 +100,81 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
...
-                    return;
+                    await context.Authentication.SignOutAsync(CookieAuthenticationDefaults.AuthenticationScheme);
+                    await context.Authentication.SignOutAsync(OpenIdConnectDefaults.AuthenticationScheme);
                 }
-
-                context.Response.ContentType = "text/html";
-                await context.Response.WriteAsync($"<html><body>Hello Authenticated User {context.User.Identity.Name}<br>{Environment.NewLine}");
-                await context.Response.WriteAsync("Claims:<br>" + Environment.NewLine);
-                foreach (var claim in context.User.Claims)
+                else if (context.Request.Path.Equals("/usersignout"))
```
Tratcher: `/signed-out`

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -139,108 +149,134 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
-        /// Handles Signout
+        /// Sign out the Relying Party from the OpenID provider 
         /// </summary>
-        /// <returns></returns>
-        protected override async Task HandleSignOutAsync(SignOutContext signout)
+        /// <returns>A task executing the sign out procedure</returns>
+        protected override async Task HandleSignOutAsync(SignOutContext context)
         {
-            if (signout != null)
+            if (context == null)
```
Tratcher: throw argumennullexception or invalidoperationexception.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -139,108 +149,134 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
+                _configuration = await Options.ConfigurationManager.GetConfigurationAsync(Context.RequestAborted);
+            }
 
-                var redirectContext = new RedirectContext(Context, Options, properties)
-                {
-                    ProtocolMessage = message
-                };
+            var message = new OpenIdConnectMessage()
+            {
+                IssuerAddress = _configuration == null ? string.Empty : (_configuration.EndSessionEndpoint ?? string.Empty),
```
Tratcher: _configuration?.EndSessionEndpoint ?? string.Empty,

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -68,7 +69,14 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
             {
                 return await HandleRemoteSignOutAsync();
             }
-            return await base.HandleRequestAsync();
+            else if (Options.SignedOutCallbackPath.HasValue && Options.SignedOutCallbackPath == Request.Path)
+            {
+                return await HandleSignOutCallbackAsync();
+            }
+            else
```
Tratcher: this else is redundant.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -139,108 +149,134 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
 
-                var redirectContext = new RedirectContext(Context, Options, properties)
-                {
-                    ProtocolMessage = message
-                };
+            var message = new OpenIdConnectMessage()
+            {
+                IssuerAddress = _configuration == null ? string.Empty : (_configuration.EndSessionEndpoint ?? string.Empty),
+
+                // Redirect back of SigneOutCallbackPath first before user agent is redirected to actual post logout redirect uri
```
Tratcher: back to

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -139,108 +149,134 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
+        /// Response to the callback from OpenId provider after session ended.
+        /// </summary>
+        /// <returns>A task executing the callback procedure</returns>
+        protected virtual Task<bool> HandleSignOutCallbackAsync()
+        {
+            StringValues protectedState;
+            if (Request.Query.TryGetValue("State", out protectedState))
+            {
+                var properties = Options.StateDataFormat.Unprotect(protectedState);
+                if (!string.IsNullOrEmpty(properties.RedirectUri))
```
Tratcher: Can unprotect return null?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -139,108 +149,134 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
+                {
+                    Response.Redirect(properties.RedirectUri);
+                    return Task.FromResult(true);
                 }
             }
+
+            Response.Redirect("/");
```
Tratcher: should we prefer Options.PostLogoutRedirectUri if set?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -139,108 +149,134 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
+        }
 
-                    await Response.Body.WriteAsync(buffer, 0, buffer.Length);
+        /// <summary>
+        /// Response to the callback from OpenId provider after session ended.
+        /// </summary>
+        /// <returns>A task executing the callback procedure</returns>
+        protected virtual Task<bool> HandleSignOutCallbackAsync()
+        {
+            StringValues protectedState;
```
Tratcher: Do we need a new event here? RedirectToPostLogoutRedirectUri?

Tratcher:
@brentschmaltz @vibronet

Tratcher:
@muratg @Eilon @vibronet
One open question: Is this a breaking behavior change? And if so, are we OK with that?
- It adds an extra redirect to the remote sign-out flow.
- It adds a new endpoint that needs to be registered with AADv2. What if customers move to AADv2 before taking this change?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
troydai: I don't know. Since it is a black box to me  so I prepare for the worst.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
muratg: Can we check the callers of the method and ensure this is never null? If so we can probably remove the null check.

troydai: The caller is `AuthenticationHandler` where none of the methods is guarded. If we chance one please, we shall change all the places. May be we need to sweep these things in a separate PR?

Ref: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs#L272

Tratcher: Except that would have null reffed right here:
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs#L269

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -139,108 +147,133 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
         }
 
         /// <summary>
-        /// Handles Signout
+        /// Sign out the Relying Party from the OpenID provider
```
PinpointTownes: nit: this comment feels a bit misleading, as this method redirects the user agent TO the identity provider. It's not involved when the signout is triggered by the identity provider itself.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -139,108 +147,133 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
 
-                    await Response.Body.WriteAsync(buffer, 0, buffer.Length);
+        /// <summary>
+        /// Response to the callback from OpenId provider after session ended.
+        /// </summary>
+        /// <returns>A task executing the callback procedure</returns>
+        protected virtual Task<bool> HandleSignOutCallbackAsync()
+        {
+            StringValues protectedState;
+            if (Request.Query.TryGetValue("State", out protectedState))
```
PinpointTownes: The "official" casing is `state` (and OAuth2/OIDC parameters are case sensitive). Does TryGetValue make a case insensitive search?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -139,108 +147,133 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
+            {
+                var properties = Options.StateDataFormat.Unprotect(protectedState);
+                if (!string.IsNullOrEmpty(properties?.RedirectUri))
```
PinpointTownes: nit: we usually use the state data format to flow user properties AND to provide anti-XSRF. Is worth adding a tiny comment explaining why correlation is not enforced here?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: There's a constant for this somewhere...

PinpointTownes: `OpenIdConnectParameterNames.State`

Alternatively, you can instantiate an `OpenIdConnectMessage` and retrieve the state via `OpenIdConnectMessage.State`.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: No point for a single property.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
troydai: will update.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
PinpointTownes: I suppose this remark was for `OpenIdConnectMessage`, not my suggestion about adding a warning about correlation?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
troydai: However in all the other methods in that class none of them check the `null`.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
troydai: will update.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
PinpointTownes: :+1: for removing the null checks in the derived classes.

PinpointTownes:
FWIW, I tested this PR with OpenIddict and it works fine :clap: 

A few remarks:
  - This is indeed a breaking change, which is quite subtile as existing code will still compile. At this point, it might be better to completely remove `PostLogoutRedirectUri` as it now has no added value and is quite confusing (with this PR, it's more like an equivalent to the return URL, but for logout). Alternatively, it should be marked as obsolete and removed in a future version.
  - A while ago, [@Tratcher suggested merging this new endpoint with `CallbackPath`](https://github.com/aspnet/Security/issues/743#issue-140815032). I think it was a good idea (at least, it would help reduce the number of endpoints we now have :smile:)
  - I think it would be nice to have a new event in `HandleSignOutCallbackAsync`.

Tratcher:
If we merge it with CallbackPath that would help us avoid one of the breaks. However, is the incoming message well marked so we could easily distinguish it from a sign-in? I don't want to create ambiguity or make the sign-in code any more complicated.

PinpointTownes:
> However, is the incoming message well marked so we could easily distinguish it from a sign-in?

Not directly as there's no mandatory "logout-specific" parameter we could use to determine whether a message is an authorization response or a logout callback.

Luckily, the identity provider MUST flow the `state` when it's sent by the relying party. In our case, we could store the message type in the state before protecting it and sending it to the identity provider. We could triage requests based on this ticket property.

The only annoying thing is how `RemoteAuthenticationHandler`/`OpenIdConnectHandler` is designed. The relevant code is in `HandleRemoteAuthenticateAsync`, which doesn't seem appropriate for handling a logout response.

Tratcher:
Yeah, a state parameter would work. You'd have to intervene about here: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L432 before (or after?) MessageReceived

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -139,108 +147,136 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
...
+        {
+            StringValues protectedState;
+            if (Request.Query.TryGetValue(OpenIdConnectParameterNames.State, out protectedState))
+            {
+                var properties = Options.StateDataFormat.Unprotect(protectedState);
 
-                    await Response.Body.WriteAsync(buffer, 0, buffer.Length);
+                var redirectContext = new RedirectContext(Context, Options, properties);
+                await Options.Events.RedirectToPostLogoutRedirectUri(redirectContext);
+                if (redirectContext.HandledResponse || redirectContext.Skipped)
```
PinpointTownes: Note: you can't return `true` for the `Skipped` case (because the whole point of `SkipToNextMiddleware()` is to bypass the built-in logic and to allow the rest of the pipeline to handle the response: returning `true` will prevent that)

troydai:
Yes, we can unify the endpoints to one. It will take a few changes:
1. Flow one additional property in state to indicate the type of the action.
2. In `OpenIdConnectHandler.HandleSignOutAsync` check the property in `status` and create `AuthenticateResult` accordingly. A new property will be added to the `AuthenticateProperties`.
3. In `RemoteAuthenticateHandler.HandleRemoveCallbackAsync` a fork will be added to handle signout flow.

I share the same concern with @PinpointTownes. It seems to me that `RemoteAuthenticateHandler` will make additional assumption about what will present in the `AuthenticateProperties` to alter the workflow. It creates additional complexity in the base class. Is it this trade off worthwhile?

Is there a case in which an identity provider only allows one callback path to just justify the needs of doing this?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
troydai: Got it. Will fix.

Tratcher:
We never got a response from AAD about when their v2 was going public.

Tratcher:
Do not make changes to RemoteAuthenticateHandler that only apply to OIDC.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs
```diff
@@ -36,6 +36,11 @@ public interface IOpenIdConnectEvents : IRemoteAuthenticationEvents
         Task RedirectToIdentityProviderForSignOut(RedirectContext context);
 
         /// <summary>
+        /// Invoked before redirecting to the post logout redirect uri.
+        /// </summary>
+        Task RedirectToPostLogoutRedirectUri(RedirectContext context);
```
Tratcher: hmm, this is a breaking change to add to the interface. Is it worth it? @muratg @Eilon?

Eilon: It is a breaking change, cannot do it in 1.1. Can either do something clever, or can add a new interface.

Tratcher: I'd rather postpone adding the event. File a 2.0 bug to add it.

Eilon: If the bug we file is to change the interface, please mark it as `breaking-change`.

PinpointTownes: If interfaces cause too much versioning pain when adding new events, then removing them to exclusively use the `OpenIdEvents` class might be an interesting option for 2.0 (as long as we keep the delegate properties + virtual methods, of course :smile:).

troydai: I'll take out the change from this PR.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
troydai: Proposed: https://github.com/aspnet/Security/issues/942

troydai:
Rebased on top of the dev branch. Waiting for CI and will merge it passes.

troydai:
I'm also working on a follow up change to add a event to `HandleSignOutCallbackAsync`

Tratcher:
You can't add events, right? https://github.com/aspnet/Security/issues/942

troydai:
Ah, yeah. I forgot that one. All right, just let the branch sleep for a while (https://github.com/aspnet/Security/pull/954 to be closed soon)

PinpointTownes:
@troydai if you want to include this event before 2.0.0, one option might be to only add it to `OpenIdConnectEvents` (but not to `IOpenIdConnectEvents`) and use a `as cast` to trigger it when the events instance inherits from `OpenIdConnectEvents`.

troydai:
The major concerns are: 1. it is not a clean design and partially breach the interface. 2. discoverability will be poor. users need insight into internal mechanism. I think we better stay put instead of introduce half-baked solution to a relative minor user case. 

If there are more evidence that the extension point is that important we can revisit this decision.

Thanks for advice.

Issue #916, opened 2016-07-22
AuthenticationTokenExtensions should have an UpdateToken


brockallen:
Some API where we can update just one token (e.g. a refresh token or an access token) and the rest of what's in there is preserved. 

This is what a client app must do now (which is requires too much internal knowledge): https://github.com/IdentityServer/IdentityServer4.Samples/blob/dev/Clients/src/MvcHybrid/Controllers/HomeController.cs#L46

HaoK:
Something like this is what you are asking for? You probably are still left with doing your own copy into a subsequent properties to send down new tokens.  

```
 public static Task UpdateTokenValueAsync(this AuthenticationProperties properties, string name, string newValue) {
   var tokens = await HttpContext.Authentication.GetTokensAsync().Where(t => t.Name == name);
   foreach (var tok in tokens) tok.Value = newValue;
 }
```

brockallen:
Well, given that there's some logic to the access token expiration, I'd almost prefer slightly higher order APIs -- one for access token with optional expiration param, one for RT. All of course could be built on the above more general API.

brockallen:
Thanks

-Brock

PR #931, opened 2016-07-29
Add new UpdateTokenValue API

HaoK:
Fixes https://github.com/aspnet/Security/issues/916

@Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
+++ b/src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
@@ -63,6 +63,26 @@ namespace Microsoft.AspNetCore.Authentication
                 : null;
         }
 
+        public static bool UpdateTokenValue(this AuthenticationProperties properties, string tokenName, string tokenValue)
+        {
+            if (properties == null)
+            {
+                throw new ArgumentNullException(nameof(properties));
+            }
+            if (tokenName == null)
+            {
+                throw new ArgumentNullException(nameof(tokenName));
+            }
+
+            var tokenKey = TokenKeyPrefix + tokenName;
+            if (!properties.Items.ContainsKey(tokenKey))
+            {
+                return false;
+            }
+            properties.Items[tokenKey] = tokenValue;
+            return true;
+        }
+
         public static IEnumerable<AuthenticationToken> GetTokens(this AuthenticationProperties properties)
         {
             if (properties == null)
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Authentication/TokenExtensions.cs
```diff
@@ -63,6 +63,26 @@ public static string GetTokenValue(this AuthenticationProperties properties, str
                 : null;
         }
 
+        public static bool UpdateTokenValue(this AuthenticationProperties properties, string tokenName, string tokenValue)
```
Tratcher: Should there be a way to add a new token as well? and/or set a token regardless of if it already existed?

HaoK: We don't really want to encourage that given the way things are setup, since we basically have to nuke the digest key.  Ideally we want them to figure out all their tokens ahead of time and call Store once

HaoK: i.e. we don't want to encourage, AddToken called in a loop

Tratcher:
Let's see, this reduces the original sample code
```
var tokens = new List<AuthenticationToken>();
tokens.Add(new AuthenticationToken { Name = OpenIdConnectParameterNames.IdToken, Value = old_id_token });
tokens.Add(new AuthenticationToken { Name = OpenIdConnectParameterNames.AccessToken, Value = new_access_token });
tokens.Add(new AuthenticationToken { Name = OpenIdConnectParameterNames.RefreshToken, Value = new_refresh_token });

var expiresAt = DateTime.UtcNow + TimeSpan.FromSeconds(tokenResult.ExpiresIn);
tokens.Add(new AuthenticationToken { Name = "expires_at", Value = expiresAt.ToString("o", CultureInfo.InvariantCulture) });

var info = await HttpContext.Authentication.GetAuthenticateInfoAsync("cookies");
info.Properties.StoreTokens(tokens);
await HttpContext.Authentication.SignInAsync("cookies", info.Principal, info.Properties);
```
to:
```
var info = await HttpContext.Authentication.GetAuthenticateInfoAsync("cookies");

info.Properties.UpdateTokenValue(OpenIdConnectParameterNames.AccessToken, new_access_token);
info.Properties.UpdateTokenValue(OpenIdConnectParameterNames.RefreshToken, new_refresh_token);
var expiresAt = DateTime.UtcNow + TimeSpan.FromSeconds(tokenResult.ExpiresIn);
info.Properties.UpdateTokenValue("expires_at", expiresAt.ToString("o", CultureInfo.InvariantCulture));

await HttpContext.Authentication.SignInAsync("cookies", info.Principal, info.Properties);
```
Nice. :shipit:

Issue #879, opened 2016-06-12
AuthorizationHandlerContext responsibility split up & thread safety


8snit:
short background - we are currently bringing our ABAC approach forward, staying on the XACML path conceptually (architecture/flow) but replacing the language element part with "code" to make it more developer friendly (the "leave rules to business experts" thing didn't work out for us so far...)

first of all, thx for the great job with the authorization abstractions, without too much tweaking we were able to get "some kind of" PDP executing c# code (resp. ScriptCs scripts) running, in addition to the out of the box framework functionality we support

A. handler processing in parallel 
B. direct routing of requirements to handlers (instead of iterating through all registered handlers)
C. triggering webhooks after approval (~ simplified obligation stuff)
D. pluggable evaluation strategies (~ combining algorithm) e.g. with fail fast (not going through all handlers every time)
E. auditing information for all succeeded/failed requirements (in order!)

what is preventing us from going straight ahead with the current (and not a patched) AspNetCore.Authorization implementation is the "not very customizable nature" of the AuthorizationHandlerContext with its two responsibilites - holding current state and evaluating it:

- one thing we propose is to introduce an interface like
```
   public interface IAuthorizationEvaluator
   {
      bool HasFailed(AuthorizationHandlerContext context);
   
      bool HasSucceeded(AuthorizationHandlerContext context);
   }
```
which is injected and used within the IAuthorizationService (instead of the HasFailed/HasSucceeded properties of the AuthorizationHandlerContext) to make the decisions (see D)

- we also vote for an additive (see E) & thread safe (see A) implementation of the Succeed/Fail calls within the AuthorizationHandlerContext (instead of _pendingRequirements.Remove), 
```
    public class AuthorizationHandlerContext
    {
        ...

        public ConcurrentQueue<IAuthorizationRequirement> FailedRequirements { get; }

        public ConcurrentQueue<IAuthorizationRequirement> SucceededRequirements { get; }

        public void Fail(IAuthorizationRequirement requirement = null)
        {
            FailedRequirements.Enqueue(requirement);
        }

        public void Succeed(IAuthorizationRequirement requirement)
        {
            if (requirement == null)
            {
                throw new ArgumentNullException(nameof(requirement));
            }

            SucceededRequirements.Enqueue(requirement);
        }
    }
```
an interface/base class for AuthorizationHandlerContext combined with a factory approach could be an alternative

looking forward to feedback and hopefully a chance to contribute :)

HaoK:
[...]

PR #934, opened 2016-08-02
Add IAuthorizationEvaluator/IAuthorizationHandlerContextFactory

HaoK:
@blowdart @divega 

Add a powerful new extensibility point, which enables plugging in different semantics for whether Authorize returns true or not.  

Not sure if we have to preserve old ctor parameters as well or not, just to be safe I added an additional ctor for back compat (but I'd prefer just to modify the original if possible)

### DIFF
--- a/src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationService.cs
+++ b/src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationService.cs
@@ -16,6 +16,7 @@ namespace Microsoft.AspNetCore.Authorization
     /// </summary>
     public class DefaultAuthorizationService : IAuthorizationService
     {
+        private readonly IAuthorizationHandlerContextFactory _contextFactory;
         private readonly IAuthorizationEvaluator _evaluator;
         private readonly IAuthorizationPolicyProvider _policyProvider;
         private readonly IList<IAuthorizationHandler> _handlers;
@@ -27,7 +28,7 @@ namespace Microsoft.AspNetCore.Authorization
         /// <param name="policyProvider">The <see cref="IAuthorizationPolicyProvider"/> used to provide policies.</param>
         /// <param name="handlers">The handlers used to fulfill <see cref="IAuthorizationRequirement"/>s.</param>
         /// <param name="logger">The logger used to log messages, warnings and errors.</param>  
-        public DefaultAuthorizationService(IAuthorizationPolicyProvider policyProvider, IEnumerable<IAuthorizationHandler> handlers, ILogger<DefaultAuthorizationService> logger) : this(policyProvider, handlers, logger, new DefaultAuthorizationEvaluator()) { }
+        public DefaultAuthorizationService(IAuthorizationPolicyProvider policyProvider, IEnumerable<IAuthorizationHandler> handlers, ILogger<DefaultAuthorizationService> logger) : this(policyProvider, handlers, logger, new DefaultAuthorizationHandlerContextFactory(), new DefaultAuthorizationEvaluator()) { }
 
         /// <summary>
         /// Creates a new instance of <see cref="DefaultAuthorizationService"/>.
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.Authorization/IAuthorizationEvaluator.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// Determines whether an authorization request was successful or not.
+    /// </summary>
+    public interface IAuthorizationEvaluator
+    {
+        /// <summary>
+        /// Returns true, if authorization has failed.
+        /// </summary>
+        /// <param name="context">The authorization information.</param>
+        /// <returns>True if authorization has failed.</returns>
+        bool HasFailed(AuthorizationHandlerContext context);
```
smbecker: Not sure that I see the value of having HasSucceeded and HasFailure especially given that I only see usage of HasSucceeded. HasFailure should always be ! HasSucceeded. So simplify the contract and remove the method. Or if the method must stay then make it an extension method to force the implementation.

HaoK: HasFailed is there for short circuiting. Both can be false since succeeded is only true if all requirements are met by default. The default implementation does not short circuit executing all handlers but that was an explicit ask to be enabled.

HaoK:
Updated with a factory for replacing the default context as well

HaoK:
Fixes https://github.com/aspnet/Security/issues/879

HaoK:
@divega says :shipit:

8snit:
thanks for following this up, in addition to the existing default behavior, it is now easy to plugin combining algorithms like 'deny-unless-permit' and 'permit-unless-deny'

the **IAuthorizationHandlerContextFactory** should help regarding thread safety, however, I'm still not sure 'where to track processed requirements', not every handler evaluates every requirement to 'permit' or 'deny' - XACML even introduces two more states: 'not applicable' and 'indeterminate' (but I think just 'indeterminate' is fine)

while it is definitely possible to track the 'processed' requirements from outside (in the concrete **IAuthorizationService** implementation), it feels more natural to me to do it within the **AuthorizationHandlerContext**, that's why I suggested something like

```
    public class AuthorizationHandlerContext
    {
       public virtual void Fail(IAuthorizationRequirement requirement = null)
       {
           ...
```

what do you think?

HaoK:
So today we basically only track whether requirements are met, they would be considered indeterminate otherwise.  Fail is meant to kill the entire request irrespective of any requirements state, so I don't really think that jives with having an overload that takes a requirement.  Perhaps the naming of these methods should not be paired (Succeed/FailRequest) might have been a better naming...

I think introducing more detailed states for your requirements is something fairly specific to your IAuthorizationService implementations, and isn't something we are going to bake into the context at this point.  We can certainly considert his for 2.0 when we can make more breaking changes.

What would you expect to be able to do differently based on the various unmet requirement states?  (Which today are all lumped together as failures)

src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationService.cs
```diff
@@ -26,7 +28,17 @@ public class DefaultAuthorizationService : IAuthorizationService
         /// <param name="policyProvider">The <see cref="IAuthorizationPolicyProvider"/> used to provide policies.</param>
         /// <param name="handlers">The handlers used to fulfill <see cref="IAuthorizationRequirement"/>s.</param>
         /// <param name="logger">The logger used to log messages, warnings and errors.</param>  
-        public DefaultAuthorizationService(IAuthorizationPolicyProvider policyProvider, IEnumerable<IAuthorizationHandler> handlers, ILogger<DefaultAuthorizationService> logger)
+        public DefaultAuthorizationService(IAuthorizationPolicyProvider policyProvider, IEnumerable<IAuthorizationHandler> handlers, ILogger<DefaultAuthorizationService> logger) : this(policyProvider, handlers, logger, new DefaultAuthorizationHandlerContextFactory(), new DefaultAuthorizationEvaluator()) { }
```
HaoK: @Eilon is it considered a breaking change for us to consume new services in ctors when we expect this service to come from DI?  It would be nice if we didn't have to add a ctor overload (which we would likely remove in 2.0)

Eilon: Yep it's a breaking change. This is the one we talked about the other day, right?

HaoK: > Yep it's a breaking change. This is the one we talked about the other day, right?
> 
> 
> 

Yeah I will check in the pr as is with the extra constructors.

HaoK:
6f15d616a8954420bcb277ed938de1b14f6dbd6f

Issue #927, opened 2016-07-28
Code correctly returning a 401 unauthorised error in RC2 now incorrectly returns a 500 error


dumbledad:
We have an ASP.Net Core web service using Azure AD B2C for bearer token authentication. Our iOS client makes an API request and if the ASP.Net Core server returns a 401 Unauthorised error with the term "SecurityTokenExpiredException" in the body then the client refreshes its token and tries again.

This works great under ASP.Net Core RC2 but now that we have moved to the release version we cannot deploy our server without breaking our clients because the server no longer correctly returns a 401 Unauthorised error and instead returns a 500 error.

The code that sets up the 401 is in Startup.cs

```csharp
public void Configure(IApplicationBuilder app, IHostingEnvironment env)
{
    app.Use(async (context, next) =>
    {
        try
        {
            await next();
        }
        catch (Exception ex)
        {
            if (context.Response.HasStarted)
            {
                throw;
            }

            int statusCode = 500;

            if (ex is SecurityTokenInvalidSignatureException ||
                ex is SecurityTokenInvalidLifetimeException ||
                ex is SecurityTokenExpiredException)
            {
                statusCode = 401;
            }
            else if (ex is ResourceNotAuthorizedException)
            {
                statusCode = 403;
            }
            else if (ex is ResourceNotFoundException || ex is UserDoesNotExistException)
            {
                statusCode = 404;
            }

            context.Response.StatusCode = statusCode;
            await context.Response.WriteAsync(new ExceptionResponse(statusCode, ex).ToString());
        }
    });
   //[...]
```

(N.B. If I step through this code in the debugger each line executes, the local variable `statusCode` is correctly set to 401, and those final two lines are run, but the error sent back by ASP.Net Core to the client is 500 and not 401.)

Has something changed going from RC2 to the ASP.Net Core release that would render this code incorrect? If so what is the new way of writing this?

(I have cross-posted this as a Stack Overflow question too [here](http://stackoverflow.com/q/38637454/575530). I hope that's OK.)

dumbledad:
I have created a Gist showing the stark difference between the text written to `context.Response.WriteAsync` and the response text received on the client. The Gist is [here](https://gist.github.com/dumbledad/9d3f70b5621eb004e5f8758c4dcdc3f4)

Eilon:
@troydai can you debug into this and see what's going on here? The SO question might have some more hints.

troydai:
maybe related: https://github.com/aspnet/Security/issues/699

[...]

PR #935, opened 2016-08-02
JwtBearer Token: Catch exception during unauthorized flow

troydai:
Issue: #927

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -164,7 +164,6 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
                 return;
             }
 
-            // REVIEW: Should this check if there was an error, and then if that error was already handled??
             var ticket = (await HandleAuthenticateOnceSafeAsync())?.Ticket;
             if (ticket != null)
             {
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -323,7 +323,6 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
                     await Options.Events.RedirectToReturnUrl(redirectContext);
                 }
             }
-
         }
 
         private static bool IsHostRelative(string path)
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -165,7 +165,7 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
             }
 
             // REVIEW: Should this check if there was an error, and then if that error was already handled??
-            var ticket = (await HandleAuthenticateOnceAsync())?.Ticket;
+            var ticket = (await HandleAuthenticateOnceSafeAsync())?.Ticket;
             if (ticket != null)
             {
                 if (_refreshIssuedUtc.HasValue)
--- a/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
@@ -185,11 +185,11 @@ namespace Microsoft.AspNetCore.Authentication.JwtBearer
 
         protected override async Task<bool> HandleUnauthorizedAsync(ChallengeContext context)
         {
-            var authResult = await HandleAuthenticateOnceAsync();
+            var authResult = await HandleAuthenticateOnceSafeAsync();
 
             var eventContext = new JwtBearerChallengeContext(Context, Options, new AuthenticationProperties(context.Properties))
             {
-                AuthenticateFailure = authResult?.Failure,
+                AuthenticateFailure = authResult?.Failure
             };
 
             // Avoid returning error=invalid_token if the error is not caused by an authentication failure (e.g missing token).
--- a/src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
@@ -234,9 +234,8 @@ namespace Microsoft.AspNetCore.Authentication
         }
 
         /// <summary>
-        /// Handle the authentication for once.
-        ///
-        /// If the authentication has been done before returns the last authentication result.
+        /// Used to ensure HandleAuthenticateAsync is only invoked once. The subsequent calls
+        /// will return the same authenticate result.
         /// </summary>
         protected Task<AuthenticateResult> HandleAuthenticateOnceAsync()
         {
@@ -249,11 +248,9 @@ namespace Microsoft.AspNetCore.Authentication
         }
 
         /// <summary>
-        /// Handle the authentication for once.
-        ///
-        /// If the authentication has been done before returns the last authentication result.
-        /// This method won't throw exception. Any exception thrown during the authentication will be convert
-        /// to a AuthenticateResult.
+        /// Used to ensure HandleAuthenticateAsync is only invoked once safely. The subsequent
+        /// calls will return the same authentication result. Any exceptions will be converted
+        /// into a failed authenticatoin result containing the exception.
         /// </summary>
         protected async Task<AuthenticateResult> HandleAuthenticateOnceSafeAsync()
         {
--- a/src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
@@ -233,13 +233,41 @@ namespace Microsoft.AspNetCore.Authentication
             }
         }
 
+        /// <summary>
+        /// Handle the authentication for once.
+        ///
+        /// If the authentication has been done before returns the last authentication result.
+        /// </summary>
         protected Task<AuthenticateResult> HandleAuthenticateOnceAsync()
         {
             if (_authenticateTask == null)
             {
                 _authenticateTask = HandleAuthenticateAsync();
             }
+
             return _authenticateTask;
+        }
+
+        /// <summary>
+        /// Handle the authentication for once.
+        ///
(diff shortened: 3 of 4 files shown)


src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -185,11 +185,20 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
 
         protected override async Task<bool> HandleUnauthorizedAsync(ChallengeContext context)
         {
-            var authResult = await HandleAuthenticateOnceAsync();
+            Exception failure = null;
```
HaoK: Do we really ever want HandleAuthenticateOnceAsync to throw?  Should we just make this failsafe and push this try/catch into HandleAuthenticateOnce, to always return at least a non null authResult with a Failure or Success.

Its pretty gross if this is what needs to be done everywhere we call HandleAuthenticateOnceAsync

troydai: Ok, I'll move it up.

troydai: There's once thing I'd to bring into this discussion.

`HandleAuthenticateAsync` should be treated same as `HandleAuthenticateOnceAsync`, because later is just former with a restriction of not to be repeatedly invoked. 

So if we set up the rule that `HandleAuthenticateOnceAsync` won't throw exception the rule should apply to `HandleAuthenticateAsync`, too. This is not the case in `JwtBearerHandler` today ([see here](https://github.com/aspnet/Security/blob/137e18f39d6f9e0932916fa9e3a8c296e8dd8978/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs#L182)).

It is actually unavoidable because `HandleAutneticateAsync` is virtual.

Should we consider a template pattern to enforce this rule? (change will be out of the scope of this issue, but worth considering).

HaoK: As you mention already we don't necessarily know what happens in HandleAuthenticate since its virtual. In general we only call the Once method I believe, so I think it's fine to just normalize our calls there as a result. Effectively the Once method can be thought of as EnsureNormalizedAuthResult (we could introduce a new method that wraps Once instead of changing the behavior but that will require more code changes for any middleware to pick up the new behavior). 

We don't know what people who are calling HandleAuthenticate explicitly are doing today since that's not what we do, so I don't think we can or should try to change the behavior of that virtual method, only the wrapper.

troydai: Tried to push up the the exception catching: ([see here](https://github.com/aspnet/Security/tree/troy-927-2)) it will break some user's scenario. Because exception which used to create the middleware ([here](https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs#L68)) is now safely captured. Exception is not thrown from middleware any more. This will break all the customer scenarios where leverage exception to break through authenticate middleware (it's not a right thing to do though).

HaoK: I think this is reasonable behavior to break, @Tratcher thoughts?  I don't think its reasonable for people to depend on random uncaught exceptions

Tratcher: In Katana we had global try/catch blocks and it made the components impossible for users to debug. We explicitly removed them in K and tried to restrict our exception handling to expected exceptions (e.g. token validation).

HaoK: Wouldn't it be enough if we simply added logging within this catch for debugging?

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
```diff
@@ -233,12 +233,28 @@ public bool ShouldHandleScheme(string authenticationScheme, bool handleAutomatic
...
+        /// Exception is thrown from the HandleAuthenticateAsync task will be handled and saved to AuthenticateResult.
+        /// </summary>
         protected Task<AuthenticateResult> HandleAuthenticateOnceAsync()
         {
             if (_authenticateTask == null)
             {
-                _authenticateTask = HandleAuthenticateAsync();
+                try
+                {
+                    _authenticateTask = HandleAuthenticateAsync().ContinueWith<AuthenticateResult>(
```
HaoK: Given that we don't consistently expect/check errors from this, we should probably Log any failures from this method

PinpointTownes:
Really curious: why not simply updating the OIDC/JWT middleware to avoid rethrowing exceptions by default? This line should return `AuthenticateResult.Fail(ex)`  IMHO: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs#L182

The biggest concern I have with the current approach is that it may hide critical exceptions.

HaoK:
I'm happy with a smaller fix to just avoid throwing as well, try/catch blocks are meh in general

troydai:
Take two here: https://github.com/aspnet/Security/pull/937

troydai:
Ping @Tratcher @HaoK updated.

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
```diff
@@ -233,15 +233,43 @@ public bool ShouldHandleScheme(string authenticationScheme, bool handleAutomatic
...
+        ///
+        /// If the authentication has been done before returns the last authentication result.
+        /// This method won't throw exception. Any exception thrown during the authentication will be convert
+        /// to a AuthenticateResult.
+        /// </summary>
+        protected Task<AuthenticateResult> HandleAuthenticateOnceSafeAsync()
+        {
+            try
+            {
+                return HandleAuthenticateOnceAsync().ContinueWith<AuthenticateResult>(
```
Tratcher: Just use async/await

Tratcher:
What about this one?
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs#L188

PinpointTownes:
I'm still unsure to understand why this approach is preferable to the other one. Why would you want to catch exceptions from events? (= user code)

HaoK:
The basic idea is that we don't want to catch exceptions during authenticates. But we do want a safe way to see if we have a ticket already like in the challenge flows.  We can treat any errors as if was no ticket in that case.

Tratcher:
Yeah, we're not trying to catch exceptions from events/user-code. We're trying to allow challenge to succeed even if auth failed.

PinpointTownes:
Don't exceptions thrown from `HandleAuthenticateAsync` indicate that something bad happened? (by exceptions, I mean real ones, not the token validation exceptions we catch)

In this case, do we really want to ignore that and proceed the request anyway?

troydai:
The problem is that the logics which is called back from response's `OnStarting` is executed in server. So it is up to server to determine how the exceptions are handled. This causes ambiguity. For example, Kestrel translate the exception to a 500 while TestServer doesn't do anything. 

The change focus on the logic invoked in the callback from `OnStarting`.

PinpointTownes:
Point taken. I guess it's just me, but I'm tempted to think that an uncaught exception that escapes from `HandleAuthenticateAsync` is an internal server error. With this PR, this exception is ignored and we ultimately end up returning a 401 response, that is supposed to indicate a client error (a lack of authentication), not a server error.

Tratcher:
Honestly that's up the the caller. They would have to catch the auth exception once before even invoking the challenge.

Tratcher:
One more place: https://github.com/aspnet/Security/blob/926c7fab4bac4b5b5ec58adc08d069edafc21132/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs#L168

troydai:
Added.

Tratcher:
:shipit: (but give @HaoK a few minutes to review first)

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -165,7 +165,7 @@ private CookieOptions BuildCookieOptions()
             }
 
             // REVIEW: Should this check if there was an error, and then if that error was already handled??
```
HaoK: While we are here, we should resolve this really old REVIEW of mine :)  I think for cookies its safe to just ignore any errors and treat them equivalent to a null ticket, so we can probably just remove this review now.  @Tratcher agreed?

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
```diff
@@ -233,15 +233,40 @@ public bool ShouldHandleScheme(string authenticationScheme, bool handleAutomatic
             }
         }
 
+        /// <summary>
+        /// Handle the authentication for once.
```
HaoK: Nit: slightly better wording might be:

Used to ensure HandleAuthenticateAsync is only invoked once, subsequent calls will return the same result.

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
```diff
@@ -233,15 +233,40 @@ public bool ShouldHandleScheme(string authenticationScheme, bool handleAutomatic
...
                 _authenticateTask = HandleAuthenticateAsync();
             }
+
             return _authenticateTask;
         }
 
+        /// <summary>
+        /// Handle the authentication for once.
+        ///
+        /// If the authentication has been done before returns the last authentication result.
```
HaoK: Maybe:

Used to ensure HandleAuthenticateAsync is only invoked once safely, subsequent calls will return the same authentication result.  Any exceptions will be converted into a failed AuthenticationResult containing the exception.

HaoK:
:shipit: once the minor wording/comment removal is considered

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
Tratcher: Agreed, we've now handled the error. Remove the comment.

PR #937, opened 2016-08-04
Catch exception thrown from AuthenticationFailed.

troydai:
Address issue #927 

Take two.

This is a smaller fix. Avoid re-thrown in the `catch` clause in JwtBearerHandler's authenticate method.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
@@ -166,20 +166,30 @@ namespace Microsoft.AspNetCore.Authentication.JwtBearer
             }
             catch (Exception ex)
             {
-                Logger.ErrorProcessingMessage(ex);
-
-                var authenticationFailedContext = new AuthenticationFailedContext(Context, Options)
+                try
                 {
-                    Exception = ex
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -166,20 +166,30 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
...
+                    if (authenticationFailedContext.CheckEventResult(out result))
+                    {
+                        return result;
+                    }
+                    else
+                    {
+                        return AuthenticateResult.Fail(ex);
+                    }
+                }
+                catch (Exception ex2)
```
Tratcher: Why are we doing this? This is strictly user code, they can handle their own exceptions.

PinpointTownes:
Curious: what was wrong with this approach?

troydai:
`HandleAuthenticatedAsync` are used in different situation, whether exceptions are allowed will be treated differently.

For authenticate flow, user exception will be thrown.
For challenge flow, a failed authentication will instead cause a 403 instead of blowing up the entire stack. Especially there are portion of the challenge scenario run in the server layer through response's OnStarting callback.

Issue #800, opened 2016-04-26
Update CookieAuthenticationHandler.ApplyHeaders to honor AuthenticationProperties.RedirectUri


PinpointTownes:
Currently, this code doesn't work as `CookieAuthenticationHandler.ApplyHeaders` doesn't honor `AuthenticationProperties.RedirectUri` (instead, you get an empty result). It would be great to fix that in the next milestone.

```csharp
public IActionResult SignOut() {
    return SignOut(new AuthenticationProperties { RedirectUri = "/" },
        CookieAuthenticationDefaults.AuthenticationScheme);
}
```

Tratcher:
Note that the use of RedirectUri varies. In cases like Challenge it is used as the final destination after authentication is complete. Using it directly for SignOut may be fine though.

PinpointTownes:
Well, AFAICT, we've removed all the problematic uses of `RedirectUri` (e.g https://github.com/aspnet/Security/pull/275), which is now basically just a `ReturnUrl` and nothing more.

muratg:
Let's investigate in 1.0.0.

troydai:
Fixed in https://github.com/aspnet/Security/pull/940/commits/6a9f1f9887ec5983da5c20b4d8b788929d7b1b32

PR #940, opened 2016-08-09
Honor AuthenticationProperties.RedirectUri in CookieAuthenticationHan…

troydai:
Address issue https://github.com/aspnet/Security/issues/800

Working on tests.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -311,14 +311,24 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
             Response.Headers[HeaderNames.CacheControl] = HeaderValueNoCache;
             Response.Headers[HeaderNames.Pragma] = HeaderValueNoCache;
             Response.Headers[HeaderNames.Expires] = HeaderValueMinusOne;
+
             if (shouldRedirectToReturnUrl && Response.StatusCode == 200)
             {
+                CookieRedirectContext redirectContext = null;
+                
                 var query = Request.Query;
                 var redirectUri = query[Options.ReturnUrlParameter];
-                if (!StringValues.IsNullOrEmpty(redirectUri)
-                    && IsHostRelative(redirectUri))
+                if (!StringValues.IsNullOrEmpty(redirectUri) && IsHostRelative(redirectUri))
                 {
-                    var redirectContext = new CookieRedirectContext(Context, Options, redirectUri, properties);
+                    redirectContext = new CookieRedirectContext(Context, Options, redirectUri, properties);
+                }
+                else if (!string.IsNullOrEmpty(properties.RedirectUri) && IsHostRelative(properties.RedirectUri))
+                {
+                    redirectContext = new CookieRedirectContext(Context, Options, properties.RedirectUri, properties);
+                }
+
+                if (redirectContext != null)
+                {
                     await Options.Events.RedirectToReturnUrl(redirectContext);
                 }
             }
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -315,20 +315,23 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
             if (shouldRedirectToReturnUrl && Response.StatusCode == 200)
             {
                 CookieRedirectContext redirectContext = null;
-                
-                var query = Request.Query;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -311,14 +311,24 @@ private CookieOptions BuildCookieOptions()
...
+
             if (shouldRedirectToReturnUrl && Response.StatusCode == 200)
             {
+                CookieRedirectContext redirectContext = null;
+                
                 var query = Request.Query;
                 var redirectUri = query[Options.ReturnUrlParameter];
-                if (!StringValues.IsNullOrEmpty(redirectUri)
-                    && IsHostRelative(redirectUri))
+                if (!StringValues.IsNullOrEmpty(redirectUri) && IsHostRelative(redirectUri))
```
PinpointTownes: IMO, the redirect_uri stored in the properties should be preferred to the return_url parameter (when it's present).

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -311,14 +311,24 @@ private CookieOptions BuildCookieOptions()
             Response.Headers[HeaderNames.CacheControl] = HeaderValueNoCache;
             Response.Headers[HeaderNames.Pragma] = HeaderValueNoCache;
             Response.Headers[HeaderNames.Expires] = HeaderValueMinusOne;
+
             if (shouldRedirectToReturnUrl && Response.StatusCode == 200)
             {
+                CookieRedirectContext redirectContext = null;
+                
                 var query = Request.Query;
                 var redirectUri = query[Options.ReturnUrlParameter];
```
PinpointTownes: Not sure why you need such a complicated approach.

Why not something like that?

```csharp
var redirectUri = properties.RedirectUri;
if (string.IsNullOrEmpty(redirect)) {
    redirectUri = query[Options.ReturnUrlParameter];
}

**Comment by PinpointTownes** [comment]
Thanks for fixing that :+1:

File: src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs (hunk 2 above)
Comment (troydai) [inline src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs]: Original code, didn't think twice. But apparently you suggested a better version. I'll updater.

File: src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs (hunk 1 above)
Comment (troydai) [inline src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs]: @Tratcher ideas?

**Comment by troydai** [comment]
Updated. Ping @Tratcher @PinpointTownes

**Comment by Tratcher** [comment]
:shipit:

File: src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs [hunk 3]
```diff
@@ -311,14 +311,27 @@ private CookieOptions BuildCookieOptions()
...
-                var redirectUri = query[Options.ReturnUrlParameter];
-                if (!StringValues.IsNullOrEmpty(redirectUri)
-                    && IsHostRelative(redirectUri))
+                CookieRedirectContext redirectContext = null;
+
+                // set redirect uri in order:
+                // 1. properties.RedirectUri
+                // 2. query parameter ReturnUrlParameter
+                var redirectUri = properties.RedirectUri;
+                if (string.IsNullOrEmpty(redirectUri) || !IsHostRelative(redirectUri))
```
PinpointTownes: Is there a particular reason to prevent one from setting an absolute URL in the authentication properties?

This seems too restrictive and inconsistent with what we do in `RemoteAuthenticationHandler.HandleRemoteCallbackAsync`: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs#L109.

troydai: @Tratcher ?

Tratcher: The value from the query can be modified externally and cannot be trusted so we don't allow it to be absolute. The value from properties is more trustworthy, we could remove the check for that one if you add a comment for the query.

PinpointTownes: @Tratcher sure, query string values clearly fall into the NTUI category :smile:
My remark was only about the redirect URI stored in the authentication ticket.

PR #943, opened 2016-08-11
Allow absolute uri in authentication properties for cookie auth redirect

troydai:
/cc @PinpointTownes @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -11,7 +11,6 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Authentication;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Http.Features.Authentication;
-using Microsoft.Extensions.Primitives;
 using Microsoft.Net.Http.Headers;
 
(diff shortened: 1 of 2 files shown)


PinpointTownes:
Nice, thanks! :clap: 

Question: is it worth logging an info/warning message when a returnUrl query string parameter is rejected?

(minor nit: you could remove the variable declaration at line 317 and use `var` at line 325)

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -314,13 +313,11 @@ private CookieOptions BuildCookieOptions()
...
-
                 // set redirect uri in order:
                 // 1. properties.RedirectUri
                 // 2. query parameter ReturnUrlParameter
                 var redirectUri = properties.RedirectUri;
-                if (string.IsNullOrEmpty(redirectUri) || !IsHostRelative(redirectUri))
+                if (string.IsNullOrEmpty(redirectUri))
                 {
                     redirectUri = Request.Query[Options.ReturnUrlParameter];
                     if (string.IsNullOrEmpty(redirectUri) || !IsHostRelative(redirectUri))
```
Tratcher: Please comment why we require these to be relative.

Tratcher:
:shipit:

troydai:
Comments are added, waiting for the CI to pass.

PR #947, opened 2016-08-17
Update OpenId Connect tests

troydai:
1. Move configuration related tests to their own class;
2. Add tests to cover missing options scenarios;
3. Add TestDefaultValues class for default values used in OpenId connect
tests.

Related: https://github.com/aspnet/Security/issues/772

/cc @Tratcher

### DIFF
--- a/test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/Infrastructure/TestDefaultValues.cs
+++ b/test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/Infrastructure/TestDefaultValues.cs
@@ -1,0 +1,10 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Authentication.Tests.OpenIdConnect.Infrastructre
+{
+    internal class TestDefaultValues
+    {
+        public static readonly string DefaultAuthority = @"https://example.com/common";
+    }
+}
--- a/test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/OpenIdConnectConfigurationTests.cs
+++ b/test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/OpenIdConnectConfigurationTests.cs
@@ -1,0 +1,121 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Authentication.Tests.OpenIdConnect.Infrastructre;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.TestHost;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Authentication.Tests.OpenIdConnect
+{
+    public class OpenIdConnectConfigurationTests
+    {
+        [Fact]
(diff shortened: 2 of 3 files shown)


troydai:
Get more people @BrennanConroy @muratg @JunTaoLuo

test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/OpenIdConnectConfigurationTests.cs
```diff
@@ -0,0 +1,121 @@
...
+            return new TestServer(builder);
+        }
+
+        private void TestConfigurationException<T>(
+            OpenIdConnectOptions options,
+            Action<T> verifyException)
+            where T : Exception
+        {
+            var builder = new WebHostBuilder()
+                .Configure(app => app.UseOpenIdConnectAuthentication(options))
```
BrennanConroy: super nit: same order for these in both locations

troydai: super nit indeed.

BrennanConroy:
:ship: :it:

test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/Infrastructure/TestDefaultValues.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Authentication.Tests.OpenIdConnect.Infrastructre
```
Tratcher: This needed a whole new namespace?

Tratcher:
:shipit:

test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/Infrastructure/TestDefaultValues.cs
troydai: I generally align the folder structure with the namespace so as to quickly figure where the source codes are. Since it is test code, I won't break the convention as no customer is directly impacted.

Tratcher: The question stands, why did it need a new folder?

BrennanConroy: Namespace is spelled wrong...

troydai: Namespace will be fixed in next PR. A new folder to separate tests from utilities. Personal preference.

HaoK: Also there already is a TestUtilities class, why can't you just put this in there instead of having a stand alone namespace + class + folder + file with one constant?

https://github.com/aspnet/Security/blob/dev/test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/TestUtilities.cs#L14

troydai: I do plan to move move stuff in the this folder. For example the `Transaction` and `TestServer` code. These classes will be shared among test classes. I do intend to have multiple test classes each have limited number of test methods. that improve readability.

Let's keep this conversation when my next PR is out, we can then iron out how the utility class layout.

PR #966, opened 2016-09-05
Increase .travis.yml consistency between repos

dougbu:
- aspnet/Universe#349
- minimize `dotnet` setup time; no need for caching
- build with `--quiet`


dougbu:
5b323e5

Issue #859, opened 2016-06-01
Authorize(Github) may return a Facebook user


Tratcher:
https://github.com/aspnet/Security/pull/823/files/59fc691f4152e6d5017176c0b700ee9834640481#r64260972
For remote providers we implemented Authenticate to delegate to the current SignInScheme provider. However, we opted not to verify that the current identity provided by the SignInScheme provider was actually produced by the given middleware. E.g. When you have OAuth set up for Facebook and Github, they share an auth cookie, so when you Authorize(Github), if Facebook is already logged in, you get Facebook. This may be very confusing to users if there are portions of their application that require credentials from different providers. To work around this they would have to sign-out of the exiting provider, redirect, and then sign-in with the new provider.

HaoK:
Just to be clear, prior to returning the SignInScheme Identity, Authorize on any remote provider would have thrown an exception?  So this is a new behavior we are defining?

Tratcher:
It used to cause an infinite loop because Authenticate(Github) always returned null, so a new Challenge(Github) would be issued. We broke the infinite loop by having it return the current SignInScheme user.

HaoK:
Okay yeah so that infinite loop behavior is the interaction between MVC's AuthorizeFilter and RemoteHandlers always returning null.  I don't think the new behavior will be too confusing since if they are sharing external cookies, they will likely not be calling the shared scheme Facebook/Google, but rather something like "External", so it would be more that you are getting the "External" identity when asking for either "Facebook"/"Google" which makes sense to me if you've configured them this way

Tratcher:
We're going to start by documenting this and see if users run into issues.

Other options:
- The app can store two identities in the cookie.

mikes-gh:
Would it be possible to store external provider cookies by provider name

eg

Identity.Google
Identity.Facebook


Or would that require too much heavy lifting.
Apologies if this is way off.

Tratcher:
We can store the provider information directly in the cookie / auth properties. We don't need a separate cookie.

PR #980, opened 2016-09-20
Discriminate between providers when sharing an auth cookie

Tratcher:
#859 @Haok

This prevents `[Authorize("Github")]` from returning an identity produced by Facebook, etc. when they share a cookie middleware.


HaoK:
Can you also add a test exercising the template flow which uses a shared External cookie as opposed to named google/facebook cookies.

Tratcher:
I don't follow, they already do use a shared cookie. https://github.com/aspnet/Security/pull/980/files#diff-f6bd2d01f6e995e0fa83a42bb849b920R1010

HaoK:
The configure code path the templates/identity use is slightly different, they use SharedAuthenticationOptions.SignInScheme, would be good to make sure that identity/template configuration works as well with this change

Tratcher:
SharedAuthenticationOptions is being used here: https://github.com/aspnet/Security/pull/980/files#diff-f6bd2d01f6e995e0fa83a42bb849b920R1103

Issue #930, opened 2016-07-29
Returning true from HandleUnauthorizedAsync doesn't prevent the other automatic handlers from being invoked


PinpointTownes:
When refactoring ASOS code, I discovered a weird bug in `AuthenticationHandler` that prevents `HandleUnauthorizedAsync` from working correctly. The following conditions must be met for this annoying bug to occur:

  - The ASP.NET Core pipeline must contain at least 2 authentication middleware (e.g cookies + a custom middleware).

  - The first authentication middleware must be configured with `AutomaticChallenge = true`. The challenge mode - automatic or manual - used for the second middleware is not important.

  - The second authentication middleware must override `HandleUnauthorizedAsync` and must return `true` (which is supposed to prevent the prior handlers from being invoked).

  - `AuthenticationManager.ChallengeAsync` must be called with the authentication scheme associed with the second middleware.

When all these conditions are met, returning `true` from `HandleUnauthorizedAsync` has no effect: the first automatic middleware is still ultimately called and overrides the challenge applied by the second one (which is the only one that should handle it).

This bug is likely caused by `AuthenticationHandler.HandleAutomaticChallengeIfNeeded`.

Unlike `AuthenticationHandler.ChallengeAsync`, this method - executed by all the authentication handlers in the pipeline when headers are sent - doesn't check whether the prior handler handled the challenge and blindly calls `HandleUnauthorizedAsync`.

```csharp
private async Task HandleAutomaticChallengeIfNeeded()
{
    if (!ChallengeCalled && Options.AutomaticChallenge && Response.StatusCode == 401)
    {
        await HandleUnauthorizedAsync(new ChallengeContext(Options.AuthenticationScheme));
    }
}
```

Ironically, the JWT middleware is not impacted by this bug because it returns `false` to allow the other middleware to handle the challenge: the `ChallengeAsync` of all the prior handlers is called, which automatically sets `ChallengeCalled`. Since the authentication scheme doesn't correspond to the prior middleware, `ChallengeAsync` ignores the challenge and things work like they are supposed to.

/cc @HaoK @Tratcher

HaoK:
Ick, I will take a look

PinpointTownes:
Okay, note for me: never say "ick" :trollface: 

![image](https://cloud.githubusercontent.com/assets/6998306/17258245/384003f8-55c6-11e6-9bad-84a87c498572.png)

[...]

PR #982, opened 2016-09-21
Always call prior handlers in challenge

HaoK:
cc @Tratcher 

FYI I have a 'bad feeling about this' change, no tests failed either as a result :/...  I added one to verify the new behavior

Fixes https://github.com/aspnet/Security/issues/930

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
@@ -311,6 +311,11 @@ namespace Microsoft.AspNetCore.Authentication
             return TaskCache.CompletedTask;
         }
 
+        /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
```diff
@@ -323,7 +328,7 @@ protected virtual Task<bool> HandleForbiddenAsync(ChallengeContext context)
         /// changing the 401 result to 302 of a login page or external sign-in location.)
         /// </summary>
         /// <param name="context"></param>
-        /// <returns>True if no other handlers should be called</returns>
+        /// <returns>The returned boolean is ignored.</returns>
```
Tratcher: "The returned boolean is no longer used."

HaoK:
e12838e38f5f8be8371c0a7b02d9ce47e0663ce0

Issue #883, opened 2016-06-15
OpenIdConnect with AAD does not return error_description


dstrockis:
If AAD gives an error during the auth pipeline, the middlware swallows up the `error_description` from AAD and doesn’t return it to the developer.  It makes it pretty difficult to debug issues when talking to AAD.  

You can repro by forgetting to include a client secret in the OIDC config, and letting the middleware attempted to redeem the auth code without a client secret. It will result in a 400 from AAD, but it's not clear from the middleware why it occurred.

Tratcher:
It fails here:
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L670
with a 400 response. This gets caught by https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L622, runs AuthenticationFailed, and re-throws. Assuming nobody catches this then the server will convert it to a 500 response.

The DeveloperExceptionPage would show something like
```
System.AggregateException: Unhandled remote failure. ---> System.Net.Http.HttpRequestException: Response status code does not indicate success: 400 (Bad Request).
   at System.Net.Http.HttpResponseMessage.EnsureSuccessStatusCode()
   at Microsoft.AspNetCore.Authentication.OpenIdConnect.OpenIdConnectHandler.<RedeemAuthorizationCodeAsync>d__21.MoveNext() in C:\git\Universe\Security\src\Microsoft.AspNetCore.Authentication.OpenIdConnect\OpenIdConnectHandler.cs:line 670
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at Microsoft.AspNetCore.Authentication.OpenIdConnect.OpenIdConnectHandler.<HandleRemoteAuthenticateAsync>d__19.MoveNext() in C:\git\Universe\Security\src\Microsoft.AspNetCore.Authentication.OpenIdConnect\OpenIdConnectHandler.cs:line 558
```

The response body with the description is not currently available. The OAuth handler does do some work to expose the response body in this scenario:
[...]

PR #984, opened 2016-09-21
Log the error inforamtion when redeem auth code

troydai:
Address issue #883

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/LoggingExtensions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/LoggingExtensions.cs
@@ -37,7 +37,8 @@ namespace Microsoft.Extensions.Logging
         private static Action<ILogger, string, Exception> _invalidLogoutQueryStringRedirectUrl;
         private static Action<ILogger, Exception> _nullOrEmptyAuthorizationResponseState;
         private static Action<ILogger, Exception> _unableToReadAuthorizationResponseState;
-        private static Action<ILogger, string, string, string, Exception> _authorizationResponseError;
+        private static Action<ILogger, string, string, string, Exception> _responseError;
+        private static Action<ILogger, string, string, string, int, Exception> _responseErrorWithStatusCode;
         private static Action<ILogger, Exception> _exceptionProcessingMessage;
         private static Action<ILogger, Exception> _accessTokenNotAvailable;
         private static Action<ILogger, Exception> _retrievingClaims;
@@ -106,10 +107,14 @@ namespace Microsoft.Extensions.Logging
                 eventId: 11,
                 logLevel: LogLevel.Debug,
                 formatString: "Unable to read the message.State.");
-            _authorizationResponseError = LoggerMessage.Define<string, string, string>(
+            _responseError = LoggerMessage.Define<string, string, string>(
                 eventId: 12,
                 logLevel: LogLevel.Error,
                 formatString: "Message contains error: '{Error}', error_description: '{ErrorDescription}', error_uri: '{ErrorUri}'.");
+            _responseErrorWithStatusCode = LoggerMessage.Define<string, string, string, int>(
+                eventId: 49,
+                logLevel: LogLevel.Error,
+                formatString: "Message contains error: '{Error}', error_description: '{ErrorDescription}', error_uri: '{ErrorUri}', status code '{StatusCode}'.");
             _updatingConfiguration = LoggerMessage.Define(
                 eventId: 13,
                 logLevel: LogLevel.Debug,
@@ -380,9 +385,14 @@ namespace Microsoft.Extensions.Logging
             _unableToReadAuthorizationResponseState(logger, null);
         }
 
-        public static void AuthorizationResponseError(this ILogger logger, string error, string errorDescription, string errorUri)
+        public static void ResponseError(this ILogger logger, string error, string errorDescription, string errorUri)
         {
-            _authorizationResponseError(logger, error, errorDescription, errorUri, null);
+            _responseError(logger, error, errorDescription, errorUri, null);
+        }
+
+        public static void ResponseErrorWithStatusCode(this ILogger logger, string error, string errorDescription, string errorUri, int statusCode)
+        {
+            _responseErrorWithStatusCode(logger, error, errorDescription, errorUri, statusCode, null);
         }
 
         public static void ExceptionProcessingMessage(this ILogger logger, Exception ex)
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
@@ -507,14 +507,7 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
                 // if any of the error fields are set, throw error null
                 if (!string.IsNullOrEmpty(authorizationResponse.Error))
                 {
-                    Logger.AuthorizationResponseError(
-                        authorizationResponse.Error,
-                        authorizationResponse.ErrorDescription ?? "ErrorDecription null",
-                        authorizationResponse.ErrorUri ?? "ErrorUri null");
-
-                    return AuthenticateResult.Fail(new OpenIdConnectProtocolException(
-                        string.Format(CultureInfo.InvariantCulture, Resources.MessageContainsError, authorizationResponse.Error,
-                        authorizationResponse.ErrorDescription ?? "ErrorDecription null", authorizationResponse.ErrorUri ?? "ErrorUri null")));
+                    return AuthenticateResult.Fail(CreateOpenIdConnectProtocolException(authorizationResponse, response: null));
                 }
 
                 if (_configuration == null && Options.ConfigurationManager != null)
@@ -590,6 +583,7 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
                     {
                         return result;
                     }
+
                     authorizationResponse = tokenResponseReceivedContext.ProtocolMessage;
                     tokenEndpointResponse = tokenResponseReceivedContext.TokenEndpointResponse;
 
@@ -684,20 +678,50 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
         }
 
         /// <summary>
-        /// Redeems the authorization code for tokens at the token endpoint
+        /// Redeems the authorization code for tokens at the token endpoint.
         /// </summary>
         /// <param name="tokenEndpointRequest">The request that will be sent to the token endpoint and is available for customization.</param>
         /// <returns>OpenIdConnect message that has tokens inside it.</returns>
         protected virtual async Task<OpenIdConnectMessage> RedeemAuthorizationCodeAsync(OpenIdConnectMessage tokenEndpointRequest)
         {
             Logger.RedeemingCodeForTokens();
+
             var requestMessage = new HttpRequestMessage(HttpMethod.Post, _configuration.TokenEndpoint);
             requestMessage.Content = new FormUrlEncodedContent(tokenEndpointRequest.Parameters);
+
             var responseMessage = await Backchannel.SendAsync(requestMessage);
-            responseMessage.EnsureSuccessStatusCode();
-            var tokenResonse = await responseMessage.Content.ReadAsStringAsync();
-            var jsonTokenResponse = JObject.Parse(tokenResonse);
-            return new OpenIdConnectMessage(jsonTokenResponse);
+
+            var contentMediaType = responseMessage.Content.Headers.ContentType?.MediaType;
+            if (string.IsNullOrEmpty(contentMediaType))
+            {
+                Logger.LogDebug($"Unexpected token response format. Status Code: {(int)responseMessage.StatusCode}. Content-Type header is missing.");
+            }
+            else if (!string.Equals(contentMediaType, "application/json", StringComparison.OrdinalIgnoreCase))
+            {
+                Logger.LogDebug($"Unexpected token response format. Status Code: {(int)responseMessage.StatusCode}. Content-Type {responseMessage.Content.Headers.ContentType}.");
+            }
+
+            // Error handling:
+            // 1. If the response body can't be parsed as json, throws.
+            // 2. If the response's status code is not in 2XX range, throw OpenIdConnectProtocolException. If the body is correct parsed, 
+            //    pass the error information from body to the exception.
+            OpenIdConnectMessage message;
+            try
+            {
+                var responseContent = await responseMessage.Content.ReadAsStringAsync();
+                message = new OpenIdConnectMessage(responseContent);
+            }
+            catch (Exception ex)
+            {
+                throw new OpenIdConnectProtocolException($"Failed to parse token response body as JSON. Status Code: {(int)responseMessage.StatusCode}. Content-Type: {responseMessage.Content.Headers.ContentType}", ex);
+            }
+
+            if (!responseMessage.IsSuccessStatusCode)
+            {
+                throw CreateOpenIdConnectProtocolException(message, responseMessage);
+            }
+
+            return message;
         }
 
         /// <summary>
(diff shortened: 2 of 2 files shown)


src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -585,6 +578,11 @@ protected virtual Task<bool> HandleSignOutCallbackAsync()
                         tokenEndpointResponse = await RedeemAuthorizationCodeAsync(tokenEndpointRequest);
                     }
 
+                    if (!string.IsNullOrEmpty(tokenEndpointResponse.Error))
```
PinpointTownes: Note: this is a behavior change as the `TokenResponseReceived` event will no longer be called for errored responses.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -691,13 +689,21 @@ private void PopulateSessionProperties(OpenIdConnectMessage message, Authenticat
...
             requestMessage.Content = new FormUrlEncodedContent(tokenEndpointRequest.Parameters);
+
             var responseMessage = await Backchannel.SendAsync(requestMessage);
-            responseMessage.EnsureSuccessStatusCode();
-            var tokenResonse = await responseMessage.Content.ReadAsStringAsync();
-            var jsonTokenResponse = JObject.Parse(tokenResonse);
-            return new OpenIdConnectMessage(jsonTokenResponse);
+            var message = new OpenIdConnectMessage(await responseMessage.Content.ReadAsStringAsync());
+
+            if (string.IsNullOrEmpty(message.Error))
```
PinpointTownes: This looks fragile and will likely cause a weird exception if the HTTP error response doesn't contain JSON (e.g because of a 500 response returned before the identity provider has a chance to send back a proper JSON error).

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -691,13 +689,21 @@ private void PopulateSessionProperties(OpenIdConnectMessage message, Authenticat
...
+            {
+                // If error doesn't present in the response content ensure the response successed.
```
PinpointTownes: > You are under arrest - you have the right to remain silent. Anything you say can and will be used against you in a court of law.

Grammar police :trollface:

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
troydai: Addressed, see the latest commit.

Overall, the focus here is that if authentication response status is 200 but with an `error` in body, do we consider it an error and interrupt the operation. The change I put in the second commit will allow the authentication to move one in this situation.

In real world, I doubt if there is any authentication provider will do this. (How to set status code for failed authentication response is not defined in the spec.)

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
troydai: Addressed in the latest commit. If the body is not a json, exception will be thrown to interrupt the authentication process.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -580,6 +573,9 @@ protected virtual Task<bool> HandleSignOutCallbackAsync()
                     ticket = authorizationCodeReceivedContext.Ticket;
                     jwt = authorizationCodeReceivedContext.JwtSecurityToken;
 
+                    // Redeem the authorization code.
+                    // The authorization response may contain error while its status code doesn't indicate failure. In this case
+                    // exception won't be thrown. The OpenIdConnectMessage will carry the error code.
```
Tratcher: This comment belongs inside RedeemAuthorizationCodeAsync

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -691,13 +688,33 @@ private void PopulateSessionProperties(OpenIdConnectMessage message, Authenticat
...
+                {
+                    throw CreateOpenIdConnectProtocolException(result);
+                }
+
+                return result;
+            }
+            else
+            {
+                throw new OpenIdConnectProtocolException(
+                    $"Failed to redeem authentication code. The authenticaiton response content type is not application/json but {responseMessage.Content.Headers.ContentType.MediaType}.");
```
Tratcher: `authorization`

Tratcher: We also need to include the status code.
"Failed to redeem the authorization code. Status code: `statusCode`; Content-Type: `content-type`.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
PinpointTownes: Sure, `OpenIdConnectMessage` will likely throw an exception if it cannot parse the response body. My remark was more about the exception message: if a server returns an empty 500 response, will the message be clear enough about the fact the server was faulty?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
PinpointTownes: Yeah, forget that, `TokenResponseReceived` is not invoked for error responses due to the `EnsureSuccessStatusCode` call, so no regression here :smile: 

> How to set status code for failed authentication response is not defined in the spec.

Actually, it is: https://tools.ietf.org/html/rfc6749#section-5.2

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -691,13 +688,33 @@ private void PopulateSessionProperties(OpenIdConnectMessage message, Authenticat
...
-            var tokenResonse = await responseMessage.Content.ReadAsStringAsync();
-            var jsonTokenResponse = JObject.Parse(tokenResonse);
-            return new OpenIdConnectMessage(jsonTokenResponse);
+
+            // Error handling:
+            // 1. If the response body is json, convert it to an OpenIdConnectMessage.
+            // 2. If the response's status code is not in 2XX range, throw OpenIdConnectProtocolException. If the body is correct parsed, 
+            //    pass the error information from body to the exception.
+            // 3. If the response body is not json, throw OpenIdConnectProtocolException.
+            if (string.Equals(responseMessage.Content.Headers.ContentType.MediaType, "applicaiton/json", StringComparison.OrdinalIgnoreCase))
```
PinpointTownes: `applicaiton/json` -> `applicaiton/json`

PinpointTownes: Also, won't this line null-ref if no `Content-Type` header is returned?

Related remark: https://github.com/aspnet/Security/issues/985#issuecomment-249035900

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
troydai: I see. It's in the OAuth spec. Thanks for pointing out. I'll make the change to the flow accordingly.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -684,20 +679,49 @@ private void PopulateSessionProperties(OpenIdConnectMessage message, Authenticat
         }
 
         /// <summary>
-        /// Redeems the authorization code for tokens at the token endpoint
+        /// Redeems the authorization code for tokens at the token endpoint.
+        /// 
+        /// The authorization response may contain error while its status code doesn't indicate failure. In this case
```
Tratcher: we decided from the spec that this isn't true, the status code should be 400 for errors

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -684,20 +679,49 @@ private void PopulateSessionProperties(OpenIdConnectMessage message, Authenticat
...
             var responseMessage = await Backchannel.SendAsync(requestMessage);
-            responseMessage.EnsureSuccessStatusCode();
-            var tokenResonse = await responseMessage.Content.ReadAsStringAsync();
-            var jsonTokenResponse = JObject.Parse(tokenResonse);
-            return new OpenIdConnectMessage(jsonTokenResponse);
+
+            var contentMediaType = responseMessage.Content.Headers.ContentType?.MediaType;
+            if (!string.IsNullOrEmpty(contentMediaType) && !string.Equals(contentMediaType, "application/json", StringComparison.OrdinalIgnoreCase))
+            {
+                Logger.LogWarning($"The token error response's content type is not application/json but {contentMediaType}.");
```
Tratcher: 3rd party caused issues should be logged as Debug or Info. Make this one Info
"Unexpected token response format. Status Code: {...}; Content-Type: '{...}'"

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -684,20 +679,49 @@ private void PopulateSessionProperties(OpenIdConnectMessage message, Authenticat
...
             requestMessage.Content = new FormUrlEncodedContent(tokenEndpointRequest.Parameters);
+
             var responseMessage = await Backchannel.SendAsync(requestMessage);
-            responseMessage.EnsureSuccessStatusCode();
-            var tokenResonse = await responseMessage.Content.ReadAsStringAsync();
-            var jsonTokenResponse = JObject.Parse(tokenResonse);
-            return new OpenIdConnectMessage(jsonTokenResponse);
+
+            var contentMediaType = responseMessage.Content.Headers.ContentType?.MediaType;
+            if (!string.IsNullOrEmpty(contentMediaType) && !string.Equals(contentMediaType, "application/json", StringComparison.OrdinalIgnoreCase))
```
Tratcher: Log for null or empty too.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -684,20 +679,49 @@ private void PopulateSessionProperties(OpenIdConnectMessage message, Authenticat
...
+            // 1. If the response body can't be parsed as json, throws.
+            // 2. If the response's status code is not in 2XX range, throw OpenIdConnectProtocolException. If the body is correct parsed, 
+            //    pass the error information from body to the exception.
+            OpenIdConnectMessage message;
+            try
+            {
+                var responseContent = await responseMessage.Content.ReadAsStringAsync();
+                message = new OpenIdConnectMessage(responseContent);
+            }
+            catch (Exception ex)
```
Tratcher: JSON parse exception?

troydai: I'd rather not in case `OpenIdConnectMessage` change the internal json parsing logic.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
troydai: yeah. will just remove.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -684,20 +678,50 @@ private void PopulateSessionProperties(OpenIdConnectMessage message, Authenticat
...
+            {
+                throw new OpenIdConnectProtocolException($"Failed to parse token error response body as JSON. Status Code: {(int)responseMessage.StatusCode}. Content-Type: {responseMessage.Content.Headers.ContentType}", ex);
```
Tratcher: token response, you don't know if it's an error yet.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -1157,5 +1184,30 @@ private string BuildRedirectUriIfRelative(string uri)
...
+
+            return new OpenIdConnectProtocolException(string.Format(
+                CultureInfo.InvariantCulture,
+                Resources.MessageContainsError,
+                message.Error,
+                description,
+                errorUri));
+        }
+
+        private OpenIdConnectProtocolException CreateOpenIdConnectProtocolException(OpenIdConnectMessage message)
```
Tratcher: Remove this overload, it's not providing anything

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/LoggingExtensions.cs
```diff
@@ -110,6 +111,10 @@ static LoggingExtensions()
                 eventId: 12,
                 logLevel: LogLevel.Error,
                 formatString: "Message contains error: '{Error}', error_description: '{ErrorDescription}', error_uri: '{ErrorUri}'.");
+            _authorizationResponseErrorWithStatusCode = LoggerMessage.Define<string, string, string, int>(
+                eventId: 12,
```
Tratcher: New event id, the call sites and content are different.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -1157,5 +1184,30 @@ private string BuildRedirectUriIfRelative(string uri)
...
         }
+
+        private OpenIdConnectProtocolException CreateOpenIdConnectProtocolException(OpenIdConnectMessage message, HttpResponseMessage response)
+        {
+            var description = message.ErrorDescription ?? "error_description is null";
+            var errorUri = message.ErrorUri ?? "error_uri is null";
+
+            if (response != null)
+            {
+                Logger.AuthorizationResponseErrorWithStatusCode(message.Error, description, errorUri, (int)response.StatusCode);
```
Tratcher: This one wasn't an authorization response, it was the token response...

troydai: The error message doesn't mention the type of the response.

```
 "Message contains error: '{Error}', error_description: '{ErrorDescription}', error_uri: '{ErrorUri}', status code '{StatusCode}'.")
```

troydai:
Updated.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/LoggingExtensions.cs
troydai: Set to `49`

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
troydai: Removed.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
troydai: Updated

troydai:
PR is approved. I'll merge once CI passes.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
PinpointTownes: I guess @Tratcher means the event is badly named: the response you receive at this point is a token response, not an authorization response.

PinpointTownes: Arf, looks like this method is shared and is called for both authorization and token response errors. Maybe a more generic name would be better? `ResponseErrorWithStatusCode`?

troydai: All right, I'll update the method name.

troydai:
Amended the PR once more to update the logging extension method name. I'm waiting on the CI and will merge this PR afterwards.

Issue #971, opened 2016-09-13
Google middleware authorization should use prompt instead of approval_prompt


hempels:
I see support in GoogleHandler.BuildChallengeUrl for approval_prompt via:
`AddQueryString(queryStrings, properties, "approval_prompt");`

First, I can't figure out how to apply that property. Setting it in CookieAuthenticationEvents.OnSigningIn doesn't seem to cause any different behavior by Google.

Second, it's not the right property. Google deprecated it in favor of prompt, which has four possible values: "none", "consent", "select_account", or "select_account consent".
see: https://developers.google.com/identity/protocols/OAuth2WebServer

I've managed to work around this limitation by adding ?prompt=select_account to the AuthorizationEndpoint in GoogleOptions, but it's a pretty ugly hack.

Tratcher:
Those property overrides can be specified in the initial call to Challenge with the AuthenticationProperties collection.

As for approval_prompt, we'll have to dig up the official list of parameters and update it.

hempels:
One note, I think the current solution is perhaps creating extra work for you guys and for consumers of the API. I struggle to come up with a reason it wouldn't make sense to just provide a parameters dictionary so any value can be passed to the target API. That approach seems to work for WebRequest headers, etc. The way this is implemented now, the library will constantly be chasing API revisions.

Eilon:
We should check the Google docs if `approval_prompt` is still the right field, and if not, update our codes.

troydai:
`approval_prompt` is renamed to `prompt` for Google API. I'll make changes in our code accordingly.

troydai:
`approval_prompt` is a legacy parameter. Though it is not in the official document, it is still working. `prompt` is a different parameter for different usage. We will keep both.

troydai:
Issue is fixed in ddeef1f9ac09c34c2a2c28e35923ae7288b1049e

I filed additional issue #990 to track @hempels 's feedback on auth query parameters.

PR #989, opened 2016-09-23
Rename Google auth endpoint parameter approval_prompt to prompt

troydai:
Issue #971

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Google/GoogleHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Google/GoogleHandler.cs
@@ -87,18 +87,20 @@ namespace Microsoft.AspNetCore.Authentication.Google
         // TODO: Abstract this properties override pattern into the base class?
         protected override string BuildChallengeUrl(AuthenticationProperties properties, string redirectUri)
         {
-            var scope = FormatScope();
+            // Google Identity Platform Manual:
+            // https://developers.google.com/identity/protocols/OAuth2WebServer
 
             var queryStrings = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Authentication.Google/GoogleHandler.cs
```diff
@@ -87,18 +87,19 @@ public GoogleHandler(HttpClient httpClient)
...
             queryStrings.Add("response_type", "code");
             queryStrings.Add("client_id", Options.ClientId);
             queryStrings.Add("redirect_uri", redirectUri);
 
-            AddQueryString(queryStrings, properties, "scope", scope);
-
+            AddQueryString(queryStrings, properties, "scope", FormatScope());
             AddQueryString(queryStrings, properties, "access_type", Options.AccessType);
-            AddQueryString(queryStrings, properties, "approval_prompt");
+            AddQueryString(queryStrings, properties, "prompt");
```
PinpointTownes: Isn't that a functional breaking change, as it changes the authentication property name?

PinpointTownes: Note: you'll also have to fix the Google tests: https://github.com/aspnet/Security/blob/dev/test/Microsoft.AspNetCore.Authentication.Test/Google/GoogleMiddlewareTests.cs

troydai: True. I'll keep both.

troydai: Tests are updated.

Tratcher:
:shipit:

Issue #988, opened 2016-09-23
How to use AuthorizationEndpoint which contains query string parameters with OAuth.


ademcaglin:
Is there an easy way to use `AuthorizationEndpoint `which contains query string parameters?  For example below code doesn't work as i expect

        var facebookOptions = new FacebookOptions
        {
              //...
              AuthorizationEndpoint = FacebookDefaults.AuthorizationEndpoint + "?param1=param1",
        };

Only way i found to solve this problem is using event:

       var facebookOptions = new FacebookOptions
        {
              //...
              Events = new OAuthEvents()
                {
                    OnRedirectToAuthorizationEndpoint = ctx =>
                    {
                        ctx.HttpContext.Response.Redirect(ctx.RedirectUri + "&param1=param1");
                        return Task.FromResult(0);
                    }
                }
        };

Tratcher:
That's it for now. RE: https://github.com/aspnet/Security/issues/988

troydai:
Related https://github.com/aspnet/Security/issues/990

PR #992, opened 2016-09-23
Changed BuildChallengeUrl in order to accept AuthorizationEndpoint wh…

ademcaglin:
…ich contains query parameters

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
@@ -15,6 +15,7 @@ using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Http.Features.Authentication;
 using Microsoft.Extensions.Primitives;
 using Newtonsoft.Json.Linq;
+using Microsoft.AspNetCore.WebUtilities;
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -206,16 +207,15 @@ protected virtual string BuildChallengeUrl(AuthenticationProperties properties,
...
+            var parameters = new Dictionary<string, string>
             {
                 { "client_id", Options.ClientId },
                 { "scope", scope },
                 { "response_type", "code" },
                 { "redirect_uri", redirectUri },
                 { "state", state },
             };
-            return Options.AuthorizationEndpoint + queryBuilder.ToString();
+            return QueryHelpers.AddQueryString(Options.AuthorizationEndpoint, parameters);
```
Tratcher: We've done this in other places in the auth middleware, we could also do it here. https://github.com/aspnet/Security/blob/5231c5a853a36f62bcd4f982660a59b2b7f1dd40/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs#L28

Tratcher:
RE: https://github.com/aspnet/Security/issues/988

Issue #973, opened 2016-09-14
CookieAuthenticationHandler, in case using SessionStore, cookieOptions.Expire is not set on renewal


JaapMosselman:
I think I found an issue in Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationHandler class, method FinishResponseAsync. There is a line of where cookieOptions.Expire property is set if IsPersistent is true. The problem is that, in case of using a SessionStore, ticket.Properties.IsPersistent is always false, because in that case ticket is assigned a new instance with clean options.
So the cookie send to the browser is now a session cookie and when the browser is closed and reopened, the user must login again.

troydai:
I'm investigating.

troydai:
Good catch. A fix is being worked on.

JaapMosselman:
Great. Actually I discovered this problem in my current project which uses [Microsoft.Owin.Security.Cookies](https://www.nuget.org/packages/Microsoft.Owin.Security.Cookies). I also created an issue at the [Katana project](https://katanaproject.codeplex.com/workitem/470), but there seems to be no activity.
After I found the problem, I just thought: lets look it the problem still exists in the aspnet core code :-)

troydai:
Thanks for reporting the issue, @JaapMosselman !

troydai:
https://github.com/aspnet/Security/commit/0152691108717bd74cc5f0406e7b745b31c8ca23

PR #995, opened 2016-09-27
Cookie Auto: set properties back to ticket when using SessionStore

troydai:
#973

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -150,6 +150,7 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
                 HttpOnly = Options.CookieHttpOnly,
                 Path = Options.CookiePath ?? (OriginalPathBase.HasValue ? OriginalPathBase.ToString() : "/"),
             };
+
             if (Options.CookieSecure == CookieSecurePolicy.SameAsRequest)
             {
                 cookieOptions.Secure = Request.IsHttps;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -188,7 +188,7 @@ private CookieOptions BuildCookieOptions()
                         new ClaimsIdentity(
                             new[] { new Claim(SessionIdClaim, _sessionKey, ClaimValueTypes.String, Options.ClaimsIssuer) },
                             Options.AuthenticationScheme));
-                    ticket = new AuthenticationTicket(principal, null, Options.AuthenticationScheme);
+                    ticket = new AuthenticationTicket(principal, ticket.Properties, Options.AuthenticationScheme);
```
Tratcher: we need to be more selective about the properties we keep in the cookie. only take what you need.

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -172,13 +174,22 @@ private CookieOptions BuildCookieOptions()
...
+                    properties.IssuedUtc = _refreshIssuedUtc;
                 }
+
                 if (_refreshExpiresUtc.HasValue)
                 {
-                    ticket.Properties.ExpiresUtc = _refreshExpiresUtc;
+                    properties.ExpiresUtc = _refreshExpiresUtc;
+                }
+
+                var cookieOptions = BuildCookieOptions();
```
Tratcher: you can move this back, only the properties change is needed.

troydai:
/cc @Tratcher updated.

As the PR is approved. I'll merge once CI passed.

PR #996, opened 2016-09-28
Fix #976: Break claims value in array into multiple claims

### LINKED ISSUE #976: OIDC handler bug in user info response handling for multiple claims of same type
Imagine user info endpoint returns this:

```
{"sub":"88421113","email":"BobSmith@email.com","email_verified":true,"role":["role1","role2"]}
```

This line: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L788

Adds the role claim with this as the value: `[ "role1", "role2" ]`. 

It should break the array into multiple claims. Here's an extension method that already does this: https://github.com/IdentityModel/IdentityModel2/blob/dev/src/IdentityModel/Client/JObjectExtensions.cs#L12

troydai:
#976

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Utility/ClaimsHelper.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Utility/ClaimsHelper.cs
@@ -31,6 +31,6 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
         }
 
         private static void AddClaimsToIdentity(JToken item, ClaimsIdentity identity, string key, string issuer)
-            => identity.AddClaim(new Claim(key, item.ToString(), ClaimValueTypes.String, issuer));
+            => identity.AddClaim(new Claim(key, item?.ToString() ?? string.Empty, ClaimValueTypes.String, issuer));
     }
 }
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Utility/ClaimsHelper.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Utility/ClaimsHelper.cs
@@ -1,0 +1,36 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Security.Claims;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
+{
+    internal static class ClaimsHelper
+    {
+        public static void AddClaimsToIdentity(
+            JObject userInformationPayload,
+            ClaimsIdentity identity,
+            string issuer)
+        {
+            foreach (var pair in userInformationPayload)
+            {
+                var array = pair.Value as JArray;
+                if (array != null)
+                {
+                    foreach (var item in array)
+                    {
+                        AddClaimsToIdentity(item, identity, pair.Key, issuer);
+                    }
+                }
+                else
+                {
+                    AddClaimsToIdentity(pair.Value, identity, pair.Key, issuer);
+                }
+            }
+        }
+
+        private static void AddClaimsToIdentity(JToken item, ClaimsIdentity identity, string key, string issuer)
+            => identity.AddClaim(new Claim(key, item.ToString(), ClaimValueTypes.String, issuer));
+    }
+}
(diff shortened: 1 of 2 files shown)


PinpointTownes:
@troydai curious: have you considered using a smarter approach like the one used by Wilson? It has a few benefits: it "preserves" the claim value type and uses a specific value type when it can't flatten the JSON payload.

troydai:
Yes.

Wilson's approach looks comprehensive, but I haven't fully evaluated it yet. The reason why I hesitate to adopt here is that the change to be brought in will be much larger than it is needed to address the target issue which for now involves array only. I tend to keep the change smaller at this stage. 

In addition, I'm wondering if completely change how we break down the claims may break customers, please let me know if it something from your perspective worth worries about.

Tratcher:
(linked for my convenience) https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/master/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs#L307

@PinpointTownes do you have scenarios for any of the data types covered by wilson right now? They look very generic and it's not clear how they map back to JSON structures. E.g. `IEnumberable<object>` or `IDictionary<string, object>`

Tratcher:
Or were you referring to this part? https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/master/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs#L392-L399

PinpointTownes:
> In addition, I'm wondering if completely change how we break down the claims may break customers, please let me know if it something from your perspective worth worries about.

Well, this change itself is already a breaking change for people who rely on the fact the OIDC middleware doesn't properly format "array claims". Preserving the claim value type wouldn't break them harder :smile: 

@Tratcher Wilson's approach is a bit complex because `JwtPayload` is a generic dictionary that accepts both primitives types (like `string`, `int` or enumerations of primitive types) and JSON.NET types (i.e types derived from `JToken`). I suppose this approach was adopted for performance, to avoid instantiating `JToken` objects when you only want to use primitive types.

Here, we always deal with `JObject`, so the related code will be much simpler. The code snippet you mentioned in your last post + https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/master/src/System.IdentityModel.Tokens.Jwt/JwtPayload.cs#L547-L582 should be enough to preserve the claim value type and to assign a JSON-specific type when we can't flatten it.

troydai:
How about address the array claims issue in the original issue first and then follow up a more comprehensive solution to the overall situation? I believe we need a little bit more time to evaluate and test the existing solution.

Tratcher:
@PinpointTownes it looks like those are additional issues beyond flattening arrays to multiple claims. I'd want to see a representative user-info dataset that took advantage of these different structures.

PinpointTownes:
@Tratcher I'm not suggesting updating the code used to flatten the claims. I'd just like the claim value types to be preserved (which is completely orthogonal).

Tratcher:
Yes, file a different bug and we'll look into it.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Utility/ClaimsHelper.cs
```diff
@@ -0,0 +1,36 @@
...
+                }
+                else
+                {
+                    AddClaimsToIdentity(pair.Value, identity, pair.Key, issuer);
+                }
+            }
+        }
+
+        private static void AddClaimsToIdentity(JToken item, ClaimsIdentity identity, string key, string issuer)
+            => identity.AddClaim(new Claim(key, item.ToString(), ClaimValueTypes.String, issuer));
```
PinpointTownes: FWIW, the existing code is most likely invalid, as `Claim`'s constructor always throws an exception for null values (but not for empty values).

PinpointTownes:
Done: https://github.com/aspnet/Security/issues/998

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Utility/ClaimsHelper.cs
troydai: Addressed in next commit. If the value is null, turns it into string.Empty.

troydai:
The PR is approved, I'll merge once the CI passed.

Issue #998, opened 2016-09-29
Update OpenIdConnectHandler.GetUserInformationAsync to preserve the claim value types


PinpointTownes:
https://github.com/aspnet/Security/pull/996#issuecomment-250236617

Eilon:
No work planned for this, but a good idea to consider.

PinpointTownes:
Open to contributions? :smile:

Tratcher:
Let us know what you come up with.

PR #1002, opened 2016-10-07
Update ClaimsHelper.AddClaimsToIdentity to infer the claim value type from the JSON token type

PinpointTownes:
Fixes https://github.com/aspnet/Security/issues/998.

image

Note: the userinfo mechanism has currently no unit test.
To ensure the changes added by this PR work as intended, I used ASOS beta7. To test it yourself, you can use this snippet:

```csharp
app.UseOpenIdConnectAuthentication(new OpenIdConnectOptions
{
    Authority = "http://localhost:3304/",
    ClientId = "client_id",
    ResponseType = "id_token token",
    GetClaimsFromUserInfoEndpoint = true,
    RequireHttpsMetadata = false
});
```

```csharp
app.UseOpenIdConnectServer(options =>
{
    options.AllowInsecureHttp = true;
    options.SigningCredentials.AddEphemeralKey();

    options.AuthorizationEndpointPath = "/connect/authorize";
    options.UserinfoEndpointPath = "/connect/userinfo";

    options.Provider.OnValidateAuthorizationRequest = context =>
    {
        context.Validate();

        return Task.CompletedTask;
    };

    options.Provider.OnHandleAuthorizationRequest = context =>
    {
        var identity = new ClaimsIdentity(context.Options.AuthenticationScheme);
        identity.AddClaim(new Claim(ClaimTypes.NameIdentifier, "Bob le Bricoleur"));

        context.Validate(new ClaimsPrincipal(identity));

        return Task.CompletedTask;
    };

    options.Provider.OnHandleUserinfoRequest = context =>
    {
        context.Claims["userinfo_array_claim"] = JArray.FromObject(new[] {
            new int[] { 1, 2, 3 },
            new int[] { 4, 5, 6 },
            new int[] { 7, 8, 9 }
        });
        context.Claims["userinfo_bool_claim"] = true;
        context.Claims["userinfo_date_claim"] = DateTimeOffset.Now;
        context.Claims["userinfo_float_claim"] = double.MaxValue;
        context.Claims["userinfo_guid_claim"] = Guid.NewGuid();
        context.Claims["userinfo_int_claim"] = int.MaxValue;
        context.Claims["userinfo_long_claim"] = long.MaxValue;
        context.Claims["userinfo_object_claim"] = JObject.FromObject(new { Name = "value" });
        context.Claims["userinfo_string_claim"] = "string";

        return Task.CompletedTask;
    };
});
```

/cc @Eilon @Tratcher @troydai

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Utility/ClaimsHelper.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Utility/ClaimsHelper.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.IdentityModel.Tokens.Jwt;
 using System.Security.Claims;
 using Newtonsoft.Json.Linq;
 
@@ -31,6 +32,50 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
         }
 
         private static void AddClaimsToIdentity(JToken item, ClaimsIdentity identity, string key, string issuer)
-            => identity.AddClaim(new Claim(key, item?.ToString() ?? string.Empty, ClaimValueTypes.String, issuer));
+            => identity.AddClaim(new Claim(key, item?.ToString() ?? string.Empty, GetClaimValueType(item), issuer));
+
+        private static string GetClaimValueType(JToken token)
+        {
+            if (token == null)
+            {
+                return JsonClaimValueTypes.JsonNull;
+            }
+
+            switch (token.Type)
+            {
+                case JTokenType.Array:
+                    return JsonClaimValueTypes.JsonArray;
+
+                case JTokenType.Boolean:
+                    return ClaimValueTypes.Boolean;
+
+                case JTokenType.Date:
+                    return ClaimValueTypes.DateTime;
+
+                case JTokenType.Float:
+                    return ClaimValueTypes.Double;
+
+                case JTokenType.Integer:
+                {
+                    var value = (long) token;
+                    if (value >= int.MinValue && value <= int.MaxValue)
+                    {
+                        return ClaimValueTypes.Integer;
+                    }
+
+                    return ClaimValueTypes.Integer64;
+                }
+
+                case JTokenType.Object:
+                    return JsonClaimValueTypes.Json;
+
+                case JTokenType.String:
+                    return ClaimValueTypes.String;
+            }
+
+            // Fall back to ClaimValueTypes.String when no appropriate
+            // claim value type can be inferred from the claim value.
+            return ClaimValueTypes.String;
+        }
     }
 }


troydai:
The PR itself looks fine. Change is simple.

I do have some points to make, and I'd like to gather more advices to help us decide if this is the best approach.

I think we'd better to strike a balance between over-process and under-process the claims' value types. We're definitely under-process today therefore application needs to do additional work to parse the claims' value correctly. However, over-process isn't necessarily helpful. It will definitely spends more resources. For example, like in the PR, do we really need to determine if a integer is `Integer` or `Integer64`? It occurs to me that application will be perfectly capable of doing it.

Are there use cases out there the JsonObject and JsonArray value types are helpful?

Tratcher:
I'd still be curios to see app code that consumes these types and how useful the different identifiers are in practice.

PinpointTownes:
> However, over-process isn't necessarily helpful. It will definitely spends more resources.

IMHO, the perf' impact should be really marginal (and it should be an allocation-free path). That said, that's something we can measure if you really want to be sure.

> Are there use cases out there the JsonObject and JsonArray value types are helpful?

> I'd still be curios to see app code that consumes these types and how useful the different identifiers are in practice.

Here's a very concrete use case, reported a while ago by an ASOS user who uses it as a proxy between another OIDC server and his client app (i.e final authentication is delegated to a third party server, but the main server is ASOS):

  - When the authorization server app - that uses the OIDC client middleware - redirects the user to the third party server and gets back an access token it uses to make a userinfo request, the claims are properly extracted from the JSON payload but the real value types are irremediably lost.

  - If you use the `ClaimsPrincipal` generated by the OIDC middleware as-is with ASOS, it will ultimately call IdentityModel with the wrong claim value types.

  - IM will serialize the JSON payload using the claim value types, so integers, floats, JSON arrays and objects will be serialized as strings.

With this PR, the claim value types are preserved, which gives a chance to IM to serialize claims correctly. Of course, it's something you can work around in your own code (by trying to parse the claim values as suggested by @troydai, but it's painful, time-consuming and not bullet-proof).

PinpointTownes:
> I'd still be curios to see app code

I can't share code under NDA, but I can recreate a sample from scratch if you need to see some code :smile:

troydai:
Ok, this is a good use case. In this sense I think the change has its value.

I have some suggestions developed on top of it:

This is essentially a json claims round-trip. Therefore the logic of JSON-to-Claims should be consistent with Claims-to-JSON. But using the same code here and IdentityModel it does guarantee that, but just for now. However in case IdentityModel changes their logic, the round-trip will be broken without an update on our site. So better solutions could be:

1. Make the JSON-to-Claim logic in IdentityModel public and adopt here through reference.
2. Open up the claim help allows customizing. 

If choose to stay with current form, we can use some test cases.

PinpointTownes:
> Make the JSON-to-Claim logic in IdentityModel public and adopt here through reference.

/cc @brentschmaltz

PinpointTownes:
@brentschmaltz ding dong :watch:

PinpointTownes:
@troydai looks like it will be hard to involve the IM folks :smile: 

Maybe it would be simpler to merge this PR so it's included in 1.1.0 and revisit how claims are handled in a future version?

troydai:
The change looks good to me. @Tratcher @muratg ?

muratg:
👍 though I'll let @Tratcher to take a look as well.

troydai:
Approved and merged. Thanks @PinpointTownes

PinpointTownes:
Thanks for merging it :smile:

Issue #884, opened 2016-06-15
Can't perform custom error handling using OpenIdConnect OnAuthenticationFailed event 


dstrockis:
I couldn’t figure out how to get the OnAuthenticationFailed event to work how I wanted it to.  My goal was to catch any exceptions that occurred during the auth pipeline & redirect to my own custom error page with information about the exception – the code is here: https://github.com/Azure-Samples/active-directory-dotnet-webapp-openidconnect-aspnetcore/blob/master/WebApp-OpenIDConnect-DotNet/Startup.cs#L78-83

When I call `HandleResponse()` in this event, the OIDC handler still tries to succeed the authentication flow, and subsequently throws an exception when it discovers the auth ticket is null (because some exception occurred during token validation).  It succeeds because of the call on this line: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L637

I believe if the OnAuthenticationFailed event is invoked, the auth flow should not succeed.

Tratcher:
HandleResponse isn't working in OnAuthenticationFailed. Use OnRemoteFailure instead. I don't think we need OnAuthenticationFailed anymore, it seems redundant with OnRemoteFailure.

Tratcher:
FYI: JwtBearer has a similar issue with OnAuthenticationFailed and I don't see an easy workaround there.

Tratcher:
@HaoK

HaoK:
Yeah we should just remove the OIDC specific OnAuthenticationFailed since we have OnRemoteFailure now...

Eilon:
We should add the newer/better event to the JWT flavor as well.

brentschmaltz:
OnAuthenticationFailed was meant to indicate that we couldn't authenticate the user. I don't think OnRemoteFailure has the same semantic. This is also a change from Katana.

Tratcher:
The other event does not apply to JWTBearer, it's from the base class for the redirecting middleware.

Tratcher:
OnRemoteFailure has the same parameters and conditions and executes immediately after OnAuthenticationFailed. The only difference is the name.

Tratcher:
Since this was not removed for RTM, we'll have to keep it and see if there's some way to fix it.

brentschmaltz:
OnAuthenticationFailed makes sense for OIDC, but maybe not for JwtBearer as the webapi call is not technically about Authentication but Authorization.

PR #1008, opened 2016-10-17
Honor OIDC's OnAuthenticationFailed HandleResponse()

Tratcher:
# 884 @HaoK this fills in a gap in the event flow. What do you think?

I still need to try it with other events, other middleware, and add tests.

### DIFF
--- a/samples/JwtBearerSample/Startup.cs
+++ b/samples/JwtBearerSample/Startup.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using Microsoft.AspNetCore.Authentication.JwtBearer;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
@@ -15,6 +16,8 @@ namespace JwtBearerSample
     {
         public Startup(IHostingEnvironment env)
         {
+            Environment = env;
+
             var builder = new ConfigurationBuilder()
                 .SetBasePath(env.ContentRootPath);
 
@@ -29,6 +32,8 @@ namespace JwtBearerSample
         }
 
         public IConfiguration Configuration { get; set; }
+
+        public IHostingEnvironment Environment { get; set; }
 
         // Shared between users in memory
         public IList<Todo> Todos { get; } = new List<Todo>();
@@ -68,7 +73,23 @@ namespace JwtBearerSample
             {
                 // You also need to update /wwwroot/app/scripts/app.js
                 Authority = Configuration["jwt:authority"],
-                Audience = Configuration["jwt:audience"]
+                Audience = Configuration["jwt:audience"],
+                Events = new JwtBearerEvents()
+                {
+                    OnAuthenticationFailed = c =>
+                    {
+                        c.HandleResponse();
+
+                        c.Response.StatusCode = 500;
+                        c.Response.ContentType = "text/plain";
+                        if (Environment.IsDevelopment())
+                        {
+                            // Debug only, in production do not share exceptions with the remote host.
+                            return c.Response.WriteAsync(c.Exception.ToString());
+                        }
+                        return c.Response.WriteAsync("An error occurred processing your authentication.");
+                    }
+                }
             });
 
             // [Authorize] would usually handle this
--- a/samples/OpenIdConnectSample/Startup.cs
+++ b/samples/OpenIdConnectSample/Startup.cs
@@ -20,6 +20,8 @@ namespace OpenIdConnectSample
     {
         public Startup(IHostingEnvironment env)
         {
+            Environment = env;
+
             var builder = new ConfigurationBuilder()
                 .SetBasePath(env.ContentRootPath);
 
@@ -34,6 +36,8 @@ namespace OpenIdConnectSample
         }
 
         public IConfiguration Configuration { get; set; }
+
+        public IHostingEnvironment Environment { get; set; }
 
         public void ConfigureServices(IServiceCollection services)
         {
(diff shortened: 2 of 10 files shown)


HaoK:
Yes this looks like a much better :)

Tratcher:
Updated with a similar approach to make JwtBearer work

Tratcher:
@HaoK I added a lot of JWT and OIDC tests. I found two oddities beyond the original bug.
https://github.com/aspnet/Security/pull/1008/files#diff-89f4ccfea34a2fdaef439f48cf2d2547R195
1. Some of the events did not include the current AuthTicket, I added it.
2. Some of the events when Handled still result in the TicketReceived event being invoked and the cookie set. I think this original code (https://github.com/aspnet/Security/pull/1008/files#diff-a771d2d9b96202c4add713e120e832dbL54) where Handled events returned a ticket was wrong, Handled was documented to say the caller generated a complete response, not that they generated a Ticket and short circuited some of the validation logic. I worry a little that changing it will break former users of Handled (if any), and also that fixing #1 above will cause the same thing to start happening for other events. Of the two I think it'd be better to break Handled behavior to discard the Ticket as it didn't work in many scenarios anyways (hence the original bug).

Thoughts?

Tratcher:
@HaoK Updated.

PinpointTownes:
@Tratcher curious: how urgent is this? I'd like to test it with the aspnet-contrib stuff to make sure the new `AuthenticateResult.Handle()` semantic doesn't break anything.

Tratcher:
@PinpointTownes how long do you need? I'd like to get it in this week.

PinpointTownes:
I'll give a try tomorrow morning.

Do you know if a recent `Microsoft.NETCore.App` package is needed to play with the ASP.NET Core 1.1.0 bits?

Tratcher:
You likely need preview1

PinpointTownes:
@Tratcher FYI, I've run all the aspnet-contrib tests and everything is green (the OAuth2 introspection/validation middleware also use `HandleResponse()` as a way to bypass the default validation logic, but they return `AuthenticateResult.Skip()`, so the behavior is not changed in practice).

> I worry a little that changing it will break former users of Handled (if any), and also that fixing #1 above will cause the same thing to start happening for other events. Of the two I think it'd be better to break Handled behavior to discard the Ticket as it didn't work in many scenarios anyways (hence the original bug).

It's indeed an important change but I agree it makes more sense than leaving it as-is. If this change is approved, I'll rework the aspnet-contrib middleware to match the new logic used by the built-in middleware.

PinpointTownes:
Note that while I think this change is positive, I'm a bit concerned about not being able to bypass the default token validation logic without handling the complete response. You might want to add some `context.Ticket != null` checks to restore the `context.HandleResponse()` behavior.

PinpointTownes:
It would also be worth adding a tiny comment somewhere to indicate that `AuthenticateResult.Handle()` doesn't really stop processing the request when the authentication middleware is invoked via `AuthenticationManager.SignInAsync`, instead of as part of automatic authentication.

Tratcher:
The trouble is that the Ticket is almost always not null, so that's not a good indicator of what HandleResponse should do. Have you actually seen anyone provide a Ticket and skip the remaining validation? If so was it for Jwt or OAuth?

The only additional behavior that HandleResponse would skip if there were a Ticket is the last section of HandleRemoteCallbackAsync: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs#L83-L115

If we really want to preserve the ability to skip from the middle of the validation logic all the way to HandleRemoteCallbackAsync then we may need a third state: SkipToNextMiddleware, SkipRemainingValidation, and HandledResponse.

PinpointTownes:
> The trouble is that the Ticket is almost always not null, so that's not a good indicator of what HandleResponse should do. Have you actually seen anyone provide a Ticket and skip the remaining validation? If so was it for Jwt or OAuth?

Yep, but only once or twice, so I suppose it's not that frequent. IIRC, it was from `OAuthValidationEvents.RetrieveToken`, the equivalent of `JwtBearerEvents.MessageReceived` in the contrib world. The cool thing with this event is that there's no ticket yet, so we can easily add a null check.

The real question is more about the intermediate events (e.g `OpenIdConnectEvents.TokenResponseReceived` or `OpenIdConnectEvents.TokenValidated`), that are called at the middle of the authentication process. FWIW, I've personally never used `HandleResponse` from these events (which doesn't mean no one else is relying on this behavior :smile:)

PinpointTownes:
Random idea: wouldn't a reference equality check be an interesting alternative to `HandleResponse` for this scenario?

If the ticket was replaced, stop the validation process and use it as-is.
If it was not replaced, continue the validation process.

Tratcher:
A ticket replacement check seems pretty weak.

I went back and looked at how this worked in Katana. A Handled event would only ever return an empty ticket. This was changed when OIDC was first moved to Core.
http://katanaproject.codeplex.com/SourceControl/latest#src/Microsoft.Owin.Security.OpenIdConnect/OpenidConnectAuthenticationHandler.cs
https://github.com/aspnet/Security/commit/49e66f03110b486f8dd9959dd2cee802d676c28f#diff-d9d4f83dd99a371877f472e7c4611b29R87

Interestingly the comment for Ticket has said the same thing the whole time: `Gets or set the <see cref="Ticket"/> to return if this event signals it handled the event.` I overlooked the comment on HandleResponse that said `Set the <see cref="Ticket"/> to trigger SignIn.`. So it seems the intended behavior was for people to null out Ticket if they didn't want SignIn to be called (like the last iteration of my tests did). That seems really strange.

Tratcher:
Hmm, logical elimination: nobody could call HandleResponse in AuthenticationFailed or TokenResponseReceived without specifying a ticket because it would have thrown before (hence #884). All the other events already had a ticket so if they called HandleResponse then that would have resulted in SignIn. I will add the ticket to TokenResponseReceived for consistency with the other events, but not to AuthenticationFailed because it would be bad for a failed auth to accidentally sign-in.  Then I'll allow ticket to be nulled out if you didn't want to call sing-in. This is the least likely to break anyone this release, and fixes the main bug in AuthFailed.

HaoK:
That sounds reasonable, and at least you've added more test coverage for this behavior now :)

Tratcher:
Updated / reverted

Tratcher:
https://github.com/aspnet/Security/issues/772

PinpointTownes:
The fact `HandleResponse()` doesn't have the same behavior when `Ticket` is null is frankly weird, but I guess it's one of those "better than nothing" compromises :shipit:

Issue #863, opened 2016-06-02
Enable IClaimsTransformer to be specified in DI instead of only options


jburbach:
I'm not sure whether to post this here or in the EF repo, but I figured I'd start here.

ASP.NET Core RC2 WebAPI/SPA using windows auth, and I'm attempting to do a very basic claims transformation by looking the ID up in my database with my DBContext and adding claims I find there to the windows principal. I am using a repository pattern with EF, all registered as Scoped lifetime, the dbcontext should as well by default, so all methods are sharing the same dbcontext on each request. I get a lot of random EF errors on my _users.FindByUsername() method that say things like 

_Invalid attempt to call ReadAsync when reader is closed.
InvalidOperationException: BeginExecuteReader requires an open and available Connection. The connection&#x27;s current state is open.
InvalidOperationException: The connection was not closed. The connection&#x27;s current state is connecting._

I also get this error at startup, but then can refresh the page and get my SPA like I should:

_InvalidOperationException: An attempt was made to use the context while it is being configured. A DbContext instance cannot be used inside OnConfiguring since it is still being configured at this point._

I think this is related to the transform running async and competing with other threads trying to access the db at the same time? I get these errors randomly under load, nothing consistent. If I take the DB lookup out of my claimstransform, zero errors. I tried making my user lookup async and awaiting it, which lessens the amount of errors but doesn't completely solve the issue.

My question is, is what I'm doing not recommended? Should I not be trying to use EF with a ClaimsTransform, or should I have a separate context just for this user lookup? Do I need to scope my dbcontext/repositories differently in DI? I'm curious what the envisioned recommended pattern is for something like this, and if this is possibly an EF Core issue that should work but doesn't.

Apologies if there is something obvious here I'm missing, I've looked through all the documentation I can find for best practices and didn't find anything that recommends something different.

```
using System;
using System.Security.Claims;
using System.Threading.Tasks;
using MyApp.Models.Repositories.Interfaces;
using Microsoft.AspNetCore.Authentication;

namespace MyApp.Authorization
{
    public class ClaimsTransformer : IClaimsTransformer
    {
        private readonly IUserRepository _users;

        public ClaimsTransformer(IUserRepository users)
        {
            _users = users;           
        }

        public async Task<ClaimsPrincipal> TransformAsync(ClaimsTransformationContext context)
        {
            System.Security.Principal.WindowsIdentity windowsIdentity = null;

            foreach (var i in context.Principal.Identities)
            {
                //windows token
[...]

PR #1014, opened 2016-10-24
Support IClaimsTransformer from DI

HaoK:
Fixes https://github.com/aspnet/Security/issues/863
- Added TransformType to ClaimsTransformationOptions, Type will be used via GetService if no instance is specified.

cc @blowdart @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication/ClaimsTransformationMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Authentication/ClaimsTransformationMiddleware.cs
@@ -5,6 +5,7 @@ using System;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.Authentication
@@ -38,7 +39,7 @@ namespace Microsoft.AspNetCore.Authentication
             var transform = Options.Transformer;
             if (transform == null && Options.TransformerType != null)
             {
-                transform = context.RequestServices.GetService(Options.TransformerType) as IClaimsTransformer;
+                transform = context.RequestServices.GetRequiredService(Options.TransformerType) as IClaimsTransformer;
             }
             var handler = new ClaimsTransformationHandler(transform, context);
             handler.RegisterAuthenticationHandler(context.GetAuthentication());
--- a/src/Microsoft.AspNetCore.Authentication/ClaimsTransformationMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Authentication/ClaimsTransformationMiddleware.cs
@@ -9,42 +9,48 @@ using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.Authentication
 {
-    public class ClaimsTransformationMiddleware<TClaimsTransformer> where TClaimsTransformer : IClaimsTransformer
+    public class ClaimsTransformationMiddleware
     {
         private readonly RequestDelegate _next;
-        private readonly IClaimsTransformer _transform;
 
         public ClaimsTransformationMiddleware(
             RequestDelegate next,
-            TClaimsTransformer transformer)
+            IOptions<ClaimsTransformationOptions> options)
         {
             if (next == null)
             {
                 throw new ArgumentNullException(nameof(next));
             }
 
-            if (transformer == null)
+            if (options == null)
             {
-                throw new ArgumentNullException(nameof(transformer));
+                throw new ArgumentNullException(nameof(options));
             }
 
-            _transform = transformer;
+            Options = options.Value;
             _next = next;
         }
 
+        public ClaimsTransformationOptions Options { get; set; }
+
         public async Task Invoke(HttpContext context)
         {
-            var handler = new ClaimsTransformationHandler(_transform, context);
+            var handler = new ClaimsTransformationHandler(Options.Transformer, context);
             handler.RegisterAuthenticationHandler(context.GetAuthentication());
             try
             {
-                if (_transform != null)
+                var transform = Options.Transformer;
+                if (transform == null && Options.TransformerType != null)
+                {
+                    transform = context.RequestServices.GetService(Options.TransformerType) as IClaimsTransformer;
+                }
+                if (transform != null)
                 {
                     var transformationContext = new ClaimsTransformationContext(context)
                     {
                         Principal = context.User
                     };
-                    context.User = await _transform.TransformAsync(transformationContext);
+                    context.User = await transform.TransformAsync(transformationContext);
                 }
                 await _next(context);
             }
@@ -53,15 +59,5 @@ namespace Microsoft.AspNetCore.Authentication
                 handler.UnregisterAuthenticationHandler(context.GetAuthentication());
             }
         }
-    }
-
-    public class ClaimsTransformationMiddleware : ClaimsTransformationMiddleware<IClaimsTransformer>
-    {
-        public ClaimsTransformationMiddleware(
-            RequestDelegate next,
-            IOptions<ClaimsTransformationOptions> options) : base(next, options.Value.Transformer)
-        { }
-
-        public ClaimsTransformationOptions Options { get; set; }
     }
 }
--- a/src/Microsoft.AspNetCore.Authentication/ClaimsTransformationMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Authentication/ClaimsTransformationMiddleware.cs
@@ -9,43 +9,42 @@ using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.Authentication
 {
-    public class ClaimsTransformationMiddleware
+    public class ClaimsTransformationMiddleware<TClaimsTransformer> where TClaimsTransformer : IClaimsTransformer
     {
         private readonly RequestDelegate _next;
+        private readonly IClaimsTransformer _transform;
 
         public ClaimsTransformationMiddleware(
             RequestDelegate next,
-            IOptions<ClaimsTransformationOptions> options)
+            TClaimsTransformer transformer)
         {
             if (next == null)
             {
                 throw new ArgumentNullException(nameof(next));
             }
 
-            if (options == null)
+            if (transformer == null)
             {
-                throw new ArgumentNullException(nameof(options));
+                throw new ArgumentNullException(nameof(transformer));
             }
 
-            Options = options.Value;
+            _transform = transformer;
             _next = next;
         }
 
-        public ClaimsTransformationOptions Options { get; set; }
-
         public async Task Invoke(HttpContext context)
         {
-            var handler = new ClaimsTransformationHandler(Options.Transformer, context);
+            var handler = new ClaimsTransformationHandler(_transform, context);
             handler.RegisterAuthenticationHandler(context.GetAuthentication());
             try
             {
-                if (Options.Transformer != null)
+                if (_transform != null)
                 {
                     var transformationContext = new ClaimsTransformationContext(context)
                     {
                         Principal = context.User
                     };
-                    context.User = await Options.Transformer.TransformAsync(transformationContext);
+                    context.User = await _transform.TransformAsync(transformationContext);
                 }
                 await _next(context);
             }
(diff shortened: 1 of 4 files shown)


HaoK:
@PinpointTownes thoughts?

PinpointTownes:
> @PinpointTownes thoughts?

Using a generic argument is an interesting approach (that could be generalized to other "services options" like the events classes).

Unfortunately, your PR has **exactly** the same limitation as the current code: you can't use scoped transformers, because the `_claimsTransformer` field is bound to the lifetime of the middleware itself (i.e singleton by definition). This means no transformer that takes a `DbContext` parameter... :sweat: 

To fix that, you'd have to move the `IClaimsTransformer` resolution to `InvokeAsync`.

HaoK:
Bleh so basically only way to fix that is to do service location, I guess we could do something like a TransformerType, if the Transform instance is null, we look at the type, and get it from DI inside of invoke.

PinpointTownes:
@HaoK yup. That's pretty much what @ilmax and I had in mind: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/issues/275#issue-156060959

1. Is `TransformerType` non-null? Use it to create a transformer using `context.RequestServices` (done from `InvokeAsync`).
2. Is `Transformer` non-null? Use it as-is.
3. Throw if both are null.

HaoK:
Why do you think the Type should take precedence over the instance?  Are we expecting both to be set at the same time?  

Today it looks like ClaimsTransformation is just a no-op if no transform is specified, I could see an argument for throwing as well, but its probably too late to make that change now

HaoK:
Updated, back to the old middleware, just with the tweak where TransformType will be used to find the service from the container if no Transformer was specified.  Fixed the test to verify a scoped transformer

PinpointTownes:
> Why do you think the Type should take precedence over the instance? Are we expecting both to be set at the same time?

Ideally, this pattern should be generalized and applied to the events classes. Yet, all the `Events` properties [are always initialized with a default value](https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs#L55), so if you check the `Events` property first, it will always be non-null. For consistency, I'd simply apply the same reasoning to the claims transformation middleware.

HaoK:
The problem with DI taking precedence is a major issue when you can have multiple instances of the middleware in the app.  This is fine if instance wins as you can always override anything in DI, but with precedence flipped, every middleware would be forced to use the same event from DI.  Events are nullable today, so if we wanted to use this pattern in the Auth middleware, the pattern could just be to set the Events to null to switch to the Events found in the container.

PinpointTownes:
> The problem with DI taking precedence is a major issue when you can have multiple instances of the middleware in the app.

I'm not sure to understand how it's a problem: for the transformer to be resolved from the DI container, one would have to set `TransformerType` explicitly.

https://github.com/aspnet/Security/issues/863#issuecomment-254304602

HaoK:
There's no problem for claims transformation, I was talking about the general pattern as applied to the Auth middlewares, where different middleware want to use the same type of event, just configured differently.  If DI always wins, there would be no way to accomplish this other than creating a different type per instance which is pretty meh.

PinpointTownes:
> If DI always wins, there would be no way to accomplish this other than creating a different type per instance which is pretty meh.

But if you leave `EventsType` to `null`, the `Events` property will be used as-is, right? It should behave exactly the same way as currently.

Or do you mean when both properties are specified?

PinpointTownes:
The downside of your approach is that it's not super intuitive. Something like that won't work:

```
app.UseJwtBearerAuthentication(new JwtBearerOptions {
    EventsType = typeof(MyUberCoolEventsClass)
});
```

One will have to nullify the `Events` property for this to work properly:

```
app.UseJwtBearerAuthentication(new JwtBearerOptions {
    Events = null,
    EventsType = typeof(MyUberCoolEventsClass)
});
```

src/Microsoft.AspNetCore.Authentication/ClaimsTransformationMiddleware.cs
```diff
@@ -39,13 +39,18 @@ public class ClaimsTransformationMiddleware
             handler.RegisterAuthenticationHandler(context.GetAuthentication());
             try
             {
-                if (Options.Transformer != null)
+                var transform = Options.Transformer;
```
PinpointTownes: The transformer is initialized too late: it must be created before line 38 is invoked, or `ClaimsTransformationHandler` won't work properly.

src/Microsoft.AspNetCore.Authentication/ClaimsTransformationMiddleware.cs
```diff
@@ -39,13 +39,18 @@ public class ClaimsTransformationMiddleware
...
+                if (transform == null && Options.TransformerType != null)
+                {
+                    transform = context.RequestServices.GetService(Options.TransformerType) as IClaimsTransformer;
```
PinpointTownes: Any particular reason not to use `GetRequiredService`? Using `GetService` + an as cast will "fail" silently and won't help one understand what's wrong with the DI registration.

HaoK:
Yeah the issue is when both are specified.  I agree that setting it to null is a bit hard to discover, but that can be addressed via doc comments at least where it can specify the resolution order (Events instance then DI container if null).  

But in general I think it feels more natural for the instance to win if specified (especially for claims transformation which has no default).  We can revisit this if/when we try to do something similar for the Auth middleware

src/Microsoft.AspNetCore.Authentication/ClaimsTransformationMiddleware.cs
HaoK: Its not required today to specify a transform, its perfectly valid to add a no-op claims transformation right now so making it required would be a breaking change for apps...

HaoK:
Updated fixing the passive authentication code path, subtle bit was that you have to actually turn off AutomaticAuthentication since no-opts the passive Authenticate call since it caches the user from the AutomaticAuthenticate call...

src/Microsoft.AspNetCore.Authentication/ClaimsTransformationMiddleware.cs
PinpointTownes: `TransformerType` is a totally new thing, so there's no breaking change. When I explicitly provide a type, I expect to get an exception if things go wrong.

PinpointTownes: > If we add a check that throws if neither transform is specified, that most certainly is a functional breaking change...

Except that's not what I'm suggesting: https://github.com/aspnet/Security/pull/1014#discussion_r84986523 :smile: 

If you `UseClaimsTransformation()` "with nothing else", this code path will never be reached (because `TransformerType` will be null).

PinpointTownes: Eureka :smile:

Alternatively, a warning message when `GetService() == null` could do the job too, if you're not a huge fan of `GetRequiredService`.

HaoK:
@Tratcher how's this look to you?

src/Microsoft.AspNetCore.Authentication/ClaimsTransformationMiddleware.cs
```diff
@@ -35,17 +36,22 @@ public class ClaimsTransformationMiddleware
 
         public async Task Invoke(HttpContext context)
         {
-            var handler = new ClaimsTransformationHandler(Options.Transformer, context);
+            var transform = Options.Transformer;
+            if (transform == null && Options.TransformerType != null)
```
Tratcher: why even specify a type? Why not just resolve IClaimsTransformer from DI?

src/Microsoft.AspNetCore.Authentication/ClaimsTransformationMiddleware.cs
```diff
@@ -35,17 +36,22 @@ public class ClaimsTransformationMiddleware
...
+            {
+                transform = context.RequestServices.GetRequiredService(Options.TransformerType) as IClaimsTransformer;
```
Tratcher: What's the scenario for scoped transformers?

src/Microsoft.AspNetCore.Authentication/ClaimsTransformationMiddleware.cs
Tratcher: As a fallback I mean

src/Microsoft.AspNetCore.Authentication/ClaimsTransformationMiddleware.cs
PinpointTownes: E.g transformers that rely on a `DbContext`. There's a big thread about that here: https://github.com/aspnet/Security/issues/863

src/Microsoft.AspNetCore.Authentication/ClaimsTransformationMiddleware.cs
PinpointTownes: If you "hardcode" the transformer type to `IClaimsTransformer`, one won't be able to use multiple transformers in the same pipeline.

https://github.com/aspnet/Security/issues/863#issuecomment-254304602

Tratcher: How many transformers do you need?

PinpointTownes: One may want to use a transformer for a Web API sub-app (via `app.Map` or `app.UseWhen`) and another one for the rest of the app.

HaoK: Yeah so what we have today is geared towards a single claims transformer instance, this PR makes it a bit easier to have multiple claims transformation middlewares, and also pull the transform from DI (needed basically only for scoped transformers)

src/Microsoft.AspNetCore.Authentication/ClaimsTransformationMiddleware.cs
HaoK: @davidfowl do you have any opposition/alternatives to this approach?

davidfowl: The type on options things is a bit weird (and new). MVC sorta does something similar but I need to think about this pattern a bit more before it spreads like wildfire 😄

Tratcher: Note we did something similar in ResponseCompression, inspired by MVC filters: https://github.com/aspnet/BasicMiddleware/blob/dev/src/Microsoft.AspNetCore.ResponseCompression/CompressionProviderCollection.cs#L35

PinpointTownes:
Any progress here? :sweat_smile:

src/Microsoft.AspNetCore.Authentication/ClaimsTransformationMiddleware.cs
HaoK: 🔥  🎆 🔥

HaoK: @Tratcher @PinpointTownes any last concerns with this PR before I merge?

PinpointTownes: @HaoK personally, I'd really like to see a POC of how it would be generalized to the events classes so we're sure we don't end up with something inconsistent.

HaoK: Um, that's a much bigger fish to fry, but I can spend a few cycles today(since it's moving day) with a prototype PR for cleaning up events in 1.2

Tratcher: Not worth it yet, we need to take the whole event system back to the drawing board.

HaoK: I know, I figure that's worth spending a bit of time thinking about today, since I don't want to check in anything right before the move, I'll play with what things look like if we nuked all the IFooEvents, kept the events types (as required base classes), and also introduce a sibling FooEventType similar to how this works (where we would use the type to GetService (if the instance is null, otherwise new up a FooEvent)

HaoK:
See https://github.com/aspnet/Security/pull/1055 for applying this more generally to all Auth events

dmccaffery:
@PinpointTownes:

In your example for ```JwtBearerOptions``` on  the ```Events``` property, does the following not work?:

```csharp
private IJwtBearerEvents events;

public IJwtBearerEvents Events
{
    get
    {
        return this.EventType == null ? this.events : null;
    }
}
```

Basically, I'm suggesting that the options themselves be smart enough not to expose both an instance and a type for DI at the same time -- and to pick which one wins based on their own specific needs.

HaoK:
Will be rolled into auth 2.0

PR #1019, opened 2016-11-01
Made ChunkingCookieManager's default chunk size public AND Updated Authentication's Base64UrlTextEncoder to use WebUtilities's Base64UrlTextEncoder logic

kichalla:
@Tratcher Couple of minor changes...I plan to keep these as separate commits when merging...


src/Microsoft.AspNetCore.ChunkingCookieManager.Sources/ChunkingCookieManager.cs
```diff
@@ -30,6 +30,11 @@ namespace Microsoft.AspNetCore.Internal
     internal class ChunkingCookieManager
     {
 #endif
+        /// <summary>
+        /// The default maximum size of characters in a cookie to send back to the client.
+        /// </summary>
+        public const int DefaultChunkSize = 4070;
```
kichalla: To enable testing

Issue #903, opened 2016-07-16
Ensuring the generated redirect URL is valid


troydai:
In issue #844, @PinpointTownes wrote:

> > The null ref was in OIDC.
> 
> On a related note, we should add more checks to ensure an authorization request or a logout request cannot be generated when the authorization/logout endpoint path resolved from the OIDC configuration or set from the `RedirectToIdentityProvider` event is null.
> 
> Currently, if the authorization endpoint is undefined, the challenge is applied to the current address (e.g ?response_type=id_token...), which can result in an endless loop. 3 OpenIddict users reported a similar behavior in their own apps.
> 
> You can easily reproduce it by cloning the [OpenIddict MVC samples](https://github.com/openiddict/openiddict-samples/tree/master/samples/Mvc) and updating the server sample to disable the authorization endpoint:
> 
> ```
> services.AddOpenIddict<ApplicationUser, IdentityRole<Guid>, ApplicationDbContext, Guid>()
>     // .EnableAuthorizationEndpoint("/connect/authorize")
>     // .EnableLogoutEndpoint("/connect/logout")
>     .EnableTokenEndpoint("/connect/token")
>     // .EnableUserinfoEndpoint("/connect/userinfo")
> 
>     // .AllowAuthorizationCodeFlow()
>     // .AllowRefreshTokenFlow()
>     .AllowPasswordFlow()
> 
>     .DisableHttpsRequirement();
> ```

Tratcher:
Note there is an existing check for this in HandleUnauthorizedAsync:
```
if (!Uri.IsWellFormedUriString(redirectUri, UriKind.Absolute))
{
    Logger.InvalidAuthenticationRequestUrl(redirectUri);
}
```
Note it only logs. This was because Katana did not apply challenges immediately, it waited until OnSendingHeaders or pipeline-unwind so any exception thrown would not be visible to user code. Core applies challenges immediately when called from user code. While pipeline unwind automatic challenges are still supported, they are not the normal flow. Also, throwing an exception has the advantage of changing this infinite redirect to a 500.

PinpointTownes:
:+1: for throwing an exception (an InvalidOperationException, maybe?)

liquidboy:
thought id add this in here for histories sake ...  
On my project we just upgraded to netcore 1.1.0  and suddenly our SignOut's were failing ...   Traced it to the changes implemented here around checking if IssuerAddress was null or empty ...   

https://twitter.com/josefajardo/status/819329380421804032

[...]

PR #1020, opened 2016-11-01
Ensure redirect uris can be generated

Tratcher:
#903 @HaoK @brentschmaltz @PinpointTownes 
This helps users discover middleware configuration errors that may otherwise result in infinite challenge redirects.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
@@ -208,6 +208,12 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
 
             message.State = Options.StateDataFormat.Protect(properties);
 
+            if (string.IsNullOrEmpty(message.IssuerAddress))
+            {
+                throw new InvalidOperationException(
+                    "Cannot redirect to the end session endpoint, the configuration may be missing or invalid.");
+            }
+
             if (Options.AuthenticationMethod == OpenIdConnectRedirectBehavior.RedirectGet)
             {
                 var redirectUri = message.CreateLogoutRequestUrl();
@@ -355,6 +361,12 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
             properties.Items.Add(OpenIdConnectDefaults.RedirectUriForCodePropertiesKey, message.RedirectUri);
 
             message.State = Options.StateDataFormat.Protect(properties);
+
+            if (string.IsNullOrEmpty(message.IssuerAddress))
+            {
+                throw new InvalidOperationException(
+                    "Cannot redirect to the authorization endpoint, the configuration may be missing or invalid.");
+            }
 
             if (Options.AuthenticationMethod == OpenIdConnectRedirectBehavior.RedirectGet)
             {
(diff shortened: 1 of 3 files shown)


PinpointTownes:
:shipit: tho' I'm not sure to understand why you include `message.IssuerAddress` in the exception message (it's always null or empty) :sweat_smile: 

Or maybe you wanted to combine this new check with `Uri.IsWellFormedUriString`?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -208,6 +208,12 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
 
             message.State = Options.StateDataFormat.Protect(properties);
 
+            if (string.IsNullOrEmpty(message.IssuerAddress))
+            {
+                throw new InvalidOperationException(
+                    $"Cannot redirect to the end session endpoint, the configuration may be missing or invalid: '{message.IssuerAddress}'");
+            }
```
HaoK: Maybe this should just say the message's issuer address is missing instead?

Tratcher: Messages and issuers are generic terms in the object model, if you look up a few lines you will see issuer assigned from an end session endpoint.

Tratcher:
@PinpointTownes woops, that was leftover from when I had tried to replace the IsWellFormedUriString check. Removing.

Tratcher:
I didn't combine this with the IsWellFormedUriString check because I wanted something that worked with FormPost too.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -208,6 +208,12 @@ public OpenIdConnectHandler(HttpClient backchannel, HtmlEncoder htmlEncoder)
 
             message.State = Options.StateDataFormat.Protect(properties);
 
+            if (string.IsNullOrEmpty(message.IssuerAddress))
+            {
+                throw new InvalidOperationException(
+                    $"Cannot redirect to the end session endpoint, the configuration may be missing or invalid.");
```
PinpointTownes: nit: `$` is not necessary as string interpolation is not used here.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -356,6 +362,12 @@ protected virtual Task<bool> HandleSignOutCallbackAsync()
 
             message.State = Options.StateDataFormat.Protect(properties);
 
+            if (string.IsNullOrEmpty(message.IssuerAddress))
+            {
+                throw new InvalidOperationException(
+                    $"Cannot redirect to the authorization endpoint, the configuration may be missing or invalid.");
```
PinpointTownes: Same remark :smile:

PinpointTownes:
> I didn't combine this with the IsWellFormedUriString check because I wanted something that worked with FormPost too.

Curious: is there any particular reason it couldn't be used for both redirections and auto-post forms?

Tratcher:
Just that the URI in the FormPost scenario isn't assembled, it's passed straight through from IssureAddress. Also note that IsWellFormedUriString is extremely strict and would block urls that most clients can handle.

PinpointTownes:
@Tratcher makes sense :ship: :it:

liquidboy:
see my post here ... this change broke our current .netcore app when we upgraded to 1.1  https://github.com/aspnet/Security/issues/903#issuecomment-272037399

PR #1041, opened 2016-11-16
Pin versions to produce a servicing build

pranavkm:


pranavkm:
cc @muratg \ @Tratcher

PR #1065, opened 2016-12-12
[Design] Authentication 2.0

HaoK:
## **TL:DR summary**

- Single `UseAuthentication` is required, which takes care of the old 'automatic' challenge, authenticate, in addition to giving each scheme a chance to handle the request like the old middlewares did.
- 'Automatic' authenticate and challenge are now a single `Default[Authenticate/Challenge/SignIn]Scheme` in `AuthenticationOptions`, rather than a flag on each middleware.  No ambiguity, altho there's one bit of magic where if none are specified and there's only a single scheme registered, we use that scheme as the default.

## Pros of the new design

```C#
   // ConfigureServices:

       _dbContext = context;
    }
    
        {
        }
}
```

## Composing configuration of a scheme's options looks like this (Identity needs this):
```
                    services.AddCookieAuthentication("Cookie1");
                    services.ConfigureSchemeHandler<CookieAuthenticationOptions>("Cookie1",
                        o => o.CookieName = "One");
```

## Claims Transformation

```
     return Task.FromResult(p);
  }
```

## Interop options (Currently exploring Partial interop)
- No interop (nuke all old code, takeover namespaces as necessary entirely)
- Partial interop (focusing on Startup.cs interop, aka make UseXyzAuth do something similar with the new stack.  PR adds a `UseLegacyAuthentication` which the old UseXyzAuth methods would call with the XyzOptions, and this would add a new `AuthenticationScheme` to the new service.
- Full interop (would need to keep the old code, and play some games inside of the IAuthenticationFeature)

## Example: Updated with twitter, google social auth + cookie sign in
```C#
public class Startup
{
     public void ConfigureServices(IServiceCollection services)
     {
             services.AddAuthentication(o =>
             {
                 o.DefaultAuthenticationScheme = CookieAuthenticationDefaults.AuthenticationScheme;
                 o.DefaultChallengeScheme = CookieAuthenticationDefaults.AuthenticationScheme;
                 o.DefaultSignInScheme = CookieAuthenticationDefaults.AuthenticationScheme;
                 o.ClaimsTransform => p => {
                     if (!p.Identities.Any(i => i.AuthenticationType == "Cookies"))
                     {
                        var id = new ClaimsIdentity("xform");
                        id.AddClaim(new Claim("xform", "yup"));
                        p.AddIdentity(id);
                     }
                     return Task.FromResult(p); 
                 }
            });
            services.AddCookieAuthentication();
            services.AddGoogleAuthentication(options =>
            {
                options.ClientId = "667949426586-0bor1qj05d9fjqkvhil6tvoupjfv46fr.apps.googleusercontent.com";
                options.ClientSecret = "tPycZp08PGQBNDz2XycEB145";
                options.SaveTokens = true;
                options.Events = new OAuthEvents()
                {
                    OnRemoteFailure = ctx =>
                    {
                        ctx.Response.Redirect("/error?FailureMessage=" + UrlEncoder.Default.Encode(ctx.Failure.Message));
                        ctx.HandleResponse();
                        return Task.FromResult(0);
                    }
                };
            });
            services.AddTwitterAuthentication(options =>
            {
                options.ConsumerKey = "BSdJJ0CrDuvEhpkchnukXZBUv";
                options.ConsumerSecret = "xKUNuKhsRdHD03eLn67xhPAyE1wFFEndFo1X2UJaK2m1jdAxf4";
                options.RetrieveUserDetails = true;
                options.SaveTokens = true;
                options.Events = new TwitterEvents()
                {
                    OnCreatingTicket = ctx =>
                    {
                        var profilePic = ctx.User.Value<string>("profile_image_url");
                        ctx.Principal.Identities.First().AddClaim(new Claim("urn:twitter:profilepicture", profilePic, ClaimTypes.Uri, ctx.Options.ClaimsIssuer));
                        return Task.FromResult(0);
                    },
                    OnRemoteFailure = ctx =>
                    {
                        ctx.Response.Redirect("/error?FailureMessage=" + UrlEncoder.Default.Encode(ctx.Failure.Message));
                        ctx.HandleResponse();
                        return Task.FromResult(0);
                    }
                };
            });

     }

     public void Configure(IApplicationBuilder app)
     {
            app.UseAuthentication(); // needed for AutomaticAuthenticate

            // Choose an authentication type
            app.Map("/login", signinApp =>
            {
                signinApp.Run(async context =>
                {
                    var authType = context.Request.Query["authscheme"];
                    if (!string.IsNullOrEmpty(authType))
                    {
                        // By default the client will be redirect back to the URL that issued the challenge (/login?authtype=foo),
                        // send them to the home page instead (/).
                        await context.ChallengeAsync(authType, new AuthenticationProperties2() { RedirectUri = "/" });
                        return;
                    }

                    //foreach (var type in context.Authentication.GetAuthenticationSchemes())
                    foreach (var type in context.RequestServices.GetRequiredService<IOptions<AuthenticationOptions2>>().Value.Schemes)
                    {
                        // TODO: display name?
                        await context.Response.WriteAsync("<a href=\"?authscheme=" + type.Name + "\">" + (type.Name ?? "(suppressed)") + "</a><br>");
                    }
                    await context.Response.WriteAsync("</body></html>");
                });
            });

            // Sign-out to remove the user cookie.
            app.Map("/logout", signoutApp =>
            {
                signoutApp.Run(async context =>
                {
                    await context.SignOutAsync(CookieAuthenticationDefaults.AuthenticationScheme);
                    await context.Response.WriteAsync("You have been logged out. Goodbye " + context.User.Identity.Name + "<br>");
                    await context.Response.WriteAsync("<a href=\"/\">Home</a>");
                    await context.Response.WriteAsync("</body></html>");
                });
            });

            // Display the remote error
            app.Map("/error", errorApp =>
            {
                errorApp.Run(async context =>
                {
                    await context.Response.WriteAsync("An remote failure has occurred: " + context.Request.Query["FailureMessage"] + "<br>");
                    await context.Response.WriteAsync("<a href=\"/\">Home</a>");
                    await context.Response.WriteAsync("</body></html>");
                });
            });


            app.Run(async context =>
            {
                // CookieAuthenticationOptions.AutomaticAuthenticate = true (default) causes User to be set
                var user = context.User;

                // Display user information
                await context.Response.WriteAsync("Hello " + (context.User.Identity.Name ?? "anonymous") + "<br>");
                foreach (var claim in context.User.Claims)
                {
                    await context.Response.WriteAsync(claim.Type + ": " + claim.Value + "<br>");
                }

                await context.Response.WriteAsync("Tokens:<br>");
                
                //await context.Response.WriteAsync("Access Token: " + await context.Authentication.GetTokenAsync("access_token") + "<br>");
                //await context.Response.WriteAsync("Refresh Token: " + await context.Authentication.GetTokenAsync("refresh_token") + "<br>");
                //await context.Response.WriteAsync("Token Type: " + await context.Authentication.GetTokenAsync("token_type") + "<br>");
                //await context.Response.WriteAsync("expires_at: " + await context.Authentication.GetTokenAsync("expires_at") + "<br>");
                await context.Response.WriteAsync("<a href=\"/logout\">Logout</a><br>");
                await context.Response.WriteAsync("</body></html>");
            });
     }
}
```


HaoK:
Other things to consider for later: (can we get rid of the accept logic in the Challenge/SignIn/SignOut) handlers with the new model do we still care about multiple handlers getting a chance at each scheme?  Would drastically simplify the context's/manager logic if we could get rid of that stuff

davidfowl:
W00t

HaoK:
So I got the cookie sample working on top of the new stack...  it became clear that the AuthenticationManager's job really is to maintain the per request cache that maps authenticationScheme to the worker instance per scheme.  

There's also a problem with how AddCookies is sugar ontop of AuthenticationOptions, as it needs to add a transient CookieAuthenticationHandler for itself...  but it doesn't have access to the services hanging where it is right now...

src/Microsoft.AspNetCore.Authentication2/HttpContextExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features.Authentication;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    internal static class HttpContextExtensions
+    {
+        internal static IHttpAuthenticationFeature GetAuthentication(this HttpContext context)
```
HaoK: We don't, vestige that was copied over and not deleted yet

HaoK:
You know, we could keep the auth APIs on HttpContext as extension methods, since to get rid of the IHttpContextAccessor we probably have to start taking HttpContext for all IAuthenticationManager Authenticate/Challenge/SignIn apis...

So we could go back to something like:

````
context.AuthenticateAsync("cookies") => context.GetRequiredService<IAuthenticationManager>().AuthenticateAsync(context, "cookies")
```

davidfowl:
@HaoK in your example in the PR description, can you also show a simple middleware with the raw calls to `ChallengeAsync` and `ForbidAsync`

davidfowl:
> You know, we could keep the auth APIs on HttpContext as extension methods, since to get rid of the IHttpContextAccessor we probably have to start taking HttpContext for all IAuthenticationManager Authenticate/Challenge/SignIn apis...


That might be cleaner although the settable `IAuthenticationManager` is questionable. The feature also exposes the `User` property.

HaoK:
Yeah I don't think we need the feature, just some extension methods to hang the manager APIs directly off HttpContext.

davidfowl:
> Yeah I don't think we need the feature, just some extension methods to hang the manager APIs directly off HttpContext.

You need to add more code to your samples:


```C#
public static class HttpAuthenticationExtensions
{
    public virtual async Task<AuthenticationTicket2> AuthenticateAsync(this HttpContext context, string authenticationScheme) 
    {
        return context.RequestServices.GetRequiredService<IAuthenticationManager>().AuthenticateAsync(context, authenticationScheme);
    }
}
```

HaoK:
Updated the sample and the example at the top of the PR with a middleware that simulates basic login flow:

1. First request (no cookie) results in Challenge(Cookies) => Redirect to default login url of "/Account/login")
2. Go to /login to sign in a cookie
3. Return to base path to see "hello oldtimer"
4. Going to /forbid results in a 302 => AccessDeniedPath

HaoK:
@davidfowl how do you feel about each auth scheme being a top level thing (that internally calls AddAuthentication with its own configuration).  Only thing that irritates me is that we end up calling AddAuthentication a bunch of times, but they are TryAdd's so its safe...  This lets the auth handlers properly add their handlers.

```
services.AddCookieAuthentication();
services.AddGoogleAuthentication(o => o.ClientId = "adlkjb");
```

HaoK:
@PinpointTownes @brockallen take a look and see how this feels, I'm working on OAuth/Remote today now that cookies is functional

brockallen:
I'm traveling this week. Won't be back until next week.

HaoK:
@davidfowl Okay so the remote middleware requires the ability to handle entire requests which cookies does not, so the next iteration will add back some kind of HandleRequest entry point where every scheme gets a chance to see if this is a request for them (otherwise let it continue)...  So much for the AuthenticationMiddleware only setting httpContext.User.  I'm now more open to your idea of making it a IStartupFilter, its pretty much mandatory for remote auth middleware

HaoK:
@Tratcher any thoughts on any improvements we can make to the control flow logic we have today?  Personally I hate the bool, so we should do something to replace: 

```
        public virtual Task<bool> HandleRequestAsync()
```

We could use the Enum we have today I guess?

```
    public enum EventResultState
    {
        /// <summary>
        /// Continue with normal processing.
        /// </summary>
        Continue,

        /// <summary>
        /// Discontinue processing the request in the current middleware and pass control to the next one.
        /// </summary>
        Skipped,

        /// <summary>
        /// Discontinue all processing for this request.
        /// </summary>
        HandledResponse
    }
```

src/Microsoft.AspNetCore.Authentication2/IAuthenticationSchemeHandler.cs
```diff
@@ -0,0 +1,212 @@
...
+        // Gives the handler access to the configuration data
+        Task InitializeAsync(AuthenticationScheme scheme, HttpContext context);
+
+        Task AuthenticateAsync(AuthenticateContext context);
+        Task ChallengeAsync(ChallengeContext context);
+        Task SignInAsync(SignInContext context);
+        Task SignOutAsync(SignOutContext context);
+    }
+
+    public abstract class AuthenticationSchemeHandler<TOptions> : IAuthenticationSchemeHandler where TOptions : class
```
Tratcher: Perf: How can we avoid re-creating these handlers on every request? Is there some way we can make them stateless and store the relevant state in the HttpContext?

PinpointTownes:
> @PinpointTownes @brockallen take a look and see how this feels.

Honestly, this sounds like an interesting approach, but I still have no idea what concrete problem(s) this huge API change is supposed to fix (I'm currently blocked by the fact the migration path will be **extremely painful**, without being sure the new design solves something that we can't fix in the existing architecture).

Curious: what's your plan for the host-provided authentication handlers?

src/Microsoft.AspNetCore.Authentication2/AuthenticationOptions.cs
```diff
@@ -0,0 +1,24 @@
...
+        public IDictionary<string, AuthenticationScheme> SchemeMap { get; } = new Dictionary<string, AuthenticationScheme>(); // case sensitive?
+
+        public void AddScheme(string name, Action<AuthenticationSchemeBuilder> configureBuilder)
+        {
+            var builder = new AuthenticationSchemeBuilder(name);
+            configureBuilder(builder);
+            SchemeMap[name] = builder.Build();
+        }
+
+        public string DefaultAuthenticationScheme { get; set; }
```
PinpointTownes: How is that an authorization concern? That would be great if we could avoid making the same mistake twice :smile:

Tratcher: `[Authorize]` is the component that calls this and it can be configured via DefaultAuthPolicy, so I think this one is redundant. Same for DefaultChallengeScheme.

PinpointTownes:
> Okay so the remote middleware requires the ability to handle entire requests which cookies does not, so the next iteration will add back some kind of HandleRequest entry point where every scheme gets a chance to see if this is a request for them (otherwise let it continue)...

Please don't make that `RemoteAuthenticationHandler`-specific, it's also something ASOS relies on (heavily): https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/blob/dev/src/AspNet.Security.OpenIdConnect.Server/OpenIdConnectServerHandler.cs#L25-L140

src/Microsoft.AspNetCore.Authentication2/AuthenticationOptions.cs
PinpointTownes: The fact the authorization filters are currently responsible of determining the applicable scheme doesn't mean it's the right design nor really an authorization concern. In Web API 2 OWIN, the separation of concerns was much better/clearer, thanks to `HostAuthenticationFilter`/`HostAuthenticationAttribute`.

PinpointTownes:
@HaoK note: I'm not 100% sure `EventResultState` is a good candidate for `HandleRequestAsync`, as the logic is really binary in this case (`AuthenticationMiddleware.InvokeAsync` calls or doesn't invoke the rest of the pipeline depending on whether true or false was returned, but there's no "default logic" that can be skipped)

davidfowl:
> Honestly, this sounds like an interesting approach, but I still have no idea what concrete problem(s) this huge API change is supposed to fix (I'm currently blocked by the fact the migration path will be extremely painful, without being sure the new design solves something that we can't fix in the existing architecture).

Migration *is* painful, no denying that. Maybe we deprecate the old system and make a new system side by side. That would let 3rd party auth middleware gradually migrate.

The concrete things this solves:
- It brings authZ and authN closer together in pattern
- Less middleware to deal with, you configure a single middleware in the pipeline (or maybe a startup filter if that's a good idea) and you configure services to enable different auth schemes.
- Adding an authentication handler to the system isn't request specific as all state is passed to the handler via the contexts.
- The concept of active/passive auth middleware isn't even a thing (arguably, it could be solved in other ways). 
- The overall design seems simpler (for now) and the layering looks better than what we currently have.

HaoK:
@PinpointTownes don't worry the request handling path is coming back, I'm looking at unifying the request flow processing that we already have today in the handlers, with the top level HandleRequest in IAuthenticationHandler, it feels a little weird that they are inconsistent today, hopefully that will result in a single Result class rather than the 2, 3, or 4? we have today...

HaoK:
@PinpointTownes I think the hosts can just follow the new pattern and add their own authentication handlers/schemes.  IIS today doesn't look to be doing anything special, WebListener looks a bit different with the explicit scehemes it handles like NTLM/Kerberos, but is there something that sticks out to you where the new pattern breaks down for the hosts?

```
   services.AddCookiesAuthentication("Cookies")
   services.AddIISIntegratedAuthentication("Windows");
```

src/Microsoft.AspNetCore.Authentication2/IAuthenticationSchemeHandler.cs
HaoK: Specifically, I'm adding back some kind of `Task<AuthenticationRequestResult> HandleRequest()` back for the remote auth flows, does this make sense to live in IAuthenticationHandler.

src/Microsoft.AspNetCore.Authentication2.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -0,0 +1,461 @@
...
+                Options.Events = context.RequestServices.GetRequiredService(Options.EventsType) as CookieAuthenticationEvents;
+            }
+
+            if (Options.Events == null)
+            {
+                Options.Events = new CookieAuthenticationEvents();
+            }
+            if (String.IsNullOrEmpty(Options.CookieName))
+            {
+                Options.CookieName = CookieAuthenticationDefaults.CookiePrefix + Scheme.Name;
```
HaoK: The handlers are initialized per request, as enabling DI activation is a goal now.  So each auth options is not really like our IOptions instances, each is actually per named scheme instance, similar to what we have today, and why I stuff it into the dictionary of each scheme...  (RIP named options)

src/Microsoft.AspNetCore.Authentication2/DefaultAuthenticationManager.cs
```diff
@@ -0,0 +1,173 @@
...
+            Schemes = schemes;
+            _accessor = context;
+        }
+
+        public IAuthenticationSchemeProvider Schemes { get; }
+
+        // handler instance cache, need to initialize once per request, manager lifetime should be per request
+        private Dictionary<string, IAuthenticationSchemeHandler> _handlerMap = new Dictionary<string, IAuthenticationSchemeHandler>();
+
+        private readonly IHttpContextAccessor _accessor;
```
HaoK: Yeah its on the todo list, possibly we modify the signature of all of the methods to explicitly pass in an HttpContext instead.

src/Microsoft.AspNetCore.Authentication2/AuthenticationOptions.cs
HaoK: Now is certainly a good time to revisit AuthZ if anything needs to move there, there's a few breaking changes I have on my todo list there, namely a similar change where `bool AuthorizeAsync` needs to become `AuthorizationResult AuthorizeAsync` similar to how we are beefing up HandleRequest

src/Microsoft.AspNetCore.Authentication2/AuthenticationSchemeBuilder.cs
```diff
@@ -0,0 +1,25 @@
...
+        public AuthenticationSchemeBuilder(string name)
+        {
+            Name = name;
+        }
+
+        public string Name { get; }
+
+        public Type HandlerType { get; set; }
+
+        // Holds things like the configured options instances for the handler
```
HaoK: Tradeoff vs generic hell, once I implement a few more auth handlers, I'll do a cleanup/rollup pass to see if we need anything at this layer but there's is an equivalent `AuthenticationSchemeHandler<TOptions>` that is similar to `AuthenticationHandler<TOptions>` today.  The options don't need to be typed at the scheme builder layer, since these untyped options are only used by the scheme handler instance later, when it can just cast it to the type it knows.  Its not related to the global singleton AuthenticationOptions which deals only with the various schemes and handler types.

PinpointTownes:
@davidfowl AFAICT, none of the points you list are really "issues" (they are more like general architecture concerns).

> The concept of active/passive auth middleware isn't even a thing (arguably, it could be solved in other ways).

Of course it is, it's just handled differently: instead of being handled at the handler level (current design), it's done in `AuthenticationMiddleware` :smile: 

> The overall design seems simpler (for now) and the layering looks better than what we currently have.

Sure, but for a reason: it doesn't support the same features set yet (e.g it doesn't seem to support "automatic challenge" nor converts raw 401 responses into "challenges").

> but is there something that sticks out to you where the new pattern breaks down for the hosts?

@HaoK no no, I'm just curious to know how the layering issue will be solved (IIRC, the host-provided handlers directly implement `IAuthenticationHandler` because they can't depend on the Security stack).

Tratcher:
@PinpointTownes We should consider removing the automatic 401 -> Challenge and require auth to be invoked explicitly. This made sense in Katana because challenges were always processed on the way out, but now they're processed immediately except for these 401s so you have two disconnected stages. This caused the original bug you complained about in #930.

PinpointTownes:
> We should consider removing the automatic 401 -> Challenge and require auth to be invoked explicitly.

I'm fine with that, but this will be blocking for "pure OWIN" applications/frameworks (e.g like NancyFX), where the ASP.NET Core authentication infrastructure is not directly available. Returning a 401 status code was a convenient way to trigger a challenge.

Tratcher:
True, but we could move that logic to the owin adapter, or provide something similar in the adapter.

HaoK:
Well if we can't have weblistener/IIS depend on the security repo, maybe we need to continue with Http.Authentication for all of the abstractions.

HaoK:
One other nice sideeffect of this design, claims transformation should be much simpler, as it will become just be a service that the IAuthenticationManager uses and automatically applies once as part of any AuthenticateAsync call.

HaoK:
Finally got basic social auth (with only twitter for now) working...
Added SocialSample2 which is a pruned down version of the existing SocialSample.  Twitter was first, because it directly derived from RemoteAuthenticationHandler, unlike the other OAuth2 ones.

Here's what the social configure services looks like now with both Cookies and twitter, and using the cookie automatically for context.User :

```C#
        public void ConfigureServices(IServiceCollection services)
        {
            services.AddCookieAuthentication(o => o.LoginPath = "/login");
            services.AddTwitterAuthentication(options =>
            {
                options.ConsumerKey = "BSdJJ0CrDuvEhpkchnukXZBUv";
                options.ConsumerSecret = "xKUNuKhsRdHD03eLn67xhPAyE1wFFEndFo1X2UJaK2m1jdAxf4";
                options.SignInScheme = CookieAuthenticationDefaults.AuthenticationScheme;
                options.RetrieveUserDetails = true;
                options.SaveTokens = true;
                options.Events = new TwitterEvents()
                {
                    OnCreatingTicket = ctx =>
                    {
                        var profilePic = ctx.User.Value<string>("profile_image_url");
                        ctx.Principal.Identities.First().AddClaim(new Claim("urn:twitter:profilepicture", profilePic, ClaimTypes.Uri, ctx.Options.ClaimsIssuer));
                        return Task.FromResult(0);
                    },
                    OnRemoteFailure = ctx =>
                    {
                        ctx.Response.Redirect("/error?FailureMessage=" + UrlEncoder.Default.Encode(ctx.Failure.Message));
                        ctx.HandleResponse();
                        return Task.FromResult(0);
                    }
                };
            });
            {
                options.DefaultAuthenticationScheme = CookieAuthenticationDefaults.AuthenticationScheme;
            });
        }

```

I also boiled down the `AddXyzAuthentication()` method into a generic method for the ones that derive from Security:

```C#
        public static IServiceCollection AddSchemeHandler<TOptions, THandler>(this IServiceCollection services, Action<TOptions> configureOptions)
            where TOptions : AuthenticationSchemeOptions, new()
            where THandler : AuthenticationSchemeHandler<TOptions>
        {
            var handlerOptions = new TOptions();
            configureOptions?.Invoke(handlerOptions);
            services.AddAuthentication(o =>
            {
                o.AddScheme(handlerOptions.AuthenticationScheme, b =>
                {
                    b.HandlerType = typeof(THandler);
                    b.Settings["Options"] = handlerOptions;
                });
            });
            services.AddTransient<THandler>();
            return services;
        }

        public static IServiceCollection AddCookieAuthentication(this IServiceCollection services, Action<CookieAuthenticationOptions> configureOptions) =>
            services.AddSchemeHandler<CookieAuthenticationOptions, CookieAuthenticationHandler>(configureOptions);

        public static IServiceCollection AddTwitterAuthentication(this IServiceCollection services, Action<TwitterOptions> configureOptions) =>
            services.AddSchemeHandler<TwitterOptions, TwitterHandler>(configureOptions);

```

HaoK:
Pushed some cleanup before starting on OAuth2 migration:
- Removed `IHttpContextAccessor` usage.  `IAuthenticationManager` apis always take HttpContext, doesn't change public surface area as they are exposed as extension methods on httpContext anyways.
- Normalized auth context and event context's to share common base classes: `BaseContext` contains HttpContext property, `BaseAuthenticationContext` contains scheme + `AuthenticationProperties`, today every context pretty much copied these two properties.

HaoK:
Google support added, migration is pretty trivial for OAuth based providers

src/Microsoft.AspNetCore.Authentication2.Cookies/CookieAuthenticationHandler.cs
PinpointTownes: I agree with @Tratcher. It's not only weird, it's dangerous (and potentially a perf' disaster if you receive thousands of requests when the app warms up, as there's no locking preventing `InitializeAsync` from being called multiple times).

There's also another BIG problem with this design approach: configuration errors cannot be detected at startup and won't prevent the app from loading successfully (so each time it will receive a new request, it will try to initialize the handlers... and fail miserably).

HaoK:
> I agree with @Tratcher. It's not only weird, it's dangerous (and potentially a perf' disaster if you receive thousands of requests when the app warms up, as there's no locking preventing InitializeAsync from being called multiple times).

Today we create a handler instance per request inside of each auth middleware, how is creating a scheme handler with associated options any different than the pattern today?  In both situations we have a per request handler instance per auth scheme, except now we have a scheme options to go along with the handler.

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs#L67
```
        {
            var handler = CreateHandler();
            await handler.InitializeAsync(Options, context, Logger, UrlEncoder);
```

> There's also another BIG problem with this design approach: configuration errors cannot be detected at startup and won't prevent the app from loading successfully (so each time it will receive a new request, it will try to initialize the handlers... and fail miserably).

While its true we can't detect errors at startup time since we aren't configuring the options until they are used in a request, we can detect initialization failures, and mark them as failed in the global authentication options if we want to prevent each request from trying to reinitialize...but isn't your app hosed anyways if your auth handlers are blowing up?  Is this that different than today when you forget to register an auth handler, you get a runtime exception saying "no auth scheme registered for foo", so with the new stack you'd get "auth scheme 'foo' failed to initialize"

PinpointTownes:
> Today we create a handler instance per request inside of each auth middleware, how is creating a scheme handler with associated options any different than the pattern today?

In the current design, options initialization is always made in the middleware constructors, that are guaranteed to be executed exactly once per middleware registration, at startup time. This is true for all the ASP.NET Core authentication middleware from cookies to OpenID Connect.

> Is this that different than today when you forget to register an auth handler, you get a runtime exception saying "no auth scheme registered for foo", so with the new stack you'd get "auth scheme 'foo' failed to initialize"

It's fundamentally different: with the current system, an application error like a broken middleware configuration is always considered as a critical error that prevents the entire app from servicing requests. If `CaptureStartupErrors` is disabled, the process will even terminate... and potentially trigger an IIS app pool alert or inform monitoring systems like SCOM).

HaoK:
> In the current design, options initialization is always made in the middleware constructors, that are guaranteed to be executed exactly once per middleware registration, at startup time. This is true for all the ASP.NET Core authentication middleware from cookies to OpenID Connect.

Fair enough I can certainly move the non events options initialization back to the singleton, I was only trying to get per request events instances, and those were hanging off of the options today but the events can be generalized to some kind of per request thingy.

HaoK:
@PinpointTownes so I took a closer look, it appears that the individual scheme options are already only created once, as its run during ConfigureServices today, basically the AuthenticationScheme metadata is computed once, but a new instance of an AuthenticationSchemeHandler is created per request, that accesses its singleton AuthenticationSchemeOptions that lives inside of the AuthenticationScheme data...

So if validation at startup time is your only remaining concern, I can certainly add an explicit Validate() call that happens as part of UseAuthentication() (before the app starts)

```
        public static IServiceCollection AddSchemeHandler<TOptions, THandler>(this IServiceCollection services, Action<TOptions> configureOptions)
            where TOptions : AuthenticationSchemeOptions, new()
            where THandler : AuthenticationSchemeHandler<TOptions>
        {
            var handlerOptions = new TOptions();
            configureOptions?.Invoke(handlerOptions);
            services.AddAuthentication(o =>
            {
                o.AddScheme(handlerOptions.AuthenticationScheme, b =>
                {
                    b.HandlerType = typeof(THandler);
                    b.Settings["Options"] = handlerOptions;
                });
            });
            services.AddTransient<THandler>();
            return services;
        }
```

PinpointTownes:
> @PinpointTownes so I took a closer look, it appears that the individual scheme options are already only created once, as its run during ConfigureServices today, basically the AuthenticationScheme metadata is computed once, but a new instance of an AuthenticationSchemeHandler is created per request, that accesses its AuthenticationSchemeOptions which lives inside of the singleton AuthenticationScheme data...

That's the whole problem: options are shared between the handlers instances of a same scheme, so they may be mutated concurrently when calling `InitializeAsync`.

HaoK:
Ah yea, ok I need to make two instances of the same middleware work now...

HaoK:
Updated:
Every scheme handler instances now create/configure their own scheme options instead of using a shared handler instance from the singleton AuthenticationScheme:
```
        public virtual Task InitializeAsync(AuthenticationScheme scheme, HttpContext context)
        {
            Scheme = scheme;
            Options = new TOptions();
            var configureOptions = scheme.Settings["ConfigureOptions"] as Action<TOptions>;
            configureOptions?.Invoke(Options);
 ```

Working on reworking validation of the TOptions

**Comment by PinpointTownes** [comment]
> Every scheme handler instance now creates/configures their own scheme options instead of using a shared handler instance from the singleton AuthenticationScheme:

Ouch, this may result in terrible performance if the configuration delegate takes time to complete like in this sample:

```csharp
app.UseOpenIdConnectServer(options => {
    options.Provider = new AuthorizationProvider();

    // Enable the authorization, logout, token and userinfo endpoints.
    options.AuthorizationEndpointPath = "/connect/authorize";
    options.LogoutEndpointPath = "/connect/logout";
    options.TokenEndpointPath = "/connect/token";
    options.UserinfoEndpointPath = "/connect/userinfo";

    // Register a new ephemeral key, that is discarded when the application
    // shuts down. Tokens signed using this key are automatically invalidated.
    // This method should only be used during development.
    options.SigningCredentials.AddEphemeralKey();

    // On production, using a X.509 certificate stored in the machine store is recommended.
    // You can generate a self-signed certificate using Pluralsight's self-cert utility:
    // https://s3.amazonaws.com/pluralsight-free/keith-brown/samples/SelfCert.zip
    //
    // options.SigningCredentials.AddCertificate("7D2A741FE34CC2C7369237A5F2078988E17A6A75");
    //
    // Alternatively, you can also store the certificate as an embedded .pfx resource
    // directly in this assembly or in a file published alongside this project:
    //
    // options.SigningCredentials.AddCertificate(
    //     assembly: typeof(Startup).GetTypeInfo().Assembly,
    //     resource: "Mvc.Server.Certificate.pfx",
    //     password: "Owin.Security.OpenIdConnect.Server");
});
```

(and the most annoying thing is that it will generate a different signing key per request, which will cause weird validation issues...)

HaoK:
Hrm, so that usage pattern is relying on the old singleton options lifetime, while the current iteration of scheme options is basically scoped.  For something like that, maybe the solution would be to split the things that don't change per request into a different options class and access those via `IOptions` in the handler?

HaoK:
Added an one time explicit pass through the schemes to validate each handler/options pair, not sure this is worth pulling out, since each handler gets a chance to handle each request, might be better to just merge this into the default handler.Initialize logic

HaoK:
Going to move JWTBearer next, and then port all the existing tests to see what actually still works :)

HaoK:
JWT ported and all tests (excluding Facebook/MsftA/OpenIDC) ported.

62/147 are currently failing

HaoK:
Down to 35/147, 401/403 behavior changes accounts for a bunch of these

HaoK:
All tests are now passing/updated, 10 are skipped for investigation:
- Claims transformation feature missing
- Automatic challenge 401/403 behavior tests

HaoK:
Porting OIDC/MSA + adding back claims transformation next

HaoK:
- Initial OIDC migration, 21 OIDC tests failing
- also moved handler service dependencies like ILoggerFactory, UrlEncoder, IDataProtector back to their ctors instead of resolving from RequestServices in InitializeHandler

HaoK:
Added back claims transformation which is a lot simpler now with the service.  The service depends on a new `IClaimsTransformation` service which has a single method: 

We call this on any successful `AuthenticateAsync` call.  We can probably hang the default delegate off the AuthenticationOptions now too:

```
     return Task.FromResult(p);
  }
```

HaoK:
`SharedAuthenticationOptions.SignInScheme` => `AuthenticationOptions.DefaultSignInScheme`

HaoK:
@davidfowl @Tratcher @PinpointTownes @brockallen I've ported all the auth middleware over, and all the tests are passing except for ones that are relying on the 401/403 status code triggering automatic challenge behavior.

I've also updated the snippet at the top with what registering the new stuff looks like.  One of the biggest differences now is that there is a handler instance per scheme for each request, along with a new recreated `AuthenticationSchemeOptions` instance as well.  To address @PinpointTownes concern about not re-initializing singletons in the options, I believe one approach could be storing the singleton in the `AuthenticationScheme` settings which continue to be singletons (1 per scheme):

```
// Create the creds before calling AddOIDC, and then store them into the singleton scheme settings
public static IServiceCollection AddOIDC(Action<OIDCOptions> configureOptions, SigningCredentials creds)
            services.AddAuthentication(o =>
            {
                o.AddScheme("OIDC", b =>
                {
                    b.HandlerType = typeof(OIDCHandler);
                    b.Settings["ConfigureOptions"] = configureOptions;
                    b.Settings["SigningCredentials"] = creds;
                });
            });
            services.AddTransient<OIDCHandler>();

       // Inside of the handler, stuff the instance back into the options
        protected async override Task<OpenIdConnectOptions> CreateOptionsAsync()
        {
            if (options.SigningCredentials == null) {
                 options.SigningCredentials = Scheme.Settings["SigningCredentials"] as SigningCredentials;
            }
            return options;
        }
```

I also applied the proposed new EventType pattern to enable scoped dependencies in the event handlers across the board as well.

hassanselim0:
So I've been trying to follow up with this issue as much as I can (starting from the issue that had a fix that cause other issues in 1.1 and all that).
But there is one thing I want to understand. In this new design, will it be possible to have multiple types of authorization (eg: Cookies for Views, and JWT for API), and just have plain `[Authorize]` attributes on controllers/actions ... AND have different challenge behavior where with Views you get a redirect and with API you get plain 401/403?
Because currently in order to do that, I disabled automatic challenge in the cookies auth middleware, and enabled it in my custom auth middleware, and from its HandleUnauthorizedAsync I call the PriorHandler's ChallengeAsync under certain conditions (to cause a page redirect in most cases except a few).

HaoK:
@hassanselim0 The prototype changes haven't rippled out to MVC/identity yet, so I'm not 100% sure how things would change there exactly.  Automatic challenge does not exist in the current iteration of this PR, so `[Authorize]` would need to configure the default authorization policy to specify what scheme's to challenge, there's no more chaining in the new design.  If you listed both Cookies and JWT, each will be challenged in the order you specify (so last one wins).  So I guess the answer is, the new design makes it impossible to rely on the side effects of the old implementation details of chaining auth handlers, so you will need to have a different policy configured for Views vs API Views (which is what we've always recommended)...

hassanselim0:
@HaoK well then if challenge will be called on all auth providers (so last one wins), can't you make it possible for the first provider's challenge to prevent the second provider's challenge from being called?
Maybe some simple form of chaining can be made cleanly on top of this new design and it wouldn't be as bad as the old design. I think this will allow you to cover all the cases that anybody would need. (some developers just love writing auto-magical code 😆)

HaoK:
>  if challenge will be called on all auth providers (so last one wins), can't you make it possible for the first provider's challenge to prevent the second provider's challenge from being called?

The list of auth schemes that are challenged is defined by the policy.  So its fully controlled and defined by the developer at this point.

hassanselim0:
ah, got it :)
Maybe later on you could figure out a way to let the developer specify some logic (in Startup?) that would decide which handler's challenge gets called and apply that logic globally on every request. I just feel that there must be a way to do this cleanly without manually specifying schemes or policies.

HaoK:
Looking how identity would plug into these changes:

Pros:
- UseIdentity no longer needed only Addidentity

Cons:
- Today IdentityOptions.Cookies contains the 4 cookie options that are used by identity and can be customized via IdentityOptions, this no longer would be possible since the Auth needs to be configured earlier.  Need a `ConfigureScheme` method which would allow further customization/override of something like a cookie scheme's options, which would wrap the existing `Action<TOptions>`?

brockallen:
I've also not had a chance to checkup on the status of this. Is it possible to register multiple instances of cookies, or OIDC? Also, is that configuration static at startup time, or can it be dynamic (e.g. from a database)?

HaoK:
Tweaked thing so the `AuthenticationOptions2` now keeps the scheme data in `AuthenticationSchemeBuilder` form so we have a chance to further configure a scheme's options.  Added `ConfigureScheme` and `ConfigureSchemeHandler<TOptions>` methods which allow this scenario:

```
                    services.AddCookieAuthentication("Cookie1");
                    services.ConfigureSchemeHandler<CookieAuthenticationOptions>("Cookie1",
                        o => o.CookieName = "One");
```

HaoK:
@brockallen Multiple instances are still supported like before.  The initialization of each scheme handler instance options is done per request, so while it is easy for you to get a DbContext to your scheme handler via ctor injection, its not really possible to get a DbContext into your `Action<TOptions>` lambda, so you would probably do your static initialization of the options, and then override with stuff from the database in the handler?

It'd look something like this I imagine, where you are registering 2 cookie instances with a custom handler:

```C#
   // ConfigureServices:

       _dbContext = context;
    }
    
        {
        }
}
```

But I'll add a test to make sure this actually works :)

PinpointTownes:
> I've ported all the auth middleware over, and all the tests are passing except for ones that are relying on the 401/403 status code triggering automatic challenge behavior, which I believe we are cutting?

Before ditching this feature, I think it would be worth reviving the OWIN security draft @Tratcher wrote a while ago to offer a standard way for OWIN apps to use the ASP.NET Core authentication architecture without being tied to the ASP.NET Core primitives.

> To address @PinpointTownes concern about not re-initializing singletons in the options, I believe one approach could be storing the singleton in the AuthenticationScheme settings which continue to be singletons (1 per scheme):

This doesn't sound like a serious solution, IMHO.

Signing credentials are not the only options that should be treated as singletons. There are also all the ticket formats/data protectors/OIDC configuration managers/HTTP clients: with your approach, all these components are now instantiated per request instead of being reused as they should. This will end up being really "allocatey".

Note that I'm not opposed to allowing the options to be resolved at runtime (it could be super useful for multitenancy: https://github.com/aspnet/Security/issues/35#issuecomment-122125509), but **I'm fairly confident** the options should be static by default, not created per request.

What about keeping using `IOptions<>` so the options manager can be replaced for multitenancy scenarios and moving the validation part in more a appropriate place? E.g in a `IConfigureOptions` class.

> Pros: UseIdentity no longer needed only Addidentity

Personally, I wouldn't consider that as a pro (for me, it's even a con): there are scenarios where you need to use the Identity services but don't want to register the cookies middleware. See https://github.com/openiddict/openiddict-samples/blob/master/samples/PasswordFlow/AuthorizationServer/Startup.cs for a concrete case where the cookies middleware would be unwanted.

FWIW, I worked on an ASOS prototype targeting your new auth' stack during the holidays. It was a rather painful experience, and unfortunately, I can't really say the benefits it offers were obvious for me :sweat_smile: 

I'll keep working on it to match your recent changes, but my intermediate conclusion is that your design doesn't make things easier to implement, despite the fact important scenarios are no longer possible (e.g being able to run an authentication middleware depending on the request path). I also suspect an important perf' regression due to the per-request options create.

PinpointTownes:
Speaking of perf' regression, I'd bet there's also a major change you didn't expect: creating a new `ConfigurationManager` per-request will prevent the OIDC middleware from caching the discovery documents. In practice, this means that the OIDC middleware will have to make 2 backchannel calls (one for the configuration, one of the JWK set) before each OIDC operation. That's terrible.

HaoK:
@PinpointTownes Thanks for playing with this, keep in mind this is just what I've been prototyping/playing with so far, as opposed to anything approaching firm decisions either way.

I think the first question to ask is whether it makes sense to switch the stack from Middleware based to service based.  I mostly like the flexibility the new stack gives us to decide whether we want things to be per request or not.  If we end up deciding we don't want per request scheme handler options instances, its easy enough to switch back to singleton via caching inside of the AuthenticationScheme anyways.

This PR will be a topic of discussion at triage this week, so we should have some more folks weighing in soon.

HaoK:
@PinpointTownes So it looks like the only reason I made the options initialization per request, was to enable getting the events from DI.  With a slight tweak of just moving where the events instance lives (from the options to somewhere in the base Handler, there will be no reason to have per request options again.  So don't be too caught up on that issue for now, I'll tweak things a bit tomorrow and the options can return to being singletons.

PinpointTownes:
> This PR will be a topic of discussion at triage this week, so we should have some more folks weighing in soon.

Nice!

> With a slight tweak of just moving where the events instance lives (from the options to somewhere in the base Handler, there will be no reason to have per request options again. 

Well, the reason to want per-request option would be to support multitenancy, as I said in my post.

> So don't be too caught up on that issue for now, I'll tweak things a bit tomorrow and the options can return to being singletons.

I'm not. I just wanted to make sure you were 100% aware of the implications this design has before making a decision :smile:

HaoK:
Updated moved the source of truth for the events instance to the handler, scheme options are now only configured once as part of the building of the authentication scheme, and stored inside of the Settings bag for each scheme.

Also simplified the initialization of the scheme handlers back down to Initialize/InitializeOptions which is expected to do the validation inline.

HaoK:
@PinpointTownes if I can give you an easier way to plug in multi tenancy style options as part of the new stack, is that a big win for you?  

It should be fairly easy by adding a little more abstractions around how the new stack resolves the scheme/handlers.  By default we would be single tenant/singletons, but it should be fairly straight forward to replace with a different service that's able to resolve different instances for the same scheme to support multi-tenancy.

PinpointTownes:
> @PinpointTownes if I can give you an easier way to plug in multi tenancy style options as part of the new stack, is that a big win for you?

If that does make the "static options" scenario a PITA, yeah, it would make everyone happy :+1: 

> By default we would be single tenant/singletons, but it should be fairly straight forward to replace with a different service that's able to resolve different instances for the same scheme to support multi-tenancy.

Do you think it could use `IOptions<>` under the hood? Or is a whole new abstraction required? It would be great not to tie it with the security stuff, so it could be potentially used by other stacks in ASP.NET Core.

HaoK:
Right now I'm just planning on looking at how named instances of scheme options would work for security, we can certainly look to see if its something we can generalize after that.  The individual scheme handlers are not using IOptions currently, basically IOptions is an alias for singleton options today, but now that we have IOptionsSnapshot which is scoped options, I would guess something like INamedOptions would be a possibility.

HaoK:
@PinpointTownes https://github.com/aspnet/Options/issues/161 will track a parallel design issue to this which is how to do named options (again)

@brockallen for your DbContext configure, the new IOptionsSnapshot actually should support this, except there's a bug where we are caching things busting the scenario: https://github.com/aspnet/Options/issues/160

HaoK:
Rebased to dev to pickup RC3 csproj merge (urghhh).  Added default challenge support for 401 => challenge for a default scheme (or a single scheme)

HaoK:
Reenabled 401 automatic challenge tests, only 5 skipped tests left

HaoK:
Renamed IAuthenticationManager2 =>IAuthenticationService which seems more appropriate now

src/Microsoft.AspNetCore.Authentication2.Cookies/CookieAppBuilderExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+            if (app == null)
+            {
+                throw new ArgumentNullException(nameof(app));
+            }
+            if (options == null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            var schemeProvider = app.ApplicationServices.GetRequiredService<IAuthenticationSchemeProvider>();
```
HaoK: Well this is the goop that will need to be added to each of the old Use methods to make things work... AddAuthentication() is also required for this to work.  But this works for cookies, and should be more or less identical across the board (stuff options instance in, jam in a service locator func to new up the handler instead of getting one from request services based on type.

@Eilon @Tratcher

src/Microsoft.AspNetCore.Authentication2.Cookies/CookieAppBuilderExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+            var sharedOptions = app.ApplicationServices.GetRequiredService<IOptions<AuthenticationOptions2>>();
+            var settings = new Dictionary<string, object>();
+            settings["Options"] = options;
+            var scheme = new AuthenticationScheme(options.AuthenticationScheme, typeof(CookieAuthenticationHandler), settings, sharedOptions.Value);
+            scheme.ResolveHandlerFunc = context => new CookieAuthenticationHandler(
```
Tratcher: Why is ResolveHandlerFunc needed? What do you do with the type param you pass into AuthenticationScheme?

Tratcher: Activator.CreateInstance?

PinpointTownes:
Omagad, so much sorcery in the recent commits :trollface:

src/Microsoft.AspNetCore.Authentication2/AuthAppBuilderExtensions.cs
```diff
@@ -23,7 +23,14 @@ public static IApplicationBuilder UseAuthentication(this IApplicationBuilder app
                 throw new ArgumentNullException(nameof(app));
             }
             
-            return app.UseMiddleware<AuthenticationMiddleware>();
+            // We only every want one instance of this middleware in the app
+            if (!app.Properties.ContainsKey(".AspNetCore.AuthenticationMiddleware"))
+            {
+                app.Properties[".AspNetCore.AuthenticationMiddleware"] = true;
```
Tratcher: We're going to have to consider how this affects pipeline branches.

src/Microsoft.AspNetCore.Authentication2.Cookies/CookieAppBuilderExtensions.cs
HaoK: Normally we pull the handler from RequestServices, we can't do this since we are building the scheme after the service provider has been created so this is the factory method for how we are going to create the handler per request...  I can switch to activator if you think that's cleaner

HaoK:
@PinpointTownes no joke, but we are attempting to minimize breaking changes to existing apps now...so there's going to be a lot of eye bleeding code hacks to make the old Startup APIs work with the new stack...

PinpointTownes:
@HaoK 

ROFL

> so there's going to be a lot of eye bleeding code hacks to make the old Startup APIs work with the new stack...

The problem with code hacks is that they are not only ugly... they often technically suck.

If I understood correctly what your recent commits are doing, this code snippet won't really work as one could imagine:

```csharp
app.UseOAuthValidation();

app.UseSignalR();

app.UseCookieAuthentication();

app.UseMvc();
```

Though it's registered after SignalR, the cookie handler will be actually invoked via `AuthenticationMiddleware`, which will be registered by the first authentication middleware in the pipeline (`app.UseOAuthValidation()` here). This will surely lead to really strange things, IMHO.

At this point, wouldn't creating a new experimental PR based on the existing architecture be easier and cleaner?

Tratcher:
Yes, there may be some behavioral oddities. The goal is to at least make the default template scenario work as expected to limit the impact on existing apps. The methods will be marked obsolete and we will encourage migration to the new system.

PinpointTownes:
> The methods will be marked obsolete and we will encourage migration to the new system.

Is it really worth the extreme confusion it will cause? If the new system is better, by all means, adopt it and break the entire world. If it's not, well, maybe it would be better to improve the existing one.

But an intermediate design that involves two different designs doesn't seem the right answer, IMHO. It will be pretty much like `app.UseGoogleAuthentication()` vs `app.UseGoogleAuthentication(new GoogleAuthenticationOptions())` vs `app.UseGoogleAuthentication(new GoogleOAuth2AuthenticationOptions())` in Katana: it will be confusing and people will keep asking "hey, what's the difference between the three methods?

HaoK:
I'd prefer a clean break but I understand why it might not be a good idea to break the world. That's a lot of customer pain to deal with. Hence these ugly looking changes.

Tratcher:
At least with obsolete extensions we can give the developer direct feedback about what changed. If we just remove the extensions then it fails to compile and they have no idea why. We may not even have to implement the extensions, just make them throw :-)

PinpointTownes:
> If we just remove the extensions then it fails to compile and they have no idea why.

I'm pretty sure that's what migration guides are for :trollface:

HaoK:
Cleaned things up a bit, pushed the goop into `app.UseLegacyAuthentication(AuthenticationSchemeOptions, Func<HttpContext, IAuthenticationSchemeHandler>)` which does a pretty good job sweeping this underneath, all google tests appear unaffected by this change, so I think as far as the new stack goes, either way of registration should behave the same.

test/Microsoft.AspNetCore.Authentication2.Test/Google/GoogleMiddlewareTests.cs
```diff
@@ -1100,7 +1103,7 @@ private static TestServer CreateServer(Action<GoogleOptions> configureOptions, F
                         };
                     });
                     services.AddCookieAuthentication(TestExtensions.CookieAuthenticationScheme);
-                    services.AddGoogleAuthentication(configureOptions);
+                    //services.AddGoogleAuthentication(configureOptions);
```
HaoK: Mixing old and new patterns of registration also don't matter since order no longer has any real meaning to the schemes...

leastprivilege:
+1 for making breaking changes in v2. People need to understand that this API works differently.

[Obsolete] is not good enough IMO

Eilon:
@PinpointTownes @leastprivilege thanks for keeping an eye on this PR and staying involved. Honestly, we're still at the point where we're playing with a bunch of different ideas. I mean, wouldn't it be great if we could preserve compatibility as much as possible *and* have the new features that this design enables? (And eliminate as many problems as possible with the old design?)

Ok, yeah, that's quite optimistic, but that's part of what this PR is about! We've had some design meetings within the team to see what we can come up with.

I agree it could be worse if we make everything *look* compatible but it behaves very differently, potentially introducing serious app bugs.

But we have to balance that with breaking dozens of NuGet packages (*ahem*, **yours**!), and countless apps that use the old system.

And then perhaps the hardest problem of it all: breaking peoples' understanding of how this stuff works!

So, we're not quite sure how this will all end up, but we love what the new design brings, and we've been reading all your reactions to it and trying to account for as much of it as possible.

I hope that once we end up on a design that we like, we'll be able to make it clear that we have an official proposal, and will of course seek a round of feedback on that (probably in a new PR, so that we can start the discussion fresh).

PinpointTownes:
> I agree it could be worse if we make everything look compatible but it behaves very differently, potentially introducing serious app bugs.

Unfortunately, I have the weird feeling that's where this PR is currently heading to.

I haven't played with the latest commits yet, but I'm sure there are also security concerns when using authentication middleware with branching. E.g, in this snippet, won't the cookie middleware execute even for non-API calls? (since it will register the cookie handler regardless of the branching and execute it via the authentication middleware registered by `app.UseJwtBearerAuthentication()`)

```csharp
app.UseJwtBearerAuthentication(...);

app.UseWhen(context => context.Request.Path != "/api", map => {
    map.UseCookieAuthentication();
});

app.UseMvc();
```

> But we have to balance that with breaking dozens of NuGet packages (ahem, yours!), and countless apps that use the old system.

Existing packages targeting ASP.NET Core 1.x will most likely break due to massive changes in .NET Core 2.0 anyway (see https://github.com/aspnet/SignalR/issues/155#issuecomment-275269398 for a concrete case), so if an important design change must be adopted, it's the perfect moment.

> So, we're not quite sure how this will all end up, but we love what the new design brings, and we've been reading all your reactions to it and trying to account for as much of it as possible.

That would be great if you could list the _concrete_ things "the new design brings", 'cause it's not clear to me yet.

HaoK:
Map is an interesting one, and its more or less gone for Auth in the new design.  But that example is another instance of bearer/cookies conflicting and a necessary workaround from the old middleware not playing nicely together.  In the new stack, there is only a tiny bit of overlap, where each handler is given a chance to handle the request, but there's no automatic challenge like conflict that you need to resolve via conditional existence of schemes.  Each scheme is part of the authentication service, and since we don't have support for different services per branch, every scheme will appear in every branch.  This isn't that different today from AuthZ, where there's no way to limit policies to a specific branch either, and since policies can reference schemes, having schemes only exist in some branches is a bit weird anyways.

PinpointTownes:
> Map is an interesting one, and its more or less gone for Auth in the new design.

Not really gone if you decide to keep your "legacy authentication" extensions :trollface: 

> But that example is another instance of bearer/cookies conflicting and a necessary workaround from the old middleware not playing nicely together.

I'm not sure I would consider that as a work around in this particular case, as it's super common to create subsections in your apps, that can possibly use different schemes.

> In the new stack, there is only a tiny bit of overlap, where each handler is given a chance to handle the request, but there's no automatic challenge like conflict that you need to resolve via conditional existence of schemes.

The problem with the sample I mentioned is not with the automatic challenge part, but with automatic authentication: since the cookie handler will be called despite the path restriction, it will populate `HttpContext.User` with the identity corresponding to the authentication cookie. Obviously, your legacy authentication extensions introduce a huge XSRF risk.

HaoK:
UseLegacyAuthentication as it stands doesn't really preserve the old behavior (and its not really trying to, its trying to implement the old API behavior using the new stack as best as it can), its just adding the new authentication middleware (if one hasn't been added already), and any schemes added by one branch will be available to all branches as soon as its run once since the service is shared.

HaoK:
Updated first comment at the top with a cleaned up up to date summary, including a section about some of the things we like about the new design (also my main focus was on fixing things that were existing issues(claims transformation, per request events, using db contexts) and also the things that were hampering our ability to make improvements, aka RIP automatic)

HaoK:
Made ISystemClock a service since this is now easy (unlike when I tried to do this before)

HaoK:
Summary of next iteration after meeting with @davidfowl @Tratcher @Eilon:

- Abandoning interop via UseLegacy, we won't touch the existing HttpAbstractions/Authentication manager
- Add new Auth abstractions package in Http (i.e. Microsoft.Extensions.Authentication.Abstractions?) with the interfaces like IAuthenticationService, IAuthenticationSchemeHandler/Provider etc.
- Take over the existing Security namespaces in a breaking way
- Nuke DefaultChallengeScheme (no naked 401 support)
- Look into ProcessRequest improvements, maybe use routing?

PinpointTownes:
> Abandoning interop via UseLegacy, we won't touch the existing HttpAbstractions/Authentication manager, so old 1.x apps and auth will still work the same

How will that work in practice? Will "legacy" middleware be visible using the new `HttpContext` extensions? E.g will the Identity login template be able to list (and use) legacy social providers?

If the two worlds can't work together, it would be way better to remove the old APIs.

> Nuke DefaultChallengeScheme (no naked 401 support)

:+1: for removing the "naked 401" handling, but if you remove `DefaultChallengeScheme`, this means you'll have to remove the parameterless `ChallengeAsync` overload, which will impact the entire stack (i.e no more parameterless `[Authorize]` attribute).

HaoK:
We can keep the DefaultChallengeScheme, I was just talking about killing the naked 401 logic that is in the new auth middleware.  

For interop basically the only thing we will support is existing 1.x packages continuing to work, so old identity +with old 1.x providers using the old APIs will continue to work.  So in practice, this just means we are preserving the HttpAuthentication interfaces, so old httpContext.Authentication APIs with the old 1.x packages and providers and middleware will continue to do what they always did.  The new stack won't affect that at all since the new IAuthenticationService doesn't depend on the old HttpAbstractions IAuthenticationManager at all.  Any kind of interop between the two will be left as a choose your own adventure exercise :)

But basically old 1.x providers will continue to work in MVC2.0 side by side with the new stack, but upgrading to 2.0 will require moving to the new stack, and all the old middleware/interfaces will be deleted/updated.

I'm doing of the incremental cleanup from the review in this PR, but I'll start a new one when I takeover the existing classes/namespaces in the next iteration.

HaoK:
Regarding identity in 2.0, and everything else, it will be updated to only use the new stack (i.e. enumerating only the authentication schemes by default).  The old auth will still be there if you registered them, but identity, and MVC for that matter won't pick them up.  For MVC/AuthorizeFilter there was some brief discussion about how we would add something like LegacyAuthFilter that you could plug in to restore talking to the old context.Authentication.Xyz.  But [Authorize] would target the new auth system only by default.

PinpointTownes:
So it will be pretty much like using Katana authentication middleware on ASP.NET Core... totally unusable in practice? If even authorization doesn't work with the legacy middleware, there's really no point keeping them in 2.0.

HaoK:
What do you mean about 'keeping them in 2.0'?  The plan is to nuke all the old code in 2.0.  But you can still use old 1.1 packages in 2.0 apps, and that scenario should still work.

PinpointTownes:
> What do you mean about 'keeping them in 2.0'? The plan is to nuke all the old code in 2.0.

Ah great. This point wasn't clear to me :trollface: 

> But you can still use old 1.1 packages in 2.0 apps, and that scenario should still work.

Hum, are you sure about that? The major version increment means things won't be guaranteed to be binary compatible. See https://github.com/aspnet/SignalR/issues/155#issuecomment-275269398 for a concrete case.

HaoK:
Yeah I believe the plan is for most to be binary compatible, but basically the interop should be limited to preserving the old HttpAbstractions authentication namespaces/interfaces.  But 2.0 Security/IIS/WebListener will depend on a new Authentication.Abstractions package which will be in the HttpAbstractions repo but in the Authentication namespace and contain all the new interfaces/extension methods.  We can choose to keep the existing names if we want now (IAuthenticationHandler vs IAuthenticationSchemeHandler etc...  I'll start the new PR tomorrow and we can revisit naming there

HaoK:
This PR is effectively abandoned now, new iteration PR is here: https://github.com/aspnet/Security/pull/1113

PR #1075, opened 2016-12-19
Modify OAuthHandler.cs

SH2015:
Modified in response to this issue :
https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/issues/126


Tratcher:
:-1: we can come up with better control flow than returning null.

Eilon:
The current code provides a more descriptive failure in case of an error, so we would like to keep that behavior.

If you see some particular value in propagating the null value, please log an issue so we can discuss the merits of it.

PR #1077, opened 2017-01-05
Upgrade to VS 2017

natemcmaster:
Upgrades the solution to VS 2017.

Also, upgrades the Microsoft.AspNetCore.ChunkingCookieManager.Sources package to a 'contentFiles' pkg instead of 'shared' sources. cref https://github.com/aspnet/Coherence-Signed/issues/379

cc @Eilon - it's unclear who should review this for this repo. Any suggestions?


src/Microsoft.Owin.Security.Interop/Properties/AssemblyInfo.cs
```diff
@@ -1,15 +1,8 @@
...
 
-using System.Reflection;
-using System.Resources;
 using System.Runtime.InteropServices;
 
-[assembly: AssemblyMetadata("Serviceable", "True")]
-[assembly: NeutralResourcesLanguage("en-us")]
-
 // The following GUID is for the ID of the typelib if this project is exposed to COM
 [assembly: Guid("a7922dd8-09f1-43e4-938b-cc523ea08898")]
```
Tratcher: can this move to the csproj? We don't really need it, but I don't know what removing it would do.

natemcmaster: AFAIK this isn't one of the assembly attributes that can be generated by the SDK. cref https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.GenerateAssemblyInfo.targets

natemcmaster:
Merged https://github.com/aspnet/Security/commit/ba8f85d9726d9810cf9c017a4f51762cd52abd9b

src/Microsoft.AspNetCore.Authentication.Cookies/__TemporarySources__/TaskCacheOfT.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
PinpointTownes: Out of curiosity, shouldn't these files be excluded from source control?

natemcmaster: This is a workaround for 'contentFiles' not yet working in MSBuild. We will be deleting these in a few weeks when we get an upgrade to NuGet that fixes the issue. cref https://github.com/aspnet/Common/issues/180

PinpointTownes: Good to know, thanks!

Issue #1082, opened 2017-01-10
Update IdentityModel dependencies to 5.1.2


Tratcher:
@blowdart @muratg recommended for 1.1.1

muratg:
Putting this in 1.1.1 for the patch release triage.

blowdart:
Not just recommended. Mandatory.

Tratcher:
https://technet.microsoft.com/library/security/3214296

Eilon:
This patch bug is **approved**. Please use the normal code review process w/ a PR and make sure the fix is in the correct branch, then close the bug and mark it as done.

PR #1092, opened 2017-01-19
Update IdentityModel dependencies

Tratcher:
Due to an MSRC. https://github.com/aspnet/Security/issues/1082

I'll do a different change for dev since it has already changed to csproj.


blowdart:
:shipit:

PR #1102, opened 2017-01-26
Update OpenId package versions

pranavkm:


PinpointTownes:
FYI, I suspect the new 2.1.3 IdentityModel bits of being impacted by [another](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/557) strong name issue.

All your builds are currently failing with this kind of error:

> Strong name signature could not be verified.  The assembly may have been tampered with, or it was delay signed but not fully signed with the correct private key. (Exception from HRESULT: 0x80131045)

Tratcher:
Yup, @brentschmaltz is working on it.

brentschmaltz:
issue has been fixed

Issue #1110, opened 2017-02-02
Two consecutive calls to SystemClock can result in smaller values.


unjello:
SystemClock substracts milliseconds to get integral values, but it does not change UtcTicks, because UtcTics have 100-nanoseconds resolution, they are left behind. `636216354734678439` becomes `636216354730008439`. That can in turn cause undefined behavior:

(from Immediate Window, _clock is Microsoft.AspNetCore.Authentication.SystemClock instance)
```
_clock.UtcNow < _clock.UtcNow
false
_clock.UtcNow < _clock.UtcNow
true
```

PR #1111, opened 2017-02-02
Truncate SystemClock to Seconds Precision

benaadams:
Resolves https://github.com/aspnet/Security/issues/1110

And because @blowdart really wants me to make a PR to a repo with Security in its name... 😉

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication/SystemSecondsClock.cs
+++ b/src/Microsoft.AspNetCore.Authentication/SystemSecondsClock.cs
@@ -7,9 +7,9 @@ using System;
 namespace Microsoft.AspNetCore.Authentication
 {
     /// <summary>
-    /// Provides access to the normal system clock.
+    /// Provides access to the normal system clock with precision in seconds.
     /// </summary>
-    public class SystemClock : ISystemClock
+    public class SystemSecondsClock : ISystemClock
     {
         /// <summary>
         /// Retrieves the current system time in UTC.
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
@@ -30,7 +30,7 @@ namespace Microsoft.AspNetCore.Builder
             SlidingExpiration = true;
             CookieHttpOnly = true;
             CookieSecure = CookieSecurePolicy.SameAsRequest;
-            SystemClock = new SystemClock();
+            SystemClock = new SystemSecondsClock();
             Events = new CookieAuthenticationEvents();
         }
 
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
@@ -30,7 +30,7 @@ namespace Microsoft.AspNetCore.Builder
             SlidingExpiration = true;
             CookieHttpOnly = true;
             CookieSecure = CookieSecurePolicy.SameAsRequest;
-            SystemClock = new SystemSecondsClock();
+            SystemClock = new SystemClock();
             Events = new CookieAuthenticationEvents();
         }
 
--- a/src/Microsoft.AspNetCore.Authentication/AuthenticationOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication/AuthenticationOptions.cs
@@ -54,6 +54,6 @@ namespace Microsoft.AspNetCore.Builder
         /// For testing purposes only.
         /// </summary>
         [EditorBrowsable(EditorBrowsableState.Never)]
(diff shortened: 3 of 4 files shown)


blowdart:
I'd consider renaming the class to SystemSecondsClock to be clearer about what it does.

blowdart:
@Tratcher thoughts?

unjello:
I was prepering pull request myself. I fixed it this way:
```
                DateTimeOffset utcNow = DateTimeOffset.UtcNow;
                return utcNow.AddTicks(-(utcNow.Ticks % 10000000));
```
I know I'm biased, but it looks a little faster and cleaner. Maybe a bit more cryptic, but I think it's worth it :)

benaadams:
@unjello is surprisingly messy

This bit`DateTimeOffset.UtcNow.Ticks` is
```
DateTimeOffset.UtcNow
-> new DateTimeOffset(DateTime.UtcNow)
.Ticks
-> ClockDateTime.Ticks
--> new DateTime((m_dateTime + Offset).Ticks, DateTimeKind.Unspecified)
```
And `DateTimeOffset.AddTicks` is
```
DateTimeOffset.AddTicks
-> new DateTimeOffset(ClockDateTime.AddTicks(ticks), Offset)
--> ClockDateTime.AddTicks
---> DateTime AddTicks
-> new DateTimeOffset(long, offset)
--> ValidateOffset
--> new DateTime
--> ValidateDate
---> new DateTime

**Comment by unjello** [comment]
@benaadams ouch. Please apply water to the burned area :) Lesson learned. Simpler the API the more mess inside I guess :)

**Comment by benaadams** [comment]
@unjello most of it should get inlined and compiled away; and the `new DateTimeOffset` will be the same, so the two changes will likely? end up at the same point...

**Comment by PinpointTownes** [comment]
FYI, this issue was discussed last year: https://github.com/aspnet/Security/issues/874.

**Comment by unjello** [comment]
@PinpointTownes thanks for reference. To be honest, I'm surprised ISystemClock is not refactored into either common interfaces and keep your implementation here, or even implementation (full-clock, seconds-clock, testing-clock that allows manual advancement) could be refactored into common space. I personally have few places where I could use it. I discovered the issue because I actually used this clock instead doing my own. I do not necessarily agree with one of the points made in #874 that I should then copy implementation. This stuff is so useful and so quite generic, that there should not be a need to copy it, and even if so - again, it would be nice to at least have an interface in a common namespace.

**Comment by PinpointTownes** [comment]
@unjello I definitely agree with you (I also use it in my own projects). The ASP.NET and BCL teams chatted about this interface a while ago but AFAIK, they've never reached a consensus (https://github.com/aspnet/Caching/issues/7).

**Comment by HaoK** [comment]
I wanted to make this a service in security for 1.0 but it was too much of a hassle with how things were setup in middleware, as part of the 2.0 changes, this might become something you replace in the container which the handlers all pickup that way, instead of the options instance today.

File: src/Microsoft.AspNetCore.Authentication/SystemSecondsClock.cs [hunk 1]
```diff
@@ -20,8 +20,8 @@ public DateTimeOffset UtcNow
             {
                 // the clock measures whole seconds only, to have integral expires_in results, and
                 // because milliseconds do not round-trip serialization formats
-                DateTimeOffset utcNow = DateTimeOffset.UtcNow;
-                return utcNow.AddMilliseconds(-utcNow.Millisecond);
+                DateTime utcNowPrecisionSeconds = new DateTime((DateTime.UtcNow.Ticks / TimeSpan.TicksPerSecond) * TimeSpan.TicksPerSecond, DateTimeKind.Utc);
```
Eilon: Not enough `var`, we cannot accept this otherwise beautiful PR 😄

src/Microsoft.AspNetCore.Authentication/SystemSecondsClock.cs
```diff
@@ -7,9 +7,9 @@
 namespace Microsoft.AspNetCore.Authentication
 {
     /// <summary>
-    /// Provides access to the normal system clock.
+    /// Provides access to the normal system clock with precision in seconds.
     /// </summary>
-    public class SystemClock : ISystemClock
+    public class SystemSecondsClock : ISystemClock
```
Eilon: Also shouldn't rename this, as it's an unnecessary breaking change.

benaadams: Reverted

src/Microsoft.AspNetCore.Authentication/SystemSecondsClock.cs
benaadams: moar var'ed

Eilon:
@HaoK - you seemed most passionate about this bug, can you do final review, then squash & merge?

blowdart:
Obviously a name change needs to go everywhere else its used.

Tratcher:
@blowdart it's not getting renamed.

blowdart:
Darn.

PR #1112, opened 2017-02-02
Fix SystemClock precision errors

### LINKED ISSUE #1110: Two consecutive calls to SystemClock can result in smaller values.

unjello:
Substracting milliseconds left nanoseconds part.

Closes #1110.


unjello:
#1111 does the job far better after all :)

Issue #1061, opened 2016-12-07
Redesign automatic authentication choice mechanism


blowdart:
The current mechanism for choosing which authentication middleware runs by default has been fragile and caused pain when more than one middleware has been set to automatically run, leading to what is, at best, undefined behaviour. The undefined behaviour changed in 1.1, but [some folks were relying on it.](https://github.com/aspnet/Security/issues/1044)

The basic idea behind automatic challenging middleware was that only one middleware should be configured to automatically challenge, however historically this was configured by a property on the individual middleware, and, as middlewares cannot inspect each other, it was too easy to set multiple middlewares to automatically challenge.

The new behaviour should take the decision of what middleware to use to issue a challenge when no middleware is specifically selected out of the hands of the middleware, and into the realm of authorization policy. It's envisaged that the default policy would contain information on the elected middleware to use to issue a challenge, and syntactic sugar added so you can set it as a property on authorization, rather than having to write a complete policy.

This leaves an open question of what happens when Challenge is called manually, without specifying which middleware to use - do we use the one from the default policy, or remove automatic challenge at that point and force a developer to specify the middleware to be used?

blowdart:
Tagging interested parties: @PinpointTownes @leastprivilege @brockallen 
MSFT: @davidfowl @DamianEdwards @HaoK @Eilon @Tratcher

leastprivilege:
https://www.youtube.com/watch?v=GvtGVoS0ObQ

brockallen:
After a bit of experience and reflection how how all of the authentication middleware works today, my sense is that the whole authentication middleware design that was done for katana is not generally appropriate here in ASP.NET Core. IOW, the whole linked list wireup per-request seems unnecessary given that we have DI and that we could inject the main logic (the handlers). There would still be the need for middleware to trigger the handler, but that's about it. API calls like Authenticate/Challenge/Signin/Signout could simply be to a service that was obtained via DI (and not via the handler chain wired up during the request). 

Perhaps then more specialized interfaces make sense to accommodate this (ICookeAuthentication, IExternalAuthentication, etc?) would be necessary for this. The design of a one-size-fits all authentication manager API seems like an unnecessary abstraction.

Not sure how earth shattering this suggestion would be in practice, but again the linked list wireup seems brittle and the source of many problems.

davidfowl:
[...]

Issue #901, opened 2016-07-15
AuthZ 2.0 Improve failure scenarios + [401 vs 403?]


HaoK:
## The two somewhat related scenarios we are trying to address:
1. How to determine which requirements failed and easily display them
2. Remove the automatic challenge behavior/logic and make 401/403 more explicit (somehow)

## How things work today:
1. `bool IAuthorizationService.AuthenticateAsync` only returns sucess or failure, its impossible to deduce what caused the failure.
2. See [AuthorizeFilter.OnAuthorization](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Authorization/AuthorizeFilter.cs#L101).

Basically the caller is responsible for building up the ClaimsPrincipal for authorization.  There's a hint on the policy via the `AuthenticationSchemes` which the caller should use for Authentication and merge together for the final principal.  The first important point here is that we no longer know what claims came from which authentication scheme (claims transformation also is a wrinkle so even if only a single scheme was used, there's no guarantee).  

Next the `IAuthorizationService` checks the requirements from the `AuthorizationPolicy` against the merged principal and if it fails, we challenge all of the schemes.  So today the Challenge(401) vs Forbid(403) behavior is completely left to the implementations of the authentication middleware for how they respond to the challenges.  The default implementation of that behavior is to return 401 if the scheme does not have an authentication ticket, and 403 if it does (i.e. if a cookie/bearer token exists, or the remote identity already has a signed in identity)

## Ideas for improvement:
1. The needed change is to enable getting back a new `AuthorizationResult` that contains the failed requirements, that should be sufficient to enable callers to display error information.  We should also add some mechanism to describe the requirement failure to make it easy to generate this data, maybe a `AuthZFailureDescriber<TRequirement>`?

2. We'd like to try and eliminate the ChallengeBehavior.Automatic and change the meaning of Challenge to always result in Unauthorized/401 behavior.  And Forbid should be called explicitly instead of something that happens to some challenges.  Callers like AuthorizeFilter would need to have enough information to make the choice.

## Scenarios to target:

1. `[Authorize("Administrator")]` protects an action, when a regular user hits the action, an error page stating that the user was forbidden because he was either not in the Adminstrator role, or did not have the Administrator claim (depending on how the policy was implemented)

2. 

## Related issues that should also be addressed: 
https://github.com/aspnet/Diagnostics/issues/316 
https://github.com/aspnet/Security/issues/860
[...]

PR #1113, opened 2017-02-03
Auth 2.0 Iteration 2

HaoK:
Test showing dynamic add/remove of schemes: https://github.com/aspnet/Security/pull/1113/commits/d92d92384f7f6cf2e5e17c975efaa0bd6ab299c0#diff-220317c67a0ea163cf751157519c7f66R29

- Will move this to HttpAbstractions once its baked (maybe the base Authentication package will need to move as well for servers to use the common infrastructure).

cc @Eilon @Tratcher @davidfowl

Follow up items:
- Revisit HandleRequest routing
- Normalize/revisit all of the events (some derive from Control some don't), mixed handling patterns.
- Scheme options validation (things like SignInScheme defaults are initalized in handler, so can't validate on ConfigureTime)

Test holes:
- Need unit tests for base OAuth and new Auth service interfaces, and when UseAuth is missing (RemoteAuth should throw if feature is missing)


HaoK:
Rather than nuking the existing auth package requiring updating everything at once, implementation of the new stack will be in a Authentication.Impl package until everything is done

- Renamed IAuthenticationSchemeHandler => IAuthenticationHandler (taking over the old interface name), shouldn't lose any clarity
- Added IAuthenticationHandlerResolver which has a single method which resolve handler instance for a given scheme + request.

brockallen:
@leastprivilege and I have a TODO item to look over and give feedback. We were hoping to have done this last week, but it didn't happen. Perhaps this weekend or next week. Thanks.

HaoK:
Sure, you guys might want to look at the old PR first for better context/info, and it has more or less the same shape, it will take a bit for me to stitch everything back together again in this PR.

HaoK:
Seems too confusing to have any dependency on the old HttpAuthentication feature, going to copy ChallengeBehavior and AuthenticationProperties instead, since we aren't supporting old and new together.

test/Microsoft.AspNetCore.Authentication.Test/Cookies/CookieMiddlewareTests.cs
```diff
@@ -851,68 +740,85 @@ private Task SignOutAsWrong(HttpContext context)
         }
 
         [Fact]
-        public async Task MapWillNotAffectChallenge()
+        public async Task MapWillAffectChallenge()
```
HaoK: Note: this is a behavior change from old stack to new stack (Map does affect redirect url for challenge now that its not tied to any middleware)

HaoK:
Cookies and tests have been updated to use the new stack, and pull system clock from DI

test/Microsoft.AspNetCore.Authentication.Test/Cookies/CookieMiddlewareTests.cs
PinpointTownes: Be careful, it's most likely not a wanted change: if `/page` is configured as the callback path and the user is actually redirected to `/login/page`, all he'll get is a 404 response, as no login page will handle the request.

I assume it similarly impacts all the RAT-based middleware that use callback paths and have built-in endpoints? (OIDC, OAuth2, Twitter...)

test/Microsoft.AspNetCore.Authentication.Test/Cookies/CookieMiddlewareTests.cs
```diff
@@ -851,68 +740,85 @@ private Task SignOutAsWrong(HttpContext context)
...
             Assert.Equal(HttpStatusCode.Redirect, transaction.Response.StatusCode);
 
             var location = transaction.Response.Headers.Location;
-            Assert.Equal("/page", location.LocalPath);
+            Assert.Equal("/login/page", location.LocalPath);
             Assert.Equal("?ReturnUrl=%2F", location.Query);
         }
 
-        [Fact]
+        [ConditionalFact(Skip = "Revisit, exception no longer thrown")]
```
PinpointTownes: Hum, why? The `InvalidOperationException` is there to inform the dev' that something is wrong with his setup. I don't think it's preferable to silently ignore the operation.

test/Microsoft.AspNetCore.Authentication.Test/Cookies/CookieMiddlewareTests.cs
```diff
@@ -739,11 +739,19 @@ private Task SignOutAsWrong(HttpContext context)
             Assert.Equal(HttpStatusCode.Redirect, transaction2.Response.StatusCode);
         }
 
-        [Fact]
-        public async Task MapWillAffectChallenge()
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public async Task MapWillAffectChallengeOnlyWithUseAuth(bool useAuth)
```
HaoK: FYI @PinpointTownes fixed the map + resolve url behavior to work similar to before (using the path of the single middleware), via a new IAuthenticationFeature like we discussed @Tratcher

samples/OpenIdConnectSample/Properties/launchSettings.json
```diff
@@ -3,8 +3,8 @@
     "windowsAuthentication": false,
     "anonymousAuthentication": true,
     "iisExpress": {
-      "applicationUrl": "http://localhost:42023",
```
Tratcher: revert, these samples need to stay on a specific port.

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationScheme.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Reflection;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public class AuthenticationScheme
+    {
+        public AuthenticationScheme(string name, Type handlerType, Dictionary<string, object> settings)
+        {
+            Name = name ?? throw new ArgumentNullException(nameof(name));
```
Tratcher: FYI: @Eilon has rejected the use of this syntax.

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationScheme.cs
```diff
@@ -0,0 +1,34 @@
...
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public class AuthenticationScheme
+    {
+        public AuthenticationScheme(string name, Type handlerType, Dictionary<string, object> settings)
```
Tratcher: Why not just take an IReadOnlyDictionary?

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationScheme.cs
blowdart: Does that catch string.empty?

src/Microsoft.AspNetCore.Authentication.Abstractions/ChallengeBehavior.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public enum ChallengeBehavior
```
Tratcher: Looking to eliminate this. Let Challenge and Forbidden be separate code paths, and Automatic would be resolved at a higher level.

test/Microsoft.AspNetCore.Authentication.Test/Cookies/CookieMiddlewareTests.cs
HaoK: I think the old test threw because it was doing an AutomaticChallenge followed by an second challenge, which no longer happens in the new stack, I'll have to debug the old test, to see what was triggering it

src/Microsoft.AspNetCore.Authentication.Abstractions/ChallengeBehavior.cs
HaoK: Yeah but we likely can't do this until there's a bunch of corresponding AuthZ changes, we should tackle that in a follow on PR specifically for AuthZ (see https://github.com/aspnet/Security/issues/901)

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationScheme.cs
HaoK: Good point, fixed

src/Microsoft.AspNetCore.Authentication.Abstractions/Microsoft.AspNetCore.Authentication.Abstractions.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <Description>ASP.NET Core common types used by the various authentication components.</Description>
+    <VersionPrefix>1.2.0</VersionPrefix>
+    <TargetFrameworks>net451;netstandard1.3</TargetFrameworks>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageTags>aspnetcore;authentication;security</PackageTags>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.DataProtection" Version="1.2.0-*" />
+    <PackageReference Include="Microsoft.AspNetCore.Http" Version="1.2.0-*" />
```
Tratcher: What did you need from .HTTP? It's unusual for an Abstractions package to have this many concrete dependencies.

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
```diff
@@ -2,19 +2,15 @@
...
 using System;
-using System.ComponentModel;
-using Microsoft.AspNetCore.Authentication;
-using Microsoft.AspNetCore.Authentication.Cookies;
 using Microsoft.AspNetCore.DataProtection;
 using Microsoft.AspNetCore.Http;
-using Microsoft.Extensions.Options;
 
-namespace Microsoft.AspNetCore.Builder
+namespace Microsoft.AspNetCore.Authentication.Cookies
```
Tratcher: We'd used this namespace to improve intellisence in Startup, it's the same namespace the extensions were in. Should these move to the DI namespace?

src/Microsoft.AspNetCore.Authentication.Abstractions/Microsoft.AspNetCore.Authentication.Abstractions.csproj
HaoK: Good catch, these were just carried over from eixsting Authentication.csproj, I switched them to just the Http.Abstractions package and things were fine.

src/Microsoft.AspNetCore.Authentication.Cookies/Events/BaseCookieContext.cs
```diff
@@ -23,5 +24,7 @@ public class BaseCookieContext : BaseContext
         }
 
         public CookieAuthenticationOptions Options { get; }
+
+        public AuthenticationScheme Scheme { get; }
```
Tratcher: Would Scheme go on BaseAuthenticationContext?

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookMiddleware.cs
```diff
@@ -1,89 +0,0 @@
...
-            {
-                throw new ArgumentNullException(nameof(sharedOptions));
-            }
-
-            if (options == null)
-            {
-                throw new ArgumentNullException(nameof(options));
-            }
-
-            if (string.IsNullOrEmpty(Options.AppId))
```
Tratcher: These checks are missing

src/Microsoft.AspNetCore.Authentication.Cookies/Events/BaseCookieContext.cs
HaoK: I tried to bridge the context's between the service and the handlers.  The service doesn't care about the scheme object, only the string, but maybe we should just have a BaseAuthenticationHandlerContext that does have the scheme.  I have a todo in the top comment to make a pass through all of our events/context objects since they have always been a bit of a hodgepodge mess

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookMiddleware.cs
HaoK: Grrr yup, I'll add a test for these as well

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
HaoK: I don't think that's necessary with lambdas, as intellisense knows the type from the signature, the namespace hack was only needed for `new`

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -216,20 +216,10 @@ private CookieOptions BuildCookieOptions()
...
-                await FinishResponseAsync();
-                //await HandleAutomaticChallengeIfNeeded();
-            }
-        }
-
         protected virtual async Task FinishResponseAsync()
         {
             // Only renew if requested, and neither sign in or sign out was called
-            if (!_shouldRefresh || SignInAccepted || SignOutAccepted)
+            if (!_shouldRefresh || SignInCalled || SignOutCalled)
```
HaoK: Test hole here, need to add a test that verifies cookies aren't renewed if SignIn or SignOut is called. (this was broken before this change)

test/Microsoft.AspNetCore.Authentication.Test/Cookies/CookieMiddlewareTests.cs
PinpointTownes: Really great news!

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookMiddleware.cs
HaoK: Moved the checks to Validate in the options which at least causes an exception, but the current behavior isn't perfect, in that it blows up the first time the scheme is used in the auth service.

Tratcher: Weren't we going to call Validate from the UseFooAuth extension too?

HaoK: Sorry I mispoke, it blows up when the Auth service is constructed as the schemes are built when SchemeProvider/HandlerResolvers are built.  So maybe that's not that bad if the auth service itself cannot be resolved due to a configuration error.

HaoK: But the callstack is huge and messy but at least the error message is clear in the argument exception.

Tratcher: Can you share the callstack/exception?

HaoK:
We also had lost the OauthOptions validations too which had no test coverage :/  I added some basic validation tests for those, also compact'ed the test structure, I nuked all the folders that had a single file, so basically only OpenIdConnect has a subfolder in Auth.Test now.

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookMiddleware.cs
HaoK:
```
typeof(System.ArgumentException): The 'ClientId' option must be provided.
Parameter name: ClientId
Stack Trace:
   at Microsoft.AspNetCore.Authentication.OAuth.OAuthOptions.Validate() in C:\GitHub\Security\src\Microsoft.AspNetCore.Authentication.OAuth\OAuthOptions.cs:line 33
   at Microsoft.Extensions.DependencyInjection.AuthenticationServiceCollectionExtensions.<>c__DisplayClass2_0`2.<AddSchemeHandler>b__1(AuthenticationSchemeBuilder b) in C:\GitHub\Security\src\Microsoft.AspNetCore.Authentication.Impl\AuthenticationServiceCollectionExtensions.cs:line 66
   at Microsoft.AspNetCore.Authentication.AuthenticationOptions.AddScheme(String name, Action`1 configureBuilder) in C:\GitHub\Security\src\Microsoft.AspNetCore.Authentication.Abstractions\AuthenticationOptions.cs:line 42
   at Microsoft.Extensions.DependencyInjection.AuthenticationServiceCollectionExtensions.<>c__DisplayClass2_0`2.<AddSchemeHandler>b__0(AuthenticationOptions o) in C:\GitHub\Security\src\Microsoft.AspNetCore.Authentication.Impl\AuthenticationServiceCollectionExtensions.cs:line 56
   at Microsoft.Extensions.Options.OptionsCache`1.CreateOptions()
   at System.Threading.LazyInitializer.EnsureInitializedCore[T](T& target, Boolean& initialized, Object& syncLock, Func`1 valueFactory)
   at Microsoft.Extensions.Options.OptionsCache`1.get_Value()
   at Microsoft.AspNetCore.Authentication.DefaultAuthenticationSchemeProvider..ctor(IOptions`1 options) in C:\GitHub\Security\src\Microsoft.AspNetCore.Authentication.Impl\DefaultAuthenticationSchemeProvider.cs:line 18
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitScoped(ScopedCallSite scopedCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitScoped(ScopedCallSite scopedCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitScoped(ScopedCallSite scopedCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.<>c__DisplayClass17_0.<RealizeService>b__0(ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService[T](IServiceProvider provider)
   at Microsoft.AspNetCore.Authentication.AuthenticationHttpContextExtensions.ChallengeAsync(HttpContext context, String scheme, AuthenticationProperties properties, ChallengeBehavior behavior) in C:\GitHub\Security\src\Microsoft.AspNetCore.Authentication.Abstractions\AuthenticationHttpContextExtensions.cs:line 23
   at Microsoft.AspNetCore.Authentication.AuthenticationHttpContextExtensions.ChallengeAsync(HttpContext context, String scheme, AuthenticationProperties properties) in C:\GitHub\Security\src\Microsoft.AspNetCore.Authentication.Abstractions\AuthenticationHttpContextExtensions.cs:line 20
   at Microsoft.AspNetCore.Authentication.AuthenticationHttpContextExtensions.ChallengeAsync(HttpContext context, String scheme) in C:\GitHub\Security\src\Microsoft.AspNetCore.Authentication.Abstractions\AuthenticationHttpContextExtensions.cs:line 17
   at Microsoft.AspNetCore.Authentication.OAuth.OAuthTests.<>c__DisplayClass0_0.<ThrowsIfClientIdMissing>b__4() in C:\GitHub\Security\test\Microsoft.AspNetCore.Authentication.Test\OAuthTests.cs:line 33
```

Tratcher: That only says OAuth, it doesn't tell you Facebook. And Facebook calls it AppId. This is going to cause confusion.

HaoK: I copied a different exception failure for the above example from the oauth tests, the facebook one says app id and has a facebookoptions.validate at the top.

Tratcher: Ah, good to know

brockallen:
Hi all -- again, no time yet to have a look, but I wanted to let you know how we're integrating with the current security middleware in IdentityServer. In short, we're registering our own authentication middleware to be able to do a decorator pattern like feature to applications hosting us. We'd like to know if we can do something similar in this v2 work. Here's the main handler we have now:

https://github.com/IdentityServer/IdentityServer4/blob/dev/src/IdentityServer4/Hosting/AuthenticationHandler.cs

PinpointTownes:
@brockallen the "authentication middleware" chain is no longer a thing in this prototype, so you won't be able to port your handler as-is. That said, I guess you could achieve something similar using the new `IAuthenticationService`, which is more or less the equivalent of `AuthenticationManager`, but DI-injected.

brockallen:
> I guess you could achieve something similar by replacing IAuthenticationService

Right, but then we'd probably need to decorate whatever implementation is provided by default. Too bad the core DI system is so lacking in so many useful features.

PinpointTownes:
> Too bad the core DI system is so lacking in so many useful features.

There a few workarounds to make the default DI stack (kinda) work with the decorator pattern, but it's indeed not ideal... I guess that's the price to pay for having full DI here :trollface: 

> To be honest, I'm not a huge fan of the "full DI" approach suggested by @HaoK and @brockallen as it will likely introduce a whole new set of issues without really solving the initial one.

https://github.com/aspnet/Security/issues/1061#issuecomment-266092809

brockallen:
Well, the original chained handler design in katana was essentially an attempt at designing a poor man's decorator pattern in a system that had no DI. Moving to a service oriented approach in a DI system with missing features is just changing the surrounding code. I think I agree that it's not addressing the fundamental requirement. I've yet to look into the other changes to see if they've done anything to make the overall design better.

HaoK:
@brockallen it looks like you could just replace the DefaultAuthenticationService with your own derived one, something like this:

```
  public class AuthenticationService : DefaultAuthenticationService, IAuthenticationService 
  {
        private readonly IdentityServerOptions _options;
        private readonly IHttpContextAccessor _context;
        private readonly ISessionIdService _sessionId;
        private readonly ILogger<AuthenticationHandler> _logger;
        private readonly IEventService _events;

        public AuthenticationService(
            IAuthenticationHandlerResolver cache, 
            IClaimsTransformation transform,
            IHttpContextAccessor context, 
            IdentityServerOptions options, 
            ISessionIdService sessionId,
            IEventService events,
            ILogger<AuthenticationService> logger) : base(cache, transform)
        {
            _context = context;
            _options = options;
            _sessionId = sessionId;
            _events = events;
            _logger = logger;
        }

        public override async Task SignInAsync(SignInContext context)
        {
            if (context.AuthenticationScheme == _options.Authentication.EffectiveAuthenticationScheme)
            {
                AugmentContext(context);
                await RaiseSignInEventAsync(context.Principal);
            }
            await base.SignInAsync(context);
        }

        private void AugmentContext(SignInContext context)
        {
            _logger.LogDebug("Augmenting SignInContext");

            context.Principal.AssertRequiredClaims();
            context.Principal.AugmentMissingClaims();

            _sessionId.CreateSessionId(context);
        }

        public override async Task SignOutAsync(SignOutContext context)
        {
            if (context.AuthenticationScheme == _options.Authentication.EffectiveAuthenticationScheme)
            {
                _sessionId.RemoveCookie();
                await RaiseSignOutEventAsync();
            }
            await base.SignOutAsync(context);
        }

        // Would need to move to your own middleware most likely
        internal async Task InitAsync()
        {
            var auth = GetAuthentication();
            _handler = auth.Handler;
            auth.Handler = this;

            await _sessionId.EnsureSessionCookieAsync();
        }

        internal void Cleanup()
        {
            var auth = GetAuthentication();
            auth.Handler = _handler;
        }

        // No longer needed, DI registration
        IHttpAuthenticationFeature GetAuthentication()
        {
            var auth = _context.HttpContext.Features.Get<IHttpAuthenticationFeature>();
            if (auth == null)
            {
                auth = new HttpAuthenticationFeature();
                _context.HttpContext.Features.Set(auth);
            }
            return auth;
        }

        private async Task RaiseSignInEventAsync(ClaimsPrincipal principal)
        {
            await _events.RaiseLoginEventAsync(principal);
        }

        private async Task RaiseSignOutEventAsync()
        {
            var principal = await _context.HttpContext.GetIdentityServerUserAsync();
            if (principal.IsAuthenticated())
            {
                await _events.RaiseLogoutEventAsync(principal);
            }
        }
    }
   }
```

brockallen:
Sure, but given our design and the design of the new auth service, who knows if the hosting app has chosen to do their own... so we'd need to decorate whatever is already registered. But in any event, good news that it won't be all that different that what we're faced with today. Thanks.

HaoK:
Yeah so the idea is the old `IAuthenticationHandler` is left alone for now and will be unchanged, the new stack basically will ignore the old stack, and can live alongside the old, they don't interact with each other, so you could potentially support both the old and new at the same time if you desired.

HaoK:
Updated adding a test exercising dyanmically adding and removing schemes, was able to remove the ugly ResolverFunc with new logic that just tries to resolve the service against RequestServices, and if that fails uses ActivatorUtilities.CreateInstance instead.  Which means the handlers don't have to be registered in advance unless they want to have a non scoped lifetime.

HaoK:
Added new flag on AuthenticationScheme.CanHandleRequests which is what the AuthenticationMiddleware will look at to build the list of handlers to give a chance during the request cycle.  RemoteAuth handlers are set to true (JwtBearer/Cookies set this flag to false).

Tratcher:
Most of the middleware handle requests on a specific route. Should middleware register a route with the service?

src/Microsoft.AspNetCore.Authentication.Impl/AuthenticationServiceCollectionExtensions.cs
```diff
@@ -46,8 +45,7 @@ public static IServiceCollection AddAuthentication(this IServiceCollection servi
             return services;
         }
 
-        // REVIEW: rename to just AddScheme?
-        public static IServiceCollection AddSchemeHandler<TOptions, THandler>(this IServiceCollection services, string authenticationScheme, Action<TOptions> configureOptions)
+        public static IServiceCollection AddScheme<TOptions, THandler>(this IServiceCollection services, string authenticationScheme, Action<TOptions> configureOptions, bool canHandleRequests)
```
PinpointTownes: What's the point of doing that? Request handling is something the handlers do by implementing `HandleRequestAsync`, it shouldn't be configurable. Specially by the end user.

HaoK: This is the infrastructure method.  AddGoogle/Cookies/FacebookAuth use this method to register themselves and set the flag.  The idea was to make it opt in for the schemes to get called by the middleware, but as @Tratcher mentioned, next I'll probably scope the blanket callback the handlers get today, and instead they would register their callback paths and the middleware would do some ghetto routing, instead of this opt in bit..  You'd still be able to do more advanced routing by replacing the AuthenticationMiddleware (or adding another customer middleware).  The motivation is to avoid creating the handlers for all the schemes until they are needed/called.

HaoK:
How does something like `IEnumerable<PathString> CallbackPaths` on AuthenticationScheme (most would only have one, OIDC has two for sign in + sign out).  They don't have to be unique, and the middleware will do the routing to the all of the appropriate scheme's HandleRequest that matched.  We'd still keep the bool signifying whether we should stop processing the request.

src/Microsoft.AspNetCore.Authentication.Impl/AuthenticationServiceCollectionExtensions.cs
PinpointTownes: This could potentially lead to really weird issues if the users directly called `AddScheme()` without using the appropriate `canHandleRequests`, but since it's justified by perf', why not... :sweat_smile:

PinpointTownes: Please don't make your custom routing logic too hard to override.

We use a specific `MatchEndpoint` event in ASOS that allows users to dynamically determine the type of request/endpoint, and I really want to preserve it in the new world: https://github.com/aspnet-contrib/AspNet.Security.OpenIdConnect.Server/blob/dev/src/AspNet.Security.OpenIdConnect.Server/OpenIdConnectServerHandler.cs#L27-L170.

HaoK: The new logic will be pushed shortly, I'm just making sure all the tests still work...  

The default routing is basically RemoteOptions single CallbackPath has become many CallbackPaths(to support OIDC's two). canHandleRequests is gone, but there's a new helper extension called AddRemoteScheme, that adds the callback paths to the scheme.

The middleware will call the IAuthenticationSchemeProvider's GetRequestHandlerSchemes with the request path to get back the list of schemes to call ProcessRequest on.  The default implementation just maps the path string to schemes.  

Looking the ASOS, you would just need to add them all as callback events, and should be good to go I think.

HaoK:
Updated: schemes must now explicitly register callback paths in their scheme data.  Schemes ProcessRequest will only be called if Request.Path matches a callback path they specify.

Will probably remove HandleRemoteAuthenticate since its no longer needed.  Also will change the base ProcessRequest to throw now that it shouldn't be invoked anymore

src/Microsoft.AspNetCore.Authentication.Impl/AuthenticationServiceCollectionExtensions.cs
PinpointTownes: > Looking the ASOS, you would just need to add them all as callback events, and should be good to go I think.

You mean as callback path? In this case, this won't work. The `MatchEndpoint` event MUST be called for every request to allow the developer to decide whether it's a request that should be handled or not, depending on his own policy (not just the path).

Not to mention that I ultimately want to support URLs like /connect/register/[dynamic routing parameter]).

HaoK: So right now you can accomplish that with a dedicated custom ASOS middleware that always calls into that schemes process request, or you can override the IAuthenticationSchemeProvider's GetRequestHandlerScheme method to always return the ASOS scheme for all requests.

PinpointTownes: Replacing `IAuthenticationSchemeProvider` is definitely not an option (it will be broken if the user decides to register his own instance).

Having to create a custom middleware just for that seems weird. I thought the whole point of this PR was to remove the "authentication middleware" concept (except the new global one, of course)? :open_mouth:

brockallen: > Replacing IAuthenticationSchemeProvider is definitely not an option (it will be broken if the user decides to register his own instance).

So many workarounds in various ASP.NET features due to a lack of a robust DI system.

HaoK: The idea was to move Auth to be a service, which has more or less been accomplished.  We wanted to move away from invoking all handlers for all requests by default (for perf and to work better with multi-tenancy), while not blocking the ability for a scheme to inspect all requests.

If replacing the default `IAuthenticationSchemeProvider` is not an option, then you are only left with adding your own call to `IAuthenticationSchemeProvider.GetSchemeAsync("ASOS").ProcessRequestAsync()` somewhere, probably via middleware

If this is ends up being a common scenario, we can always add a new options flag like InvokeAllSchemesForAllRequests (aka SlowMoMode = true) for the single Auth middleware to call them all like before.

HaoK: Or maybe we just need a slightly less ghetto routing point that's separate from IAuthenticationSchemeProvider so you can replace that without affecting other things.

PinpointTownes: Sounds like a real PITA just for an hypothetical perf' gain that wasn't even measured 😄

HaoK: Well its kind of required for this https://github.com/aspnet/Security/issues/1132 once we commit to letting people add hundreds/thousands of schemes, we can't really call them all for every request.

PinpointTownes: Why would you add hundreds/thousands of schemes? It doesn't make much sense, even for multi-tenant scenarios (in this case, you'd want a single scheme handler, but with potentially unlimited options, e.g coming from a DB)

brockallen: > Why would you add hundreds/thousands of schemes

I have a customer that uses ws-fed with ~2K external STSs. We could not do it with Katana, so the protocol work had to be done by hand.

PinpointTownes: By scheme, I mean scheme handler. In your case, you'd have only one scheme handler (WS-Fed), with 2K different sets of options, resolved dynamically.

HaoK: The scheme settings are all singletons right now, so if each tenant has a different ClientId/Secret, they will need their own scheme.

HaoK: We have a whole new can of worms once we try to enable reading options from the DB, as we have no async, and lots of our assumptions (options being singleton) go out the window as soon as we start to do dynamic per request options binding (against databases no less...and we are using super slow reflection binding code against configuration as well)...

PinpointTownes: @HaoK point taken, tho' I'm not sure why you'd want to support scenarios with "hundreds/thousands of schemes" if there's no real support in the options stack (I'm not sure having hundreds of schemes + hundreds of handlers + hundreds of options instances is really _scalable_).

Not being able to invoke my handler for every path is really annoying for me. As a compromise, would you consider bringing back your per-scheme "promiscuous mode"? (aka `CanHandleRequest` :smile:)

Tratcher:
How terrible would it be to pull in actual routing?

HaoK:
Well we could add something like "routes.MapAuthenticationRoutes()".  We'd still have to have some internal plumbing in the schemes to store the routes for this method to consume.  So the required startup pattern would then become:

```
    app.UseRouter(routes => routes.MapAuthenticationRoutes());
    app.UseAuthentication();
```

Doesn't look too unreasonable, but is it reasonable for us to require routing for auth to work?

Tratcher:
Not bad. How well would that work with dynamically added/removed auth schemes?

HaoK:
I'm not sure, depends on how well routing supports dynamically adding routes :)  Let me see

PinpointTownes:
Forking my comment as it was hidden by GitHub:

@HaoK point taken, tho' I'm not sure why you'd want to support scenarios with "hundreds/thousands of schemes" if there's no real support in the options stack (I'm not sure having hundreds of schemes + hundreds of handlers + hundreds of options instances is really _scalable_).

Not being able to invoke my handler for every path is really annoying for me. As a compromise, would you consider bringing back your per-scheme "promiscuous mode"? (aka `CanHandleRequest` :smile:)

HaoK:
I'm not an routing expert, but it looks if we forced a pattern on the callback paths to fit into a route constraint, say `/auth/signin/[scheme]` we could easily plug in something that dispatched dynamically.  There might be other ways as well.

HaoK:
@PinpointTownes its going to end up being a question of how easy we want to make it to have a global handler like that.  We certainly could have a handleAllRequests flag on AuthenticationScheme that defaults to false.

HaoK:
Started cleaning up some of the legacy handler processing. 

`AuthenticateResult.Nothing/None()` is the replacement for `Skip()` since we no longer really 'skip' anything.  The concept only exists a bit in the Remote/ProcessRequest codepath which keeps the concept of ResponseHandled/StopProcessing (the new name for context.SkipToNextMiddleware() which now returns a `None()/Nothing` result

I'm considering introducing a new ProcessRequestResult which would let us elimate the funky request processing that shows up in AuthenticateResult today since that's the main return type for Auth as a whole.

PinpointTownes:
> We certainly could have a handleAllRequests flag on AuthenticationScheme that defaults to false.

That would be much appreciated :sweat_smile: 

> The concept only exists a bit in the Remote/ProcessRequest codepath which keeps the concept of ResponseHandled/StopProcessing (the new name for context.SkipToNextMiddleware() which now returns a None()/Nothing result

Hum, but then, what's the difference between `HandleResponse()` and `StopProcessing()`?

HaoK:
Difference between HandledResponse and StopProcessing is whether ProcessRequest returns true/false, same as today's handledResponse/Skip.  Controls whether the middleware continues to invoke other handlers or not via the ghetto routing :)

PinpointTownes:
I'm confused now.

When `StopProcessing` is called from an event, `HandleRequestAsync` is supposed to return `false`, right? (e.g https://github.com/aspnet/Security/blob/d80b0d3f26a85bfa2125fe4f9003fcad8cf851b9/src/Microsoft.AspNetCore.Authentication.Impl/RemoteAuthenticationHandler.cs#L116)

Yet, returning `false` asks the authentication middleware to continue invoking the remaining authentication middleware and possibly, invoking the rest of the ASP.NET Core pipeline: https://github.com/aspnet/Security/blob/d80b0d3f26a85bfa2125fe4f9003fcad8cf851b9/src/Microsoft.AspNetCore.Authentication.Impl/AuthenticationMiddleware.cs#L49-L52

So AFAICT, your new `StopProcessing()` method actually means "skip the rest of the authentication handler but continue invoking the other middleware and the rest of the pipeline"? IMHO, `Skip` was way clearer.

HaoK:
Fair enough, it was mostly the SkipToNextMiddleware that I was renaming, since the emphasis should just be that its skipping any further processing in this handler.  What happens after that is the normal execution.

We can keep it as Continue/Handled/Skip as the three states which correspond to (false, true, false for the AuthMiddleware invoking the rest of the pipeline)

HaoK:
The confusing part is that StopProcessing is the correct name from the perspective of the individual handler's usage of this state, since it lives in the BaseControlContext.  I'll leave that method name still as ProcessingCompleted since for both HandledResponse + Skipped, that means true for the purposes of the handler who is using the context.  All of this smells a bit to me which is why I'm playing around with this area right now...

HaoK:
Ok I revereted the StopProcessing state back to Skip.  So the net result of these changes are:

AuthenticateResult.Skip => AuthenticateResult.None to signify no failure, no ticket, nothing
BaseControlContext.CheckEventResult => ProcessingCompleted which is what all of the handlers call on event contexts to determine if they should stop procesing.
Skip remains the verb on the event context objects.

HaoK:
Note: Identity can consider breaking the dependency on Http with this change, as it can switch to using IAuthenticationService directly without going through HttpContext.  The only other dependency would be on using the RequestAborted as the default cancellation token.  I'm not sure anyone is actually using that functionality the cancellation tokens today, so we could either remove it, or make it settable in some way.

HaoK:
Some implications if we try to use the routing middleware instead of our own ghetto routing:
- How to expose the route configuration to the schemes /signin-{Scheme} by default? Scheme a required paramater and AuthHandler would simply lookup the scheme handler and invoke ProcessRequest
```
    schemeBuilder.AddRoute("/signin-{Scheme}"); // Route that would work for all schemes?
    schemeBuilder.AddRoute("/fixedCustom"); // A fixed route would specify the current scheme
```

- We'd have to find some way to plug into the mvc routes, i.e.
```
    app.UseAuthentication();
    app.UseMvc(routes => {
         routes.MapRoutes("mvcstuff");
         routes.MapAuthenticationRoutes(); // uses the configured auth schemes to add routes
    });
```

Or Without MVC, this overload could add routing and configure it?
```
    app.UseAuthentication(routes => { }); // automatically would call routes.MapAuthenticationRoutes()
```

- We'd lose the ability to 'Skip' and let the normal request processing continue.  Once routing selects a route handler, the middleware chain is broken.  So there would be no support for an auth handler to inspect all requests (have to add your own middleware)

HaoK:
See https://github.com/aspnet/Security/pull/1144 for an iteration using IOptionsFactory which provides named options (and potentially validation).  It also split the ProcessRequest into a different interface IAuthenticationRequestHandler which is used instead of CallbackPaths to determine which schemes get called by the authentication middleware

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationScheme.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Reflection;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public class AuthenticationScheme
+    {
+        public AuthenticationScheme(string name, Type handlerType, IEnumerable<PathString> callbackPaths, IReadOnlyDictionary<string, object> settings)
+        {
+            Name = name ?? throw new ArgumentNullException(nameof(name));
```
Tratcher: @davidfowl has rejected any use of this new syntax. @Eilon has not made an official ruling...

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationScheme.cs
```diff
@@ -0,0 +1,38 @@
...
+            HandlerType = handlerType ?? throw new ArgumentNullException(nameof(handlerType));
+            Settings = settings ?? throw new ArgumentNullException(nameof(settings));
+            CallbackPaths = callbackPaths ?? Enumerable.Empty<PathString>();
+        }
+
+        public string Name { get; }
+        public Type HandlerType { get; }
+
+        /// <summary>
+        /// If true, the AuthenticationMiddleware will call the handler's HandleRequestAsync method.
```
Tratcher: Wrong comment?

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationScheme.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// </summary>
+        public IEnumerable<PathString> CallbackPaths { get; }
+
+        // Holds things like the configured options instances for the handler
+        // Also replacement for AuthenticationDescription
+        public IReadOnlyDictionary<string, object> Settings { get; }
```
Tratcher: This is really weird. Are we using it for anything besides Options and Description? Why not just expose those as first class properties?

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationTicket.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        /// <summary>
+        /// Gets the authentication type.
+        /// </summary>
+        public string AuthenticationScheme { get; private set; }
+
+        /// <summary>
+        /// Gets the claims-principal with authenticated user identities.
+        /// </summary>
+        public ClaimsPrincipal Principal{ get; private set; }
```
Tratcher: formatting

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationTicket.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AuthenticationTicket"/> class
+        /// </summary>
+        /// <param name="principal">the <see cref="ClaimsPrincipal"/> that represents the authenticated user.</param>
+        /// <param name="properties">additional properties that can be consumed by the user or runtime.</param>
+        /// <param name="authenticationScheme">the authentication middleware that was responsible for this ticket.</param>
+        public AuthenticationTicket(ClaimsPrincipal principal, AuthenticationProperties properties, string authenticationScheme)
+        {
+            AuthenticationScheme = authenticationScheme;
+            Principal = principal ?? throw new ArgumentNullException(nameof(principal));
```
Tratcher: rejected syntax

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationToken.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public class AuthenticationToken
+    {
+        public string Name { get; set; }
+        public string Value { get; set; }
```
Tratcher: Make immutable and set via constructor?

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            get
+            {
+                return _schemes;
+            }
+        }
+
+        public IDictionary<string, AuthenticationSchemeBuilder> SchemeMap { get; } = new Dictionary<string, AuthenticationSchemeBuilder>(); // case sensitive?
+
+        public void AddScheme(string name, Action<AuthenticationSchemeBuilder> configureBuilder)
```
Tratcher: Nevermind, I found it on IAuthenticationSchemeProvider

HaoK: Yeah its a bit weird to remove it here, but we could add it if some one wanted to rip out/replace someone else's scheme during configure services timeframe...

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs
```diff
@@ -0,0 +1,69 @@
...
+                throw new InvalidOperationException("Scheme already exists: " + name);
+            }
+
+            var builder = new AuthenticationSchemeBuilder(name);
+            configureBuilder(builder);
+            _schemes.Add(builder);
+            SchemeMap[name] = builder;
+        }
+
+        public void ConfigureScheme(string name, Action<AuthenticationSchemeBuilder> configureBuilder)
```
HaoK: Its for identity/other people to configure schemes...  this uglyness is why I went with named options, you should probably look at https://github.com/aspnet/Security/pull/1144 which is effectively iteration 3 and the current branch I'm working on.  I'll address these issues overthere

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationScheme.cs
HaoK: Many of the comments were already obsolete, we'll need to do a full scrub as part of this

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationScheme.cs
HaoK: No, but I nuked AuthenticationDescription which was similar and unused mostly.  We can put the DisplayName as a property on the scheme itself or inside of here like Description used to...

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// </summary>
+        public IEnumerable<AuthenticationSchemeBuilder> Schemes
+        {
+            get
+            {
+                return _schemes;
+            }
+        }
+
+        public IDictionary<string, AuthenticationSchemeBuilder> SchemeMap { get; } = new Dictionary<string, AuthenticationSchemeBuilder>(); // case sensitive?
```
HaoK: Yeah I'm not sure what it should be, do we want to be case sensitive?  Seems more annoying than helpful

samples/SocialSample/Startup.cs
```diff
@@ -329,20 +330,20 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
...
                 // This is what [Authorize] calls
-                // var user = await context.Authentication.AuthenticateAsync(AuthenticationManager.AutomaticScheme);
+                // var user = await context.AuthenticateAsync(AuthenticationManager.AutomaticScheme);
 
                 // This is what [Authorize(ActiveAuthenticationSchemes = MicrosoftAccountDefaults.AuthenticationScheme)] calls
-                // var user = await context.Authentication.AuthenticateAsync(MicrosoftAccountDefaults.AuthenticationScheme);
+                // var user = await context.AuthenticateAsync(MicrosoftAccountDefaults.AuthenticationScheme);
 
                 // Deny anonymous request beyond this point.
                 if (user == null || !user.Identities.Any(identity => identity.IsAuthenticated))
```
HaoK: its basically equivalent, either services.AuthenticateAsync(context), or context.AuthenticateAsync(), the context methods are just doing GetRequiredService

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticateResult.cs
```diff
@@ -0,0 +1,90 @@
...
+            }
+            return new AuthenticateResult() { Ticket = ticket };
+        }
+
+        public static AuthenticateResult Handle()
+        {
+            return new AuthenticateResult() { Handled = true };
+        }
+
+        public static AuthenticateResult None()
```
HaoK: Skip makes sense when its part of the pipeline, like for handle request in the remote case.  It doesn't really make sense when you ask for an identity to get skip...  Empty or Nothing is really what we want here, basically its when there's No cookie, or no token...  NotFound, NoPrincipal also work

samples/SocialSample/Startup.cs
```diff
@@ -280,16 +279,18 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
...
 
                     context.Response.ContentType = "text/html";
                     await context.Response.WriteAsync("<html><body>");
                     await context.Response.WriteAsync("Choose an authentication scheme: <br>");
-                    foreach (var type in context.Authentication.GetAuthenticationSchemes())
+                    var schemeProvider = context.RequestServices.GetRequiredService<IAuthenticationSchemeProvider>();
+                    foreach (var provider in await schemeProvider.GetAllSchemesAsync())
                     {
-                        await context.Response.WriteAsync("<a href=\"?authscheme=" + type.AuthenticationScheme + "\">" + (type.DisplayName ?? "(suppressed)") + "</a><br>");
+                        // REVIEW: we lost access to display name (which is buried in the handler options)
```
HaoK: Yeah its fallen into a gap, since they hang off the scheme options, but those are configured too late, the schemes need to be registered as part of ConfigureServices.  We can just have the handler set the schemes display name after its options are initialized i guess

src/Microsoft.AspNetCore.Authentication.Impl/DefaultAuthenticationService.cs
```diff
@@ -0,0 +1,139 @@
...
+            if (handler == null)
+            {
+                throw new InvalidOperationException($"No authentication handler is configured to handle the scheme: {authenticationScheme}");
+            }
+
+            var signOutContext = new SignOutContext(httpContext, authenticationScheme, properties);
+            await handler.SignOutAsync(signOutContext);
+        }
+
+        public virtual Task SignInAsync(HttpContext httpContext, string authenticationScheme, ClaimsPrincipal principal)
```
Tratcher: Why isn't this an extension?

src/Microsoft.AspNetCore.Authentication.Impl/DefaultAuthenticationService.cs
```diff
@@ -0,0 +1,139 @@
...
+
+            if (principal == null)
+            {
+                throw new ArgumentNullException(nameof(principal));
+            }
+
+            return SignInAsync(httpContext, authenticationScheme, principal, properties: null);
+        }
+
+        public virtual Task ForbidAsync(HttpContext httpContext, string authenticationScheme)
```
Tratcher: Extension?

src/Microsoft.AspNetCore.Authentication.Impl/DefaultClaimsTransformation.cs
```diff
@@ -0,0 +1,19 @@
...
+
+using System.Security.Claims;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    /// <summary>
+    /// Default claims transformation is a no-op.
+    /// </summary>
+    public class DefaultClaimsTransformation : IClaimsTransformation
```
Tratcher: NoOpClaimsTransformation

src/Microsoft.AspNetCore.Authentication.Impl/DefaultAuthenticationService.cs
```diff
@@ -0,0 +1,139 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public class DefaultAuthenticationService : IAuthenticationService
```
Tratcher: AuthenticationService

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs
Tratcher: They have always been Ordinal. No reason to change them right now.

src/Microsoft.AspNetCore.Authentication.Impl/DefaultAuthenticationSchemeProvider.cs
```diff
@@ -0,0 +1,133 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    /// <summary>
+    /// Builds the actual AuthenticationScheme instances from the AuthenticationOptions.
+    /// </summary>
+    public class DefaultAuthenticationSchemeProvider : IAuthenticationSchemeProvider
```
HaoK: Good call, dropped Default prefix across the board

HaoK:
Replaced with https://github.com/aspnet/Security/pull/1151

samples/SocialSample/Startup.cs
```diff
@@ -45,100 +44,68 @@ public Startup(IHostingEnvironment env)
...
+            services.AddFacebookAuthentication(o =>
             {
-                AppId = Configuration["facebook:appid"],
-                AppSecret = Configuration["facebook:appsecret"],
-                Scope = { "email" },
-                Fields = { "name", "email" },
-                SaveTokens = true,
+                o.AppId = Configuration["facebook:appid"];
+                o.AppSecret = Configuration["facebook:appsecret"];
+                o.Scope.Clear();
```
HaoK: Clears removed

src/Microsoft.AspNetCore.Authentication.Impl/DefaultAuthenticationService.cs
HaoK: Yeah unneeded these already exists as extensions on httpContext, I'll remove

src/Microsoft.AspNetCore.Authentication.Impl/DefaultAuthenticationService.cs
HaoK: Yup nuked

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationScheme.cs
Eilon: We did make a ruling on this and we hate it. No throw expressions please 😄

HaoK: Blame VS :)  I'll expand these again

samples/SocialSample/Startup.cs
```diff
@@ -329,20 +330,20 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
                 var user = context.User;
 
                 // This is what [Authorize] calls
-                // var user = await context.Authentication.AuthenticateAsync(AuthenticationManager.AutomaticScheme);
+                // var user = await context.AuthenticateAsync(AuthenticationManager.AutomaticScheme);
 
                 // This is what [Authorize(ActiveAuthenticationSchemes = MicrosoftAccountDefaults.AuthenticationScheme)] calls
```
HaoK: That was a name that slipped through, was supposed to be renamed to AuthenticationScheme to match the policy :(

PR #1117, opened 2017-02-07
Change method name from Success to Succeed in AuthenticateResult

cnblogs-dudu:
`AuthenticateResult.Success(ticket);` -> `AuthenticateResult.Succeed(ticket);`


Tratcher:
We're unlikely to accept this as it's a breaking change to the public API surface with no functional benefits.

Issue #969, opened 2016-09-10
Google handler should include claim for profile image url


hempels:
Google returns the user's profile url as well as the profile image url. Currently only the profile url is included as a claim. The image is useful data and should be included as a claim also.

Eilon:
We generally don't include extra data because not all apps need them, and can make the cookie size get excessively large. However, we made it fairly easy to add the right scopes (if necessary), and then map the claims into the data you request.

hempels:
It is possible to map the claims, but having gone through the process, I'm not convinced it's easy. The method is not at all intuitive and is entirely undocumented apart from sample code. While I applaud the effort, as an end user the current solution is far too time consuming. I'm finding this to be fairly consistently the case with ASP.NET Core. My hope is that some of this prodding will prompt you guys to consider not just whether you've made something possible, but also whether you've made it useable.

Eilon:
@hempels understood, so I'm re-opening this issue. Can you give an example of what you found to be not easy, and perhaps propose a syntax that you would find to be easier to use?

muratg:
We could potentially consider adding a feature where a user app can map a JSON filed to a claim to make this easier.

Tratcher:
Idea:
`options.AddClaimMap(jsonKey, claimName, claimType);`

HaoK:
We should just generalize the functionality that's currently hardcoded in Facebook/Google handler that's copying claims over into the default claim map, and let that be the source of truth for what's done before the creating ticket event is fired.

Tratcher:
Similar issue mapping custom challenge parameters: https://github.com/aspnet/Security/issues/990

StummeJ:
@hempels do you remember what sample code explained how to do this custom mapping?

Issue #1024, opened 2016-11-04
Reduce claims in ClaimsIdentity after completing OIDC protocol legs


brentschmaltz:
OIDC is a multi-leg protocol that can have three legs ( code -> access_token -> user_info) flowing through the three endpoints. Authorize, Token and UserInfo. The IdentityProvider uses the payload to provide links between the legs. Once the sequence of messages is complete and deemed valid, some of those claims have little value, such as: nonce, at_hash, c_hash. These could be removed from the ClaimsIdentity presented to the application layer. This would reduce the size of cookies.

Tratcher:
why are those things even claims? They're not consumed as claims in the ODIC middleware are they?

brockallen:
These things are part of the id_token payload (which are co-mingled with the user's claims) and have been expressed as claims ever since katana.

PinpointTownes:
> and have been expressed as claims ever since katana.

Because essentially, they are, as the JWT format doesn't make any distinction between protocol claims (nonce, expiration date, access token/code hashes, etc.) and user claims.

brockallen:
> Because essentially, they are, as the JWT format doesn't make any distinction between protocol claims (nonce, expiration date, access token/code hashes, etc.) and user claims.

Which is exactly what this issue is requesting. :)

HaoK:
Would hooking up a claims transformation to strip out the unwanted claims be enough for this scenario?

PinpointTownes:
@HaoK wouldn't claims transformation come too late? The OIDC/remote middleware immediately calls the cookie middleware via `SignInAsync`, so I'm not sure you can plug the CT middleware here.

Personally, I'm not sure the OIDC middleware is the best place for this change. I think it would be better to handle that directly in the JWT middleware, using a specific validation options.

HaoK:
It's probably a gap right now that we don't have any claims transformation involved in the remote authentication handler flows before sign in passes the claims to another scheme...

HaoK:
claims transformation is always needing love :(

PinpointTownes:
> claims transformation is always needing love :(

We all need love :love:

That said, if it's a feature you like, it should be quite easy to add it. We just need to call the transformer here: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication/ClaimsTransformationHandler.cs#L67

Tratcher:
@PinpointTownes that only works if you put the CTM in the pipeline between cookies and OIDC.

[...]

PR #1124, opened 2017-02-15
Make claim mapping more configurable

Tratcher:
#969 
@HaoK @blowdart @PinpointTownes 

Thoughts?


PinpointTownes:
> Thoughts?

Not sure I like it, but I can't think of a better approach, so... :trollface:

Tratcher:
I was looking at how this would apply to OIDC and realized that right now all possible data from the user-info is being added as claims. No wonder OIDC has cookie size issues. Would it make sense to replace that system with this new explicit mapping? It would likely require a little more setup for devs, and I don't know if there are any default mappings it would make sense to include.

IdentityModel already includes all of the claims from the JWTs, right?

/cc: @brentschmaltz

HaoK:
Not in love with the resolver name, I'll chew on this a bit

src/Microsoft.AspNetCore.Authentication.Google/GoogleOptions.cs
```diff
@@ -25,6 +27,13 @@ public GoogleOptions()
             Scope.Add("openid");
             Scope.Add("profile");
             Scope.Add("email");
+
+            ClaimResolvers.Add(ClaimTypes.NameIdentifier, "id");
+            ClaimResolvers.Add(ClaimTypes.Name, "displayName");
+            ClaimResolvers.AddNested(ClaimTypes.GivenName, "name", "givenName");
+            ClaimResolvers.AddNested(ClaimTypes.Surname, "name", "familyName");
```
HaoK: What does nested do exactly?

HaoK:
Yeah so my initial instinct is that this feels pretty heavy, but maybe if we just tweak a few names it won't feel so complicated.

PinpointTownes:
> IdentityModel already includes all of the claims from the JWTs, right?

Yep.

Related ticket: https://github.com/aspnet/Security/issues/1024

src/Microsoft.AspNetCore.Authentication/ClaimResolver.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Security.Claims;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public abstract class ClaimResolver<T>
```
HaoK: Is this going to be used outside of OAuth?  Should we target this against OAuth/JObject for now?

src/Microsoft.AspNetCore.Authentication.Google/GoogleOptions.cs
HaoK: Just brainstorming some different ways to express this:
```
   JObjectTransform.Map(ClaimTypes.NameIdentifier, "id");
   JObjectTransform.Map(ClaimTypes.Name, "displayName");
   JObjectTransform.MapNested(ClaimTypes.GivenName, "name", "givenName");
   JObjectTransform.MapNeste(ClaimTypes.Surname, "name", "familyName");
```

HaoK: `JsonClaimsMap.Map[Nested]`

HaoK: `ClaimsResolver.Map`

src/Microsoft.AspNetCore.Authentication.MicrosoftAccount/MicrosoftAccountOptions.cs
```diff
@@ -30,9 +30,7 @@ public MicrosoftAccountOptions()
             ClaimResolvers.Add(ClaimTypes.Name, "displayName");
             ClaimResolvers.Add(ClaimTypes.GivenName, "givenName");
             ClaimResolvers.Add(ClaimTypes.Surname, "surname");
-            // Last one wins, if present
-            ClaimResolvers.Add(ClaimTypes.Email, "userPrincipalName");
-            ClaimResolvers.Add(ClaimTypes.Email, "mail");
+            ClaimResolvers.AddCustom(ClaimTypes.Email, user => user.Value<string>("mail") ?? user.Value<string>("userPrincipalName"));
```
HaoK: Honestly I almost prefer if we didn't have the sugar and only exposed the lambda syntax to be clear, its super obvious what's going on that way and we can expose helpers as needed to make the nested stuff nicer

src/Microsoft.AspNetCore.Authentication/ClaimResolver.cs
Tratcher: Trying to see if it applies to Twitter, OIDC, etc.. We'll see.

src/Microsoft.AspNetCore.Authentication.MicrosoftAccount/MicrosoftAccountOptions.cs
Tratcher: Really? I thought the new version was supper clean and easy. It sounds like you want some more explicit names though.

brentschmaltz:
@Tratcher @HaoK @PinpointTownes I am not clear on what problem are we trying to solve. Pinpoint referred us to https://github.com/aspnet/Security/issues/1024 . Which combined with https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/550 would reduce the size of cookies. I am not a fan of mapping claims in the JWT as the claim types have strong definitions https://tools.ietf.org/html/rfc7519#section-4.1

Tratcher:
@brentschmaltz see #969. We often get requests to map additional user data to claims during auth. Today if you want to add any you have to hook the events do everything yourself. This is too much for basic users and they ask to add it for them.  We don't want to maintain an exhaustive list, especially since they keep changing. This PR is to make it easier for devs to add/change claim mappings themselves. It makes our own implementation cleaner too.

src/Microsoft.AspNetCore.Authentication.MicrosoftAccount/MicrosoftAccountOptions.cs
HaoK: Well for code like this I'd prefer if it was self describing, i.e. if it was JsonPayloadClaimsMap.Add("ClaimsTypes.Email", "userPrincipalName) that has much more context

HaoK:
My first suggestion for naming is: Resolver => Mapper which gets you the additional Apply => Map.  I'd suggest Transform but that will collide with the real ClaimsTransformation feature... unless we call it PayloadClaimsTransform :)

HaoK:
ClaimMappings or PayloadClaimMappings also would read well with the current collection/AddXyz methods.

brentschmaltz:
@Tratcher perhaps I am misunderstanding, what will the ClaimReslover do? Add claims OR transform?

Tratcher:
@brentschmaltz it adds claims from the User Info

Tratcher:
Renamed. The funny thing is that I'd started with Map.

samples/SocialSample/Startup.cs
```diff
@@ -227,48 +230,16 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
...
-                            context.Identity.AddClaim(new Claim(
-                                "urn:github:url", link,
-                                ClaimValueTypes.String, context.Options.ClaimsIssuer));
-                        }
+                        context.ResolveClaims(user);
                     }
                 }
-            });
+            };
+            githubOptions.ClaimMaps.AddJsonKeyMap(ClaimTypes.NameIdentifier, "id");
```
HaoK: Since the property is ClaimsMaps, can maybe simplify to AddJsonKey?

samples/SocialSample/Startup.cs
```diff
@@ -227,48 +230,16 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
...
+            githubOptions.ClaimMaps.AddJsonKeyMap(ClaimTypes.Name, "login");
+            githubOptions.ClaimMaps.AddJsonKeyMap("urn:github:name", "name");
+            githubOptions.ClaimMaps.AddJsonKeyMap(ClaimTypes.Email, "email", ClaimValueTypes.Email);
```
HaoK: Can call this AddJsonSubKey if you want to preserve the different method for subsections if you want?

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
```diff
@@ -45,90 +45,7 @@ public FacebookHandler(HttpClient httpClient)
...
-            {
-                identity.AddClaim(new Claim(ClaimTypes.Name, name, ClaimValueTypes.String, Options.ClaimsIssuer));
-            }
-
-            var timeZone = FacebookHelper.GetTimeZone(payload);
-            if (!string.IsNullOrEmpty(timeZone))
-            {
-                identity.AddClaim(new Claim("urn:facebook:timezone", timeZone, ClaimValueTypes.String, Options.ClaimsIssuer));
-            }
+            context.ResolveClaims();
```
HaoK: MapClaims?

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookOptions.cs
```diff
@@ -30,6 +32,21 @@ public FacebookOptions()
             Fields.Add("email");
             Fields.Add("first_name");
             Fields.Add("last_name");
+
+            ClaimMaps.AddJsonKeyMap(ClaimTypes.NameIdentifier, "id");
+            ClaimMaps.AddNestedJsonKeyMap("urn:facebook:age_range_min", "age_range", "min");
```
HaoK: AddNestedJsonKeyMap => AddJsonSubKeyMap?

HaoK: I think subkey is more common than nested.

samples/SocialSample/Startup.cs
Tratcher: MapFromJsonKey? MapFromJsonSubKey, MapFromCustomJson...?

HaoK: I don't think the From adds anything does it?  Map[Custom]Json[Sub]Key works for me

Tratcher:
Drat, to add this to Twitter and OIDC I have to clone all the classes. They all depend on Newtonsoft.Json but don't share any dependencies that require it. The alternative is to make Twitter and OIDC depend on OAuth, which makes sense from a protocol perspective but not really from a packaging perspective.

HaoK:
I figured as much, oh well :(

HaoK:
You could just put the shared classes in base Auth, we've done it before in the past I believe

Tratcher:
Not without adding the JSON dependency to the base class.

HaoK:
Yuck, can we shared source at least?

Tratcher:
Not for public types...

Tratcher:
Changed Twitter to depend on OAuth. Mostly harmless.

I could do the same for OIDC, but there's a larger question on if we want to use this pattern for OIDC. Right now OIDC maps all user data to claims by default. In hindsight this is overkill and you end up with a bunch of junk claims. OIDC also gets claims from the JWT and needs to de-duplicate (with preference given to the JWT claims). I'd need a MapUnique... variant.

OIDC defines some standard claims: http://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
Of those, I think I'd pick out these defaults:
- sub, name, given_name, family_name, profile, email

@PinpointTownes, this was your pet feature in OIDC, what do you think about scaling it back?

PinpointTownes:
> @PinpointTownes, this was your pet feature in OIDC, what do you think about scaling it back?

Well, I find this "include everything" feature really convenient but I agree we have to be pragmatic and realize that the current approach used in OIDC has also its own downsides and contributes to make final authentication cookies utterly big, with all the consequences we know (e.g cookies chunks ignored by browsers or request headers refused by hosts because they are too large...).

Before going further, maybe it would be interesting to test a "blacklisted claims" approach and see if removing protocol claims like `aud`/`iss`/`nonce`/`nbf`/`iat`/`exp` would help reduce the cookies size with OIDC.

No matter what you decide (opt-in or opt-ou), I think it would be better to adopt the same design everywhere, for consistency.

HaoK:
Yep most important thing is for all the CreateTickets to consistently use the same ClaimMap pattern.

Tratcher:
I converted OIDC to use this pattern and mapped several default claims. I added a new mapper to avoid duplicating claims.

I also did an experiment using this approach to delete claims. It seems like a bit of a hack, but it's convenient. (re: https://github.com/aspnet/Security/issues/1024). This got my test cookie (an outlook.com account via AAD) down from 5,267 bytes broken into 2 chunks to 3,567 bytes with no chunking.

PinpointTownes:
> This got my test cookie (an outlook.com account via AAD) down from 5,267 bytes broken into 2 chunks to 3,567 bytes with no chunking.

This sounds really large. How many claims do you have?

I suspect using the WS-Fed claim types [here] doesn't help.
FWIW, there's an ardent debate in the AzureAD repo about the IdentityModel claims mapping feature: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/550.

Disabling this feature by default and using the JWT claim types returned by the OP as-is would likely help reduce the size of the authentication cookies, but of course, this decision has its own downsides, as every external middleware will now have its own claim types, depending on the standard they implement.

Tratcher:
I had 21 claims before and 10 after. Yes, the long claim types contribute quite a bit to the size. Using the short oidc terms for ClaimType doesn't help (yet) because the claims coming from the JWT take precedence.

Tratcher:
@HaoK I think this is ready for a formal review.

The only open question is if this also satisfies https://github.com/aspnet/Security/issues/1024 around removing some of the OIDC protocol claims that IM adds by default. @PinpointTownes, @brockallen, @leastprivilege, @brentschmaltz, is this a workable solution?
See: https://github.com/aspnet/Security/pull/1124/files#diff-5f153b21e61d008634c479494de719d6

samples/SocialSample/Startup.cs
```diff
@@ -227,48 +230,16 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
...
-                            context.Identity.AddClaim(new Claim(
-                                "urn:github:url", link,
-                                ClaimValueTypes.String, context.Options.ClaimsIssuer));
-                        }
+                        context.MapClaims(user);
                     }
                 }
-            });
+            };
+            githubOptions.ClaimMaps.MapJsonKey(ClaimTypes.NameIdentifier, "id");
```
HaoK: Would we lose any meaning if we called the property Claims since all of the methods are called Map?  Just looks a bit odd for it to be ClaimsMaps.Map everywhere?

src/Microsoft.AspNetCore.Authentication.OAuth/DeleteClaimMapper.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Linq;
+using System.Security.Claims;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    /// <summary>
+    /// A JsonClaimMapper that deletes all claims from the given ClaimsIdentity with the given ClaimType.
+    /// </summary>
+    public class DeleteClaimMapper : JsonClaimMapper
```
HaoK: A delete mapper makes me think we should choose a different word.

like ClaimAction/Operation? with Map/Delete becoming something paired?
Copy/Delete or Add/Remove

```
Claims.CopyJsonKey(ClaimTypes.NameIdentifier, "id");
Claims.Delete("ClaimTypeToDelete");
```

src/Microsoft.AspNetCore.Authentication.OAuth/JsonClaimMapperCollectionExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Security.Claims;
+using Microsoft.AspNetCore.Authentication.OAuth;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public static class JsonClaimMapperCollectionExtensions
```
HaoK: I'm not sure I like this as extension methods on an IEnumerable.

What do you think about making the public API into a ClaimsManager instead, with similar API?  The property is then just named ClaimsManager.

```C#
   ClaimsManager.CopyJsonKey(ClaimsTypes.NameIdentifier, "id");
   ClaimsManager.Delete(ClaimsTypes.ReallyLongStuffIDon'tWant);
```

You can still have these methods add some kind of strongly typed Action that maps to the Map classes, or directly add an Action with the signature of Map.  I don't think the extensiblity changes much as they could still add extension methods to the Manager just like the collection.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
```diff
@@ -55,6 +56,24 @@ public OpenIdConnectOptions(string authenticationScheme)
             Events = new OpenIdConnectEvents();
             Scope.Add("openid");
             Scope.Add("profile");
+
+            ClaimMaps.DeleteClaim("nonce");
```
HaoK: Maybe add sugar to pass in a params string[] of claims to delete?

src/Microsoft.AspNetCore.Authentication.OAuth/JsonClaimMapperCollectionExtensions.cs
Tratcher: What does the IEnumerable have to do with anything? The extension methods are for the JsonClaimMapperCollection and they all wrap the Add method.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
Tratcher: We try to avoid params in public APIs. We'll see how often this gets used externally.

src/Microsoft.AspNetCore.Authentication.OAuth/JsonClaimMapperCollectionExtensions.cs
HaoK: Sorry what I meant was instead of a JsonClaimMapperCollection, maybe that should be called something like ClaimsManager which implies that its got more than just basic add/remove

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
HaoK: What's the rationale against params?  We certainly have them in several places today already for exactly this situation, truncating a lot of duplicated lines, maybe you don't even really need DeleteClaim, just a single params DeleteClaims

Tratcher:
More renames :-). ClaimMappers are now ClaimActions, I also made a few changes to make sure that the actions run even when user-info retrieval is disabled (Twitter, OIDC), and that the rules are hardened against null user info.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
Tratcher: When you have params on a method it makes it hard to ever add other inputs to that method.

src/Microsoft.AspNetCore.Authentication.OAuth/JsonClaimMapperCollectionExtensions.cs
Tratcher: It only has basic AddRemove. I renamed it to ClaimActionCollection and the property is called ClaimActions.
ClaimActions.Map...
ClaimActions.DeleteClaim...

PinpointTownes:
> @PinpointTownes, @brockallen, @leastprivilege, @brentschmaltz, is this a workable solution?

I'd bet a lot of money @leastprivilege won't like the fact the JWT claims are magically converted to their WS-Fed equivalent, specially since he'll now have to disable that in two places: in IdentityModel and in ASP.NET Core :sweat_smile:

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
```diff
@@ -55,6 +56,24 @@ public OpenIdConnectOptions(string authenticationScheme)
...
             Scope.Add("openid");
             Scope.Add("profile");
+
+            ClaimActions.DeleteClaim("nonce");
+            ClaimActions.DeleteClaim("aud");
+            ClaimActions.DeleteClaim("iss");
+            ClaimActions.DeleteClaim("iat");
+            ClaimActions.DeleteClaim("nbf");
+            ClaimActions.DeleteClaim("exp");
+            ClaimActions.DeleteClaim("c_hash");
```
PinpointTownes: You can also add `at_hash`, `auth_time`, `acr`, `amr` and `azp`.

Tratcher:
@PinpointTownes what do you mean? I used the short names: https://github.com/aspnet/Security/pull/1124/files#diff-5f153b21e61d008634c479494de719d6R71.

PinpointTownes:
@Tratcher ah, my bad, I had not realized you had changed that in the latest commits :sweat_smile: 

Now, the question is: do we want to generalize that to the other middleware?

Also, do we want to have a global claims transformer that would convert all the claim types to a unique standard, so the application code wouldn't have to deal with multiple specs?

Tratcher:
Sounds like the ClaimsTransformation component :-)

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
HaoK: That's what overloads are for :)  I'd have the params method for this usage alone personally...

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/ClaimActionCollectionUniqueExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// This no-ops if the ClaimsIdentity already contains a Claim with the given ClaimType.
+        /// This no-ops if the key is not found or the value is empty.
+        /// </summary>
+        /// <param name="collection"></param>
+        /// <param name="claimType">The value to use for Claim.Type when creating a Claim.</param>
+        /// <param name="jsonKey">The top level key to look for in the json user data.</param>
+        /// <param name="valueType">The value to use for Claim.ValueType when creating a Claim.</param>
+        public static void MapUniqueJsonKey(this ClaimActionCollection collection, string claimType, string jsonKey, string valueType)
+        {
+            collection.Add(new UniqueJsonKeyClaimAction(claimType, valueType, jsonKey));
```
HaoK: So my earlier point about the Collection vs Manager wasn't so much about that name, but that it seems unnecessary to have public types that are only used in the extension method.  Why can't we just bake the logic directly into Method instead of having the method and a type (that's just a class around a Action.  Cleans up the public surface area, and I don't see much use for the action types outside of the methods.  Which lead me to ask why we shouldn't just call it a Manager with the methods, instead of a List of action thingies...

src/Microsoft.AspNetCore.Authentication.MicrosoftAccount/MicrosoftAccountHandler.cs
```diff
@@ -35,39 +35,7 @@ public MicrosoftAccountHandler(HttpClient httpClient)
...
-                identity.AddClaim(new Claim(ClaimTypes.Surname, surname, ClaimValueTypes.String, Options.ClaimsIssuer));
-                identity.AddClaim(new Claim("urn:microsoftaccount:surname", surname, ClaimValueTypes.String, Options.ClaimsIssuer));
-            }
-
-            var email = MicrosoftAccountHelper.GetEmail(payload);
-            if (!string.IsNullOrEmpty(email))
-            {
-                identity.AddClaim(new Claim(ClaimTypes.Email, email, ClaimValueTypes.String, Options.ClaimsIssuer));
-            }
+            context.MapClaims();
```
HaoK: MapClaims => RunActions/InvokeActions/ExecuteActions/FireWhenReady?

HaoK: InvokeClaimsActions

src/Microsoft.AspNetCore.Authentication.MicrosoftAccount/MicrosoftAccountOptions.cs
```diff
@@ -23,6 +25,12 @@ public MicrosoftAccountOptions()
             TokenEndpoint = MicrosoftAccountDefaults.TokenEndpoint;
             UserInformationEndpoint = MicrosoftAccountDefaults.UserInformationEndpoint;
             Scope.Add("https://graph.microsoft.com/user.read");
+
+            ClaimActions.MapJsonKey(ClaimTypes.NameIdentifier, "id");
```
HaoK: Now that mapper is gone from everywhere else, maybe Import/Copy is better, since map is kinda vague what we are mapping.

CopyJsonKey
ImportJsonKey

src/Microsoft.AspNetCore.Authentication.OAuth/ClaimAction.cs
```diff
@@ -0,0 +1,42 @@
...
+        public string ValueType { get; }
+
+        /// <summary>
+        /// Exhamine the given userData json, determine if the requisite data is present, and optionally add it
+        /// as a new Claim on the ClaimsIdentity.
+        /// </summary>
+        /// <param name="userData">The source data to exhamine. This value may be null.</param>
+        /// <param name="identity">The identity to add Claims to.</param>
+        /// <param name="issuer">The value to use for Claim.Issuer when creating a Claim.</param>
+        public abstract void Run(JObject userData, ClaimsIdentity identity, string issuer);
```
HaoK: Yeah I think rather than making this something classes implement, this could just be an `Action<JObject, ClaimsIdentity, string>Action` property, which the extension methods set.  So the change could be to move the Run implementations into the extension methods as property setters on the single ClaimsAction class

HaoK:
I like the usage, definitely way better than what we had before.

src/Microsoft.AspNetCore.Authentication.MicrosoftAccount/MicrosoftAccountOptions.cs
Tratcher: I think Map still works. Map JSON Key to Claim

PinpointTownes:
> Sounds like the ClaimsTransformation component :-)

Yeah, but it's totally useless OTB if you don't provide your own handler.

I had more something like "convert all these JWT/SWT/WS-Fed/custom claims to the `ClaimsTypes` names I'm used to" in mind :smile:

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/ClaimActionCollectionUniqueExtensions.cs
Tratcher: Yes you could implement all of the rules with MapCustom, but that comes down to a style choice. The classes provide some minor code cleanliness benefits like inherited properties, better stack traces, avoids closures, etc..

HaoK: Fair enough, but do the classes need to be public?  Are you expecting them to be used other than in the explicit methods?

Tratcher: I'd expect some new types to be derived from the ones provided. We could dump them into an infrastructure namespace (e.g. Oath.Claims).

HaoK: Cool moving them all into their own namespace works

natemcmaster:
@Tratcher don't forget to update API usage in downstream repos, like MusicStore.

Issue #1035, opened 2016-11-11
Provide an easy way to disable telemetry in the OpenID Connect middleware


PinpointTownes:
A (**really intrusive**) form of telemetry is [about to be added to `OpenIdConnectMessage`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/529) so that all authorization and logout requests will now expose the IdentityModel assembly version to the remote OIDC server and to the end users (which has security implications, but whatever...).

Please consider adding a `DisableTelemetry` option in the OIDC middleware so we can easily disable that "feature" once the Wilson PR is merged.

Tratcher:
@blowdart?

blowdart:
Should happen, yes.

leastprivilege:
Opt-in. Not opt-out. 

Get Outlook for iOS




On Fri, Nov 11, 2016 at 1:06 PM +0000, "Barry Dorrans" <notifications@github.com> wrote:












Should happen, yes.

brentschmaltz:
@blowdart @Tratcher is there time to put in a DisableTelemetry feature before 1.1 OR do you feel as 
@leastprivilege telemetry should be off by default? If so the identity layer could be off by default.

PinpointTownes:
FYI, I started an email thread with Scott Guthrie, Scott Hunter and Stuart Kwan about telemetry in general and this specific PR was mentioned. @polita kindly emailed me back and shared more info about this feature:

> In the unfortunate circumstance that we find a security issue in our code, our first step will be to find a patch for it and release a new package. Our next step (or perhaps in tandem) is to alert all customers who are vulnerable. This telemetry enables us to know who to call to patch vulnerabilities. Hackers don’t need to be told who has a vulnerable package; they’ll try everyone until they find someone they can target. If we can inform customers of vulnerabilities in their apps, we can help customers respond to vulnerabilities, in many cases before the vulnerability is widely known.
 
> I understand your concern. For this reason, I’ve created a switch that allows app owners to disable telemetry altogether if they prefer. App developers can handle RedirectToIdentityProvider and set DisableTelemetryParameters on the ProtocolMessage.

Eilon:
@brentschmaltz @polita is the IdentityModel telemetry feature even in the 1.1 release? I thought we had final builds of it already?

polita:
[...]

PR #1140, opened 2017-03-09
Add an opt-out DisableTelemetry option in the OpenID Connect middleware

PinpointTownes:
Fixes https://github.com/aspnet/Security/issues/1035.

/cc @Eilon @Tratcher @brentschmaltz @polita

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
@@ -241,5 +241,12 @@ namespace Microsoft.AspNetCore.Builder
         /// This is disabled by default.
         /// </summary>
         public bool SkipUnrecognizedRequests { get; set; } = false;
+
+        /// <summary>
+        /// Indicates whether telemetry should be disabled. When this feature is enabled,
+        /// the assembly version of the Microsoft IdentityModel packages is sent to the
+        /// remote OpenID Connect provider as an authorization/logout request parameter.
+        /// </summary>
+        public bool DisableTelemetry { get; set; }
     }
 }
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
```diff
@@ -241,5 +241,12 @@ public new IOpenIdConnectEvents Events
         /// This is disabled by default.
         /// </summary>
         public bool SkipUnrecognizedRequests { get; set; } = false;
+
+        /// <summary>
+        /// Indicates whether telemetry should be disabled. When this feature is disabled,
```
Eilon: Perhaps change the second sentence to be worded with positives instead of the double-negative? E.g. "When this feature is enabled, blah blah is sent to the blah blah". I just think it's a bit easier to read that way.

PinpointTownes: Fixed :+1:

test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/OpenIdConnectChallengeTests.cs
```diff
@@ -38,6 +38,20 @@ public class OpenIdConnectChallengeTests
...
+        {
+            var settings = new TestSettings(opt => opt.DisableTelemetry = true);
+
+            var server = settings.CreateTestServer();
+            var transaction = await server.SendAsync(ChallengeEndpoint);
+
+            var res = transaction.Response;
+            Assert.Equal(HttpStatusCode.Redirect, res.StatusCode);
+            Assert.DoesNotContain(OpenIdConnectParameterNames.SkuTelemetry, res.Headers.Location.Query);
+            Assert.DoesNotContain(OpenIdConnectParameterNames.VersionTelemetry, res.Headers.Location.Query);
```
Tratcher: Could use one positive test they verifies the presence of these parameters. Could be added to an existing test.

PinpointTownes: Done. FYI, I didn't update `ChallengeIssuedCorrectlyForFormPost` because the telemetry parameters are never sent when opting for POST authorization requests.

PR #1142, opened 2017-03-11
Auth 2.0 with OptionsFactory (named options)

HaoK:


HaoK:
Recreating with proper parent branch

PR #1144, opened 2017-03-11
Auth 2.0 using OptionsFactory (named options)

HaoK:
- New `IOptionsFactory` which provides named options in addition to validation, both of which are triggered via IOptionsFactory.Get("name").  Allows normal options configuration for schemes, Validation is just a different set of Actions that run after the Configures

IOptionsFactory Usage:
```C#
  services.ConfigureAll<GoogleOptions>(o => o.SignInScheme = "GoogleCookie");
  services.Configure<GoogleOptions>("Google", o => ClientId = "avsdlkj");
  services.Configure<GoogleOptions>("Google2" p => ClientId = "two");
  services.ValidateAll<GoogleOptions>(o => if (o.ClientId == null) throw new ArgumentException());

  optionsFactory.Get<GoogleOptions>("Google"); // OK
  optionsFactory.Get<GoogleOptions>("Default"); // throws ArgumentException
```
- split HandleRequest into its own `IAuthenticationRequestHandler` interface instead, CallbackPaths routing is disabled for now (all IAuthenticationRequestHandler schemes are given a chance to handle the request)

## **TL:DR Auth 2.0 summary**

- Single `UseAuthentication` is required, which takes care of the old 'automatic' authenticate, in addition to giving each scheme a chance to handle the request like the old middlewares did.
- 'Automatic' authenticate and challenge are now a single `Default[Authenticate/SignIn]Scheme` in `AuthenticationOptions`, rather than a flag on each middleware.  No ambiguity, altho there's one bit of magic where if none are specified and there's only a single scheme registered, we use that scheme as the default.

## Pros of the new design

```C#
   // ConfigureServices:

       _dbContext = context;
    }
    
        {
        }
}
```

## Claims Transformation

```
     return Task.FromResult(p);
  }
```


HaoK:
@Tratcher I Pushed the duplicated DataProtectionProvider code in all the RemoteAuths up to base to clean things up

HaoK:
@PinpointTownes do you prefer this iteration that is more similar to the old behavior, where any scheme that implements IAuthenticationRequestHandler will see all requests?  I don't have an easy way right now to implement the old callback ghetto routing with the switch to using named options

PinpointTownes:
> @PinpointTownes do you prefer this iteration that is more similar to the old behavior, where any scheme that implements IAuthenticationRequestHandler will see all requests?

I'll take a deeper look tomorrow, but it looks really promising, I like it :smile:

src/Microsoft.AspNetCore.Authentication.Impl/AuthenticationMiddleware.cs
```diff
@@ -41,12 +41,14 @@ public AuthenticationMiddleware(RequestDelegate next, IAuthenticationSchemeProvi
...
+
                 // Give each scheme a chance to handle the request if it registered this path
-                var handlers = context.RequestServices.GetRequiredService<IAuthenticationHandlerResolver>();
+                var handlers = context.RequestServices.GetRequiredService<IAuthenticationHandlerProvider>();
                 foreach (var scheme in await Schemes.GetRequestHandlerSchemes(context.Request.Path))
                 {
-                    var handler = await handlers.ResolveHandlerAsync(context, scheme.Name);
-                    if (await handler.HandleRequestAsync())
+                    var handler = await handlers.GetHandlerAsync(context, scheme.Name) as IAuthenticationRequestHandler;
+                    if (await handler?.HandleRequestAsync())
```
PinpointTownes: FYI: this line will null-ref if `handler` returns `null`.

HaoK: Oops thanks fixed

src/Microsoft.AspNetCore.Authentication.Abstractions/IAuthenticationService.cs
```diff
@@ -12,9 +12,6 @@ public interface IAuthenticationService
         Task<AuthenticateResult> AuthenticateAsync(HttpContext context, string scheme);
         Task ChallengeAsync(HttpContext context, string scheme, AuthenticationProperties properties, ChallengeBehavior behavior);
         Task ForbidAsync(HttpContext context, string scheme, AuthenticationProperties properties);
```
Tratcher: Until we sort out Challenge vs Forbid, Forbid is not a 1st class thing, it's a subset of Challenge.

HaoK:
Rebased to dev replaced with https://github.com/aspnet/Security/pull/1151

PR #1151, opened 2017-03-15
Authentication 2.0 Iteration 3 (with Names/OptionFactory)

HaoK:
- New `IOptionsFactory` which provides named options in addition to validation, both of which are triggered via IOptionsFactory.Get("name").  Allows normal options configuration for schemes, Validation is just a different set of Actions that run after the Configures

IOptionsFactory Usage:
```C#
  services.ConfigureAll<GoogleOptions>(o => o.SignInScheme = "GoogleCookie");
  services.Configure<GoogleOptions>("Google", o => ClientId = "avsdlkj");
  services.Configure<GoogleOptions>("Google2" p => ClientId = "two");
  services.ValidateAll<GoogleOptions>(o => if (o.ClientId == null) throw new ArgumentException());

  optionsFactory.Get<GoogleOptions>("Google"); // OK
  optionsFactory.Get<GoogleOptions>("Default"); // throws ArgumentException
```
- split HandleRequest into its own `IAuthenticationRequestHandler` interface instead, CallbackPaths routing is disabled for now (all IAuthenticationRequestHandler schemes are given a chance to handle the request)

## **TL:DR Auth 2.0 summary**

- Single `UseAuthentication` is required, which takes care of the old 'automatic' authenticate, in addition to giving each scheme a chance to handle the request like the old middlewares did.
- 'Automatic' authenticate and challenge are now a single `Default[Authenticate/SignIn]Scheme` in `AuthenticationOptions`, rather than a flag on each middleware.  No ambiguity, altho there's one bit of magic where if none are specified and there's only a single scheme registered, we use that scheme as the default.

## Pros of the new design

```C#
   // ConfigureServices:

       _dbContext = context;
    }
    
        {
        }
}
```

## Claims Transformation

```
     return Task.FromResult(p);
  }
```


HaoK:
@brockallen @PinpointTownes we've decided today that we are going ahead with this change, the implementation details are still subject to change, but the high level switch from middleware to services/DI/NamedOptions is a go.  Definitely let us know if any of your important scenarios are affected, or if you have any ideas for quality of life improvements in the base auth handler classes, since the entire auth code base is opened up for surgery right now...

HaoK:
Note to self: need to investigate adding schemes and configuring them outside of ConfigureServices

PinpointTownes:
> @brockallen @PinpointTownes we've decided today that we are going ahead with this change, the implementation details are still subject to change, but the high level switch from middleware to services/DI/NamedOptions is a go.

Thanks for the info, I'll take a look ASAP. When do you plan to merge this PR?

HaoK:
No firm ETA on merge, need to start working on the changes in all the other repos (Identity/servers), which likely will require some more changes here.  But we can start settling on implementation choices now.  Also the options changes will go in first, so I'd start there if you are interested on how that lands: https://github.com/aspnet/Options/pull/176

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -55,26 +55,118 @@ public class OpenIdConnectHandler : RemoteAuthenticationHandler<OpenIdConnectOpt
...
+            await base.InitializeAsync(scheme, context);
+            Events = Events ?? new OpenIdConnectEvents();
+
+            SignOutScheme = Options.SignOutScheme;
+            if (string.IsNullOrEmpty(SignOutScheme))
+            {
+                SignOutScheme = SignInScheme;
+            }
+
+            if (Options.StateDataFormat == null)
```
PinpointTownes: @HaoK correct me if I'm wrong, but `Options` is a shared instance, right? (at least, when using the default options factory). So this means that the options could be modified concurrently, as `InitializeAsync` is invoked per request. Isn't that risky as hell?

HaoK: Yeah so that's why I'm trying to move all the Options manipulations into 'validations' that occur as part of OptionsFactory.Get("schemeName")

HaoK: We can also switch to using the form of 'scoped' options for the schemes, IOptionsSnapshot for example which wouldn't be shared.

PinpointTownes: Makes sense. Thanks for the clarification.

HaoK: Now that I think about it, that's probably the easiest fix to this issue since the scheme options don't want to share

src/Microsoft.AspNetCore.Authentication.Impl/AuthenticationHandler.cs
```diff
@@ -0,0 +1,229 @@
...
+
+        protected ISystemClock Clock { get; }
+
+        protected IOptionsFactory<TOptions> OptionsFactory { get; }
+
+        /// <summary>
+        /// The handler calls methods on the events which give the application control at certain points where processing is occurring. 
+        /// If it is not provided a default instance is supplied which does nothing when the methods are called.
+        /// </summary>
+        protected virtual object Events { get; set; }
```
PinpointTownes: How bad would it be if `AuthenticationHandler` had a `TEvents` generic argument?

HaoK: 🤢

HaoK: But yeah the events stuff needs some love, I guess every handler does have an events class today.

HaoK: The events is a bit weird to live on options, maybe we can move it into the AuthenticationScheme instead, as something that gets fed into `Intialize(IOptionsFactory<TOptions>, TEvent events, HttpContext)`

HaoK: I've always felt configuring the events as part of the options makes it look ginormous.  Can we do better if we tease it apart/have a new pattern for hooking events that's not tied to options configuration?

HaoK: So just brainstorming alternatives, imagine we created an options type per event, then we could 'resolve' our events via `IOptionsFactory<EventType>("authSchemeName")`, so hooking an event would become "services.Configure<OAuthCreateTicketEventOptions>("github", o => DoSomething()")

PinpointTownes: > `services.Configure<OAuthCreateTicketEventOptions>("github", o => DoSomething())`

Not sure I like this syntax. Sounds way more complicated than:

```csharp
services.AddGoogleAuthentication(options =>
{
    options.Events.OnCreatingTicket = context =>
    {
        // ...
    };
});

Comment (HaoK) [inline src/Microsoft.AspNetCore.Authentication.Impl/AuthenticationHandler.cs]: Honestly though, now that we made it easy to replace the type.  I think its time to ask the question if the pattern should be instead:
```
   class MyGoogleAuthHandler : GoogleHandler
   {
         override OnCreatingTicket() { ... }
   }    
   

   services.AddGoogleAuthentication<MyGoogleAuthHandler>(options => )
```

That's what the new system lets us do instead...way better in my mind.  I can't express how much I hate the events stuff...

Comment (HaoK) [inline src/Microsoft.AspNetCore.Authentication.Impl/AuthenticationHandler.cs]: Its a pure wart of not being able to replace handler types before...

Comment (PinpointTownes) [inline src/Microsoft.AspNetCore.Authentication.Impl/AuthenticationHandler.cs]: >  can't express how much I hate the events stuff...

Any particular reason?

Not every event is exposed as an overridable method (the OAuth2 generic handler is pretty much an exception, because we allow both models: creating derived handlers like the social providers we have in aspnet-contrib, and configuring it via the events model).

Comment (HaoK) [inline src/Microsoft.AspNetCore.Authentication.Impl/AuthenticationHandler.cs]: Are you sure every event isn't overridable?  Isn't that the only purpose for the IXyzEvents to exist?  basically its the list of virtual methods that would exist if you could plug in your own handler.

Comment (PinpointTownes) [inline src/Microsoft.AspNetCore.Authentication.Impl/AuthenticationHandler.cs]: > Are you sure every event isn't overridable?

The events are all overridable, but not the infrastructure code that call them: the `OAuthHandler.CreateTicketAsync` method you used in your sample is quite an exception, as it's the only handler that was really designed to be subclassed.

Comment (HaoK) [inline src/Microsoft.AspNetCore.Authentication.Impl/AuthenticationHandler.cs]: So moving to a virtual method would allow us to nuke all of the context classes, and just pass them as parameters (or let them access handler directly since its now in the handler, which they couldn't even do before).

Comment (PinpointTownes) [inline src/Microsoft.AspNetCore.Authentication.Impl/AuthenticationHandler.cs]: I personally like the fact the public hooks are separated from the rest of the infrastructure code, so I'd really love this events pattern to be present in 2.0.0, but heh, your PR, your choice :smile: 

(also, I'm not a huge fan of having to override a class to just to do something that can be done in a 2-line event delegate)

Comment (HaoK) [inline src/Microsoft.AspNetCore.Authentication.Impl/AuthenticationHandler.cs]: Fair enough, @Tratcher had the same opinion, there's so many other things on my list right now, so its most likely not going to change soon

File: src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationScheme.cs [hunk 3]
```diff
@@ -0,0 +1,33 @@
...
+            Settings = settings ?? throw new ArgumentNullException(nameof(settings));
+        }
+
+        // TODO: add display name?
+        public string Name { get; }
+        public Type HandlerType { get; }
+
+        // Holds things like the configured options instances for the handler
+        // Also replacement for AuthenticationDescription
+        public IReadOnlyDictionary<string, object> Settings { get; }
```
Tratcher: Still a weird property bag

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationScheme.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        public AuthenticationScheme(string name, Type handlerType, IReadOnlyDictionary<string, object> settings)
+        {
+            Name = name ?? throw new ArgumentNullException(nameof(name));
+            if (!typeof(IAuthenticationHandler).IsAssignableFrom(handlerType))
+            {
+                throw new ArgumentException("handlerType must implement IAuthenticationSchemeHandler.");
+            }
+            HandlerType = handlerType ?? throw new ArgumentNullException(nameof(handlerType));
+            Settings = settings ?? throw new ArgumentNullException(nameof(settings));
```
Tratcher: syntax

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationSchemeBuilder.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            Name = name;
+        }
+
+        public string Name { get; }
+
+        public Type HandlerType { get; set; }
+
+        // Holds things like the configured options instances for the handler
+        public Dictionary<string, object> Settings { get; set; } = new Dictionary<string, object>(); // casing?
```
Tratcher: Ordinal

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationTicket.cs
```diff
@@ -20,12 +19,8 @@ public class AuthenticationTicket
         /// <param name="authenticationScheme">the authentication middleware that was responsible for this ticket.</param>
         public AuthenticationTicket(ClaimsPrincipal principal, AuthenticationProperties properties, string authenticationScheme)
         {
-            if (principal == null)
-            {
-                throw new ArgumentNullException(nameof(principal));
-            }
             AuthenticationScheme = authenticationScheme;
-            Principal = principal;
+            Principal = principal ?? throw new ArgumentNullException(nameof(principal));
```
Tratcher: syntax

src/Microsoft.AspNetCore.Authentication.Abstractions/IAuthenticationHandler.cs
```diff
@@ -0,0 +1,31 @@
...
+        // Gives the handler access to the configuration data
+        Task InitializeAsync(AuthenticationScheme scheme, HttpContext context);
+
+        Task<AuthenticateResult> AuthenticateAsync(AuthenticateContext context);
+        Task ChallengeAsync(ChallengeContext context);
+        Task SignInAsync(SignInContext context);
+        Task SignOutAsync(SignOutContext context);
+    }
+
+    public interface IAuthenticationRequestHandler : IAuthenticationHandler
```
Tratcher: Split file

src/Microsoft.AspNetCore.Authentication.Abstractions/IAuthenticationSchemeProvider.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public interface IAuthenticationSchemeProvider
```
Tratcher: Where did the implementation go?

HaoK: I nuked the old Authentication package and moved Impl into there

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationScheme.cs
HaoK: Yeah this can go away now, since we can add/remove options dynamically I don't think this is needed at all.  Fleshing out the add/remove auth scheme using custom options in the next update

HaoK:
Updated with some cleanup, nuked settings bag on AuthenticationScheme, RIP ?? throw

HaoK:
So in terms of final homes:

Authentication.Abstractions => Http Repo (keeps its own package)
Should we split Authentication into Authentication.Core (lives in Http), and Authentication(.Security)?  Basically the core interfaces/implementations for AddAuthentication live in Core, and the base class stuff we use for the Security auth packages would stay in this repo...

So...
Authentication.Abstractions (HttpAbstractions)
Authentication.Core (HttpAbstractions)
Authentication (Security)

Tratcher:
Which repos need to depend on Authentication.Core?

samples/SocialSample/Startup.cs
```diff
@@ -228,14 +191,42 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
...
+
+            services.AddAuthentication(options => options.DefaultAuthenticationScheme = CookieAuthenticationDefaults.AuthenticationScheme);
+        }
+
+        public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
+        {
+            loggerfactory.AddConsole(LogLevel.Information);
+
+            // Simple error page to avoid a repo dependency.
+            app.Use(async (context, next) =>
```
Tratcher: Can we just have Security depend on Diagnostics so we can stop doing this stuff? I don't think it creates any cycles, it just bumps security up the chart one level.

samples/SocialSample/Startup.cs
```diff
@@ -228,14 +191,42 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
...
+                        throw;
+                    }
+                    context.Response.StatusCode = 500;
+                    await context.Response.WriteAsync(ex.ToString());
+                }
+            });
+
+            app.UseAuthentication();
+
+            if (string.IsNullOrEmpty(Configuration["facebook:appid"]))
```
Tratcher: Move this check to ConfigureServices

src/Microsoft.AspNetCore.Authentication.Abstractions/IAuthenticationSchemeProvider.cs
Tratcher: It's not in this PR anywhere.

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAppBuilderExtensions.cs
```diff
@@ -1,50 +0,0 @@
...
-            return app.UseMiddleware<CookieAuthenticationMiddleware>();
-        }
-
-        /// <summary>
-        /// Adds the <see cref="CookieAuthenticationMiddleware"/> middleware to the specified <see cref="IApplicationBuilder"/>, which enables cookie authentication capabilities.
-        /// </summary>
-        /// <param name="app">The <see cref="IApplicationBuilder"/> to add the middleware to.</param>
-        /// <param name="options">A <see cref="CookieAuthenticationOptions"/> that specifies options for the middleware.</param>
-        /// <returns>A reference to this instance after the operation has completed.</returns>
-        public static IApplicationBuilder UseCookieAuthentication(this IApplicationBuilder app, CookieAuthenticationOptions options)
```
Tratcher: Bring these back as Obsolete Error with a fwlink.

src/Microsoft.AspNetCore.Authentication.Abstractions/SignInContext.cs
```diff
@@ -0,0 +1,20 @@
...
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public class SignInContext : BaseAuthenticationContext
+    {
+        public SignInContext(HttpContext context, string authenticationScheme, ClaimsPrincipal principal, AuthenticationProperties properties)
+            : base(context, authenticationScheme, properties)
+        {
+            Principal = principal ?? throw new ArgumentNullException(nameof(principal));
```
Tratcher: syntax

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationOptions.cs
```diff
@@ -38,25 +56,20 @@ public class RemoteAuthenticationOptions : AuthenticationOptions
         /// Gets or sets the authentication scheme corresponding to the middleware
         /// responsible of persisting user's identity after a successful authentication.
         /// This value typically corresponds to a cookie middleware registered in the Startup class.
-        /// When omitted, <see cref="SharedAuthenticationOptions.SignInScheme"/> is used as a fallback value.
+        /// When omitted, <see cref="SignInScheme"/> is used as a fallback value.
```
Tratcher: This comment is now referencing itself?

src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationScheme.cs
```diff
@@ -2,32 +2,33 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
-using System.Linq;
 using System.Reflection;
-using Microsoft.AspNetCore.Http;
```
HaoK: @Tratcher they are here, but perhaps hard to find in the 178 diff...  maybe easier to just view via commits

samples/SocialSample/Startup.cs
HaoK: Seems fine to me, I'll add it

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
```diff
@@ -71,8 +71,17 @@ protected AuthenticationHandler(IOptionsService<TOptions> optionsFactory, ILogge
         /// <returns></returns>
         public virtual Task InitializeAsync(AuthenticationScheme scheme, HttpContext context)
         {
-            Scheme = scheme ?? throw new ArgumentNullException(nameof(scheme));
-            Context = context ?? throw new ArgumentNullException(nameof(context));
+            if (scheme == null)
```
HaoK: @Tratcher See AuthenticationHandler here

HaoK:
@Tratcher I'll break this PR up into 2 new PRs to make things more managable, I'll start the PR for Abstractions/Core in HttpAbstractions and open a new one for only the Security changes

HaoK:
Updated removing the stuff that is being moved to HttpAbstractions via: https://github.com/aspnet/HttpAbstractions/pull/798

Hopefully easier to review

PinpointTownes:
@HaoK hum, is there a particular reason to move so many types to `HttpAbstractions`? Of course, `IAuthenticationHandler` and the types used to register the handler in the DI need to live there for layering reasons, but is it really useful to move things like `AuthenticationMiddleware` there?

Tratcher:
@PinpointTownes I think they're needed to write tests for HttpSysServer and IISIntegration

PinpointTownes:
`AuthenticationMiddleware`'s role is extremely limited: it simply adds a special feature to track the original paths, calls `IAuthenticationSchemeProvider` to get the default sign-in scheme (which is kinda the equivalent of "active authentication") and calls `handler.HandleRequestAsync()` for "promiscuous handlers". AFAICT, none of the HttpSysServer/IISIntegration tests couldn't live without this type. Do you have any specific example in mind?

HaoK:
The middleware isn't critical, I can move that back along with UseAuthentication to Security.  What's critical is the core services/AddAuthenticationCore as you can't even use Auth without those.  I'll move the Middleware/Use out now

HaoK:
Updated, UseXyz's are back as obsolete/throw not Supported.  Also brought back the AuthMiddleware from HttpAbstractions.

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAppBuilderExtensions.cs
```diff
@@ -12,26 +12,26 @@ namespace Microsoft.AspNetCore.Builder
     public static class CookieAppBuilderExtensions
     {
         /// <summary>
-        /// Obsolete, see TODO: fwdlink
+        /// Obsolete, see https://go.microsoft.com/fwlink/?linkid=845470
         /// </summary>
         /// <param name="app">The <see cref="IApplicationBuilder"/> to add the middleware to.</param>
         /// <returns>A reference to this instance after the operation has completed.</returns>
         [Obsolete]
```
Tratcher: Put the link in the Obsolete comment

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAppBuilderExtensions.cs
```diff
@@ -16,7 +16,7 @@ public static class CookieAppBuilderExtensions
         /// </summary>
         /// <param name="app">The <see cref="IApplicationBuilder"/> to add the middleware to.</param>
         /// <returns>A reference to this instance after the operation has completed.</returns>
-        [Obsolete]
+        [Obsolete("See https://go.microsoft.com/fwlink/?linkid=845470")]
```
Tratcher: Hmm, it's only a warning by default, you have to set IsError true
https://msdn.microsoft.com/en-us/library/system.obsoleteattribute.iserror(v=vs.110).aspx

HaoK:
Replaced with https://github.com/aspnet/Security/pull/1170

PR #1157, opened 2017-03-22
Change compilation targets

pranavkm:


src/Microsoft.Owin.Security.Interop/Microsoft.Owin.Security.Interop.csproj
```diff
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <Description>A compatibility layer for sharing authentication tickets between Microsoft.Owin.Security and Microsoft.AspNetCore.Authentication.</Description>
-    <TargetFramework>net451</TargetFramework>
+    <TargetFramework>net46</TargetFramework>
```
PinpointTownes: Isn't the bar now a bit high for a package meant to be consumed by OWIN/Katana apps that are often hosted on servers that don't have the latest (and greatest) .NET Framework bits?

pranavkm: @PinpointTownes the plan is to target what `netstandard2.0` supports which is `net461` (https://github.com/dotnet/standard/blob/master/docs/netstandard-20/README.md#net-framework-461-supporting-net-standard-20). If this were a standalone package, perhaps we could keep the TFM, but since DataProtection is getting bumped up too, this has to be retargeted.

@DamianEdwards thoughts?

pranavkm:
@PinpointTownes I'll let @DamianEdwards follow up separately. Merging this in for now to unblock my work.

PR #1170, opened 2017-03-31
Iteration 4 - Using Auth 2.0/Options 2.0 packages

HaoK:
Previous PR: https://github.com/aspnet/Security/pull/1151

@Tratcher clean PR to start with, rebased and using the Http Abstractions/Options 2.0 interfaces


src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -55,26 +55,118 @@ public class OpenIdConnectHandler : RemoteAuthenticationHandler<OpenIdConnectOpt
...
+
+                        Options.MetadataAddress += ".well-known/openid-configuration";
+                    }
+
+                    if (Options.RequireHttpsMetadata && !Options.MetadataAddress.StartsWith("https://", StringComparison.OrdinalIgnoreCase))
+                    {
+                        throw new InvalidOperationException("The MetadataAddress or Authority must use HTTPS unless disabled for development by setting RequireHttpsMetadata=false.");
+                    }
+
+                    Options.ConfigurationManager = new ConfigurationManager<OpenIdConnectConfiguration>(Options.MetadataAddress, new OpenIdConnectConfigurationRetriever(),
```
PinpointTownes: @HaoK hum, I thought all the initialization stuff was supposed to be done in the options?
'cause AFAICT, this iteration still suffers from the same concurrency issues as the previous PRs :sweat_smile:

HaoK: Whoops, I forgot we needed validation/fixup in options for that reason, well good thing you reminded me why I need to push for that feature :)

HaoK:
Updated PR with some refactoring/cleaning up around Options/Events initialization.

@Tratcher I think this has everything we need for the initial iteration (other than the AddXyz() empty overloads that use Config from DI, I only added that to Google so far)

src/Microsoft.AspNetCore.Authentication/AuthenticationSchemeOptions.cs
```diff
@@ -36,5 +36,15 @@ public virtual void Validate()
         /// If set, will be used as the service type to get the Events instance instead of the property.
         /// </summary>
         public Type EventsType { get; set; }
+
+        /// <summary>
+        /// Used to ensure that the options are only initialized once.
+        /// </summary>
+        public bool Initialized { get; set; }
```
PinpointTownes: Yurk :astonished: 

Is there a particular reason this initialization logic is not something handled directly by the options stack?

src/Microsoft.AspNetCore.Authentication/AuthenticationSchemeOptions.cs
```diff
@@ -36,5 +36,15 @@ public virtual void Validate()
...
+
+        /// <summary>
+        /// Used to prevent concurrent access during intialization.
+        /// </summary>
+        public object InitalizeLock { get; } = new object();
```
PinpointTownes: nit: typo :smile:

src/Microsoft.AspNetCore.Authentication/AuthenticationSchemeOptions.cs
HaoK: Well auth is doing something a bit different than normal options usages, as handlers are doing fixup after all the Configures have been run.  If we bring back some kind of lifetime/second validation pass we could do it in options directly, but for now, we have to roll our own...

src/Microsoft.AspNetCore.Authentication/AuthenticationSchemeOptions.cs
HaoK: Doh, fixed

src/Microsoft.AspNetCore.Authentication/AuthenticationSchemeOptions.cs
PinpointTownes: I guess it would be clearer (and probably safer?) if options was responsible of the entire initialization process, 'cause that's likely the most appropriate place to determine whether options are singleton (= initialization should happen only once) or scoped (= initialization should happen for every instance).

PinpointTownes: I took a look at the options changes, and there's indeed nothing (yet) that could help with that. Have you considered introducing an `IOptionsValidator<T>`/`IOptionsInitializerT>` that would be run after all the `IConfigureOptions<T>`? (or you could bring back ordered setups).

HaoK: Yeah Validator/Intializer was there in the origional options PR but cut due to not wanting to rush that feature into options, we still have time to revisit where responsibility for single intialization goes before we ship, but for this initial checkin, its going to live in security

samples/OpenIdConnectSample/Startup.cs
```diff
@@ -41,8 +41,39 @@ public Startup(IHostingEnvironment env)
...
+                    }
+                };
+            });
+
+
             services.AddAuthentication(sharedOptions =>
-                sharedOptions.SignInScheme = CookieAuthenticationDefaults.AuthenticationScheme);
+            {
+                sharedOptions.DefaultAuthenticationScheme = CookieAuthenticationDefaults.AuthenticationScheme;
+                sharedOptions.DefaultSignInScheme = CookieAuthenticationDefaults.AuthenticationScheme;
```
Tratcher: OIDC as DefaultChallengeScheme

samples/OpenIdConnectSample/Startup.cs
```diff
@@ -158,7 +164,7 @@ await context.Authentication.SignOutAsync(OpenIdConnectDefaults.AuthenticationSc
                 {
                     // This is what [Authorize] calls
                     // The cookie middleware will intercept this 401 and redirect to /login
-                    await context.Authentication.ChallengeAsync();
+                    await context.ChallengeAsync(CookieAuthenticationDefaults.AuthenticationScheme);
```
Tratcher: `ChallengeAsync()`
This comment is wrong, this sample has no `/login`. It may have been copied from SocialAuth. This sample should challenge OIDC implicitly.

samples/OpenIdConnectSample/Startup.cs
```diff
@@ -169,7 +175,7 @@ await context.Authentication.SignOutAsync(OpenIdConnectDefaults.AuthenticationSc
                 // Authenticated, but not authorized
                 if (context.Request.Path.Equals("/restricted") && !user.Identities.Any(identity => identity.HasClaim("special", "true")))
                 {
-                    await context.Authentication.ChallengeAsync();
+                    await context.ChallengeAsync(CookieAuthenticationDefaults.AuthenticationScheme);
```
Tratcher: `ChallengeAsync()`

samples/OpenIdConnectSample/Startup.cs
```diff
@@ -72,32 +103,7 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
                 }
```
Tratcher: UseDeveloperExceptionPage in all samples.

samples/SocialSample/Startup.cs
```diff
@@ -228,14 +198,19 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
...
-            githubOptions.ClaimActions.MapJsonKey(ClaimTypes.NameIdentifier, "id");
-            githubOptions.ClaimActions.MapJsonKey(ClaimTypes.Name, "login");
-            githubOptions.ClaimActions.MapJsonKey("urn:github:name", "name");
-            githubOptions.ClaimActions.MapJsonKey(ClaimTypes.Email, "email", ClaimValueTypes.Email);
-            githubOptions.ClaimActions.MapJsonKey("urn:github:url", "url");
-            app.UseOAuthAuthentication(githubOptions);
+                };
+            });
+
+            services.AddAuthentication(options => options.DefaultAuthenticationScheme = CookieAuthenticationDefaults.AuthenticationScheme);
```
Tratcher: Cookies as DefaultChallengeScheme

Tratcher: Move this to the top for clarity

samples/SocialSample/Startup.cs
```diff
@@ -247,16 +222,18 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
...
 
                     context.Response.ContentType = "text/html";
                     await context.Response.WriteAsync("<html><body>");
                     await context.Response.WriteAsync("Choose an authentication scheme: <br>");
-                    foreach (var type in context.Authentication.GetAuthenticationSchemes())
+                    var schemeProvider = context.RequestServices.GetRequiredService<IAuthenticationSchemeProvider>();
+                    foreach (var provider in await schemeProvider.GetAllSchemesAsync())
                     {
-                        await context.Response.WriteAsync("<a href=\"?authscheme=" + type.AuthenticationScheme + "\">" + (type.DisplayName ?? "(suppressed)") + "</a><br>");
+                        // REVIEW: we lost access to display name (which is buried in the handler options)
```
Tratcher: This should be less of a problem if IISIntegration does not register WindowsAuth even when it's not active.

samples/SocialSample/Startup.cs
```diff
@@ -296,20 +273,20 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
                 var user = context.User;
 
                 // This is what [Authorize] calls
-                // var user = await context.Authentication.AuthenticateAsync(AuthenticationManager.AutomaticScheme);
+                // var user = await context.AuthenticateAsync(AuthenticationManager.AutomaticScheme);
```
Tratcher: AutomaticScheme is removed

samples/SocialSample/Startup.cs
```diff
@@ -296,20 +273,20 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
...
-                // var user = await context.Authentication.AuthenticateAsync(MicrosoftAccountDefaults.AuthenticationScheme);
+                // var user = await context.AuthenticateAsync(MicrosoftAccountDefaults.AuthenticationScheme);
 
                 // Deny anonymous request beyond this point.
                 if (user == null || !user.Identities.Any(identity => identity.IsAuthenticated))
                 {
                     // This is what [Authorize] calls
                     // The cookie middleware will intercept this 401 and redirect to /login
-                    await context.Authentication.ChallengeAsync();
+                    await context.ChallengeAsync(CookieAuthenticationDefaults.AuthenticationScheme);
```
Tratcher: `ChallengeAsync()`

samples/SocialSample/Startup.cs
```diff
@@ -324,11 +301,11 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
                 }
 
                 await context.Response.WriteAsync("Tokens:<br>");
-
-                await context.Response.WriteAsync("Access Token: " + await context.Authentication.GetTokenAsync("access_token") + "<br>");
-                await context.Response.WriteAsync("Refresh Token: " + await context.Authentication.GetTokenAsync("refresh_token") + "<br>");
-                await context.Response.WriteAsync("Token Type: " + await context.Authentication.GetTokenAsync("token_type") + "<br>");
-                await context.Response.WriteAsync("expires_at: " + await context.Authentication.GetTokenAsync("expires_at") + "<br>");
+                
+                await context.Response.WriteAsync("Access Token: " + await context.GetTokenAsync(CookieAuthenticationDefaults.AuthenticationScheme, "access_token") + "<br>");
```
Tratcher: Why do you need the auth scheme here? Can't it use DefaultAuthScheme?

src/Microsoft.AspNetCore.Authentication/baseline.net45.json
```diff
@@ -2474,7 +2474,7 @@
         },
         {
           "Kind": "Method",
-          "Name": "SkipToNextMiddleware",
+          "Name": "StopProcessing",
```
Tratcher: revert the baseline files, disable APICheck, and file a followup bug

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
```diff
@@ -4,29 +4,38 @@
...
+using Microsoft.Extensions.Options;
 
-namespace Microsoft.AspNetCore.Authentication
-{
-    /// <summary>
-    /// Base class for the per-request work performed by most authentication middleware.
-    /// </summary>
-    /// <typeparam name="TOptions">Specifies which type for of AuthenticationOptions property</typeparam>
-    public abstract class AuthenticationHandler<TOptions> : IAuthenticationHandler where TOptions : AuthenticationOptions
+namespace Microsoft.AspNetCore.Authentication {
```
Tratcher: nit: new lines

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
```diff
@@ -38,16 +47,6 @@ protected HttpResponse Response
...
 
-        protected PathString OriginalPathBase { get; private set; }
-
-        protected PathString OriginalPath { get; private set; }
-
-        protected ILogger Logger { get; private set; }
-
-        protected UrlEncoder UrlEncoder { get; private set; }
-
-        public IAuthenticationHandler PriorHandler { get; set; }
```
Tratcher: Should we leave all of the public/protected properties as Obsolete Error with the fwlink?

Tratcher: Easier to diff if you revert the order change

Security.sln
```diff
@@ -54,286 +52,436 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "OpenIdConnect.AzureAdSample
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.ChunkingCookieManager.Sources.Test", "test\Microsoft.AspNetCore.ChunkingCookieManager.Sources.Test\Microsoft.AspNetCore.ChunkingCookieManager.Sources.Test.csproj", "{51563775-C659-4907-9BAF-9995BAB87D01}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Authentication", "src\Microsoft.AspNetCore.Authentication\Microsoft.AspNetCore.Authentication.csproj", "{BC0D4B56-1A5B-4D88-AFBF-68C0F2D545FB}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Extensions.Options", "..\Options\src\Microsoft.Extensions.Options\Microsoft.Extensions.Options.csproj", "{151B7C1C-8E47-48CC-B04C-7414001D0F7B}"
```
HaoK: Yeah I forgot to push the updated sln, doing that now

samples/SocialSample/Startup.cs
HaoK: That overload is gone right now, could add it back if we really want.

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
HaoK: I don't think its worth preserving things in the Handler just for obsolete, since they will see the error already from all of the UseXyzs

HaoK:
Updated samples feedback, I'm not sure I love DefaultAuthenticationScheme, I thought I had named it DefaultAuthenticateScheme, which maps more closely to what it means... thoughts?

HaoK:
Ah I see what happened, its GetDefaultAuthenticateSchemeAsync in the real interface IAuthenticationSchemeProvider, I just forgot to rename the property on the options.  So we need to be consistent one way or the other.  I'll submit a PR in HTTP

samples/OpenIdConnect.AzureAdSample/Startup.cs
```diff
@@ -37,68 +37,55 @@ public Startup(IHostingEnvironment env)
...
-                            context.ProtocolMessage.Code, new Uri(currentUri), credential, resource);
+                            context.ProtocolMessage.Code, new Uri(currentUri), credential, Resource);
 
                         context.HandleCodeRedemption(result.AccessToken, result.IdToken);
                     }
-                }
+                };
+            });
+
+            services.AddAuthentication(sharedOptions =>
```
Tratcher: Let's put these defaults at the top of each sample so they don't get lost under all of the options noise.

samples/OpenIdConnect.AzureAdSample/Startup.cs
```diff
@@ -37,68 +37,55 @@ public Startup(IHostingEnvironment env)
...
+            {
+                sharedOptions.DefaultAuthenticationScheme = CookieAuthenticationDefaults.AuthenticationScheme;
+                sharedOptions.DefaultSignInScheme = CookieAuthenticationDefaults.AuthenticationScheme;
+                sharedOptions.DefaultChallengeScheme = OpenIdConnectDefaults.AuthenticationScheme;
             });
+        }
+
+        public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
+        {
+            loggerfactory.AddConsole(Microsoft.Extensions.Logging.LogLevel.Information);
```
Tratcher: namespace?

samples/OpenIdConnectSample/Startup.cs
```diff
@@ -96,8 +65,25 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
                         }
                         return c.Response.WriteAsync("An error occurred processing your authentication.");
                     }
-                }
+                };
+            });
+
+
+            services.AddAuthentication(sharedOptions =>
```
Tratcher: Move defaults to top

samples/OpenIdConnectSample/Startup.cs
```diff
@@ -144,32 +130,31 @@ await context.Authentication.SignOutAsync(OpenIdConnectDefaults.AuthenticationSc
                     return;
                 }
 
-                // CookieAuthenticationOptions.AutomaticAuthenticate = true (default) causes User to be set
+                // DefaultAuthenticationScheme causes User to be set
```
Tratcher: DefaultAuthenticateScheme

samples/SocialSample/Startup.cs
```diff
@@ -45,174 +44,151 @@ public Startup(IHostingEnvironment env)
...
             {
                 // User-Secrets: https://docs.asp.net/en/latest/security/app-secrets.html
                 // See below for registration instructions for each provider.
                 throw new InvalidOperationException("User secrets must be configured for each authentication provider.");
             }
 
+            services.AddAuthentication(options =>
+            {
+                options.DefaultAuthenticationScheme = CookieAuthenticationDefaults.AuthenticationScheme;
+                options.DefaultChallengeScheme = CookieAuthenticationDefaults.AuthenticationScheme;
```
Tratcher: DefaultSignInScheme

samples/SocialSample/Startup.cs
```diff
@@ -292,24 +273,24 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
 
             app.Run(async context =>
             {
-                // CookieAuthenticationOptions.AutomaticAuthenticate = true (default) causes User to be set
+                // Setting DefaultAuthenticateScheme causes User to be set
                 var user = context.User;
 
                 // This is what [Authorize] calls
-                // var user = await context.Authentication.AuthenticateAsync(AuthenticationManager.AutomaticScheme);
+                // var user = await context.AuthenticateAsync(DefaultAuthenticateScheme);
```
Tratcher: `AuthenticateAsync()`

samples/SocialSample/Startup.cs
Tratcher: Yeah, this is clunky

samples/OpenIdConnect.AzureAdSample/Startup.cs
```diff
@@ -37,68 +37,55 @@ public Startup(IHostingEnvironment env)
...
 
                         context.HandleCodeRedemption(result.AccessToken, result.IdToken);
                     }
-                }
+                };
+            });
+
+            services.AddAuthentication(sharedOptions =>
+            {
+                sharedOptions.DefaultAuthenticationScheme = CookieAuthenticationDefaults.AuthenticationScheme;
```
Tratcher: DefaultAuthenticateScheme

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
```diff
@@ -56,193 +56,104 @@ protected string CurrentUri
...
+        protected virtual async Task InitializeEventsAsync()
         {
-            return TaskCache.CompletedTask;
-        }
-
-        private async Task HandleAutomaticChallengeIfNeeded()
-        {
-            if (!ChallengeCalled && Options.AutomaticChallenge && Response.StatusCode == 401)
+            Events = Options.Events;
+            if (Options.EventsType != null)
```
Tratcher: While it's strange for someone to set both EventsType and Events, it's even stranger that EventsType is preferred.

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
```diff
@@ -56,193 +56,104 @@ protected string CurrentUri
...
-        {
-            return Request.Scheme + "://" + Request.Host + OriginalPathBase + targetPath;
-        }
 
-        private static async Task OnStartingCallback(object state)
-        {
-            var handler = (AuthenticationHandler<TOptions>)state;
-            await handler.FinishResponseOnce();
-        }
+            // REVIEW: should we validate options only once inside of the lock?
```
Tratcher: I think it needs to be outside. If the options are wrong then you need to fail every request, not just the first one.

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
```diff
@@ -56,193 +56,104 @@ protected string CurrentUri
...
-            OriginalPath = Request.Path;
-            Logger = logger;
-            UrlEncoder = encoder;
-
-            RegisterAuthenticationHandler();
-
-            Response.OnStarting(OnStartingCallback, this);
 
-            if (ShouldHandleScheme(AuthenticationManager.AutomaticScheme, Options.AutomaticAuthenticate))
+            Options = OptionsSnapshot.Get(Scheme.Name) ?? new TOptions();
```
Tratcher: I think cookie auth is the only one that works with default options. We'd be better off throwing a specific error here and requiring providers to register their options.

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
```diff
@@ -56,193 +56,104 @@ protected string CurrentUri
...
-            {
-                await FinishResponseOnce();
-            }
-            finally
-            {
-                UnregisterAuthenticationHandler();
-            }
-        }
+        /// <returns>A new instance of the events instance.</returns>
+        protected virtual Task<object> CreateEventsAsync() => Task.FromResult<object>(null);
```
Tratcher: Not null, return a new object or NoEventsDefined

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
```diff
@@ -56,193 +56,104 @@ protected string CurrentUri
...
-        public bool ShouldHandleScheme(string authenticationScheme, bool handleAutomatic)
+        protected string BuildRedirectUri(string targetPath)
         {
-            return string.Equals(Options.AuthenticationScheme, authenticationScheme, StringComparison.Ordinal) ||
-                (handleAutomatic && string.Equals(authenticationScheme, AuthenticationManager.AutomaticScheme, StringComparison.Ordinal));
+            return Request.Scheme + "://" + Request.Host + OriginalPathBase + targetPath;
         }
 
-        public async Task AuthenticateAsync(AuthenticateContext context)
+        public async Task<AuthenticateResult> AuthenticateAsync(AuthenticateContext context)
```
Tratcher: Context is not used? It's not even null checked.

src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs
```diff
@@ -2,90 +2,64 @@
...
-            UrlEncoder = encoder;
-
-            if (string.IsNullOrEmpty(Options.ClaimsIssuer))
-            {
-                // Default to something reasonable
-                Options.ClaimsIssuer = Options.AuthenticationScheme;
-            }
-
-            _next = next;
+            _next = next ?? throw new ArgumentNullException(nameof(next));
```
Tratcher: syntax

src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs
```diff
@@ -2,90 +2,64 @@
...
-        public ILogger Logger { get; set; }
-
-        public UrlEncoder UrlEncoder { get; set; }
+        public IAuthenticationSchemeProvider Schemes { get; set; }
 
         public async Task Invoke(HttpContext context)
         {
-            var handler = CreateHandler();
-            await handler.InitializeAsync(Options, context, Logger, UrlEncoder);
+            var oldFeature = context.Features.Get<IAuthenticationFeature>();
```
Tratcher: Why are you expecting there to be an old feature? There can only be one auth middleware now, correct?

src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs
```diff
@@ -2,90 +2,64 @@
...
+                    {
+                        context.User = result.Principal;
+                    }
                 }
-                catch (Exception)
+
+                // REVIEW: alternatively could depend on a routing middleware to do this
+
+                // Give any IAuthenticationRequestHandler schemes a chance to handle the request
+                var handlers = context.RequestServices.GetRequiredService<IAuthenticationHandlerProvider>();
```
Tratcher: Move HandleRequestAsync processing above defaultAuthenticate. HandleRequestAsync may short circuit the request.

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
HaoK: Fair enough, I'll nuke the review

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
HaoK: Yeah we have to do this because we always new up the events today.

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
Tratcher: Those are different audiences. These would be for implementors, not consumers.

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
Tratcher: I just meant we should change it to `if (Events == null && Options.EventsType != null)`

HaoK: Yeah so Events is always not null in general, as the default is an empty events object, if we added that check, they'd have to explicitly null out events and set eventstype.

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
HaoK: So its pretty much an error in any implementation of IOptionsSnapshot for Get to return null, so we could actually just assume that it always returns not null.

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
HaoK: Sweet, means we can nuke the type then

src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs
HaoK: "Should be one" as opposed to 'can only be one'.  Really we just need at least one.  But I'm fine with dropping the save/restore goop

HaoK: Is it important to null out features after we are done?  Or can I nuke the try/finally entirely?

samples/OpenIdConnect.AzureAdSample/Startup.cs
HaoK: Its because Azure also has a log level so this is disambiguiating

src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs
HaoK: Moved

src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs
```diff
@@ -2,90 +2,72 @@
...
+                    var result = await context.AuthenticateAsync(defaultAuthenticate.Name);
+                    if (result?.Principal != null)
+                    {
+                        context.User = result.Principal;
+                    }
                 }
+
+                await _next(context);
+            }
+            finally
```
Tratcher: Yes, you can remove the finally and the nulling out.

src/Microsoft.AspNetCore.Authentication.Twitter/Events/BaseTwitterContext.cs
```diff
@@ -1,27 +1,30 @@
...
+        /// <param name="properties">The AuthenticationProperties</param>
+        public BaseTwitterContext(HttpContext context, AuthenticationScheme scheme, TwitterOptions options, AuthenticationProperties properties)
+            : base(context, scheme.Name, properties)
         {
             Options = options;
         }
 
         public TwitterOptions Options { get; }
+
+        public AuthenticationScheme Scheme { get; }
```
Tratcher: It's weird that BaseAuthenticationContext has the scheme name, but BaseTwitterContext has the full scheme. BaseAuthenticationContext is being used for both events and Authenticate/Challenge/etc.? That's an awkward cross-over. How about seperating them to BaseAuthenticationContext and BaseAuthenticationEventContext?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -804,7 +892,7 @@ private void PopulateSessionProperties(OpenIdConnectMessage message, Authenticat
 
             var userInformationReceivedContext = await RunUserInformationReceivedEventAsync(ticket, message, user);
             AuthenticateResult result;
-            if (userInformationReceivedContext.CheckEventResult(out result))
+            if (userInformationReceivedContext.ProcessingCompleted(out result))
```
Tratcher: Verb? CheckProcessingCompleted?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -335,21 +423,21 @@ protected virtual Task<bool> HandleSignOutCallbackAsync()
...
             {
                 Logger.RedirectToIdentityProviderHandledResponse();
-                return true;
+                return;
             }
             else if (redirectContext.Skipped)
             {
                 Logger.RedirectToIdentityProviderSkipped();
-                return false;
+                return;
```
Tratcher: We'll have to revisit what skip means, this change makes it the same as Handled.

src/Microsoft.AspNetCore.Authentication.OAuth/Microsoft.AspNetCore.Authentication.OAuth.csproj
```diff
@@ -12,8 +12,8 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.AspNetCore.Authentication\Microsoft.AspNetCore.Authentication.csproj" />
-    <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="$(AspNetCoreVersion)" PrivateAssets="All" />
     <PackageReference Include="Newtonsoft.Json" Version="$(JsonNetVersion)" />
+    <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="$(AspNetCoreVersion)" PrivateAssets="All" />
```
Tratcher: revert

src/Microsoft.AspNetCore.Authentication.OAuth/Events/OAuthCreatingTicketContext.cs
```diff
@@ -39,53 +40,33 @@ public class OAuthCreatingTicketContext : BaseContext
...
-            if (user == null)
-            {
-                throw new ArgumentNullException(nameof(user));
-            }
-
-            TokenResponse = tokens;
-            Backchannel = backchannel;
-            User = user;
-            Options = options;
+            TokenResponse = tokens ?? throw new ArgumentNullException(nameof(tokens));
```
Tratcher: syntax

src/Microsoft.AspNetCore.Authentication.JwtBearer/Microsoft.AspNetCore.Authentication.JwtBearer.csproj
```diff
@@ -9,8 +9,8 @@
   </PropertyGroup>
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.AspNetCore.Authentication\Microsoft.AspNetCore.Authentication.csproj" />
-    <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="$(AspNetCoreVersion)" PrivateAssets="All" />
     <PackageReference Include="Microsoft.IdentityModel.Protocols.OpenIdConnect" Version="$(IdentityModelOpenIdVersion)" />
     <PackageReference Include="System.Security.Claims" Version="$(CoreFxVersion)" />
+    <PackageReference Include="Microsoft.Extensions.TaskCache.Sources" Version="$(AspNetCoreVersion)" PrivateAssets="All" />
```
Tratcher: revert

src/Microsoft.AspNetCore.Authentication.JwtBearer/Events/BaseJwtBearerContext.cs
```diff
@@ -2,24 +2,21 @@
...
+        public BaseJwtBearerContext(HttpContext context, AuthenticationScheme scheme, JwtBearerOptions options)
             : base(context)
         {
-            if (options == null)
-            {
-                throw new ArgumentNullException(nameof(options));
-            }
-
-            Options = options;
+            Options = options ?? throw new ArgumentNullException(nameof(options));
```
Tratcher: syntax

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -1,32 +1,96 @@
...
+            base.InitializeOptions();
+
+            if (String.IsNullOrEmpty(Options.CookieName))
+            {
+                Options.CookieName = CookieAuthenticationDefaults.CookiePrefix + Scheme.Name;
+            }
+            if (Options.TicketDataFormat == null)
+            {
+                var provider = Options.DataProtectionProvider ?? Context.RequestServices.GetRequiredService<IDataProtectionProvider>();
+                var dataProtector = provider.CreateProtector(typeof(CookieAuthenticationHandler).FullName, Scheme.Name, "v2");
```
HaoK: Our Cookie Interop tests don't seem to catch this, are we missing a variation?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -53,28 +53,117 @@ public class OpenIdConnectHandler : RemoteAuthenticationHandler<OpenIdConnectOpt
...
+                    {
+                        throw new InvalidOperationException("The MetadataAddress or Authority must use HTTPS unless disabled for development by setting RequireHttpsMetadata=false.");
+                    }
+
+                    Options.ConfigurationManager = new ConfigurationManager<OpenIdConnectConfiguration>(Options.MetadataAddress, new OpenIdConnectConfigurationRetriever(),
+                        new HttpDocumentRetriever(Backchannel) { RequireHttps = Options.RequireHttpsMetadata });
+                }
+            }
+
+            if (Options.ConfigurationManager == null)
```
HaoK: Yep moving the last check

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
HaoK: IsProcessingComplete?

src/Microsoft.AspNetCore.Authentication.Twitter/Events/BaseTwitterContext.cs
HaoK: I still want to see if we can improve the events code somehow, I'll file an issue to revisit this in the cleanup pass.

HaoK: Filed https://github.com/aspnet/Security/issues/1181 for this

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
Tratcher: They're hardcoding the dataprotection: https://github.com/aspnet/Security/blob/dev/test/Microsoft.Owin.Security.Interop.Test/CookieInteropTests.cs#L36

Tratcher: We also only have interop tests between Katana and Core, not between Core 1 and Core 2.

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -33,11 +94,11 @@ internal class JwtBearerHandler : AuthenticationHandler<JwtBearerOptions>
...
             {
                 // Give application opportunity to find from a different location, adjust, or reject token
-                var messageReceivedContext = new MessageReceivedContext(Context, Options);
+                var messageReceivedContext = new MessageReceivedContext(Context, Scheme, Options);
 
                 // event can set the token
-                await Options.Events.MessageReceived(messageReceivedContext);
-                if (messageReceivedContext.CheckEventResult(out result))
+                await Events.MessageReceived(messageReceivedContext);
+                if (messageReceivedContext.IsProcessingComplete(out result))
```
Tratcher: Still not a verb... `IsFoo` is used for bool properties.

src/Microsoft.AspNetCore.Authentication/AuthAppBuilderExtensions.cs
```diff
@@ -12,9 +12,9 @@ namespace Microsoft.AspNetCore.Builder
     public static class AuthAppBuilderExtensions
     {
         /// <summary>
-        /// Adds the <see cref="AuthenticationMiddleware"/> middleware to the specified <see cref="IApplicationBuilder"/>, which enables authentication capabilities.
+        /// Adds the <see cref="AuthenticationMiddleware"/> handler to the specified <see cref="IApplicationBuilder"/>, which enables authentication capabilities.
```
Tratcher: One too many replacements...

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
Tratcher: https://github.com/aspnet/Security/pull/1170

Tratcher: https://github.com/aspnet/Security/issues/1182

HaoK: Ughh, should have picked a better generic purpose rather than just use the type name, i.e. AspNet.Core.Cookie, oh well, hardcoding back to the old full name...

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
HaoK: Yeah this is internal plumbing right?  This typically isn't called by event handlers is it?

HaoK:
Okay I think everything major has been addressed here, MVC/Identity are pretty much done as well.  Working on MusicStore next.

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <returns></returns>
+        public static IServiceCollection AddFacebookAuthentication(this IServiceCollection services)
+            => services.AddFacebookAuthentication(FacebookDefaults.AuthenticationScheme, _ => { });
+
+        public static IServiceCollection AddFacebookAuthentication(this IServiceCollection services, Action<FacebookOptions> configureOptions) 
+            => services.AddFacebookAuthentication(FacebookDefaults.AuthenticationScheme, configureOptions);
+
+        public static IServiceCollection AddFacebookAuthentication(this IServiceCollection services, string authenticationScheme, Action<FacebookOptions> configureOptions)
+        {
+            services.AddSingleton<IConfigureOptions<FacebookOptions>, FacebookConfigureOptions>();
```
PinpointTownes: So, the magical `IConfiguration`-based binding will be used even if you specify a configuration delegate? Not sure I'm a huge fan of that... specially since this binder doesn't seem foolproof :smile:

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -1,32 +1,97 @@
...
+        /// </summary>
+        protected new CookieAuthenticationEvents Events
+        {
+            get { return (CookieAuthenticationEvents)base.Events; }
+            set { base.Events = value; }
+        }
+
+        public override async Task InitializeAsync(AuthenticationScheme scheme, HttpContext context)
+        {
+            await base.InitializeAsync(scheme, context);
```
PinpointTownes: The cool thing with the previous design is that we didn't have to call any of the base methods in the handler stuff. Any particular not to reason to preserve this pattern in 2.0?

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -1,32 +1,97 @@
...
+
+        /// <summary>
+        /// Creates a new instance of the events instance.
+        /// </summary>
+        /// <returns>A new instance of the events instance.</returns>
+        protected override Task<object> CreateEventsAsync() => Task.FromResult<object>(new CookieAuthenticationEvents());
+
+        protected override void InitializeOptions()
+        {
+            base.InitializeOptions();
```
PinpointTownes: (same remark)

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookExtensions.cs
HaoK: Well, your delegate runs after the binding so it 'wins'.  Consider the AddXyz stuff early preview stuff we want people to try and give feedback on, the behavior will almost certainly evolve from here.  Also in general the binding will be against an empty configuration section so its basically a default no-op (unless you happen to actually define a matching section to opt into the behavior)

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
HaoK: You mean if you override Initialize and don't call base?

HaoK: Or are you just asking for us to make it impossible to call base via a separate virtual for children to implement.

InitializeAsync(scheme, context) => // do stuff, then call protected virtual InitializeAsync() { }

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookExtensions.cs
PinpointTownes: > Well, your delegate runs after the binding so it 'wins'.

Only if the DI container respects the services registration order, right? (`IConfigureOptions.Order` is no longer a thing in 1.0/2.0, so it entirely relies on the DI capabilities)

It would be way clearer if this feature was more explicit (e.g by flowing the `IConfiguration` as a parameter). It would also the developer to select the nodes he/she wants:

```csharp
services.AddFacebookAuthentication(Configuration.GetSection("security:authentication:facebook"));
```

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
PinpointTownes: > You mean if you override Initialize and don't call base?

Yep. This way, the chances of messing with the default initialization logic are low if you don't forget to call the base method.

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookExtensions.cs
HaoK: That already exists, if you want to bind against a section, you can just call `Configure(Configuration.GetSection("facebook"))`.

DI respecting registration order is an implicit contract we already have today with options, especially without order, we must use registration order as order...

PinpointTownes: > That already exists, if you want to bind against a section, you can just call 

I know. My point is that the automatic configuration import should be explicit. If I provide a delegate, I don't expect the system to combine it with another source, specially since it's not consistent (cookies and OAuth2 don't use it)

> DI respecting registration order is an implicit contract we already have today with options, especially without order, we must use registration order as order...

A requirement that was not clear at all in 1.0 and caused much much pain when using custom containers (https://github.com/aspnet/Hosting/issues/785 / https://github.com/aspnet/DependencyInjection/issues/433). But okay, point taken.

PinpointTownes: > specially since it's not consistent (cookies and OAuth2 don't use it)

But OIDC uses it, hem.

It's even worse than what I imagined:

```
// Uses the OpenIdConnect node resolved from IConfiguration.
services.AddOpenIdConnectAuthentication();

// Uses the configuration delegate... PLUS the OpenIdConnect node resolved from IConfiguration (instead of custom-scheme).
services.AddOpenIdConnectAuthentication("custom-scheme", options => { ... });

Comment (HaoK) [inline src/Microsoft.AspNetCore.Authentication.Facebook/FacebookExtensions.cs]: Yeah that's where I started as well with (AddXyz() and AddXyz(o => { }) being disjoint, but given that its not possible to actually do things like bind events in the config, the delegate is going to be required most of the time, so it would be a pretty big pit of failure if they were disjoint...

I agree that might be a potential problem with no way to turn off the binding.

Comment (HaoK) [inline src/Microsoft.AspNetCore.Authentication.Facebook/FacebookExtensions.cs]: Yeah so the convention totally doesn't work when its not a 'default' section, which is why I didn't add it for cookies.  If multiple OIDC handlers are going to be used frequently, it doesn't make a good candidate either.

Comment (PinpointTownes) [inline src/Microsoft.AspNetCore.Authentication.Facebook/FacebookExtensions.cs]: > I agree that might be a potential problem with no way to turn off the binding.

Not just a potential problem: it's actually **completely broken** when you have multiple schemes of the same type (e.g multiple OIDC or JWT middleware), because the options system has no way to tell that `JwtBearerConfigureOptions` registered with the first `AddJwtBearerAuthentication()` call should go with the scheme A and not the scheme B (all the `JwtBearerConfigureOptions` apply to all the registered schemes).

To fix that, you'd have to introduce "named configure options".

Comment (HaoK) [inline src/Microsoft.AspNetCore.Authentication.Facebook/FacebookExtensions.cs]: Well the named configure options already exist, but this 'default' binding section really won't work with multiple schemes of the same type, that's why I didn't push it everywhere, its makes the most sense for singleton frameworks like identity/MVC.  

Its possible that this won't work well for auth at all if all of the schemes are likely to have multiple instances of the same type.

**Comment by Tratcher** [comment]
Looks like there's still several smaller comments un-addressed, but otherwise you're ready to go.

File: src/Microsoft.AspNetCore.Authentication.Facebook/FacebookExtensions.cs (hunk 44 above)
Comment (HaoK) [inline src/Microsoft.AspNetCore.Authentication.Facebook/FacebookExtensions.cs]: For the first iteration I'll just decouple the AddXyz() from the AddXyz(o => { }) again for now, we can figure out how to make things consistent later

Comment (PinpointTownes) [inline src/Microsoft.AspNetCore.Authentication.Facebook/FacebookExtensions.cs]: @HaoK yeah, sounds better :+1:

File: src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectExtensions.cs [hunk 47]
```diff
@@ -16,14 +16,16 @@ public static class OpenIdConnectExtensions
         /// <param name="services"></param>
         /// <returns></returns>
         public static IServiceCollection AddOpenIdConnectAuthentication(this IServiceCollection services)
-            => services.AddOpenIdConnectAuthentication(OpenIdConnectDefaults.AuthenticationScheme, _ => { });
+        {
+            services.AddSingleton<IConfigureOptions<OpenIdConnectOptions>, OpenIdConnectConfigureOptions>();
```
PinpointTownes: Wait, if you call this method to register an OIDC provider using `IConfiguration` + one of the other overloads to register another provider using a configuration delegate, this "configure options" instance will also apply to the other instances, right?

HaoK: No, these only configure the 'default' instance today, so the one named "OpenIdConnect" or whatever the default scheme name for each auth handler.

```
 public OpenIdConnectConfigureOptions(IConfiguration config) :
            base(OpenIdConnectDefaults.AuthenticationScheme,
                options => config.GetSection(OpenIdConnectDefaults.AuthenticationScheme).Bind(options))
        { }
``

HaoK: The only way to apply a configure to all named instances is to use ConfigureAll/specify null name, the overloads that don't take name, use the empty string which means the 'default' instance.

HaoK: But for auth we always have the scheme name, so we can always just use that.

PinpointTownes: Ah, cool then. Other question: what happens if I call `AddOpenIdConnectAuthentication()` twice? (which is likely not a good idea). Will the "configure options" be run twice? Wouldn't using `TryAddSingleton` make sense here?

HaoK: TryAddSingleton isn't appropriate for options since they are meant to be layered on, for example that means if you Configure before you call Add, the bind wouldn't get added, which would be super weird.

Binding twice isn't wrong, bit weird, and you'd have weirdness like, Bind, Configure, Bind would likely stomp all over stuff since the binder is a bit eager (and dumb)

PinpointTownes: Okay. Anyway, this shouldn't be a problem, since the second call to `AddOpenIdConnectAuthentication()` should throw immediately.

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
HaoK: I'm not sure we can do too much to avoid requiring the call to base, because we do have a chain in the base classes today (Base => Remote => OAuth, etc), other than adding another virtual InitializeYourOptions() which is really the same as this one, but less likely to clobber :)

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
HaoK: Initialize on the other hand, can be refactored a bit, with a new parameterless protected virtual InitializeHandler() that's called after Options/Events have all been initialized.

HaoK:
Okay I think that's the last change I have planned for this iteration (other than anything we will need from the other repos), I made the Initialize implementation non virtual, and have it call a new protected virtual InitializeHandler(), moved the two overrides we have today (Twitter/Cookies) to override that one instead, eliminating the need to call base in those situations.

PinpointTownes:
@HaoK thanks for the great work (and for being so patient :trollface:)

Please ping me when consistent CI packages are available, I'd love to try them ASAP, to make sure everything I need is there.

HaoK:
Will do, thanks as always for all the great feedback @PinpointTownes  :)

HaoK:
@PinpointTownes 24524 should have these changes now.

PR #1180, opened 2017-04-14
AddXyz() binds to default config section

HaoK:
I didn't add these overloads for Cookies/generic OAuth since those are more likely to have multiple instances, but I can if we want.

cc @glennc @ajcvickers @Tratcher @divega @davidfowl

### DIFF
--- a/samples/OpenIdConnect.AzureAdSample/Startup.cs
+++ b/samples/OpenIdConnect.AzureAdSample/Startup.cs
@@ -72,7 +72,7 @@ namespace OpenIdConnect.AzureAdSample
             });
 
             services.AddAuthentication(sharedOptions =>
-                sharedOptions.DefaultAuthenticationScheme = CookieAuthenticationDefaults.AuthenticationScheme);
+                sharedOptions.DefaultAuthenticateScheme = CookieAuthenticationDefaults.AuthenticationScheme);
         }
 
         public void Configure(IApplicationBuilder app, ILoggerFactory loggerfactory)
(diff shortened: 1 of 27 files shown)


HaoK:
@Tratcher I'm going to merge this into the mega security changes PR

HaoK:
I'll keep this PR open to keep it obvious what the changes specifically for the AddXyz overloads were

PinpointTownes:
@HaoK so these overloads require (manually) registering an `IConfiguration` instance in the DI container, right? Wouldn't passing the `IConfiguration` as an argument be clearer? Do you know how well the config binder will work with complex types like the token validation parameters class used by the JWT middleware?

Curious: do you plan to adopt this design in other ASP.NET Core components?

HaoK:
Yeah so hosting is adding configuration to DI now, so there's the concept of a default configuration.  As a result we can add a default AddXyz() overloads to anything where it makes sense to auto bind to a default config section in that configuration.

See https://github.com/aspnet/Hosting/commit/0ab882b6d3d2ceb40eedddc00316fc4f1b7eaecc

HaoK:
RE binding complex arguments, I'm sure it will throw or fail.  We likely will need to give the binder lots more love as a result of lightning things up this way...

Tratcher:
What would the config need to look like for each provider? How much can realistically be bound?

Tratcher:
I see the minimum binding in the tests. Can you add to the tests all properties you can currently bind to?

HaoK:
Added the config binding to the core AddXyz, so its always added by default, with the Action overriding.

Added all the properties that can be set to the test.  The one we might want to fix is the ability to bind to PathString which currently doesn't work with the binder.

Merging this to the main PR, if there are any other concerns we can deal with them there

HaoK:
Rolled into https://github.com/aspnet/Security/pull/1170

PR #1184, opened 2017-04-20
Set DisplayName for OAuth + Twitter

HaoK:
@Tratcher


Tratcher:
Can you add this back to the Social sample?

PinpointTownes:
Interesting, I was about to open a new ticket to reintroduce `DisplayName`.

The problem with this PR is that there are now 2 disconnected properties: one in the scheme handler configuration, and one in the scheme options. FWIW, I'm not a huge fan of using the scheme as the display name by default. Why not hardcoding the provider name, just like in 1.0.0?

HaoK:
Yeah we can iterate on how this lands for preview 2, but I need to get working templates this week, so I'm doing the bare minimum to put out the fires :)

We probably want overloads for AddXyz that let you specify the display name I guess longer term.

HaoK:
Oh yeah and the display name in the handler configuration is going to get nuked...

HaoK:
4f206558505b32c5ec0a686764ec30a8f912860e

Issue #1176, opened 2017-04-11
Add IAuthorizationHandlerProvider interface


rposener:
Is it possible to Map Policies to Requirements at Runtime?  E.g. I would like to secure parts of an app using policies ("Job Maintenance") but allow administrative user in my app to map them to certain requirements (stored in DB) and updated during runtime.  For example ("In Partners Group" or "Is Manager of Job") might be mapped to the "Job Maintenance" policy.

Is there a way to provide this kind of mapping during runtime?  Or change the ones that are currently loaded except to restart the app?

rposener:
Looks like I can just implement my own IAuthorizationPolicyProvider and register it.  Will give this a shot.  Would be good to see this updated on docs.microsoft.com

blowdart:
For your example you wouldn't need to head down that route, you'd just go off to your database during the evaluation of the policy, and pull down the group members, and the job manager during HandleRequirementAsync().

rposener:
Not sure I follow - how do I allow an admin user change what the requirements are to meet a certain policy?  E.g. I secure an area of the app to "Job Maintenance" policy.  But don't want to hard code that it's always the "Manager of the Job" or the "Partners in the Group".  I want the Administrative User of my app to make those decisions for their company.

blowdart:
So you still apply the Job maintenance policy to the area.

Then have a single requirement for that policy.

In that requirement go off to the database, figure out what the actual requirements are, and evaluate them.

rposener:
If I'm following - why use the AuthorizationService at all?  What's the point if I'm create my own policy, requirement, and mapping them together myself.  Guess I'm missing the point here.  The IAuthorizationPolicyProvider is registered to the services with TryAdd for a reason is it not?

Seems that what you are suggesting violates SRP - a requirement that then has to act as a policy and map which requirements must be met and then also evaluate all those requirements?  Requirements should be re-usable, and making a huge requirement to evaluate what requirements really diminishes reuse.

blowdart:
You could go either way. The Provider was there because people weren't happy with just have policies. It gives the somewhat limited ability to, for example, build policies based on policy name, replacing the parameters in custom attributes folks used before.

As for SRP, well that's a choice :) But unless you're restricting to a preset list of requirements in your database you're going to have to build them on the fly anyway, from whatever your ruleset requests, which makes them not reusable. Depends on the complexity of the rules.

rposener:
[...]

PR #1193, opened 2017-04-28
AuthZ 2.0 - [Breaking] AuthorizeResult/Failure

HaoK:
@blowdart @ajcvickers @Tratcher 

Needed for: https://github.com/aspnet/Security/issues/901

First step needed to surface why authorization fails, returning a result which can contain failure information instead of just a bool.

Also added IAuthorizationHandlerProvider, and reworked IAuthorizationEvaluator to work with the new API.

Fixes: https://github.com/aspnet/Security/issues/1176


src/Microsoft.AspNetCore.Authorization.Policy/Microsoft.AspNetCore.Authorization.Policy.csproj
```diff
@@ -0,0 +1,30 @@
...
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageTags>aspnetcore;authorization</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.AspNetCore.Authorization\Microsoft.AspNetCore.Authorization.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.Abstractions" Version="$(AspNetCoreVersion)" />
```
HaoK: AuthZ currently doesn't depend on Http or AuthN, if we are ok taking these dependencies in the main AuthZ package, we don't need to split out the authorize helper logic

src/Microsoft.AspNetCore.Authorization/AuthorizeFailure.cs
```diff
@@ -0,0 +1,46 @@
...
+
+using System.Collections.Generic;
+using System.Security.Claims;
+
+namespace Microsoft.AspNetCore.Authorization
+{
+    /// <summary>
+    /// Encapsulates a failure result of <see cref="IAuthorizationService.AuthorizeAsync(ClaimsPrincipal, object, IEnumerable{IAuthorizationRequirement})"/>.
+    /// </summary>
+    public class AuthorizeFailure
```
PinpointTownes: All these `verb` + `Result` classes sound a bit weird. Why not `AuthorizationFailure`?

HaoK: I started with that, but we went with AuthenticateResult (there that was meant to signify that it was a result for the Authenticate method).  I could go either way here since I've had it both ways in this code already :)

src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationPolicyResult.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace Microsoft.AspNetCore.Authorization
+{
+    public class AuthorizationPolicyResult
+    {
+        private AuthorizationPolicyResult() { }
+
+        /// <summary>
+        /// If true, means the callee should challenge and try again.
+        /// </summary>
+        public bool Challenged { get; private set; }
```
HaoK: Challenged => RequiresAuthentication?

HaoK:
Updated the PolicyEvaluator to have the new logic where when authorization fails, forbid is returned if any schemes were successfully authenticated.

src/Microsoft.AspNetCore.Authorization.Policy/PolicyEvaluator.cs
```diff
@@ -0,0 +1,89 @@
...
+                foreach (var scheme in policy.AuthenticationSchemes)
+                {
+                    var result = await context.AuthenticateAsync(scheme);
+                    if (result != null && result.Succeeded)
+                    {
+                        newPrincipal = SecurityHelper.MergeUserPrincipal(newPrincipal, result.Principal);
+                    }
+                }
+
+                if (newPrincipal == null)
```
Tratcher: != null?

HaoK: Doh yeah, looks like I am missing tests for the Authenticate method too :)

HaoK:
- Renamed AuthorizeResult/Failure => AuthorizationResult/Failure
- Eliminated the policy authentication result class trying to reuse AuthenticateResult directly instead.
- added tests for authentication of policy evaluator

src/Microsoft.AspNetCore.Authorization.Policy/PolicyEvaluator.cs
```diff
@@ -0,0 +1,89 @@
...
+                    context.User = newPrincipal;
+                    return AuthenticateResult.Success(new AuthenticationTicket(newPrincipal, string.Join(";", policy.AuthenticationSchemes)));
+                }
+                else
+                {
+                    context.User = new ClaimsPrincipal(new ClaimsIdentity());
+                    return AuthenticateResult.None();
+                }
+            }
+            return AuthenticateResult.None();
```
HaoK: We could explicitly check the result of AuthenticateAsync to determine if we want to return success/none instead to control the behavior of forbid/challenge if authorization fails later.

HaoK:
@Tratcher updated to react to the breaking auth changes in HttpAbstractions

HaoK:
e940cdb36bc7617af3d762ff9cc853a8fef55b0a

Issue #1201, opened 2017-05-04
Bring back Microsoft.Owin.Security.Interop and tests


Tratcher:
These got removed during the netcoreapp2.0 conversion due to complicated dependencies. Figure out how to bring them back.

Tratcher:
Offline discussion:
- Bring this back for preview1
- DataProtection goes to netstandard2.0 (not netcoreapp2.0)
- Invert the dependency between Hosting and DataProtection by moving the one extension method to the Extensions.Hosting.Abstractions package.

PinpointTownes:
On a related note, have you considered moving the DP core APIs to `Microsoft.Extensions.*` and having the hosting adapter in a `Microsoft.AspNetCore.*` package, so OWIN/Katana apps using the interop packages wouldn't have to reference all the ASP.NET Core primitives, that are not used at all in this case?

Tratcher:
In hindsight the DataProtection packages should have been named MS.Extensions.DP, but renaming them is not a break we're planning to make right now.

Tratcher:
Problem: We can bring back the interop package now, but the test project doesn't work anymore because it targets both Core and net46 scenarios.
Proposal: Postpone to preview2 and split the test to two full functional tests apps, one for Katana/net46 and one for Core/netcoreapp2.0 and an orchestration test project that sends cookies from one to the other.

PinpointTownes:
> Problem: We can bring back the interop package now, but the test project doesn't work anymore because it targets both Core and net46 scenarios.

What's the problem, exactly? This test project used to target `net451` (`net452` would be required with the new xUnit version), so why would it be suddenly impossible to run it?

Tratcher:
It was pulling in MS.A.A.Cookies to make sure it could round trip the cookies, and that project no longer supports net451.

PinpointTownes:
Arf... :disappointed:

Eilon:
We'll get it out for Preview2, there just wasn't time for Preview1. Preview2 isn't very far out 😄

PinpointTownes:
That's fine, I had not planned to start using 2.0 before preview2 anyway (and with the recent events, I'm not sure it's even worth it, TBH).

PinpointTownes:
Curious: will the RTM version of this package target Katana 3 or Katana 4?

Tratcher:
I'm targeting Katana 3.0.1 right now (same as before). None of these features are affected by the changes in Katana 4, so I don't see the need to raise the dependency.

[...]

PR #1202, opened 2017-05-04
Migrate to netcoreapp2.0

Tratcher:
https://github.com/aspnet/Security/issues/1201 I had to remove Microsoft.Owin.Security.Interop and tests due to their crossed dependencies between ASP.NET Core and Katana.


src/Microsoft.AspNetCore.Authentication.Cookies/Microsoft.AspNetCore.Authentication.Cookies.csproj
```diff
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <Description>ASP.NET Core middleware that enables an application to use cookie based authentication.</Description>
-    <TargetFramework>netstandard1.3</TargetFramework>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
```
PinpointTownes: Hum, does that mean that all the ASP.NET Core projects will no longer target .NET Standard? :open_mouth:

PinpointTownes: Obviously, this also means that we won't be able to use the security packages on .NET Desktop, since it's not cross-compiled for net46 and no longer targets .NET Standard. Meh.

Eilon: Yes, for ASP.NET Core 2, most of the libraries will target .NET Core 2 in order to use new APIs that are not yet available in a .NET Standard TFM. Code that needs to target multiple platforms, such as Microsoft.Extensions.*, Entity Framework Core, and a few other libraries, will continue to use .NET Standard.

PinpointTownes: That's insane. Not being able to use the ASP.NET Core packages on .NET Desktop will be blocking for many people. Is this change discussed/announced somewhere? Why not cross-compiling the few packages that could take advantage of `netcoreapp2.0` instead of retargeting everything?

Eilon: Please refer to https://github.com/aspnet/Home/issues/2022#issuecomment-299536123 for some more details on this.

PR #1206, opened 2017-05-08
Switch to services.Initialize<TOptions>

HaoK:
Goal is to eliminate the virtual AuthenticationHandler.InitializeOptions and the double check locking logic.

Depends on: https://github.com/aspnet/Options/pull/187

Doesn't seem too bad but I had to add a global ensure dataprotection initializer to make things work, definitely is a bit subtle.
 
I only did Cookies + OAuth and its children for a proof of concept first step.  I'll do OIDC next.

cc @Tratcher @ajcvickers

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
@@ -72,9 +72,7 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
         public CookieSecurePolicy CookieSecure { get; set; }
 
         /// <summary>
-        /// If set this will be used by the CookieAuthenticationMiddleware for data protection.
-        /// he default data protection service is based on machine key when running on ASP.NET, 
-        /// and on DPAPI when running in a different process.
+        /// If set this will be used by the CookieAuthenticationHandler for data protection.
         /// </summary>
         public IDataProtectionProvider DataProtectionProvider { get; set; }
 
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
@@ -4,6 +4,7 @@
 using System;
 using Microsoft.AspNetCore.DataProtection;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.Authentication.Cookies
 {
@@ -72,6 +73,8 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
 
         /// <summary>
         /// If set this will be used by the CookieAuthenticationMiddleware for data protection.
+        /// he default data protection service is based on machine key when running on ASP.NET, 
+        /// and on DPAPI when running in a different process.
         /// </summary>
         public IDataProtectionProvider DataProtectionProvider { get; set; }
 
@@ -129,9 +132,7 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
 
         /// <summary>
         /// The TicketDataFormat is used to protect and unprotect the identity and other properties which are stored in the
-        /// cookie value. If it is not provided a default data handler is created using the data protection service contained
-        /// in the IApplicationBuilder.Properties. The default data protection service is based on machine key when running on ASP.NET, 
-        /// and on DPAPI when running in a different process.
+        /// cookie value. If not provided one will be created using <see cref="DataProtectionProvider"/>.
         /// </summary>
         public ISecureDataFormat<AuthenticationTicket> TicketDataFormat { get; set; }
 
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieExtensions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieExtensions.cs
@@ -3,6 +3,10 @@
 
 using System;
 using Microsoft.AspNetCore.Authentication.Cookies;
+using Microsoft.AspNetCore.DataProtection;
+using Microsoft.Extensions.Options;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.AspNetCore.Authentication;
 
 namespace Microsoft.Extensions.DependencyInjection
 {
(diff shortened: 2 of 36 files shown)


src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
```diff
@@ -72,6 +73,8 @@ public string CookieName
 
         /// <summary>
         /// If set this will be used by the CookieAuthenticationMiddleware for data protection.
+        /// he default data protection service is based on machine key when running on ASP.NET,
```
Tratcher: You can just remove this comment, that's an implementation detail of DataProtection.

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
```diff
@@ -48,6 +48,8 @@ protected HttpResponse Response
         /// </summary>
         protected virtual object Events { get; set; }
 
+        protected virtual string ClaimsIssuer => Options.ClaimsIssuer ?? Scheme.Name;
```
HaoK: So we have to decide whether we require all AddXyz's to chaining to a new AddAuthenticationHandler method which adds an Initializer that just defaults ClaimsIssuer to scheme.Name, or add this new property on the handler which does the defaulting.

I'm leaning towards the property.  But we still are going to have chaining for each virtual level of InitializeOptions we have today, so there's going to be an AddRemoteAuthenticationHandler in the next iteration.

HaoK:
Updated with OIDC/Twitter changes, one nice sideeffect is some of the dependencies for the handlers are no longer needed, namely dataprotection/shared AuthenticationOptions, as those can now just live in the options initializers

src/Microsoft.AspNetCore.Authentication/AuthenticationServiceCollectionExtensions.cs
```diff
@@ -74,5 +74,30 @@ public static IServiceCollection AddAuthentication(this IServiceCollection servi
             where TOptions : AuthenticationSchemeOptions, new()
             where THandler : AuthenticationHandler<TOptions>
             => services.AddScheme<TOptions, THandler>(authenticationScheme, displayName, configureScheme: null, configureOptions: configureOptions);
+
+        public static IServiceCollection AddRemoteScheme<TOptions, THandler>(this IServiceCollection services, string authenticationScheme, string displayName, Action<TOptions> configureOptions)
+            where TOptions : RemoteAuthenticationOptions, new()
+            where THandler : AuthenticationHandler<TOptions>
```
Tratcher: RemoteAuthenticationHandler?

PinpointTownes:
I really love what's happening here :smile: 

That said, is there any particular reason we can't get rid of the initialization lock in `AuthenticationSchemeOptions` with this approach? :open_mouth:

HaoK:
Yeah that's the plan @PinpointTownes to get rid of that lock, I just got interrupted and need to finish removing whatevers left.  I'll update the PR with the final iteration today

PinpointTownes:
Brilliant!

src/Microsoft.AspNetCore.Authentication.Cookies/CookieExtensions.cs
```diff
@@ -15,7 +19,62 @@ public static class CookieExtensions
         public static IServiceCollection AddCookieAuthentication(this IServiceCollection services, Action<CookieAuthenticationOptions> configureOptions) =>
             services.AddCookieAuthentication(CookieAuthenticationDefaults.AuthenticationScheme, configureOptions);
 
-        public static IServiceCollection AddCookieAuthentication(this IServiceCollection services, string authenticationScheme, Action<CookieAuthenticationOptions> configureOptions) =>
-            services.AddScheme<CookieAuthenticationOptions, CookieAuthenticationHandler>(authenticationScheme, configureOptions);
+        public static IServiceCollection AddCookieAuthentication(this IServiceCollection services, string authenticationScheme, Action<CookieAuthenticationOptions> configureOptions)
+        {
+            // Makes sure that DataProtectionProvider is always set.
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<IInitializeOptions<CookieAuthenticationOptions>, EnsureDataProtection>());
+            services.Initialize<CookieAuthenticationOptions>(authenticationScheme, options =>
```
PinpointTownes: Instead of having the initialization in two places (in a delegate and in a private class), maybe it would be worth consolidating the two and moving everything to a public class? (e.g `CookieAuthenticationInitializer`)

HaoK: Hrm, do we want to allow people to replace/customize this behavior?  They already can via adding their own Initializer after, or Configuring it before.  I guess it doesn't hurt to leave it public, altho on an unrelated note, all of our handlers are still internal which probably should change now that they actually can be extended

HaoK:
Updated with final changes to JwtBearer and removed locking, also combined initializers into a single public type per Auth package.  Will merge Options PR + this next week

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthInitializer.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                options.Backchannel = new HttpClient(options.BackchannelHttpHandler ?? new HttpClientHandler());
+                options.Backchannel.DefaultRequestHeaders.UserAgent.ParseAdd("Microsoft ASP.NET Core OAuth handler");
+                options.Backchannel.Timeout = options.BackchannelTimeout;
+                options.Backchannel.MaxResponseContentBufferSize = 1024 * 1024 * 10; // 10 MB
+            }
+
+            if (options.StateDataFormat == null)
+            {
+                if (options.DataProtectionProvider == null)
```
PinpointTownes: @HaoK now that the two initialization routines have been merged, I guess this can be safely removed, no?

src/Microsoft.AspNetCore.Authentication.Cookies/CookieExtensions.cs
PinpointTownes: > Hrm, do we want to allow people to replace/customize this behavior?

Actually, that's not the scenario I had in mind. I was just wondering if making the initialization logic wouldn't prevent folks from registering the handler manually (i.e without using the extension), e.g if you don't want the `IConfiguration` importation thingy. But since the handlers are not public, this is not possible yet, even if you make the initializers public.

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthInitializer.cs
HaoK: Yep good catch, thanks I'll remove

HaoK:
1f5a27e20adf5438adcd9caf8587230307b685ce

Issue #908, opened 2016-07-20
Support for Cookie "SameSite" Flag


mariusschulz:
Are there any plans to implement support for the cookie [`SameSite` flag](https://en.wikipedia.org/wiki/HTTP_cookie#SameSite_cookie) within the cookie authentication middleware? The flag helps mitigate CSRF attacks and is currently recognized by Chrome and Opera, and possibly [more browsers soon](https://www.chromestatus.com/feature/4672634709082112).

I'm thinking about a new property on `CookieAuthenticationOptions`, as illustrated here:

``` cs
app.UseCookieAuthentication(new CookieAuthenticationOptions
{
    SameSite = SameSitePolicy.Always // or SameSitePolicy.None
    // ...
});
```

I think this feature has a very good cost-benefit ratio. The implementation should be relatively straightforward, and it wouldn't hurt to have an additional protection measure against CSRF besides antiforgery tokens.

Tratcher:
This was previously discussed here: https://github.com/aspnet/Security/issues/831#issuecomment-222013829

SameSite is still in draft form: https://tools.ietf.org/html/draft-west-first-party-cookies-07

We may add support for it when the spec is finished.

Tratcher:
@blowdart has requested this be re-triaged.

Tratcher:
Depends on https://github.com/aspnet/HttpAbstractions/issues/710

Eilon:
Need to also make sure we update the Cookie Policy middleware to allow setting this.

Eilon:
And Security-related cookies from this repo, plus also Antiforgery cookies would have SameSite enabled by default. We'd have to evaluate all other cookies created by ASP.NET Core to see what they should have by default.

Eilon:
Antiforgery work tracked here: https://github.com/aspnet/Antiforgery/issues/127

PR #1222, opened 2017-05-19
Add SameSitePolicy to CookiePolicyMiddleware

JunTaoLuo:
Addresses #908 Reacting to https://github.com/aspnet/HttpAbstractions/pull/843. Still need to update defaults.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
@@ -4,7 +4,6 @@
 using System;
 using Microsoft.AspNetCore.DataProtection;
 using Microsoft.AspNetCore.Http;
-using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.Authentication.Cookies
 {
@@ -23,6 +22,7 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
             ReturnUrlParameter = CookieAuthenticationDefaults.ReturnUrlParameter;
             ExpireTimeSpan = TimeSpan.FromDays(14);
             SlidingExpiration = true;
+            CookieSameSite = SameSiteMode.Strict;
             CookieHttpOnly = true;
             CookieSecure = CookieSecurePolicy.SameAsRequest;
             Events = new CookieAuthenticationEvents();
@@ -56,6 +56,12 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
         /// Determines the path used to create the cookie. The default value is "/" for highest browser compatibility.
         /// </summary>
         public string CookiePath { get; set; }
+
+        /// <summary>
+        /// Determines if the browser should allow the cookie to be attached to same-site or cross-site requests. The
+        /// default is Strict, which means the cookie is only allowed to be attached to same-site requests.
+        /// </summary>
+        public SameSiteMode CookieSameSite { get; set; }
 
         /// <summary>
         /// Determines if the browser should allow the cookie to be accessed by client-side javascript. The
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
@@ -899,6 +899,7 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
                 new CookieOptions
                 {
                     HttpOnly = true,
+                    SameSite = Http.SameSiteMode.Lax,
                     Secure = Request.IsHttps,
                     Expires = Clock.UtcNow.Add(Options.ProtocolValidator.NonceLifetime)
                 });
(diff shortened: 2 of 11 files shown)


src/Microsoft.AspNetCore.CookiePolicy/CookiePolicyMiddleware.cs
```diff
@@ -151,6 +151,22 @@ private void ApplyPolicy(CookieOptions options)
...
+                        if (options.SameSite == SameSiteEnforcementMode.None)
+                        {
+                            options.SameSite = SameSiteEnforcementMode.Lax;
+                        }
+                        break;
+                    case SameSitePolicy.AlwaysStrict:
+                        options.SameSite = SameSiteEnforcementMode.Strict;
+                        break;
+                    default:
+                        throw new InvalidOperationException();
```
nil4: Would it be easier for a developer to diagnose this case if an error message were provided here?

Tratcher: Poicy.SameSite.ToString(). This is more of an assert for if we add more enum values in the future.

src/Microsoft.AspNetCore.CookiePolicy/SameSitePolicy.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.CookiePolicy
+{
+    public enum SameSitePolicy
+    {
+        None = 0,
+        LaxOrStrict,
```
Tratcher: Names need work. The policy specifies the minimum required level. Maybe you don't need a new enum, you can use the existing one and just name the option SameSiteMinimum.

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
```diff
@@ -23,6 +22,7 @@ public CookieAuthenticationOptions()
             ReturnUrlParameter = CookieAuthenticationDefaults.ReturnUrlParameter;
             ExpireTimeSpan = TimeSpan.FromDays(14);
             SlidingExpiration = true;
+            CookieSameSite = SameSiteEnforcementMode.None;
```
blowdart: Probably. Strict would break, for example, links to a protected resource from another site. But given we try to be secure by default ... yea, strict.

JunTaoLuo:
🆙 📅

JunTaoLuo:
ping

JunTaoLuo:
This is currently blocking the build since MVC functional tests depends on these changes to pass. Otherwise we need to revert the HttpAbstractions commit.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -899,6 +899,7 @@ private void WriteNonceCookie(string nonce)
                 new CookieOptions
                 {
                     HttpOnly = true,
+                    SameSite = Http.SameSiteMode.Lax,
```
Tratcher: using?

src/Microsoft.AspNetCore.CookiePolicy/SameSitePolicy.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.CookiePolicy
+{
+    public enum MinimumSameSiteStrictnessPolicy
```
Tratcher: MinimumSameSitePolicy?
Since the values match the underlying enum, do we actually need this one?

JunTaoLuo: Ah yes, it used to be different since there were None, LaxOrStrict, and AlwaysStrict. But now it makes sense to just name it differently.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
JunTaoLuo: This is to differentiate between the one in M.A.Http and M.Net.Http.Headers. Both usings are already present and required.

JunTaoLuo: Brings up a good point, Should I be using the one from M.A.Http or M.Net.Http.Headers when I have the choice? Is there a preference for any particular one?

Tratcher: Stick with the aspnet one.

JunTaoLuo: Yup that's the one I try to use as much as possible.

Issue #1224, opened 2017-05-20
AddOpenIdConnectAuthentication doesn't work with 2.0.0-preview2-25108


vRITHNER:
Hi,
Any app I try to create with aspnetcore 2.0.0-preview2-25108 doesn't work because

> ArgumentException: Options.ClientId must be provided
> Parameter name: ClientId
> Microsoft.AspNetCore.Authentication.OpenIdConnect.OpenIdConnectOptions.Validate()
> 
> Stack Query Cookies Headers
> ArgumentException: Options.ClientId must be provided
>  Parameter name: ClientId

Even your sample 
`OpenIdConnectSample`
doesn't work

Tratcher:
While I did find a number of other issues with the samples (https://github.com/aspnet/Security/pull/1226), I'm not reproducing your issue in the latest dev branch. It sounds like a legitimate configuration error. OpenIdConnectSample does not include the ClientId config, it needs to be added on your own machine. I have mine set as an environment variable.

vRITHNER:
Hi, 
Thanks for the tip.

I've a request about the other sample with Azure AD connect; it could be great to have access (read-only) to a Azure sample instance to see all settings; I tried to replicate the sample with my own Azure instance and I surely made a mistake somewhere because when I click login, nothing happen...

Txs
-Vince

Tratcher:
Have you tried these docs?
https://azure.microsoft.com/en-us/resources/samples/active-directory-dotnet-webapp-openidconnect-aspnetcore/

vRITHNER:
FYI, still not working with latest template 

```
<ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.0-preview2-25149" />
    <PackageReference Include="Microsoft.EntityFrameworkCore.Tools" Version="2.0.0-preview2-25149" PrivateAssets="All" />
    <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="2.0.0-preview2-25149" PrivateAssets="All" />
  </ItemGroup>

  <ItemGroup>
    <DotNetCliToolReference Include="Microsoft.EntityFrameworkCore.Tools.DotNet" Version="2.0.0-preview2-25149" />
    <DotNetCliToolReference Include="Microsoft.Extensions.SecretManager.Tools" Version="2.0.0-preview2-25149" />
[...]

PR #1226, opened 2017-05-22
Make samples work. Fix AddOAuthAuthentication extension.

Tratcher:
@HaoK many of the samples are broken. I noticed while investigating https://github.com/aspnet/Security/issues/1224 (which doesn't repro in dev).


HaoK:
:shipit: looks like we are missing test coverage of AddOAuth :(

Issue #1231, opened 2017-05-24
OAuth authentication broken due to SameSite cookie policy


anurse:
The `SameSite=strict` cookie default added in https://github.com/aspnet/Security/issues/908 breaks OAuth handlers setting Cookies on signin because the user action was taken on a different origin.

The problem only when the user actually has to log in or otherwise interact with their OAuth provider. In that case, user action is initiated in the domain of the OAuth provider, so when the browser redirects back to the relying party, which then sets a `SameSite=strict` cookie, the browser refuses to send it on the next request. The flow is something like this (for app at `http://app` and provider at `http://provider`:

1. Browser sends `GET http://app`
2. `302` redirect to `http://provider`
3. Browser sends `GET http://provider`
4. `200` response with login page
5. User completes login
6. Browser sends `POST http://provider`
7. `302` redirect to `http://app/signin-oauth?code=...&state=...`
8. Browser sends `GET http://app/signin-oauth?code=...&state=...`
9. `302` redirect to `http://app` **with** `Set-Cookie` header with `samesite=strict`
10. Browser sends `GET http://app` **however** since the last document rendered to the user had an origin of `provider`, it refuses to send the cookie set by step 9.

This leads to a cycle in most ASP.NET OAuth configurations because it keeps trying to authenticate via a series of `302`s (because the provider a) already has a login cookie and b) is the origin site, so it keeps getting it's login cookie).

This should repro trivially by running any of the social login samples in an incognito window (forcing you to login).

Assuming I'm not missing something simple, **all OAuth-based authentication providers** are broken by this new default. We need to make some kind of fix/change in preview 2.

/cc @blowdart @Tratcher @muratg

anurse:
To clarify, I believe only Chrome is broken right now because Chrome is the only browser properly adhering to SameSite, but as other browsers come online they will be broken.

muratg:
cc @JunTaoLuo

anurse:
FYI: The way this manifests in the SocialSample sample is that the user is not logged in when the page first loads after a log in. If you refresh the page, you are logged in. However, if automatic challenging is on, you'll get into a redirect loop, since the cookie wasn't sent, so the handler thinks it needs to log in again.

JunTaoLuo:
We obviously need to at least set the samesite attribute to lax to keep things working. The spec for same site does not mention redirects so I'm not sure why the origin of the last document rendered is used to determine whether to send the cookie.

anurse:
[...]

PR #1232, opened 2017-05-25
Update default settings for SameSite

JunTaoLuo:
Addresses #1231  Need to test this with the auth samples before merging.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
@@ -22,7 +22,8 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
             ReturnUrlParameter = CookieAuthenticationDefaults.ReturnUrlParameter;
             ExpireTimeSpan = TimeSpan.FromDays(14);
             SlidingExpiration = true;
-            CookieSameSite = SameSiteMode.Strict;
+            // To support OAuth authentication, a lax mode is required, see https://github.com/aspnet/Security/issues/1231.
+            CookieSameSite = SameSiteMode.Lax;
             CookieHttpOnly = true;
             CookieSecure = CookieSecurePolicy.SameAsRequest;
             Events = new CookieAuthenticationEvents();
@@ -59,7 +60,8 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
 
         /// <summary>
         /// Determines if the browser should allow the cookie to be attached to same-site or cross-site requests. The
-        /// default is Strict, which means the cookie is only allowed to be attached to same-site requests.
+        /// default is Lax, which means the cookie is only allowed to be attached to cross-site requests using safe
+        /// HTTP methods and same-site requests.
         /// </summary>
         public SameSiteMode CookieSameSite { get; set; }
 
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationOptions.cs
```diff
@@ -22,7 +22,7 @@ public CookieAuthenticationOptions()
             ReturnUrlParameter = CookieAuthenticationDefaults.ReturnUrlParameter;
             ExpireTimeSpan = TimeSpan.FromDays(14);
             SlidingExpiration = true;
-            CookieSameSite = SameSiteMode.Strict;
+            CookieSameSite = SameSiteMode.Lax;
```
anurse: Can you put a comment in indicating why we're using `lax`? Even just a link to the issue would be sufficient. Just something so that somebody doesn't come along 6 months later and decide to change it :)

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -203,7 +203,7 @@ protected virtual void GenerateCorrelationId(AuthenticationProperties properties
             var cookieOptions = new CookieOptions
             {
                 HttpOnly = true,
-                SameSite = SameSiteMode.Lax,
+                SameSite = SameSiteMode.Strict,
```
PinpointTownes: Hum, does that **really** work with the OIDC client when using `response_mode=form_post`? (in this case, the IdP returns an auto-submit HTML form pointing to the callback URL).

PinpointTownes: BTW, I'm curious, what's the point of using a same-site policy for the correlation/anti-XSRF cookies?

PinpointTownes: Consistency is great and should always win, but be careful: `RemoteAuthenticationHandler` is a very generic base class that can be used to implement custom protocols (I personally wrote an OpenID 2.0 middleware leveraging it). That would be annoying to break scenarios that do fancy things with the callback dance... this same-site thingy tends to be really zealous :smile:

src/Microsoft.AspNetCore.Authentication.Twitter/TwitterHandler.cs
```diff
@@ -161,7 +161,7 @@ public TwitterHandler(IOptionsSnapshot<TwitterOptions> options, ILoggerFactory l
             var cookieOptions = new CookieOptions
             {
                 HttpOnly = true,
-                SameSite = SameSiteMode.Lax,
+                SameSite = SameSiteMode.Strict,
```
PinpointTownes: Same remark as above. Twitter - that implements OAuth1 - uses GET/302 responses to return the user back to the client app. If `Strict` doesn't work with the final application cookie due to the cross-origin 302 dance - as explained by @anurse - it will likely not work here.

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
PinpointTownes: > You're right, Strict isn't going to work.

@Tratcher I'm not sure `Lax` will work with `response_mode=form_post` since POST is not considered as a "safe" method by the specification.

@JunTaoLuo did `Lax` work for you?

JunTaoLuo:
Tested OIDCSample and OIDC.AzureAdSample. This is ready to merge

PR #1241, opened 2017-05-31
AddAzureAd[B2C] OIDC skin

HaoK:
- Add AddOpenIdConnect<TOptions> overload
- AddAzureAd[B2C] uses new overload
- AzureAd options use options initializers + config binding to set the default authority

cc @danroth27 @Tratcher


src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -25,9 +25,21 @@
...
+        public OpenIdConnectHandler(IOptionsSnapshot<OpenIdConnectOptions> options, ILoggerFactory logger, HtmlEncoder htmlEncoder, UrlEncoder encoder, ISystemClock clock)
+            : base(options, logger, htmlEncoder, encoder, clock)
+        { }
+    }
+
+    /// <summary>
+    /// A per-request authentication handler for OpenIdConnect.
+    /// </summary>
+    /// <typeparam name="TOptions">The options type used to configure the handler.</typeparam>
+    public class OpenIdConnectHandler<TOptions> : RemoteAuthenticationHandler<TOptions>
```
PinpointTownes: If you decide to go with this approach, maybe it would be worth modifying all the authentication middleware to use the same pattern, for consistency?

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdB2CInitializeOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+
+        public AzureAdB2CInitializeOptions() { }
+
+        public void Initialize(string name, AzureAdB2COptions options)
+        {
+            if (string.IsNullOrEmpty(options.Authority))
+            {
+                // Bind to only to any of the AzureAdB2C policy schemes
+                if (name == AzureDefaults.AzureAdB2CEditProfileAuthenticationScheme)
```
Tratcher: This is weird.

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdB2CInitializeOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+    internal class AzureAdB2CInitializeOptions : IInitializeOptions<AzureAdB2COptions>
+    {
+
+        public AzureAdB2CInitializeOptions() { }
+
+        public void Initialize(string name, AzureAdB2COptions options)
+        {
+            if (string.IsNullOrEmpty(options.Authority))
+            {
+                // Bind to only to any of the AzureAdB2C policy schemes
```
Tratcher: Typo

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdB2COptions.cs
```diff
@@ -0,0 +1,49 @@
...
+                    // Handle the error code that Azure AD B2C throws when trying to reset a password from the login page 
+                    // because password reset is not supported by a "sign-up or sign-in policy"
+                    if (context.Failure is OpenIdConnectProtocolException && context.Failure.Message.Contains("AADB2C90118"))
+                    {
+                        // If the user clicked the reset password link, redirect to the reset password route
+                        context.Response.Redirect("/Account/ResetPassword");
+                    }
+                    else if (context.Failure is OpenIdConnectProtocolException && context.Failure.Message.Contains("access_denied"))
+                    {
+                        context.Response.Redirect("/");
```
Tratcher: Infinite loop if the home page requires auth?

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdInitializeOptions.cs
```diff
@@ -0,0 +1,21 @@
...
+    internal class AzureAdInitializeOptions : IInitializeOptions<AzureAdOptions>
+    {
+
+        public AzureAdInitializeOptions() { }
+
+        public void Initialize(string name, AzureAdOptions options)
+        {
+            if (string.IsNullOrEmpty(options.Authority))
+            {
+                options.Authority = $"{options.Instance}{options.TenantId}";
```
Tratcher: And if Instance or TenantId are missing?

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+            => services.AddAzureAdAuthentication(AzureDefaults.AzureAdAuthenticationScheme, configureOptions);
+
+        public static IServiceCollection AddAzureAdAuthentication(this IServiceCollection services, string authenticationScheme, Action<AzureAdOptions> configureOptions)
+        {
+            services.AddSingleton<ConfigureDefaultOptions<AzureAdOptions>, AzureAdConfigureOptions>();
+            services.AddSingleton<IInitializeOptions<AzureAdOptions>, AzureAdInitializeOptions>();
+            return services.AddOpenIdConnectAuthentication(authenticationScheme, configureOptions);
+        }
+
+        public static IServiceCollection AddAzureAdB2CAuthentication(this IServiceCollection services)
```
Tratcher: Split AAD and B2C to different files everywhere. Treat them like they're in their own assemblies.

src/Microsoft.AspNetCore.Authentication.AzureAd/Microsoft.AspNetCore.Authentication.AzureAd.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\common.props" />
+
+  <PropertyGroup>
+    <Description>ASP.NET Core contains middleware to support Google's OpenId and OAuth 2.0 authentication workflows.</Description>
```
Tratcher: Copied from Google (which also looks out of date)?

src/Microsoft.AspNetCore.Authentication.AzureAd/Microsoft.AspNetCore.Authentication.AzureAd.csproj
```diff
@@ -0,0 +1,19 @@
...
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageTags>aspnetcore;authentication;security</PackageTags>
```
Tratcher: AAD, B2C

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdB2CInitializeOptions.cs
HaoK: Yeah Dan had to walk me through the scenario, basically there's some number of magic policies that B2C supports, there are 3 we currently support, and the idea is there's a single OIDC section in config we bind against, which is applied to each of the named schemes that map to the B2C policies (SignIn/ResetPassword/EditProfile)...  this is better than what the templates were doing before, but yeah its just a slightly cleaner hack :)

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdB2COptions.cs
HaoK: yeah @danroth27 hates this code too, were we given this code Dan?

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdInitializeOptions.cs
HaoK: Its no worse than before right?  This is just sitting up a better default for authority, but OIDC requires an authority by default right?

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdB2CInitializeOptions.cs
danroth27: Instead of special casing these three policies I'd prefer to have a generic Policies dictionary where the key is the authentication scheme name for that policy and the value is the policy ID you setup with Azure AD B2C.

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdB2COptions.cs
danroth27: This error handling code should live in the templates, not in the framework. It's based on the latest Azure AD B2C samples and docs, but I don't think it is generic enough to be part of the framework.

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdB2COptions.cs
```diff
@@ -0,0 +1,49 @@
...
+                        context.Response.Redirect("/Home/Error");
+                    }
+                    return Task.FromResult(0);
+                }
+            };
+        }
+
+        public string Instance { get; set; }
+        public string Domain { get; set; }
+        public string PolicyId { get; set; }
```
danroth27: What's the `PolicyId` property for? Remove?

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdB2COptions.cs
```diff
@@ -0,0 +1,49 @@
...
+        public string EditProfilePolicyId { get; set; }
```
danroth27: Instead of these three properties how about `public IDictionary<string, string> Policies { get; }`?

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Authentication.AzureAd;
+using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Options.Infrastructure;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public static class AzureExtensions
```
danroth27: These methods cover the web app case, but we need to also cover the Web API case where we setup JWT bearer authentication.

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdB2COptions.cs
PinpointTownes: @danroth27 the `access_denied` error is totally standard so this could definitely live in the OIDC/OAuth2 middleware: https://github.com/aspnet/Security/issues/710.

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdB2COptions.cs
HaoK: Doesn't work we need to know the properties ahead of time to make the binding work properly, feel free to file an issue to look into improving this more later, because its still a mega hack to even get these 3 to work today

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdB2CInitializeOptions.cs
HaoK: We need to revisit all of this again later post preview 2 when we have more time, it ended up being a mega hack still as you see here

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdB2COptions.cs
danroth27: > we need to know the properties ahead of time to make the binding work 

Are you referring to the config binding? The default config binder supports binding to dictionaries.

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdInitializeOptions.cs
Tratcher: If Instance is missing but TenantId is present then your authority is just TenantId, which breaks down. Check them in advance.

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdB2COptions.cs
HaoK: The premise of things today is we use AzureDefaults constants with the 3 magic values, once that becomes an arbitrary number of dynamic property names, constants don't work, and you basically will end up with some kind of AzureB2CPolicyManager service, which is likely where we end up eventually (just not in this PR)

danroth27: We can align the auth schemes in the templates without using a constant.

I don't understand why you would need a AzureB2CPolicyManager or what it would do.

HaoK: Its only job would be to bind/manage this dictionary of policies ids, and produce the correct OIDC auth schemes...aka deal with all this smoke and mirrors

danroth27:
@vibronet @SaeedAkhter-MSFT It would be great if you guys could take a look at this. We're adding some helpers for making it easier to setup auth with Azure AD and Azure AD B2C.

src/Microsoft.AspNetCore.Authentication.AzureAd/Microsoft.AspNetCore.Authentication.AzureAd.csproj
HaoK: Updated all descriptions

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdB2COptions.cs
HaoK: Yep leftover, good catch removed

src/Microsoft.AspNetCore.Authentication.Google/Microsoft.AspNetCore.Authentication.Google.csproj
```diff
@@ -3,7 +3,7 @@
   <Import Project="..\..\build\common.props" />
 
   <PropertyGroup>
-    <Description>ASP.NET Core contains middleware to support Google's OpenId and OAuth 2.0 authentication workflows.</Description>
+    <Description>Supports Google's OpenId and OAuth 2.0 authentication workflows.</Description>
```
Tratcher: OpenId was removed years ago.

HaoK:
Updated splitting AdB2c and Ad constants as well.  Will add some overloads for WebAPI next

HaoK:
Ok, bearer is added, so final surface area:

## AzureAd:
`AddAzureAdAuthentication`
`AddAzureAdBearerAuthentication`

Use `AzureAdDefaults.[Bearer]AuthenticationScheme`

Binds to config section: `Microsoft:AspNetCore:Authentication:Schemes:AzureAd[Bearer]`

## AzureAdB2C:
`AddAzureAdB2CAuthentication`
`AddAzureAdB2CBearerAuthentication`

Use `AzureAdB2cDefaults.BearerAuthenticationScheme` and the 3 policy named constant schemes.

Binds to config section: `Microsoft:AspNetCore:Authentication:Schemes:AzureAdB2C[Bearer]`

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdB2COptions.cs
Tratcher: So is this moving back to the template?

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+        public static IServiceCollection AddAzureAdAuthentication(this IServiceCollection services, Action<AzureAdOptions> configureOptions)
+        {
+            services.AddSingleton<ConfigureDefaultOptions<AzureAdOptions>, AzureAdConfigureOptions>();
+            services.AddSingleton<IInitializeOptions<AzureAdOptions>, AzureAdInitializeOptions>();
+            return services.AddOpenIdConnectAuthentication(AzureAdDefaults.AuthenticationScheme, configureOptions);
+        }
+
+        public static IServiceCollection AddAzureAdBearerAuthentication(this IServiceCollection services)
+        {
+            services.AddAzureAdAuthentication();
```
Tratcher: Why is Bearer dependent on OIDC?

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+            }
+        }
+
+        private class BearerConfigureOptions : ConfigureDefaultOptions<JwtBearerOptions>
+        {
+            private readonly IConfiguration _config;
+
+            public BearerConfigureOptions(IConfiguration config) => _config = config;
+
+            public override void Configure(string name, JwtBearerOptions options)
```
Tratcher: restrict by scheme name?

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdExtensions.cs
HaoK: It needs to read from the OIDC options to copy the authority/audience...weird yes

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdExtensions.cs
HaoK: Yeah good idea, I'll fix

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdB2COptions.cs
HaoK: My preference is to leave it in here for preview2, people can always override if they need to tweak, and we can move it into templates post preview2, but I'm working on the templates update now, so depending on how the ends up

src/Microsoft.AspNetCore.Authentication.AzureAd/AzureAdExtensions.cs
Tratcher: Can you not use Bearer without the OIDC middleware? If it's just dependent on the options then copy that part.

HaoK: Oh yeah, I see the bug, yeah it doesn't need the OIDC auth no...

PR #1248, opened 2017-06-02
React to Options, pubternal ConfigureDefaultOptions

HaoK:
Depends on https://github.com/aspnet/Options/pull/192

@Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationPostConfigureOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationPostConfigureOptions.cs
@@ -10,21 +10,21 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
     /// <summary>
     /// Used to setup defaults for all <see cref="CookieAuthenticationOptions"/>.
     /// </summary>
-    public class CookieAuthenticationInitializer : IInitializeOptions<CookieAuthenticationOptions>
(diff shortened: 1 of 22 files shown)


PinpointTownes:
So we have `ConfigureDefaultOptions`, `IConfigureOptions` and `IPostConfigureOptions`? :sweat_smile: 

At this point, wouldn't having a single interface/class w/ an `Order` property be cleaner?

HaoK:
We had order originally in Options early on, it was far from a panacea, it was just convention based with ints instead of names...

ConfigureDefaultOptions is not meant for external use, hence the infrastructure namespace/no interface.  People should go through the normal IConfigureOptions route.  PostConfigure is when you want to do stuff after people have configured things (aka fixup/validation/normalize set up some defaults).  Its much harder to mess ordering up since PostConfigures should used sparingly...

That said, it wouldn't surprise me if we eventually add more lifecycles if there's a compelling need...

HaoK:
@Tratcher this change is just hopefully making it clear that what was formerly known as IInitializeOptions, actually runs AFTER configure...  Lots of people assumed it would run first :(

Tratcher:
Finalizer? Validator?

HaoK:
Yeah this was discussed ad nauseam with the PMs yesterday and they settled on the clear (but ugly) name

HaoK:
db9ae6263233e1d95245cb911614d0e20a12edfa

PR #1250, opened 2017-06-05
Remove ConfigureDefaultOptions

HaoK:
@Tratcher @ajcvickers


Tratcher:
Going to remove the Binder dependency too?

HaoK:
Yeah good catch

HaoK:
Updated

PinpointTownes:
Does that mean that the "bind the scheme handlers config from DI-injected `IConfiguration`" plan is abandoned?

davidfowl:
Yes for 2.0 it's being abandoned. We'll revisit in 2.1

PinpointTownes:
Good to know. Out of curiosity, why abandoning it for 2.0? Not reliable enough?

davidfowl:
> Good to know. Out of curiosity, why abandoning it for 2.0? Not reliable enough?

We want to design it so that it's possible to opt out. At a high level, we want things to be coupled to `IOptions<T>` but not `IConfiguration`. Another layer in the stack should be responsible for adding the  `IConfigureOptions<T>` that binds from configuration, say in this case, `Security.Configuration`. 
Having this added by default when you call `AddXX` means the core component is coupled to `IConfiguration` which we're trying to avoid. 
Also it ends up making very simple middleware like this https://github.com/aspnet/BasicMiddleware/pull/241 require services, now require services which IMO is a non starter.

PinpointTownes:
Thanks for the explanation David, it makes total sense.

HaoK:
@davidfowl longer term I was thinking about just generalizing the ConfigBinder pattern to support custom binding via TypeConverters, so its always:

`AddXyz(o => config.Bind("Microsoft:AspNetCore:Xyz", o);`

1. Simple options that support binding using the binder today are unchanged.
2. Complex binding done via a custom bind can be done via a TypeConverter that can do IConfiguration => XyzOptions, add a top level branch in Bind<T> which will look for a type converter that can convert from IConfiguration to T
```
[TypeConverter(typeof(XyzConverter))] // Add a converter
public class XyzOptions { }

public class XyzConverter : TypeConverter
    {
        public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
        {
          return CustomBind((IConfiguration)value);
        }
    }
}
```

HaoK:
6d4fa4e91c08437ed8f6ea56deda810bf13738e6

PR #1256, opened 2017-06-09
[Prototype] Revisit the events model

HaoK:
@PinpointTownes this PR was from your branch + renames, I've only done Cookies for now, but that should demonstrate the idea, I'll take a look at OpenIdConnect as well now to make sure it still is consistent there.


HaoK:
There's some friction with the new HandlerContext and the intermediate events like CreatingTicket

HaoK:
So after doing a few of the conversions, it looks to me like it might be helpful to add a interface around each of our handlers that simply types the Options appropriately so we can constrain the events and access them... basically:

```
    public interface IOAuthHandler : IAuthenticationHandler
    {
        OAuthOptions Options { get; }
    }
```

HaoK:
That said in 2.1, we'd like follow pattern and add IXyzHandler when we make all of our handlers use this pattern.

HaoK:
Ok @PinpointTownes I'm done working on this PR for this week, I'll get notifications from github so I can answer any questions/help with any issues as you discover them.

Twitter/JwtBearer/OIDC/tests/samples are the things that need to be updated next.

src/Microsoft.AspNetCore.Authentication.Cookies/Events/CookieResultContext.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Authentication.Cookies
+{
+    public class CookieResultContext : ResultContext<CookieAuthenticationHandler>
```
PinpointTownes: It's a bit unfortunate to have to create another base class just to expose the options/scheme as first class properties :cry:

PinpointTownes:
> That said in 2.1, we'd likely follow this pattern and add IXyzHandler when we make all of our handlers use this pattern.

Just to expose the options? Sounds like a very weird pattern to me :fearful:

PinpointTownes:
@HaoK so concretely, the only advantage we have when exposing the handler in the context classes is that we get an access to the logger? (we could definitely add a property in `BaseContext` if you think it's worth it).

Not sure this tiny advantage outweighs the very annoying cons.

HaoK:
Well what actually drove the handler side of things was mostly due to making the base events generic, if you are going to make them generic, they should be generic on THandler not TOptions.

HaoK:
I mean today we have a bunch of base context's that effectively are doing the same thing that IXyzHandler would be doing, BaseXyzContext

src/Microsoft.AspNetCore.Authentication.OAuth/Events/OAuthCreatingTicketContext.cs
```diff
@@ -13,58 +13,48 @@ namespace Microsoft.AspNetCore.Authentication.OAuth
...
         /// </summary>
+        /// <param name="handler">The authentication handler.</param>
         /// <param name="ticket">The <see cref="AuthenticationTicket"/>.</param>
         /// <param name="context">The HTTP environment.</param>
         /// <param name="scheme">The authentication scheme.</param>
-        /// <param name="options">The options used by the authentication middleware.</param>
         /// <param name="backchannel">The HTTP client used by the authentication middleware</param>
         /// <param name="tokens">The tokens returned from the token endpoint.</param>
         public OAuthCreatingTicketContext(
+            IOAuthHandler handler,
```
Tratcher: @HaoK Introducing yet another abstraction here seems to defeat the point of exposing the handlers.

HaoK: Yeah given how little time we have, maybe we should just do the rename + related cleanup, but not push generics into the base events.  Keep the existing BaseXyzContext's that provide the correctly typed Options property.  But we can still push the Scheme down.

PinpointTownes: It also defeats the point of making the `IAuthenticationHandler` methods implemented explicitly in `AuthenticationHandler`, 'cause in this case, you can access them via `context.Handler.ChallengeAsync()`.

PinpointTownes:
> Well what actually drove the handler side of things was mostly due to making the base contexts generic, if you are going to make them generic, they should be generic on THandler not TOptions.

Could you please explain what's wrong with having a `TOptions` generic parameter?

HaoK:
> Could you please explain what's wrong with having a TOptions generic parameter?

There's nothing intrinsically wrong with the TOptions generic, but the combination of HttpContext + Options + Scheme matches an IAuthenticationHandler exactly now.  Something just seems off still with the events...

For example we still have the gross
```
        protected new OAuthEvents Events
        {
            get { return (OAuthEvents)base.Events; }
            set { base.Events = value; }
        }
```

HaoK:
So lets set the HandlerContext aside for now, keep your generic Options and leave it as BaseContext

@PinpointTownes forget my changes for now, and lets just apply the less controversial naming changes first, I'll noodle around with events a bit...

BaseContext [Instead of HandlerContext]
ResultContext : BaseContext
HandleRequestContext : BaseContext
RemoteResultContext : HandleRequestContext

And I think you can try just adding a virtual AuthenticationProperties to BaseContext so you can use it for Challenge/SignOut, and then just override it to return the Properties in the Ticket

HaoK:
So I noodled around with the events again, and I guess we can just fix one of the things that has always bothered me a different way, even if we can't push the Handler into the base context now, we can at least clean up all of the Context constructors to take their respective handlers and use the properties (make them public if needed), i.e.

`var context = new OAuthCreatingTicketContext(ticket, Context, Scheme, Options, Backchannel, tokens);`
becomes:
`var context = new OAuthCreatingTicketContext(this, ticket, tokens);` and have the ctor extract the properties.

This way it'll be easier to expose the handlers in the future since the context code will already be factored for that.

Issue #1011, opened 2016-10-22
Allow access to PolicyMap in AuthorizationOptions


atuzovic:
This request is related to AuthorizationOptions class in Microsoft.AspNetCore.Authorization library.

Due to architectural reasons, we need to access registered policies that are stored in AuthorizationOptions class as:
`private IDictionary<string, AuthorizationPolicy> PolicyMap { get; } = new Dictionary<string, AuthorizationPolicy>(StringComparer.OrdinalIgnoreCase);
`

This class has method allows retrieval of single policy by its name:

```
    public AuthorizationPolicy GetPolicy(string name)
        {
            if (name == null)
            {
                throw new ArgumentNullException(nameof(name));
            }

            return PolicyMap.ContainsKey(name) ? PolicyMap[name] : null;
        }
```

What we need is a similar method that would return all policies, something as simple as:

```
        public IDictionary<string, AuthorizationPolicy> GetPolicies()
        {
            return PolicyMap;
        }
```

Or, even easier version - PolicyMap property can be changed to have public gettter.

If this request is approved, I'd like to create pull request and make change myself.

Eilon:
@atuzovic can you share more info about why you need to list them? This is probably something we can implement, but the more we know about the scenario we can make sure we offer the right solution.

atuzovic:
Scenario that we want to support is to implement visibility management on certain components in AngularJS based on server-side policies.

When the user logs in, names of the policies that are satisfied are returned in the response.

On the client side, AngularJS components are secured like this:
`<some-component policy="RequiredPolicyName"></some-component>
`

In order for us to return policies that are satisfied, we need to access them from the code and validate if they are satisfied which we currently cannot.

HaoK:
@atuzovic feel free to submit a PR for this if you like as a starting point for discussion

HaoK:
We can do this in 2.1 as its not a breaking change

Eilon:
Closing because we think this change along would be insufficient for proper case-insensitivity support.

hikalkan:
Did you finally decided to not implement this?

I think it's essential for the case @atuzovic explained: https://github.com/aspnet/Security/issues/1011#issuecomment-264634640
And it's a very common scenario for a SPA.

PR #1261, opened 2017-06-12
Expose PolicyMap on options.

HaoK:
Fixes https://github.com/aspnet/Security/issues/1011

Pretty trivial just makes a private field public.

@Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Authorization/AuthorizationOptions.cs
+++ b/src/Microsoft.AspNetCore.Authorization/AuthorizationOptions.cs
@@ -11,7 +11,10 @@ namespace Microsoft.AspNetCore.Authorization
     /// </summary>
     public class AuthorizationOptions
     {
-        private IDictionary<string, AuthorizationPolicy> PolicyMap { get; } = new Dictionary<string, AuthorizationPolicy>(StringComparer.OrdinalIgnoreCase);
+        /// <summary>
+        /// Maps policy names to their AuthorizationPolicy.
+        /// </summary>
+        public IDictionary<string, AuthorizationPolicy> PolicyMap { get; } = new Dictionary<string, AuthorizationPolicy>(StringComparer.OrdinalIgnoreCase);
 
         /// <summary>
         /// Determines whether authentication handlers should be invoked after a failure.


src/Microsoft.AspNetCore.Authorization/AuthorizationOptions.cs
```diff
@@ -11,7 +11,10 @@ namespace Microsoft.AspNetCore.Authorization
     /// </summary>
     public class AuthorizationOptions
     {
-        private IDictionary<string, AuthorizationPolicy> PolicyMap { get; } = new Dictionary<string, AuthorizationPolicy>(StringComparer.OrdinalIgnoreCase);
```
Tratcher: This makes everything mutable and bypasses the argument validation on AddPolicy. What if you added GetPolicies that returned an IEnumerable instead?

davidfowl: This seems leaky since it's really an implementation detail of the default policy provider.

HaoK: Yeah this is a bit leaky since we aren't adding it to the actual interface (and I don't think we necessarily want to either, since enumerating all legal policies names isn't something that's required today).  But it seems fine for the concrete Options to expose access to its dictionary.  

How about we just expose it as a `IReadOnlyDictionary<string, AuthorizationPolicy>`?

Tratcher: IDictionary does not implement IReadOnlyDictionary

HaoK: Yeah but we can just create one when someone access it via the property

HaoK: I guess we should make it a method call then, GetPolicyMap() to try and signal that its a read only copy

HaoK: Note: the equivalent PolicyMap in AuthenticationOptions today is public, so whatever we decide to do here, we should probably be consistent: https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Authentication.Abstractions/AuthenticationOptions.cs#L22

Eilon:
Closing per discussion with @DamianEdwards and @davidfowl .

PR #1262, opened 2017-06-13
[Fixes #1133] Limit the path on the nonce and correlation id cookies

### LINKED ISSUE #1133: Consider limiting the cookies we use for nonce and correlationId to the paths that we use them on
Currently the handle of /, if there are incomplete logins they cause trouble and they stay on the browser cookie cache incrementing the size of each request. We should look into setting their path to the sign-in callback endpoint so that incomplete login flows don't increase the size of requests sent to the server.

javiercn:


src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -332,7 +332,18 @@ protected virtual Task<bool> HandleSignOutCallbackAsync()
             if (Options.ProtocolValidator.RequireNonce)
             {
                 message.Nonce = Options.ProtocolValidator.GenerateNonce();
-                WriteNonceCookie(message.Nonce);
+                var cookieOptions = new CookieOptions
```
Tratcher: Can you move this back inside WriteNonceCookie so it's out of the way?

Tratcher:
Did you intentionally skip Twitter? How about CookiePolicy?

PinpointTownes:
Why not events instead of options delegates?

Tratcher:
An interesting thought, but we want a consistent pattern across the stack and most other components don't use events.

javiercn:
@Tratcher I specifically left out anything that is not part of the OpenIdConnect middleware hierarchy.
I took a quick look at the repo and seems like twitter is the only other middleware that emits a cookie, we can path that one too if we choose to do so.

Here is the query that I used: https://github.com/aspnet/Security/search?utf8=%E2%9C%93&q=Cookies.Append&type=

As for the Action<HttpContext,CookieOptions> addition for other cookies I'm going to leave that to whomever does https://github.com/aspnet/HttpAbstractions/issues/853. That said, I think only twitter should have that option. We shouldn't do this in the cookiepolicy middleware as it would broaden the scope of the middleware a lot (and would require us to think more carefully of the consequences).

Issue #1089, opened 2017-01-17
Consider revisiting OpenIdConnectOptions.PostLogoutRedirectUri in 2.0.0


PinpointTownes:
> This is indeed a breaking change, which is quite subtile as existing code will still compile. At this point, it might be better to completely remove `PostLogoutRedirectUri` as it now has no added value and is quite confusing (with this PR, it's more like an equivalent to the return URL, but for logout). Alternatively, it should be marked as obsolete and removed in a future version.

https://github.com/aspnet/Security/pull/925#issuecomment-237384461

Now that the OIDC middleware has a dedicated "post logout endpoint", this property doesn't make much sense. I suggest removing `PostLogoutRedirectUri` in 2.0.0 to re-align the signout logic with the login flow (i.e honor `AuthenticationProperties.RedirectUri` when it's specified by the dev' and fall back to the current URL when it's not).

/cc @Tratcher

Tratcher:
Returning to the current url doesn't make sense as a fallback, it would likely cause a signout redirect loop.

PinpointTownes:
> Returning to the current url doesn't make sense as a fallback

That's what the OIDC middleware does in the 1.1.0 bits when `PostLogoutRedirectUri` is set to `null`:

```csharp
// Get the post redirect URI.
var properties = new AuthenticationProperties(signout.Properties);
if (string.IsNullOrEmpty(properties.RedirectUri))
{
    properties.RedirectUri = BuildRedirectUriIfRelative(Options.PostLogoutRedirectUri);
    if (string.IsNullOrWhiteSpace(properties.RedirectUri))
    {
        properties.RedirectUri = CurrentUri;
    }
}
```

That said, you have a valid point. Maybe it would be better to simply rename `PostLogoutRedirectUri` to a less confusing name (i.e that doesn't conflict with the OIDC parameter name).

Eilon:
How about the property setter for `PostLogoutRedirectUri` should throw if it is set to null-or-empty?

I don't think we'd get rid of this property entirely because it has some use. And renaming it would be a minimal improvement with high cost. We could certainly update the doc comments on the property if what we have now is misleading.

PinpointTownes:
> How about the property setter for PostLogoutRedirectUri should throw if it is set to null-or-empty?

Why not. Removing it and redirecting the user to the root when no `AuthenticationProperties.RedirectUri` is specified could be another option.

> We could certainly update the doc comments on the property if what we have now is misleading.

[...]

PR #1263, opened 2017-06-13
Rename PostLogoutRedirectUri to avoid spec confusion

Tratcher:
@PinpointTownes 
The usage of PostLogoutRedirectUri changed in 1.1 so that it no longer aligned with the spec term field. SignedOutCallbackPath has been used since 1.1 for the post_logout_redirect_uri protocol field. Renaming PostLogoutRedirectUri now to avoid confusion.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
@@ -167,15 +167,16 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
 
         /// <summary>
         /// The request path within the application's base path where the user agent will be returned after sign out from the identity provider.
+        /// See post_logout_redirect_uri from http://openid.net/specs/openid-connect-session-1_0.html#RedirectionAfterLogout.
         /// </summary>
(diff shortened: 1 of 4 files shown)


Tratcher:
https://github.com/aspnet/Security/issues/1089

PinpointTownes:
Why not something radically different like `DefaultSignOutReturnUrl`? (for symmetry, we could also have a `DefaultSignInReturnUrl`)

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
```diff
@@ -167,15 +167,16 @@ public new OpenIdConnectEvents Events
 
         /// <summary>
         /// The request path within the application's base path where the user agent will be returned after sign out from the identity provider.
+        /// See post_logout_redirect_uri.
```
PinpointTownes: Is this comment supposed to be helpful? 'cause it's... quite succinct :trollface:

Tratcher:
The tool didn't flag this as a breaking change because you already moved the whole options type to a different namespace.

HaoK:
Oh yeah, we had them in the Builder namespace before lolz...  yay

PinpointTownes:
@Tratcher you didn't like the names I proposed? :trollface:

Issue #1181, opened 2017-04-17
Auth 2.0 Cleanup: Revisit cleanup events/context


HaoK:
Sharing context for core IAuthenticationHandler and the events might be a bit awkward, also see if there's any improvements that can be made for the events as a whole

PinpointTownes:
@HaoK now that the authentication scheme is not longer stored in the options, retrieving it from the events is next to impossible. What do you think of introducing a `Scheme` property in `BaseContext`?

```csharp
public class BaseContext
{
    protected BaseContext(HttpContext context, string scheme);

    public string Scheme { get; }
}
```

If `BaseContext` was moved to this repo, we could even expose it as a strongly typed property:

```csharp
public class BaseContext
{
    protected BaseContext(HttpContext context, AuthenticationScheme scheme);

    public AuthenticationScheme Scheme { get; }
}
```

PinpointTownes:
@HaoK another question: do you remember why you've decided to rename `BaseControlContext.SkipToNextMiddleware()` to `Skip()`? The new name is not as clear as the old one, and doesn't really reflect what happens in practice.

HaoK:
For events, we have: https://github.com/aspnet/Security/issues/1181 to revisit exactly that question.

The whole Skip/Handled thing is a bit weird, and what makes sense in some uses (SkipToNextMiddleware) makes sense when its being run in the HandleRequest of a callback for an auth request, but doesn't make sense anywhere else.  We can use the same events issue to track revisiting that as well.

PinpointTownes:
> For events, we have: #1181 to revisit exactly that question.

So this is the right place? :trollface: 

> The whole Skip/Handled thing is a bit weird.

Weird... but insanely powerful :sweat_smile: 

> What makes sense in some uses (SkipToNextMiddleware) makes sense when its being run in the HandleRequest of a callback for an auth request, but doesn't make sense anywhere else.

Sure, but doesn't the `BaseControlContext` name imply that the event should be used only where we have flow control (i.e from `HandleRequestAsync`) and where `Skip` really means "skip the default logic and invoke the rest of the pipeline"?

HaoK:
Possibly, I think we were using it in a variety events, not just ones that are invoked through the request processing pipeline.

PinpointTownes:
@Eilon I see you moved this item to preview3. If it can help make it happen sooner, I'd love to work on it and come up with a formal API change :sweat_smile:

Eilon:
[...]

PR #1243, opened 2017-05-31
[Prototype] Revisit the events model

PinpointTownes:
Things I've changed:

  - `AuthenticateResult` was split into 2 classes: `AuthenticationResult` (living in the abstractions repo) and `RemoteAuthenticationResult` (living here).
  - `AuthenticateResult.Handle()`/`Skip()` was moved to `RemoteAuthenticationResult`.
  - The base context classes are now generic and expose the options + the scheme.
  - The specialized OIDC/JWT/cookies base classes are gone thanks to the new generic classes.
  - `HandleResponse()`/`SkipToNextMiddleware()` are no longer available from events that don't offer real flow control (basically all the events that are not called from `HandleRemoteAuthenticationAsync()`.
  - Events that used to expose `HandleResponse()`/`SkipToNextMiddleware()` for something else that real flow control now have `CompleteAuthentication`/`SkipAuthentication`/`RejectAuthentication` (or the equivalent for challenge/signout.

Please note that it's a very early prototype :smile: 

/cc @HaoK @Tratcher

PR #1266, opened 2017-06-15
Pinpoint's Grand Eventure

HaoK:
Based on https://github.com/aspnet/Security/pull/1243

Fixes https://github.com/aspnet/Security/issues/1181

Clean PR for review starting point

@Tratcher @PinpointTownes


src/Microsoft.AspNetCore.Authentication.Cookies/Events/CookieValidatePrincipalContext.cs
```diff
@@ -19,55 +19,19 @@ public class CookieValidatePrincipalContext : BaseCookieContext
...
         /// </summary>
         public bool ShouldRenew { get; set; }
-
-        /// <summary>
-        /// Called to replace the claims principal. The supplied principal will replace the value of the 
-        /// Principal property, which determines the identity of the authenticated request.
-        /// </summary>
-        /// <param name="principal">The <see cref="ClaimsPrincipal"/> used as the replacement</param>
-        public void ReplacePrincipal(ClaimsPrincipal principal)
-        {
```
HaoK: Likely need to bring these back, this will break Identity/cookie stamp validator

src/Microsoft.AspNetCore.Authentication/Microsoft.AspNetCore.Authentication.csproj
```diff
@@ -11,10 +11,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Authentication.Core" Version="$(AspNetCoreVersion)" />
+    <ProjectReference Include="..\..\..\HttpAbstractions\src\Microsoft.AspNetCore.Authentication.Core\Microsoft.AspNetCore.Authentication.Core.csproj" />
```
HaoK: Will remove

test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs
```diff
@@ -370,11 +370,40 @@ private Task SignOutAsWrong(HttpContext context)
                 o.SlidingExpiration = false;
                 o.Events = new CookieAuthenticationEvents
                 {
-                    OnValidatePrincipal = ctx =>
+                    OnValidatePrincipal = async ctx =>
                     {
-                        ctx.RejectPrincipal();
```
HaoK: reminder: to review more closely

test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs
```diff
@@ -370,11 +370,40 @@ private Task SignOutAsWrong(HttpContext context)
...
+        {
+            var server = CreateServer(o =>
+            {
+                o.ExpireTimeSpan = TimeSpan.FromMinutes(10);
+                o.SlidingExpiration = false;
+                o.Events = new CookieAuthenticationEvents
+                {
+                    OnValidatePrincipal = async ctx =>
+                    {
+                        ctx.None();
```
HaoK: Probably need a sugar method to alias None here.

test/Microsoft.AspNetCore.Authentication.Test/JwtBearerTests.cs
```diff
@@ -396,7 +396,7 @@ public class JwtBearerTests
                 {
                     OnMessageReceived = context =>
                     {
-                        context.Skip();
+                        context.None();
```
HaoK: None isn't a great name, probably need to pick something else, or add sugar for it as appropriate

version.props
```diff
@@ -2,6 +2,6 @@
 <Project>
     <PropertyGroup>
         <VersionPrefix>2.0.0</VersionPrefix>
-        <VersionSuffix>preview3</VersionSuffix>
+        <VersionSuffix>preview3-25644</VersionSuffix>
```
HaoK: Revert

HaoK:
FYI @PinpointTownes I do agree with you that this looks much better!  I think the only main thing left is to finalize on names, None was always a pretty meh name, so I think that one for sure needs to be changed, or sugar'd over at least.

PinpointTownes:
@HaoK you already know my POV: `SkipToNextMiddleware()`/`HandleResponse()`/`CompleteAuthentication()`/`RejectAuthentication()`/`SkipAuthentication` - this last one could be renamed to `IgnoreAuthentication()` - are the right mix between clarity and shortness (a two-word method name is ultra reasonable!).

Trying to shorten them will only make them less clear than they are already...

src/Microsoft.AspNetCore.Authentication.Cookies/Events/CookieValidatePrincipalContext.cs
PinpointTownes: `context.CompleteAuthentication(ticket)`.

test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs
Tratcher: What does this even do?

HaoK: There used to be an explicit method called ctx.RejectPrincipal();

HaoK: That would just null out Principal

Tratcher: And you need a method to null out the principal because...? Why not just null it out?
RejectPrincipal would be much clearer.

samples/JwtBearerSample/Startup.cs
```diff
@@ -59,7 +59,7 @@ public void ConfigureServices(IServiceCollection services)
                 {
                     OnAuthenticationFailed = c =>
                     {
-                        c.HandleResponse();
+                        c.None();
```
Tratcher: Yeah, this needs work. "Discontinue" "IveTakenCareOfEverything" "Stop"

samples/OpenIdConnect.AzureAdSample/Startup.cs
```diff
@@ -70,7 +70,7 @@ public void ConfigureServices(IServiceCollection services)
                         var authContext = new AuthenticationContext(Authority, AuthPropertiesTokenCache.ForCodeRedemption(context.Properties));
 
                         var result = await authContext.AcquireTokenByAuthorizationCodeAsync(
-                            context.ProtocolMessage.Code, new Uri(currentUri), credential, Resource);
```
Tratcher: I understand why this was changed, but ProtocolMessage needs to stay. Alias it if you want to, but ProtocolMessage is the thing in the event that gets used and will break the most consumers. Changing the context names won't affect most people when they just use the lambadas.

HaoK: I'd rather revert than add alias's, but I could be convinced either way, @PinpointTownes ?

samples/JwtBearerSample/Startup.cs
HaoK: Pinpoint called this one Skip/IgnoreAuthentication, c.Ignore() works for me as, seems like a strict improvement over None/Nothing anyways...  => Ignore()/Ignored

test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs
HaoK: Yeah I think we should keep the explicit Refresh/RejectPrincipal we have today, they'll just be updated underneath the covers if needed to do the right thing

samples/OpenIdConnect.AzureAdSample/Startup.cs
PinpointTownes: If I really wanted to waste time convincing you, @HaoK, that would be for the context method names... :trollface: 

`ProtocolMessage` is not ideal, but at least, it's self-explanatory.

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -419,7 +424,7 @@ private static bool IsHostRelative(string path)
             }
 
             var loginUri = Options.LoginPath + QueryString.Create(Options.ReturnUrlParameter, redirectUri);
-            var redirectContext = new CookieRedirectContext(Context, Scheme, Options, BuildRedirectUri(loginUri), properties);
+            var redirectContext = new CookieRedirectContext(Context, Scheme, Options, properties, BuildRedirectUri(loginUri));
```
Tratcher: These order changes seem random

PinpointTownes: It's for consistency with the other events, where `AuthenticationProperties` generally precedes parameters that are specific to the context class (e.g `OAuthRedirectToAuthorizationContext`).

That said, if you don't like it, feel free to revert it.

src/Microsoft.AspNetCore.Authentication.JwtBearer/Events/JwtBearerChallengeContext.cs
```diff
@@ -40,5 +42,15 @@ public JwtBearerChallengeContext(HttpContext context, AuthenticationScheme schem
         /// WWW-Authenticate header. This property is always null unless explicitly set.
         /// </summary>
         public string ErrorUri { get; set; }
+
+        /// <summary>
+        /// If true, will skip any default logic for this event.
+        /// </summary>
+        public bool Skipped { get; private set; }
```
PinpointTownes: Using `EventResultState` had one merit... getting rid of all these bool properties.

PinpointTownes:
For the record, the differences between my PR and @HaoK's names can be seen here: https://github.com/PinpointTownes/Security/pull/1/files

src/Microsoft.AspNetCore.Authentication.JwtBearer/Events/JwtBearerChallengeContext.cs
HaoK: This could still live in BaseContext, but since none of the other events check for skip, doesn't really make sense to move it there now.  In the future we could move this concept to be something that all events support if we wanted.

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
HaoK: Eh, maybe random, but its more work to revert them for no benefit, so I say leave it.

HaoK:
This PR will now depend on https://github.com/aspnet/HttpAbstractions/pull/873 which we will use to change AuthenticateResult

PinpointTownes:
This PR wasn't big enough for you? :trollface:

HaoK:
I can't juggle two PRs at once, so might as well conflate them :)

src/Microsoft.AspNetCore.Authentication.Cookies/Events/CookieValidatePrincipalContext.cs
HaoK: Yeah that's what ReplacePrincipal will be calling

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -155,20 +155,25 @@ private void RequestRefresh(AuthenticationTicket ticket)
                 return result;
             }
 
-            var context = new CookieValidatePrincipalContext(Context, Scheme, result.Ticket, Options);
+            var context = new CookieValidatePrincipalContext(Context, Scheme, Options, result.Ticket);
             await Events.ValidatePrincipal(context);
 
-            if (context.Principal == null)
+            if (context.ShouldRenew && context.Ticket != null)
```
HaoK: Reverting these changes, this way all of the cookie events only use BaseContext consistently.

HaoK: I'll file a follow up bug to look into normalizing cookies (adding concept of skip across the board)

HaoK: https://github.com/aspnet/Security/issues/1268

test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs
HaoK: Removing this test now, as there's no skip concept, you can manipulate only the principal from ValidateContext, no control flow to deal with...

test/Microsoft.AspNetCore.Authentication.Test/FacebookTests.cs
```diff
@@ -182,10 +181,6 @@ public class FacebookTests
                 app => app.UseAuthentication(),
                 services =>
                 {
-                    services.AddAuthentication(options =>
-                    {
-                        options.DefaultSignInScheme = "External";
-                    });
```
HaoK: @Tratcher @PinpointTownes the benefit of the split is here, we no longer need to specify DefaultSignInScheme when only a single cookie is added, we can leave SignInScheme null on the options, as it just works...

HaoK:
Rolled https://github.com/aspnet/Security/pull/1270 into this PR

src/Microsoft.AspNetCore.Authentication/Events/EventResultState.cs
```diff
@@ -11,13 +11,18 @@ public enum EventResultState
         Continue,
```
Tratcher: is this enum still used?

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -69,7 +68,7 @@ public virtual Task<bool> ShouldHandleRequestAsync()
                 {
                     return true;
                 }
-                else if (authResult.Nothing)
+                else if (authResult.Skipped || authResult.Nothing)
```
Tratcher: Why the duplication?

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationOptions.cs
```diff
@@ -23,11 +23,6 @@ public override void Validate()
             {
                 throw new ArgumentException(Resources.FormatException_OptionMustBeProvided(nameof(CallbackPath)), nameof(CallbackPath));
             }
-
-            if (string.IsNullOrEmpty(SignInScheme))
-            {
-                throw new ArgumentException(Resources.FormatException_OptionMustBeProvided(nameof(SignInScheme)), nameof(SignInScheme));
```
Tratcher: I suspect there are a bunch of log statements and resources that can be removed now.

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -244,21 +247,34 @@ private CookieOptions BuildCookieOptions()
...
+            {
+                throw new ArgumentNullException(nameof(user));
+            }
+
+            properties = properties ?? new AuthenticationProperties();
+            await HandleSignInAsync(user, properties);
+            Logger.SignedIn(Scheme.Name);
+        }
+
+        protected virtual async Task HandleSignInAsync(ClaimsPrincipal user, AuthenticationProperties properties)
```
HaoK: No I just copied the old methods over, I'll collapse this into the single public virtual SignIn

src/Microsoft.AspNetCore.Authentication.Cookies/Events/CookieSignedInContext.cs
```diff
@@ -9,32 +9,47 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
...
+        public override AuthenticationProperties Properties
+        {
+            get => Ticket?.Properties;
+
+            set
+            {
+                if (Ticket != null)
+                {
+                    Ticket = new AuthenticationTicket(Principal, value, Scheme.Name);
+                }
```
PinpointTownes: The base `Properties` property is not used here (we don't need it, because the ticket is always set from the constructor).

The only case where you'd reach the "else" clause is if you set `Ticket` to null AND set `Properties` once the ticket is null, which doesn't make much sense. Not sure what's best between no-op and throwing.

src/Microsoft.AspNetCore.Authentication/Events/EventResultState.cs
HaoK: No, thought I removed it already, perhaps it got merged back

src/Microsoft.AspNetCore.Authentication.Cookies/Events/CookieValidatePrincipalContext.cs
```diff
@@ -19,55 +19,40 @@ public class CookieValidatePrincipalContext : BaseCookieContext
...
             {
                 throw new ArgumentNullException(nameof(ticket));
             }
 
-            if (options == null)
-            {
-                throw new ArgumentNullException(nameof(options));
-            }
-
             Principal = ticket.Principal;
```
PinpointTownes: That's what this event used to do, but @HaoK reverted the changes...

HaoK: I had no issue with these having tickets, these don't need any of the control flow methods which came along with the AuthenticationContext

HaoK: But sure I'll add back the ticket

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
PinpointTownes: The old `AuthenticateResult.None()/Nothing` was split into `AuthenticateResult.None()/Nothing` (= no authentication ticket because authentication was skipped) and `RemoteAuthenticateResult.Skip()/Skipped` (no authentication ticket because one of the events opted for "skipping to next middleware).

We need to check both here.

src/Microsoft.AspNetCore.Authentication.Cookies/Events/CookieValidatePrincipalContext.cs
PinpointTownes: > control flow

If by control flow, you mean the ability to replace the ticket or skip/ignore authentication, that's exactly what `ReplacePrincipal()` and `RejectPrincipal()` do, anyway.

PinpointTownes: They just have different names.

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
HaoK: When I rename Nothing today, I'm going to take another look to see if we can simplify this into a single check again

samples/JwtBearerSample/Startup.cs
PinpointTownes: Blame @HaoK for using over-simplistic names. I had used `SkipAuthentication()`, but something like `IgnoreAuthentication()` or `BypassAuthentication()` would be fine too.

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationOptions.cs
HaoK: Man updating this is logigng stuff terrible, is really the logging pattern we are supposed to use everywhere?

src/Microsoft.AspNetCore.Authentication.Cookies/Events/CookieValidatePrincipalContext.cs
HaoK: So this event is named "ValidatePrincipal", not ValidateTicket, so today its only giving the event the opportunity to manipulate the principal in the ticket, while we can certainly switch this into something like a ValidateTicket, that's not something I'm trying to do in this PR.  So we don't really need the original ticket, we return a new ticket from context.Principal and context.Properties.

HaoK: Anyways, I don't have any problem changing this/revisiting any of this... just not in 2.0, we can update cookie events to be consistent in 2.1

src/Microsoft.AspNetCore.Authentication.Cookies/Events/CookieSignedInContext.cs
HaoK: I'm just going to revert this back to how things were before, BaseContext has Properties, these add the Principal, build a ticket from these two... much cleaner

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
HaoK: Well, looks like I'm reverting them now :)

HaoK:
Because we clearly didn't have enough events, I tweaked things a bit:

```
BaseContext (HttpContext/Request/Response/options/Scheme)
PropertiesContext : BaseContext adds AuthenticationProperties
PrincipalContext : PropertiesContext adds ClaimsPrincipal
AuthenticationContext : PrincipalContext adds the Flow/AuthenicateResult logic/methods
RedirectContext : PropertiesContext adds RedirectUri (I deleted all CookieRedirect/OAuth/Twitter events classes and use this generic event directly)
```

src/Microsoft.AspNetCore.Authentication.Cookies/Events/CookieValidatePrincipalContext.cs
HaoK: Tickets are gone from all of cookies again, they derive mostly from PrincipalContext, or PropertiesContext and only expose those directly.

HaoK:
Renamed None => Ignore, looks better, Remote still checks both which I think is fine, Skipped/Ignore have some slightly semantic differences as they come from different base context's

HaoK:
One more rename, going to rename AuthenticationContext => Result context since the usage is driven around whether context.Result != null

samples/JwtBearerSample/Startup.cs
```diff
@@ -59,7 +57,7 @@ public void ConfigureServices(IServiceCollection services)
                 {
                     OnAuthenticationFailed = c =>
                     {
-                        c.HandleResponse();
+                        c.Ignore();
```
Tratcher: Still not a fan. What is being ignored? The event? The stuff we do inside the event? The default logic after the event? The whole request?

Ignore makes sense for most of your internal uses, but it doesn't make sense in the events. I still think Handle was more accurate for these. I've handled the event, the error, the response, etc..

src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
```diff
@@ -0,0 +1,103 @@
...
+        /// Adds a <see cref="RemoteAuthenticationHandler{TOptions}"/> based <see cref="AuthenticationScheme"/> that supports remote authentication
+        /// which can be used by <see cref="IAuthenticationService"/>.
+        /// </summary>
+        /// <typeparam name="TOptions">The <see cref="RemoteAuthenticationOptions"/> type to configure the handler."/>.</typeparam>
+        /// <typeparam name="THandler">The <see cref="RemoteAuthenticationHandler{TOptions}"/> used to handle this scheme.</typeparam>
+        /// <param name="authenticationScheme">The name of this scheme.</param>
+        /// <param name="displayName">The display name of this scheme.</param>
+        /// <param name="configureOptions">Used to configure the scheme options.</param>
+        /// <returns>The builder.</returns>
+        public virtual AuthenticationBuilder AddRemoteScheme<TOptions, THandler>(string authenticationScheme, string displayName, Action<TOptions> configureOptions)
```
Tratcher: Still not sure why these aren't extension methods. They don't include any unique/private functionality.

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -69,7 +62,7 @@ public virtual Task<bool> ShouldHandleRequestAsync()
                 {
                     return true;
                 }
-                else if (authResult.Nothing)
+                else if (authResult.Skipped || authResult.Ignored)
```
Tratcher: Still smells redundant

PinpointTownes: The explanation is here: https://github.com/aspnet/Security/pull/1266#discussion_r123054352.

samples/JwtBearerSample/Startup.cs
PinpointTownes: `Handle` is not ideal because it's easily confused with `HandleResponse`, that is not available for the JWT bearer middleware (since it doesn't handle requests, it can't terminate them, unlike the OIDC middleware).

If you don't like the `Ignore` verb, something like `StopAuthentication()` would likely fit.

src/Microsoft.AspNetCore.Authentication/Events/HandleRequestContext.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <summary>
+        /// Discontinue all processing for this request and return to the client.
+        /// The caller is responsible for generating the full response.
+        /// </summary>
+        public void HandleResponse() => Result = HandleRequestResult.Handle();
+
+        /// <summary>
+        /// Discontinue processing the request in the current handler.
+        /// </summary>
+        public void SkipRequest() => Result = HandleRequestResult.SkipRequest();
```
PinpointTownes: `SkipRequest()` seems less clear than `SkipToNextMiddleware()` (and it introduces a breaking change in user code that could be easily avoided).

src/Microsoft.AspNetCore.Authentication/Events/RemoteAuthenticationContext.cs
```diff
@@ -0,0 +1,61 @@
...
+                if (Ticket != null)
+                {
+                    Ticket = new AuthenticationTicket(Principal, value, Scheme.Name);
+                }
+            }
+        }
+
+        public void Success(AuthenticationTicket ticket) => Result = HandleRequestResult.Success(ticket);
+
+        public void Ignore() => Result = HandleRequestResult.Ignore();
```
PinpointTownes: I'm pretty sure we had already concluded that single-verb methods in these contexts were not descriptive enough?

It must be crystal-clear that what's ignored/stopped here is just the authentication logic, not the request itself.

src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
HaoK: So in general if we didn't own the class, or we wanted to hang methods off an interface sure.  But given that we own the builder, what's the benefit of making them extension methods rather than normal methods here?

src/Microsoft.AspNetCore.Authentication/Events/HandleRequestContext.cs
HaoK: This is intentional as its not longer correct, returning skip does not skip to the next middleware anymore, it simply skips this handler, and lets any other request handlers get a chance (unlike Handled/HandleResponse).  Skipping to the next middleware would mean skipping to the next middleware after UseAuthentication.

SkipHandler works if you prefer?

PinpointTownes: > This is intentional as its not longer correct, returning skip does not skip to the next middleware anymore.

Ultimately, it does: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs#L44

There's fundamentally no difference between how `HandleRequestAsync` works in 1.x and how it works in 2.0: returning `true` from `HandleRequestAsync` immediately stops the request processing and returning `false` allows the authentication middleware to invoke the other handlers until one of them returns `true`: if none returns `true`, the next middleware in the pipeline is invoked.

samples/JwtBearerSample/Startup.cs
HaoK: So these need to be named as a pair to make sense.

`context.[Response]Handled()` // Response was sent, terminates the request all the way to UseAuth
vs 
'context.Stop/IgnoreHandler()' // Stop/Ignore just this handler

HaoK: ResponseHandled/StopHandling might be a reasonable pair

src/Microsoft.AspNetCore.Authentication/Events/HandleRequestContext.cs
HaoK: Yes the behavior hasn't changed much, but since these names were so tied to the old semantics, they are broken as a result.

PinpointTownes: `SkipToNextMiddleware()` still makes sense because in most cases, folks who use this API want to invoke the other middleware (e.g to render an error page or run some logic in a MVC controller), not an other `IAuthenticationRequestHandler`.

If you really want to introduce a breaking change in user code, `SkipToNextMiddleware()` could renamed to ``SkipToNextComponent()`.

Or you can rename `IAuthenticationRequestHandler` to something that has the middleware in the name and avoid the breaking change.

HaoK: Nah, the ToNextWhatever is why we have to change it now.  They want to skip this handler, putting what might be next is why we have to rename it now.

SkipHandler is pretty clear given that this is inside of a `AddXyz(o => )`

Tratcher: SkipHandler is good enough. SkipAuthHandler would be my second choice.

samples/JwtBearerSample/Startup.cs
HaoK: So now I think we can re-use the same SkipHandler name here even though it does something slightly different than the other ones

PinpointTownes: So we're back to the previous state where `HandleResponse()`/`SkipToNextMiddleware()` had multiple meanings? Oh please no, that was the whole point of my PR. You really hate me that much? :trollface:

HaoK: yeah so this 3rd state which is opt out of default, but continue is pretty hard to name and also pretty confusing to explain.

HaoK: I'm not even sure why we need to have this.  So looking at the internal usages of Ignore today, its the OIDC.SkipUnrecognizedRequests(which otherwise would Fail).  Why can't we just return Skip instead here?  

This seems totally not worth the complexity it brings, if we want to keep this as a quirk in OIDC events, fine, but I don't think this needs to spread.

PinpointTownes: > I'm not even sure why we need to have this.

We're really repeating the same things again and again :sweat_smile: 
This sample demonstrates one important scenario: skipping the default logic and taking control of the challenge logic (e.g to return a JSON payload, like what FB does for failed requests).

HaoK: Why wouldn't the sample still work via the c.Skip() method that lives in the event?

Tratcher: Skip what?

HaoK: I don't have any issue with Skip being a top level supported flag for all events, but it sure isn't critical to have now, we can keep it where we had it before, but we certainly are not going to block over this.

Given that we normalize Skip/Ignore in RemoteAuthenticationResult anyways, I think we should just combine these.  Do any tests even fail if I combine the two?

HaoK: This event has a Skip property as well :)

```
        /// <summary>
        /// Skips any default logic for this challenge.
        /// </summary>
        public void Skip() => Skipped = true;

```

HaoK: I'm going to whack this Ignore to see if anything breaks

PinpointTownes: > Why wouldn't the sample still work via the c.Skip() method that lives in the event?

@HaoK there's no such thing. If you mean `SkipToNextMiddleware()`/`SkipHandler()`, it's no longer used in the JWT bearer middleware because it doesn't implement `IAuthenticationRequestHandler` and so can't have any influence on the request processing. I replaced both `SkipToNextMiddleware()`/`SkipHandler()` by `SkipAuthentication()` in my PR.

PinpointTownes: > This event has a Skip property as well :)

That's what you get for ruining my PR. Duplicated stuff :smile_cat:

HaoK: Well we can just rename Skip to be SkipHandler to match

PinpointTownes: No, that's not what the method would do.

HaoK: This one could be SkipChallenge

HaoK: assuming the doc comments are correct :)

PinpointTownes: I thought you didn't want to use the `Skip` verb so folks can't mix up this API with `SkipHandler()`, that has a totally different meaning?

HaoK: True, so 'StopChallenge()`

Tratcher: @HaoK let's meet up and sort this out.

HaoK: Names we decided on:

Ticket/Result:
Success
NoResult
Fail

RequestHandler:
SkipHandler
HandleResponse

Challenge/JwtBearer Skips become HandleResponse as well preserving what they were called before.

PinpointTownes: You have no idea how much I regret spending time on this PR when I see those names :face_with_head_bandage:

samples/JwtBearerSample/Startup.cs
```diff
@@ -59,7 +57,9 @@ public void ConfigureServices(IServiceCollection services)
                 {
                     OnAuthenticationFailed = c =>
                     {
-                        c.HandleResponse();
+                        // Ignore/Stop/Abort/
```
Tratcher: cleanup

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -199,12 +204,7 @@ public override Task<bool> HandleRequestAsync()
...
             await Events.RedirectToIdentityProviderForSignOut(redirectContext);
-            if (redirectContext.HandledResponse)
-            {
-                Logger.RedirectToIdentityProviderForSignOutHandledResponse();
-                return;
-            }
-            else if (redirectContext.Skipped)
+            if (redirectContext.Handled)
             {
                 Logger.RedirectToIdentityProviderForSignOutSkipped();
```
Tratcher: Removed the wrong log statement?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -343,12 +344,7 @@ protected virtual Task<bool> HandleSignOutCallbackAsync()
...
             await Events.RedirectToIdentityProvider(redirectContext);
-            if (redirectContext.HandledResponse)
-            {
-                Logger.RedirectToIdentityProviderHandledResponse();
-                return;
-            }
-            else if (redirectContext.Skipped)
+            if (redirectContext.Handled)
             {
                 Logger.RedirectToIdentityProviderSkipped();
```
Tratcher: Wrong log

src/Microsoft.AspNetCore.Authentication/Events/RemoteAuthenticationContext.cs
```diff
@@ -0,0 +1,59 @@
...
+        public AuthenticationProperties Properties
+        {
+            get => Ticket?.Properties;
+            set
+            {
+                if (Ticket != null)
+                {
+                    Ticket = new AuthenticationTicket(Principal, value, Scheme.Name);
+                }
+            }
```
Tratcher: This set needs an else.

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -69,7 +62,7 @@ public virtual Task<bool> ShouldHandleRequestAsync()
                 {
                     return true;
                 }
-                else if (authResult.Nothing)
+                else if (authResult.Skipped || authResult.None)
```
Tratcher: This maps to SkipHandler and NoResult?

HaoK: Yup, the two in between states

src/Microsoft.AspNetCore.Authentication/Events/RemoteAuthenticationContext.cs
HaoK: Else what though, throw?  I think we should just nuke the Properties set, and change all of our code to stuff a properties into a ticket themselves before calling this.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
HaoK: Yup, good catch!

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
HaoK: Heh so I removed Handled, and kept skip, then renamed Skip back to Handled...  nuked the log extensions for Skip

src/Microsoft.AspNetCore.Authentication/Events/RemoteAuthenticationContext.cs
Tratcher: Yes, remove it or throw InvalidOperationException

HaoK:
ff9f145a8e89c9756ea12ff10c6d47f2f7eb345f

PinpointTownes:
@HaoK these API changes are so massive that you should consider writing an announcement/ticket listing the most important ones.

HaoK:
Yeah its on my todo list, I'll try to get to this next week

PR #1267, opened 2017-06-16
Split SIgnIn/SignOut

HaoK:
@Tratcher 

Reacts to https://github.com/aspnet/HttpAbstractions/pull/873

I'll merge this into the events PR, but easier to review these seperately first

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -14,7 +14,10 @@ using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNetCore.Authentication.Cookies
 {
-    public class CookieAuthenticationHandler : AuthenticationHandler<CookieAuthenticationOptions>
+    public class CookieAuthenticationHandler : 
+        AuthenticationHandler<CookieAuthenticationOptions>, 
+        IAuthenticationSignInHandler, 
+        IAuthenticationSignOutHandler
     {
         private const string HeaderValueNoCache = "no-cache";
         private const string HeaderValueMinusOne = "-1";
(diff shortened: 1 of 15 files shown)


src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -249,7 +252,19 @@ private CookieOptions BuildCookieOptions()
...
+        public async Task SignInAsync(ClaimsPrincipal user, AuthenticationProperties properties)
+        {
+            if (user == null)
+            {
+                throw new ArgumentNullException(nameof(user));
+            }
+
+            properties = properties ?? new AuthenticationProperties();
+            await HandleSignInAsync(user, properties);
+            Logger.SignedIn(Scheme.Name);
```
PinpointTownes: So we must now do that in EVERY handler? What a bummer :-1:

HaoK: You actually have a handler that implements sign in?

PinpointTownes: ASOS does (exactly like `OAuthAuthorizationServerMiddleware` used to).

We use that for both the authorization and token endpoints: http://kevinchalet.com/2017/01/30/implementing-simple-token-authentication-in-aspnet-core-with-openiddict/

HaoK: So yeah if you want to log that you signed in, you need to log it yourself (and granted the logging extensions code is ridiculously gross)

HaoK: But I imagine you can just add a Logger.Log at the top of your method is the only change for you, and implement the interface method directly instead of HandleSignIn

PinpointTownes: At least, do we get a sensible advantage if we split `IAuthenticationHandler`? :sweat_smile:

HaoK: SignIn/SignOut default will be more likely to be correct in applications with only a single cookie configured  So OIDC + Cookie won't need to specify DefaultSignInScheme anymore, well once we actually change RemoteAuthenticationHandler to stop requiring SignInScheme now that it will work.

PinpointTownes: Heh, pretty sure I said "advantage"... making the default scheme selection more magical than it is already is not what I'd call an advantage.

I can already hear my users say "oh, my cookies+Facebook app worked fine before I added OpenIddict and now, I have an exception"... :cry:

HaoK:
Rolled into https://github.com/aspnet/Security/pull/1266

Issue #1269, opened 2017-06-18
Remove top level methods on IServiceCollection that configure authentication


davidfowl:
`services.AddScheme` is too general of a name to put on the top level container:

![image](https://user-images.githubusercontent.com/95136/27261329-d09ab052-53f5-11e7-8bc6-a525d4841819.png)

Is it even meaningful to add a scheme without adding the authentication service? Why isn't it an extension method off on an `IAuthenticationBuilder` or something similar.

```C#
services.AddAuthentication()
              .AddScheme(...)
```

HaoK:
Not a bad idea to switch to a builder pattern.  Do we want to move everything under, i.e.?

`services.AddAuthentication[Core]().AddScheme/Xyz()`

HaoK:
So something like:
`Authentication.Core`  has `void services.AddAuthenticationCore()` // unchanged
`Authentication` has `AuthenticationBuilder services.AddAuthentication()`
Authentication.Xyz` => `AuthenticationBuilder.AddXyz()`

davidfowl:
I was also thinking we should hang the services like cookie, jwt, etc off the builder.

services.AddAuthentication()
              .AddCookies(...);

HaoK:
Yup that's what I meant with the AddXyz.

PR #1270, opened 2017-06-19
Switch to builder pattern for AddXyzAuth

HaoK:
New pattern:

```C#
services.AddAuthentication()
    .AddCookie()
    .AddGoogle(o => { })
    .AddOpenIdConnect(o => { })
```

Will remove old AddXyz methods in a followup PR after everything has been updated to use the new builder methods.

Fixes https://github.com/aspnet/Security/issues/1269

@davidfowl @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieExtensions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieExtensions.cs
@@ -2,16 +2,30 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using Microsoft.AspNetCore.Authentication.Cookies;
-using Microsoft.AspNetCore.DataProtection;
-using Microsoft.Extensions.Options;
-using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Authentication.Cookies;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Extensions.DependencyInjection
 {
     public static class CookieExtensions
     {
+        public static AuthenticationBuilder AddCookie(this AuthenticationBuilder builder) => builder.AddCookie(CookieAuthenticationDefaults.AuthenticationScheme);
+
+        public static AuthenticationBuilder AddCookie(this AuthenticationBuilder builder, string authenticationScheme) => builder.AddCookie(authenticationScheme, configureOptions: null);
+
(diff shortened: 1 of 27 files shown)


HaoK:
Will roll this into https://github.com/aspnet/Security/pull/1270

HaoK:
Note: this pattern is only for Security based auth schemes/handlers (which should be the majority of cases).  But the AddAuthenticationCore/host implementations of IAuthenticationHandler will not be affected by any of this as things stand.

src/Microsoft.AspNetCore.Authentication.Cookies/CookieExtensions.cs
```diff
@@ -2,16 +2,30 @@
...
-using Microsoft.Extensions.Options;
 using Microsoft.Extensions.DependencyInjection.Extensions;
-using Microsoft.AspNetCore.Authentication;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Extensions.DependencyInjection
 {
     public static class CookieExtensions
     {
+        public static AuthenticationBuilder AddCookie(this AuthenticationBuilder builder) => builder.AddCookie(CookieAuthenticationDefaults.AuthenticationScheme);
```
Tratcher: wrap long lines?

src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        /// <summary>
+        /// Adds a <see cref="AuthenticationScheme"/> which can be used by <see cref="IAuthenticationService"/>.
+        /// </summary>
+        /// <typeparam name="TOptions">The <see cref="AuthenticationSchemeOptions"/> type to configure the handler."/>.</typeparam>
+        /// <typeparam name="THandler">The <see cref="AuthenticationHandler{TOptions}"/> used to handle this scheme.</typeparam>
+        /// <param name="authenticationScheme">The name of this scheme.</param>
+        /// <param name="configureOptions">Used to configure the scheme options.</param>
+        /// <returns>The builder.</returns>
+        public virtual AuthenticationBuilder AddScheme<TOptions, THandler>(string authenticationScheme, Action<TOptions> configureOptions)
```
Tratcher: virtuals rather than extensions?

HaoK: Well my only concern was making sure we can modify the builder's public API in 2.1 given how late this is coming so that's why its not an IAuthenticationBuilder interface.

So I just defaulted these to public virtual

HaoK:
Yeah I'll update the announcement as part of the wave of updating all of the old usages which should be fun (not)

HaoK:
Merging this into the events PR as that's going to be the largest boulder I'm rolling around, any further changes will be applied there

HaoK:
Rolled into https://github.com/aspnet/Security/pull/1266

PR #1275, opened 2017-06-23
Switch to Principal/Properties instead of Ticket in events

HaoK:
All the state copying after every event is pretty gross

@Tratcher thoughts?

I added sugar for Success so events can avoid newing up a ticket.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
@@ -135,10 +135,9 @@ namespace Microsoft.AspNetCore.Authentication.JwtBearer
 
                         Logger.TokenValidationSucceeded();
 
-                        var ticket = new AuthenticationTicket(principal, new AuthenticationProperties(), Scheme.Name);
                         var tokenValidatedContext = new TokenValidatedContext(Context, Scheme, Options)
                         {
-                            Ticket = ticket,
+                            Principal = principal,
                             SecurityToken = validatedToken
                         };
 
@@ -147,17 +146,16 @@ namespace Microsoft.AspNetCore.Authentication.JwtBearer
                         {
                             return tokenValidatedContext.Result;
                         }
-                        ticket = tokenValidatedContext.Ticket;
 
                         if (Options.SaveToken)
                         {
-                            ticket.Properties.StoreTokens(new[]
+                            tokenValidatedContext.Properties.StoreTokens(new[]
                             {
                                 new AuthenticationToken { Name = "access_token", Value = token }
                             });
                         }
 
-                        return AuthenticateResult.Success(ticket);
+                        return AuthenticateResult.Success(new AuthenticationTicket(tokenValidatedContext.Principal, tokenValidatedContext.Properties, Scheme.Name));
                     }
                 }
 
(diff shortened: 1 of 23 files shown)


src/Microsoft.AspNetCore.Authentication/Events/RemoteAuthenticationContext.cs
```diff
@@ -18,37 +18,29 @@ public abstract class RemoteAuthenticationContext<TOptions> : HandleRequestConte
...
-            get => Ticket?.Properties;
-            set
-            {
-                if (Ticket != null)
-                {
-                    Ticket = new AuthenticationTicket(Principal, value, Scheme.Name);
-                }
-            }
-        }
+        public void Success() => Success(new AuthenticationTicket(Principal, Properties, Scheme.Name));
```
Tratcher: This will throw if Principal is null, correct?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -474,6 +474,7 @@ protected virtual Task<bool> HandleSignOutCallbackAsync()
                     return messageReceivedContext.Result;
                 }
                 authorizationResponse = messageReceivedContext.ProtocolMessage;
+                // REVIEW: should we let them set Principal here?
```
Tratcher: No.

src/Microsoft.AspNetCore.Authentication/Events/RemoteAuthenticationContext.cs
HaoK: Yeah Ticket throws with null principals

src/Microsoft.AspNetCore.Authentication/Events/RemoteAuthenticationContext.cs
```diff
@@ -18,37 +18,29 @@ public abstract class RemoteAuthenticationContext<TOptions> : HandleRequestConte
...
 
         public void Success(AuthenticationTicket ticket) => Result = HandleRequestResult.Success(ticket);
```
HaoK: Good point, yeah lets make them set the properties

HaoK:
This does make the OAuth/Result based events differ quite a bit from the HandleRequest/HandleRemote events, but I guess that's ok.

HaoK:
Actually maybe we should just be consistent and switch to ClaimsPrincipal/Properties for the ResultContext derived context's as well.

HaoK:
I updated ResultContext to match, it now also no longer has a Ticket, only a Result + Principal + Properties

src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -147,17 +146,16 @@ protected new JwtBearerEvents Events
...
                         {
-                            ticket.Properties.StoreTokens(new[]
+                            tokenValidatedContext.Properties.StoreTokens(new[]
                             {
                                 new AuthenticationToken { Name = "access_token", Value = token }
                             });
                         }
 
-                        return AuthenticateResult.Success(ticket);
+                        return AuthenticateResult.Success(new AuthenticationTicket(tokenValidatedContext.Principal, tokenValidatedContext.Properties, Scheme.Name));
```
Tratcher: tokenValidatedContext.Sucess(), return tokenValidatedContext.Result?

HaoK: Yeah I had that initially but it felt weird, but I can change it back

HaoK:
Rolling into https://github.com/aspnet/Security/pull/1266

PR #1284, opened 2017-06-29
Replace configure method on TwitterOptions and OpenIdConnectOptions with CookieBuilder

natemcmaster:
Addresses https://github.com/aspnet/HttpAbstractions/issues/853.

Uses the new CookieBuilder API added here: https://github.com/aspnet/HttpAbstractions/pull/882

Part 1- look for a separate PR to remove duplication from CookieAuthenticationOptions.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
@@ -275,8 +275,7 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
         /// <returns>A task executing the callback procedure</returns>
         protected virtual Task<bool> HandleSignOutCallbackAsync()
         {
-            StringValues protectedState;
-            if (Request.Query.TryGetValue(OpenIdConnectParameterNames.State, out protectedState))
+            if (Request.Query.TryGetValue(OpenIdConnectParameterNames.State, out StringValues protectedState))
             {
                 var properties = Options.StateDataFormat.Unprotect(protectedState);
                 if (!string.IsNullOrEmpty(properties?.RedirectUri))
@@ -505,8 +504,7 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
                     return HandleRequestResult.Fail(Resources.MessageStateIsInvalid);
                 }
 
-                string userstate = null;
-                properties.Items.TryGetValue(OpenIdConnectDefaults.UserstatePropertiesKey, out userstate);
+                properties.Items.TryGetValue(OpenIdConnectDefaults.UserstatePropertiesKey, out string userstate);
                 authorizationResponse.State = userstate;
 
                 if (!ValidateCorrelationId(properties))
@@ -859,8 +857,7 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
 
             if (!string.IsNullOrEmpty(message.ExpiresIn))
             {
-                int value;
-                if (int.TryParse(message.ExpiresIn, NumberStyles.Integer, CultureInfo.InvariantCulture, out value))
+                if (int.TryParse(message.ExpiresIn, NumberStyles.Integer, CultureInfo.InvariantCulture, out int value))
                 {
                     var expiresAt = Clock.UtcNow + TimeSpan.FromSeconds(value);
                     // https://www.w3.org/TR/xmlschema-2/#dateTime
@@ -885,21 +882,12 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
                 throw new ArgumentNullException(nameof(nonce));
             }
 
-            var options = new CookieOptions
-            {
-                HttpOnly = true,
-                SameSite = Http.SameSiteMode.None,
-                Path = OriginalPathBase + Options.CallbackPath,
-                Secure = Request.IsHttps,
-                Expires = Clock.UtcNow.Add(Options.ProtocolValidator.NonceLifetime)
-            };
-
-            Options.ConfigureNonceCookie?.Invoke(Context, options);
+            var cookieOptions = Options.NonceCookie.Build(Context, Clock.UtcNow);
 
             Response.Cookies.Append(
-                OpenIdConnectDefaults.CookieNoncePrefix + Options.StringDataFormat.Protect(nonce),
+                Options.NonceCookie.Name + Options.StringDataFormat.Protect(nonce),
                 NonceProperty,
-                options);
+                cookieOptions);
         }
 
         /// <summary>
@@ -918,23 +906,14 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
 
             foreach (var nonceKey in Request.Cookies.Keys)
             {
-                if (nonceKey.StartsWith(OpenIdConnectDefaults.CookieNoncePrefix))
+                if (nonceKey.StartsWith(Options.NonceCookie.Name))
                 {
                     try
                     {
-                        var nonceDecodedValue = Options.StringDataFormat.Unprotect(nonceKey.Substring(OpenIdConnectDefaults.CookieNoncePrefix.Length, nonceKey.Length - OpenIdConnectDefaults.CookieNoncePrefix.Length));
+                        var nonceDecodedValue = Options.StringDataFormat.Unprotect(nonceKey.Substring(Options.NonceCookie.Name.Length, nonceKey.Length - Options.NonceCookie.Name.Length));
                         if (nonceDecodedValue == nonce)
                         {
-                            var cookieOptions = new CookieOptions
-                            {
-                                HttpOnly = true,
-                                Path = OriginalPathBase + Options.CallbackPath,
-                                SameSite = Http.SameSiteMode.None,
-                                Secure = Request.IsHttps
-                            };
-
-                            Options.ConfigureNonceCookie?.Invoke(Context, cookieOptions);
-
+                            var cookieOptions = Options.NonceCookie.Build(Context, Clock.UtcNow);
                             Response.Cookies.Delete(nonceKey, cookieOptions);
                             return nonce;
                         }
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -919,22 +916,18 @@ private string ReadNonceCookie(string nonce)
...
+                            if (Options.NonceCookie.Path == null)
                             {
-                                HttpOnly = true,
-                                Path = OriginalPathBase + Options.CallbackPath,
-                                SameSite = Http.SameSiteMode.None,
-                                Secure = Request.IsHttps
-                            };
-
-                            Options.ConfigureNonceCookie?.Invoke(Context, cookieOptions);
+                                cookieOptions.Path = OriginalPathBase + Options.CallbackPath;
```
HaoK: Maybe just do this instead? `cookieOptions.Path = cookieOptions.Path ?? OriginalPathBase + Options.CallbackPath;`

There's also a subtle bug here, since even if Path == null, the Builder could set the cookie.Path, which would get blasted away

natemcmaster: Yes, that's why this deliberately checks  `if (Options.NonceCookie.Path == null)` first.

HaoK: Yeah so what if they have CustomCookieBuilder which sets cookieOptions.Path = "/mypath" in Build(), this code will still stomp on the path.  The check should be checking the real cookieOptions.Path, not the builder property

natemcmaster: Hmm. This is problematic. The [default ctor for CookieOptions](https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Features/CookieOptions.cs#L18) always sets Path to "/". If someone implements a custom cookie builder, it will be hard for us to detect if they indend to overwrite the default value or not.

```c#
var builder = new CookieBuilder();
  // builder.Name = null
var options = builder.Build(context)
  // options.Name = "/"

 // builder.Name == null means we can assume the user does not indent to change 
 // the default path, therefore we set it to OriginalPathBase + Options.CallbackPath
```
```c#
var builder = new MyCustomCookieBuilder();
  // builder.Name = null
var options = builder.Build(context)
  // options.Name = "/mypath"

  // builder.Name == null means ????
```

HaoK: We are talking about Path not name right?

HaoK: Or we can just check for options.Path == "/" since we can't really assume anything based on the property alone.

```
if (cookieOptions.Path == "/") {
    cookieOptions.Path = OriginalPathBase + Options.CallbackPath;
}
```

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -886,21 +886,18 @@ private void WriteNonceCookie(string nonce)
...
-            var options = new CookieOptions
-            {
-                HttpOnly = true,
-                SameSite = Http.SameSiteMode.None,
-                Path = OriginalPathBase + Options.CallbackPath,
-                Secure = Request.IsHttps,
-                Expires = Clock.UtcNow.Add(Options.ProtocolValidator.NonceLifetime)
-            };
+            var cookieOptions = Options.NonceCookie.Build(Context);
+            cookieOptions.Expires = Clock.UtcNow.Add(Options.ProtocolValidator.NonceLifetime);
```
Tratcher: This isn't honoring the Expiration from CookieBuilder. Only set it if not null?

Any way to resolve Options.ProtocolValidator.NonceLifetime in the builder? Then you could incorporate it in a way that could be overridden.

src/Microsoft.AspNetCore.Authentication.Twitter/TwitterHandler.cs
```diff
@@ -50,7 +49,8 @@ public TwitterHandler(IOptionsSnapshot<TwitterOptions> options, ILoggerFactory l
         {
             AuthenticationProperties properties = null;
             var query = Request.Query;
-            var protectedRequestToken = Request.Cookies[StateCookie];
+            var cookieName = Options.StateCookie.Name ?? StateCookieName;
```
Tratcher: Yeah, this gets simpler if we prevent Name from being null.

Tratcher:
Add support for the correlation cookie, or file a followup bug since this isn't currently configurable.
https://github.com/aspnet/Security/blob/ff9f145a8e89c9756ea12ff10c6d47f2f7eb345f/src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs#L190

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
HaoK: Yup, just add a Property in a OpenIdConnectCookieBuilder for it

HaoK:
Should just make the change in this PR for correlation cookie

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
natemcmaster: I figured out a way to do that. I override the build method to detect the value of Path before. Should allow user to override.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
natemcmaster: Updated

src/Microsoft.AspNetCore.Authentication.Twitter/TwitterHandler.cs
natemcmaster: Updated.

natemcmaster:
Rebased and added the correlation cookie change to this PR.

src/Microsoft.AspNetCore.Authentication/Internal/RequestPathCookieBuilder.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// </summary>
+        protected virtual string AdditionalPath { get; }
+
+        public override CookieOptions Build(HttpContext context, DateTimeOffset expiresFrom)
+        {
+            // check if the user has overridden the default value of path. If so, use that instead of our default value.
+            var path = Path;
+            if (path == null)
+            {
+                var originalPathBase = context.Features.Get<IAuthenticationFeature>()?.OriginalPathBase ?? context.Request.PathBase;
```
HaoK: Should we consider having the handlers set a OriginalPathBase property on the CookieBuilder instead of having the builder pull it from the context?

src/Microsoft.AspNetCore.Authentication/Internal/RequestPathBaseCookieBuilder.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// </summary>
+        protected virtual string AdditionalPath { get; }
+
+        public override CookieOptions Build(HttpContext context, DateTimeOffset expiresFrom)
+        {
+            // check if the user has overridden the default value of path. If so, use that instead of our default value.
+            var path = Path;
+            if (path == null)
+            {
+                var originalPathBase = context.Features.Get<IAuthenticationFeature>()?.OriginalPathBase ?? context.Request.PathBase;
```
HaoK: Commenting again in the right version: Should we make this a property on RequestPathBaseCookieBuilder that the handlers set instead of making the builder pull it out?

natemcmaster: Possibly racy. The builder is shared across multiple contexts.

natemcmaster:
:bell:

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -212,16 +201,15 @@ protected virtual bool ValidateCorrelationId(AuthenticationProperties properties
                 throw new ArgumentNullException(nameof(properties));
             }
 
-            string correlationId;
-            if (!properties.Items.TryGetValue(CorrelationProperty, out correlationId))
+            if (!properties.Items.TryGetValue(CorrelationProperty, out string correlationId))
             {
-                Logger.CorrelationPropertyNotFound(CorrelationPrefix);
+                Logger.CorrelationPropertyNotFound(RemoteAuthenticationOptions.CorrelationPrefix);
```
HaoK: Shouldn't this be something like Options.CorrelationCookie.Name now?

natemcmaster: Yup. Updated.

src/Microsoft.AspNetCore.Authentication.Twitter/TwitterHandler.cs
```diff
@@ -80,16 +78,9 @@ public TwitterHandler(IOptionsSnapshot<TwitterOptions> options, ILoggerFactory l
...
-                HttpOnly = true,
-                SameSite = SameSiteMode.Lax,
-                Secure = Request.IsHttps
-            };
-
-            Options.ConfigureStateCookie?.Invoke(Context, cookieOptions);
+            var cookieOptions = Options.StateCookie.Build(Context, Clock.UtcNow);
 
-            Response.Cookies.Delete(StateCookie, cookieOptions);
+            Response.Cookies.Delete(Options.StateCookie.Name, cookieOptions);
```
Tratcher: I'm trying to think if a user would Build their options any differently if they knew they were building a delete cookie.

PR #1285, opened 2017-06-30
Add CookieBuilder to CookieAuthenticationOptions and obsolete the duplicated properties

natemcmaster:
Addresses https://github.com/aspnet/HttpAbstractions/issues/853.

Uses the new CookieBuilder API added here: https://github.com/aspnet/HttpAbstractions/pull/882


src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -176,23 +176,15 @@ private void RequestRefresh(AuthenticationTicket ticket)
...
-                SameSite = Options.CookieSameSite,
-                HttpOnly = Options.CookieHttpOnly,
-                Path = Options.CookiePath ?? (OriginalPathBase.HasValue ? OriginalPathBase.ToString() : "/"),
-            };
+            var cookieOptions = Options.Cookie.Build(Context);
+            // ignore the 'Expires' value as this will be computed elsewhere
+            cookieOptions.Expires = null;
 
-            if (Options.CookieSecure == CookieSecurePolicy.SameAsRequest)
+            if (Options.Cookie.Path == null && OriginalPathBase.HasValue)
```
natemcmaster: Same issue here as being discussed https://github.com/aspnet/Security/pull/1284/files/6b218ab9f3a1439a5c038f49332ef16cbaedc712#r124940460

Tratcher:
@blowdart for a usability check

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
blowdart: Is this doing the same thing? I'd still expect, if a cookie does have a path but I say SameAsRequest it'd match the Secure flag to the request protocol.

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -176,23 +176,15 @@ private void RequestRefresh(AuthenticationTicket ticket)
...
         {
-            var cookieOptions = new CookieOptions
-            {
-                Domain = Options.CookieDomain,
-                SameSite = Options.CookieSameSite,
-                HttpOnly = Options.CookieHttpOnly,
-                Path = Options.CookiePath ?? (OriginalPathBase.HasValue ? OriginalPathBase.ToString() : "/"),
-            };
+            var cookieOptions = Options.Cookie.Build(Context);
+            // ignore the 'Expires' value as this will be computed elsewhere
```
HaoK: Didn't we add Expiration support to Build for Security, or is it being used elsewhere?

natemcmaster: The behavior here is funky, and FWIW one of the reasons I was suggesting we leave `Expiration` off CookeBuilder. This handler takes into consideration `CookieSignInContext.Properties.ExpiresUtc` and `CookieSignInContext.Properties.IsPersistent` to calculate `CookieOptions.Expires`. So, the value of `.Expires` set by the builder needs to be unset to avoid changing behavior.

https://github.com/aspnet/Security/blob/namc/cookie-options/test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs#L141

It would be good to find ways to clean this up without breaking existing behavior.

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
natemcmaster: The diff github is showing is a little misleading. The behavior your asking about moved into the CookieBuilder:

See https://github.com/aspnet/HttpAbstractions/blob/199b0fa212aa03a011f3c18575f0ab5102527c1a/src/Microsoft.AspNetCore.Http.Abstractions/CookieBuilder.cs#L97

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
HaoK: This is fine, I just wanted to make sure we are using the expires overload somewhere, and it looks like we are with the nonce/correlation cookies so its cool.  Otherwise I'd have suggested removing the unused overload.

natemcmaster:
Updated. Rebased on #1284 so I could re-use RequestPathBaseCookieBuilder

natemcmaster:
:bell: updated to remove CookieAuthenticationOptions.CookieSameSite completely, which was never shipped as an RTM api

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -283,14 +270,14 @@ private CookieOptions BuildCookieOptions()
 
             if (!signInContext.Properties.ExpiresUtc.HasValue)
             {
-                signInContext.Properties.ExpiresUtc = issuedUtc.Add(Options.ExpireTimeSpan);
+                signInContext.Properties.ExpiresUtc = issuedUtc.Add(Options.Cookie.Expiration ?? default(TimeSpan));
```
Tratcher: This fallback to `default(TimeSpan)` doesn't make sense, you end up issuing an already expired cookie.

Issue #1293, opened 2017-07-05
Clarify behavior when CookieAuthenticationOptions.Cookie.Expiration is set to null


natemcmaster:
CookieAuthenticationOptions.Cookie.Expiration (added https://github.com/aspnet/Security/pull/1285) is a nullable `TimeSpan`, but `CookieBuilder.Expiration` is a nullable `TimeSpan`. We should either ensure our handler behaves correctly when `Expiration` is null, or prevent it from being null. 

cc @Tratcher

Tratcher:
Where? In doc comments? Samples? Or make it throw?

natemcmaster:
For others new to this thread, let me add a little context. The reason for the `ExpireTimeSpan` API is that it represents the expiration of the _ticket_, not the _cookie_ that contains the ticket. 

Example: we haven't figure out how to handle something like this.
```c#
Cookie.Expiration = TimeSpan.FromDays(9999);
ExpiresTimeSpan = TimeSpan.FromMinutes(30);
```
This would create an auth ticket valid for the next 30 minutes, but the cookie will be stored by the browser 9999 days. That means in a month, the browser will still send the cookie, but auth won't accept the ticket contained in the cookie. Until 2.0.0, this wasn't a problem because users couldn't configure cookie expiration manually. We always inferred cookie expiration from ticket expiration.

natemcmaster:
@Tratcher I think it would be useful to assert that Cookie.Expiration >= ExpireTimeSpan. If not, throw. There is nothing wrong with Cookie.Expiration >= ExpireTimeSpan, it's just means you have a cookie that will never work and doesn't get cleaned up right away. But on the other hand, if someone configures Cookie.Expiration but not ExpireTimeSpan, they will get an error.

Tratcher:
Cookie.Expiration is currently ignored. It never makes it to the browser. https://github.com/aspnet/Security/blob/a53bf093a7d86b35e019c80515c92d7626982325/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs#L181

Auth cookies have a separate flag for persisting cookies (via setting their expiration) and it's intentionally only available per request because it requires user consent.
https://github.com/aspnet/Security/blob/a53bf093a7d86b35e019c80515c92d7626982325/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs#L222

Setting Cookie.Expiration is an error in the current design, we just don't validate it.

[...]

PR #1296, opened 2017-07-05
Revert obsoleting CookieAuthenticationOptions.ExpireTimeSpan

natemcmaster:
`CookieAuthenticationOptions.Cookie.Expiration` is conceptually different from `CookieAuthenticationOptions.ExpireTimeSpan`.

Un-oboleting the property and updating xmldocs.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -270,14 +270,14 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
 
             if (!signInContext.Properties.ExpiresUtc.HasValue)
             {
-                signInContext.Properties.ExpiresUtc = issuedUtc.Add(Options.Cookie.Expiration ?? default(TimeSpan));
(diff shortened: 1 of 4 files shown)


HaoK:
Changes look fine, but are we missing some test coverage since shouldn't a test have to be updated? :)  They probably have always been missing, but maybe we should take this opportunity to add 1-2 basic tests that actually enforce the behavior we expect?

natemcmaster:
A bunch tests in CookieTests.cs were updated. Which behavior should we test for that isn't already covered?

HaoK:
I mean the Cookie.Expires property I guess, or is that the one we basically ignore?

natemcmaster:
Yeah, that's the one we currently ignore. I left this unchanged:  https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs#L180-L181

HaoK:
Okay, worth adding a single test to capture that we ignore CookieBuilder.Expires so we don't regress this accidentally in the future?

natemcmaster:
Added a test.

natemcmaster:
Btw, let's use https://github.com/aspnet/Security/issues/1293 to clarify what CookieBuilder.Expiration should mean. We can address this after 2.0.0. For now, we'll just ignore its value.

Issue #1276, opened 2017-06-26
ClaimActions.DeleteClaim in OpenIdConnectOptions


brentschmaltz:
Will an entry in DeleteClaim result in removal of the claim from the ClaimsIdentity?
If so, we may want to consider keeping, amr and acr. 
Some of the claims, such as 'nonce' have little value one the OIDC message has been validated. 
amr and acr are not used to validate the message and may be of interest to the application layer.

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs#L54

Tratcher:
Yes, that's what it's for.  amr and acr are both optional. Are they actually implemented by AAD, ADFS, Google?

brentschmaltz:
@Tratcher these claims {amr, acr} can be of interest to the application layer. So I wouldn't remove them by default.

The others seem fine.

PR #1297, opened 2017-07-05
Keep OIDC amr and acr claims

Tratcher:
#1276 @brentschmaltz recommended keeping these claims as they may be useful to the consumer.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
@@ -48,8 +48,6 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
             ClaimActions.DeleteClaim("nonce");
             ClaimActions.DeleteClaim("aud");
             ClaimActions.DeleteClaim("azp");
-            ClaimActions.DeleteClaim("acr");
-            ClaimActions.DeleteClaim("amr");
             ClaimActions.DeleteClaim("iss");
             ClaimActions.DeleteClaim("iat");
             ClaimActions.DeleteClaim("nbf");


PinpointTownes:
If you start making an exception for `amr` and `acr`, folks will want the same treat for `iss`/`aud` and the other ones. These claims can be lengthy (specially `acr`), so not sure it's really a good idea to include them back.

Curious: how bad is it to add them back in app code using the claims mapping API?

brentschmaltz:
@PinpointTownes What I am trying to call out is that amr and acr are of interest to the application layer. iss, aud are validated by the runtime.

PinpointTownes:
> @PinpointTownes What I am trying to call out is that amr and acr are of interest to the application layer

I'm not denying that. Potentially every "protocol claim" can be of interest. The real question is "is your scenario frequent enough to justify treating `amr` and `acr` differently"? I've developed many OIDC apps and IMHO, the cases where you want to access them from controllers/modules/whathever are quite rare (it's more frequent to validate the `acr` claim directly using the events model).

Folks who need these claims could re-active them using @Tratcher's new claims API, anyway.

brentschmaltz:
@PinpointTownes it is not about if there should be / could be something of interest. What I am saying is that amr / acr are not validate by the runtime and are of interest.
Contrast that with the c_hash or nonce which are validated and only of interest when validating the protocol message. Those can be dropped.

PinpointTownes:
> @PinpointTownes it is not about if there should be / could be something of interest.

IMHO, it's definitely the most important question, I'm afraid.

Storing claims is not cheap. They end up being persisted in cookies and every claim contributes to making them bigger and bigger. It's particularly annoying because browsers like Safari are known to be insanely restrictive (see http://browsercookielimits.squawky.net/ for more information).

I'm not convinced re-including protocol claims that are used once in a blue moon is really worth it, specially since it shouldn't be hard to change using the claims actions API.

brentschmaltz:
@PinpointTownes I humbly disagree for amr and acr. So let me say this a different way, definitely there is no value in remembering the nonce. There may be value in remembering the amr / acr.  How many bytes are we talking about here? Not very many really.

For jwt's turning off long name mapping (name -> https://www.... ) will be a big win.

PinpointTownes:
> How many bytes are we talking about here? Not very many really.

When you have a 4kb per-domain limit (e.g on Safari), every byte counts. `acr` is implementation specific and can be potentially lengthy. Add the claim type, the value type, the number of claim properties (they are stored in the ticket) and the overhead caused by encryption... and boom, the limit is already reached.

What's wrong with having to opt-in for these claims?

brentschmaltz:
@PinpointTownes You didn't answer the question: how many bytes are we talking about? Claim type is 'amr', no claim properties. Don't know how encryption comes into play. normally encryption doesn't increase the size. A wrapper may for authenticated encryption.

You can set a filter on JwtSecurityTokenHandler to kick out any claims you want.

https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/master/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs#L77

Clearing the inbound map will remove the claim type mapping.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/master/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs#L67

PinpointTownes:
> You didn't answer the question: how many bytes are we talking about?

As I already said, `acr` is implementation-specific. You can store everything you want there. The IdentityServer folks use that for tenant/provider info and I've used it many times to store the external provider URL, which can be a long string.

I'll make some testing tomorrow.

> You can set a filter on JwtSecurityTokenHandler to kick out any claims you want.

That's no longer the recommended approach in ASP.NET Core 2.0.

PinpointTownes:
> no claim properties

We still have to store a `int32` containing the number of properties, even if it's 0. Take a look at the code if you're not familiar: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication/Data/TicketSerializer.cs#L240

Tratcher:
From a usability perspective it will be easier for a user to delete these claims than it will be for them to re-add them after we delete them. The ClaimActions get access to the user info endpoint data (if any) and the ClaimsIdentity, not the JWT.
https://github.com/aspnet/Security/blob/bd19ba9533cfd3ffd81e4ef2c874109e8b3dbccd/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L823
https://github.com/aspnet/Security/blob/bd19ba9533cfd3ffd81e4ef2c874109e8b3dbccd/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L667

PinpointTownes:
> The ClaimActions get access to the user info endpoint data (if any) and the ClaimsIdentity, not the JWT.

Really? That sucks! :sweat: 

If you can't change/improve that in RTM, consider at least adding an `IncludeProtocolClaims` option to disable the built-in claims exclusion.

brentschmaltz:
@Tratcher @PinpointTownes even if it not simple to get back to the jwt, I would still recommend removing the obvious claims: nonce, c_hash, at_hash, aud, iss.

The claimsIdentity usually has a copy of the token in bootstrapcontext. Did that go away?

Tratcher:
The bootstrap context should still be there, you set it, not us.

brentschmaltz:
@Tratcher thanks. It is trigger by this setting: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/master/src/Microsoft.IdentityModel.Tokens/TokenValidationParameters.cs#L481 I was just checking that the default hasn't changed.

Does this property get serialized by cookie MW?

Tratcher:
Yes: https://github.com/aspnet/Security/blob/bd19ba9533cfd3ffd81e4ef2c874109e8b3dbccd/src/Microsoft.AspNetCore.Authentication/Data/TicketSerializer.cs#L95

brockallen:
The difference between iss and amr is that iss is OIDC protocol, whereas amr is metadata about the user's authentication. I agree amr is useful to keep, whereas iss is not.

Tratcher:
Actually, if we leave the Delete Actions there then users can remove the actions via `oidcOptions.ClaimActions.Remove("acr");`
https://github.com/aspnet/Security/blob/bd19ba9533cfd3ffd81e4ef2c874109e8b3dbccd/src/Microsoft.AspNetCore.Authentication.OAuth/Claims/ClaimActionCollection.cs#L27

So it becomes a question of frequency. How often are servers going to provide these claims, and how often are apps going to need them as claims outside of the OIDC events?

brentschmaltz:
I agree that amr is the most probable to be used. For example AzureAD has a different value for public vs confidential clients after redeeming a code -> token.

brockallen:
For everyone's benefit, here are examples of what might go into amr: https://tools.ietf.org/html/rfc8176

Tratcher:
I'm going to close this PR. The ClaimActions are easy enough for a developer to adjust, or they can consume these claims in the OIDC events before the ClaimActions run.

PinpointTownes:
FWIW, I find `oidcOptions.ClaimActions.Remove("acr");` extremely confusing. It sounds like it registers an action that removes the `acr` claim (it does exactly the opposite, actually).

Maybe `RemoveAction` would be less confusing?

brockallen:
> FWIW, I find oidcOptions.ClaimActions.Remove("acr"); extremely confusing.

It's the same as: `oidcOptions.Scopes.Remove("profile")`

PinpointTownes:
@brockallen no, not at all. `oidcOptions.Scopes.Remove("profile")` really means "remove the profile scope".

`oidcOptions.ClaimActions.Remove("acr");` means "remove all the actions associated with the acr claim". In this specific case, this means "remove the action that removes the acr claim" or in other words, "preserve the acr claim".

brockallen:
Oh, I guess I've not been following these "Claim Actions" concepts. Sounds more cumbersome than just an even that fires right before the cookie is issued to allow control over the mapping.

Issue #1188, opened 2017-04-25
Provide AuthenticationProperties from state in OnRemoteFailure


Tratcher:
https://github.com/IdentityServer/IdentityServer4/issues/1087#issuecomment-297140577
When an remote login error occurs you may need to examine the state that flows with that challenge to understand how to handle it. For error scenarios we don't currently deserialize the state for you, and doing it yourself is pretty awkward.

anitchanana:
I am trying authentication with google as external provider but it is not asking for permissions and directly logging me in.
Earlier it used to ask for the permissions and i had the option to Allow/Deny.

Do you know if there are any changes in that process ?

anitchanana:
@Tratcher ,
thanks for the help earlier.
I am trying to integrate twitter now, but when i deny app the permission , it returnes with
url?denied=mmHukAAAAAAA0Yp2AAABW7PLTo

how can i get return url from there ?

I am using this code :

twitterOptions.Events = new TwitterEvents()
{
OnRemoteFailure = ctx =>
{
var authProperties = twitterOptions.StateDataFormat.Unprotect(ctx.Request.Query["denied"]);
ctx.Response.Redirect("/Account/login");
ctx.HandleResponse();
return Task.FromResult(0);
}
};

Can you please provide me insight into this. ?

Tratcher:
Twitter is special as it's still using OAuth1a. The AuthProperties aren't serialized into the url, they're only stored in the cookie.
https://github.com/aspnet/Security/blob/cccfbb7957d8bd132a2eceb362a98304d34b28f1/src/Microsoft.AspNetCore.Authentication.Twitter/TwitterHandler.cs#L134-L152

You get them back like this:
https://github.com/aspnet/Security/blob/cccfbb7957d8bd132a2eceb362a98304d34b28f1/src/Microsoft.AspNetCore.Authentication.Twitter/TwitterHandler.cs#L42-L51

anitchanana:
Thanks
Thats unusual but for last couple of days, with google authentication, they are not displaying consent screen to Allow/Deny app. Have there been any changes to this process. ?

Anything that you have got to know ?

Tratcher:
The consent screen is only show the first time you use/approve an app. To see it again you have to revoke the app's privileges from your account. There is also a `prompt` parameter you can add.
[...]

Issue #1178, opened 2017-04-13
OnAuthenticationFailed should be called when the OIDC authorization request returns an error


danroth27:
When I make an OIDC authorization request and the response is an error (ex. a post back with an error property) I expect the OnAuthenticationFailed event to be fired so that I can see the protocol message and potentially handle the event. Currently I have to handle this case using the OnRemoteFailure event, which only gives you the failure exception, and not the actual protocol message details.

I ran into this trying to port some Azure AD B2C sample code from Katana to ASP.NET Core. When you try to reset your password with Azure AD B2C you get back an error response with a particular error code in the error description that you need to handle. It's a bit ugly to do that with the OnRemoteFailure event.

Eilon:
@Tratcher let's take a look at this for 2.0.0.

brockallen:
IIRC there's another similar issue already open related to this // @Tratcher

Tratcher:
@brockallen https://github.com/aspnet/Security/issues/884 ?

brockallen:
Actually, this is the one I was thinking of: https://github.com/aspnet/Security/issues/1188

Tratcher:
Offline discussion: how can we flow information into OnRemoteFailure in a more structured way? #1188 asks for the auth properties. @danroth27 is asking for the structured failure response. Right now the failures are rendered as an exception message:
https://github.com/aspnet/Security/blob/bd19ba9533cfd3ffd81e4ef2c874109e8b3dbccd/src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs#L51
https://github.com/aspnet/Security/blob/bd19ba9533cfd3ffd81e4ef2c874109e8b3dbccd/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L1211

brockallen:
There are several places where just throwing is not really helpful. One big one is in the JwtSecurityTokehNalder itself -- what a bad design. I don't know/recall if the OIDC MW suppresses that exception, but it'd be nice for it to.

So, then, how about an "ErrorPageUrl", akin to how the cookie MW has login and access denied path properties to redirect the user to? As for capturing the error info, another all purpose error event seems like an idea.... but yikes, that's a lot of events.

Tratcher:
[...]

PR #1299, opened 2017-07-06
Add AuthenticationProperties to HandleRequestResult and RemoteFailureContext

Tratcher:
#1188 Requires https://github.com/aspnet/HttpAbstractions/pull/889
User state for the login is flown via auth properties, but these aren't available in the OnRemoteFailure event. To add them we need to flow them via the AuthResult.

If this works out then we may use it to flow additional data for #1178. The difference is that the data in #1178 isn't a consistent data structure across handlers.

/cc: @brockallen

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
@@ -491,13 +491,10 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
                 return HandleRequestResult.Fail("No message.");
             }
 
+            AuthenticationProperties properties = null;
             try
             {
-                AuthenticationProperties properties = null;
-                if (!string.IsNullOrEmpty(authorizationResponse.State))
-                {
-                    properties = Options.StateDataFormat.Unprotect(authorizationResponse.State);
-                }
+                properties = ReadPropertiesAndClearState(authorizationResponse);
 
                 var messageReceivedContext = await RunMessageReceivedEventAsync(authorizationResponse, properties);
                 if (messageReceivedContext.Result != null)
@@ -521,8 +518,7 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
                         return HandleRequestResult.Fail(Resources.MessageStateIsNullOrEmpty);
                     }
 
-                    // if state exists and we failed to 'unprotect' this is not a message we should process.
-                    properties = Options.StateDataFormat.Unprotect(authorizationResponse.State);
+                    properties = ReadPropertiesAndClearState(authorizationResponse);
                 }
 
                 if (properties == null)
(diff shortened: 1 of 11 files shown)


brockallen:
More for @anitchanana than for me to review.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -475,6 +475,12 @@ protected virtual Task<bool> HandleSignOutCallbackAsync()
                 authorizationResponse = messageReceivedContext.ProtocolMessage;
                 properties = messageReceivedContext.Properties;
 
+                // if any of the error fields are set, throw error null
+                if (!string.IsNullOrEmpty(authorizationResponse.Error))
```
PinpointTownes: Unwanted change?

Tratcher: I was trying to make the order of operations consistent between OAuth and OIDC.

PinpointTownes: IMHO, that would be better to fix the OAuth2 handler to validate the `state`/XSRF cookie before handling the error.

Tratcher: Fair. I was worried about the state being dropped for some error responses and thus hiding the error response with a correlation failure. Not sure how often that happens in practice.

Tratcher:
*bump*
I've filled out the tests and addressed some of the ordering issues.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -491,12 +491,19 @@ public override Task<bool> HandleRequestAsync()
...
             }
 
+            AuthenticationProperties properties = null;
             try
             {
-                AuthenticationProperties properties = null;
                 if (!string.IsNullOrEmpty(authorizationResponse.State))
                 {
                     properties = Options.StateDataFormat.Unprotect(authorizationResponse.State);
+
```
HaoK: Nit: private helper method for this Unprotect/decode clear state block that's repeated?

src/Microsoft.AspNetCore.Authentication/Events/RemoteFailureContext.cs
```diff
@@ -11,6 +11,7 @@ namespace Microsoft.AspNetCore.Authentication
     /// </summary>
     public class RemoteFailureContext : HandleRequestContext<RemoteAuthenticationOptions>
     {
+        [Obsolete]
```
HaoK: Should this have some text to go along with the warning?

Tratcher: *shrug* we don't really expect anyone to use these directly.

HaoK: Why bother with obsolete at all, its not incorrect to use this ctor is it?

Tratcher: It's obsolete as the first step towards getting rid of it, it's not used anymore.

Tratcher:
Updated

PR #1302, opened 2017-07-07
Fix tests to work with new default

HaoK:
@Tratcher

### DIFF
--- a/test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs
+++ b/test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs
@@ -1247,7 +1247,7 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
             => CreateServerWithServices(s =>
             {
                 s.AddSingleton<ISystemClock>(_clock);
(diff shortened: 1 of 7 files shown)


test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs
```diff
@@ -1247,7 +1247,7 @@ private TestServer CreateServer(Action<CookieAuthenticationOptions> configureOpt
             => CreateServerWithServices(s =>
             {
                 s.AddSingleton<ISystemClock>(_clock);
-                s.AddAuthentication().AddCookie(configureOptions);
+                s.AddAuthentication(o => o.DefaultScheme = CookieAuthenticationDefaults.AuthenticationScheme).AddCookie(configureOptions);
```
Tratcher: An `s.AddAuthentication(CookieAuthenticationDefaults.AuthenticationScheme)` overload for this would be nice.

HaoK: Yeah @danroth27 wants it too, I'll add it now since it makes fixing all the tests easier :)

PR #1303, opened 2017-07-07
Add AddAuthentication(defaultScheme) overload

HaoK:
@Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication/AuthenticationServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNetCore.Authentication/AuthenticationServiceCollectionExtensions.cs
@@ -27,6 +27,9 @@ namespace Microsoft.Extensions.DependencyInjection
             return new AuthenticationBuilder(services);
         }
 
+        public static AuthenticationBuilder AddAuthentication(this IServiceCollection services, string defaultScheme)
+            => services.AddAuthentication(o => o.DefaultScheme = defaultScheme);
+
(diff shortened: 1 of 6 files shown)


PinpointTownes:
That's a terrible idea. Folks will be tempted to call this middleware multiple times, resulting in all the `IAuthenticationRequestHandler` being invoked multiple times! :scream:

HaoK:
Not quite, all that would result in is the DefaultScheme being set to the last one.  This is Add, not Use

PinpointTownes:
Arf, my bad, doing too many at the same time :sweat_smile: 

So the only risk is that the first call will be ignored?

HaoK:
Its not different than today, AddAuthentication(o => o.DoStuff()).AddAuthentication(o => o.DoMoreStuff()).

DoStuff() and DoMoreStuff() are called, so whatever side effects happen...happen :)

PinpointTownes:
What I mean is that if you do `services.AddAuthentication("abc")` and then `services.AddAuthentication("xyz")`, the second call will override the first one? So the default scheme will be `xyz`, not `abc`?

If that how it works, then that's fine :smile:

HaoK:
Yup, its just sugar for o => o.DefaultScheme = "abc"

PinpointTownes:
Great, thanks, and sorry for the false alert. I'll stop doing too many things at the same time :trollface:

HaoK:
Eh welcome to my world right now :)

PR #1314, opened 2017-07-12
Add sample for refresh tokens

Tratcher:
Not actually planning to merge this yet, but I wanted to put it up here for discussion.

Not sure if there's a way to standardize this across our providers. Google followed the spec, but the others look harder. Twitter and Github tokens don't expire so there's no refresh. Facebook has their own refresh flow. I haven't found the right MSA docs yet.
https://developers.facebook.com/docs/facebook-login/access-tokens/expiration-and-extension

### DIFF
--- a/samples/OpenIdConnectSample/Startup.cs
+++ b/samples/OpenIdConnectSample/Startup.cs
@@ -1,6 +1,8 @@
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
+using System.Net.Http;
 using System.Text.Encodings.Web;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authentication;
@@ -11,7 +13,9 @@ using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
 using Microsoft.IdentityModel.Protocols.OpenIdConnect;
+using Newtonsoft.Json.Linq;
 
 namespace OpenIdConnectSample
 {
@@ -53,6 +57,7 @@ namespace OpenIdConnectSample
                 o.ClientSecret = Configuration["oidc:clientsecret"]; // for code flow
                 o.Authority = Configuration["oidc:authority"];
                 o.ResponseType = OpenIdConnectResponseType.CodeIdToken;
+                o.SaveTokens = true;
                 o.GetClaimsFromUserInfoEndpoint = true;
                 o.Events = new OpenIdConnectEvents()
                 {
@@ -73,19 +78,21 @@ namespace OpenIdConnectSample
             });
         }
 
-        public void Configure(IApplicationBuilder app)
+        public void Configure(IApplicationBuilder app, IOptionsMonitor<OpenIdConnectOptions> optionsMonitor)
         {
             app.UseDeveloperExceptionPage();
             app.UseAuthentication();
 
             app.Run(async context =>
             {
+                var response = context.Response;
+
                 if (context.Request.Path.Equals("/signedout"))
                 {
-                    await WriteHtmlAsync(context.Response, async res =>
+                    await WriteHtmlAsync(response, async res =>
                     {
                         await res.WriteAsync($"<h1>You have been signed out.</h1>");
-                        await res.WriteAsync("<a class=\"btn btn-link\" href=\"/\">Sign In</a>");
+                        await res.WriteAsync("<a class=\"btn btn-default\" href=\"/\">Home</a>");
                     });
                     return;
                 }
@@ -93,10 +100,10 @@ namespace OpenIdConnectSample
                 if (context.Request.Path.Equals("/signout"))
                 {
                     await context.SignOutAsync(CookieAuthenticationDefaults.AuthenticationScheme);
-                    await WriteHtmlAsync(context.Response, async res =>
+                    await WriteHtmlAsync(response, async res =>
                     {
-                        await context.Response.WriteAsync($"<h1>Signed out {HtmlEncode(context.User.Identity.Name)}</h1>");
-                        await context.Response.WriteAsync("<a class=\"btn btn-link\" href=\"/\">Sign In</a>");
+                        await res.WriteAsync($"<h1>Signed out {HtmlEncode(context.User.Identity.Name)}</h1>");
+                        await res.WriteAsync("<a class=\"btn btn-default\" href=\"/\">Home</a>");
                     });
                     return;
                 }
@@ -115,19 +122,22 @@ namespace OpenIdConnectSample
                 if (context.Request.Path.Equals("/Account/AccessDenied"))
                 {
                     await context.SignOutAsync(CookieAuthenticationDefaults.AuthenticationScheme);
-                    await WriteHtmlAsync(context.Response, async res =>
+                    await WriteHtmlAsync(response, async res =>
                     {
-                        await context.Response.WriteAsync($"<h1>Access Denied for user {HtmlEncode(context.User.Identity.Name)} to resource '{HtmlEncode(context.Request.Query["ReturnUrl"])}'</h1>");
-                        await context.Response.WriteAsync("<a class=\"btn btn-link\" href=\"/signout\">Sign Out</a>");
+                        await res.WriteAsync($"<h1>Access Denied for user {HtmlEncode(context.User.Identity.Name)} to resource '{HtmlEncode(context.Request.Query["ReturnUrl"])}'</h1>");
+                        await res.WriteAsync("<a class=\"btn btn-default\" href=\"/signout\">Sign Out</a>");
+                        await res.WriteAsync("<a class=\"btn btn-default\" href=\"/\">Home</a>");
                     });
                     return;
                 }
 
                 // DefaultAuthenticateScheme causes User to be set
-                var user = context.User;
+                // var user = context.User;
 
                 // This is what [Authorize] calls
-                // var user = await context.AuthenticateAsync();
+                var userResult = await context.AuthenticateAsync();
+                var user = userResult.Principal;
+                var props = userResult.Properties;
 
                 // This is what [Authorize(ActiveAuthenticationSchemes = OpenIdConnectDefaults.AuthenticationScheme)] calls
                 // var user = await context.AuthenticateAsync(OpenIdConnectDefaults.AuthenticationScheme);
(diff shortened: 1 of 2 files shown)


samples/SocialSample/Startup.cs
```diff
@@ -228,6 +231,99 @@ public void Configure(IApplicationBuilder app)
...
+                        context.Response.ContentType = "text/html";
+                        await context.Response.WriteAsync("<html><body>");
+                        await context.Response.WriteAsync("No refresh_token is available.<br>");
+                        await context.Response.WriteAsync("<a href=\"/\">Home</a>");
+                        await context.Response.WriteAsync("</body></html>");
+                    }
+
+                    if (string.Equals(GoogleDefaults.AuthenticationScheme, user.Identities.First().AuthenticationType))
+                    {
+                        var handler = (AuthenticationHandler<GoogleOptions>)await authSchemes.GetHandlerAsync(context, GoogleDefaults.AuthenticationScheme);
```
HaoK: Nice, guess we enabling these kind of scenarios with the IAuthenticationHandlerProvider as a free side effect of moving to services :)

HaoK: Although this line is a bit ugly, could use some lipstick :)

samples/SocialSample/Startup.cs
```diff
@@ -228,6 +231,99 @@ public void Configure(IApplicationBuilder app)
...
+                        authProperties.UpdateTokenValue("access_token", payload.Value<string>("access_token"));
+                        if (int.TryParse(payload.Value<string>("expires_in"), NumberStyles.Integer, CultureInfo.InvariantCulture, out var seconds))
+                        {
+                            var expiresAt = DateTimeOffset.UtcNow + TimeSpan.FromSeconds(seconds);
+                            authProperties.UpdateTokenValue("expires_at", expiresAt.ToString("o", CultureInfo.InvariantCulture));
+                        }
+                        await context.SignInAsync(user, authProperties);
+
+                        context.Response.ContentType = "text/html";
+                        await context.Response.WriteAsync("<html><body>");
```
HaoK: nit: so many context.Response, store it in a var?  Think of all the poor chars that will be saved!

samples/SocialSample/Startup.cs
```diff
@@ -228,6 +231,99 @@ public void Configure(IApplicationBuilder app)
...
+
+                        var pairs = new Dictionary<string, string>()
+                        {
+                            { "client_id", handler.Options.ClientId },
+                            { "client_secret", handler.Options.ClientSecret },
+                            { "grant_type", "refresh_token" },
+                            { "refresh_token", refreshToken }
+                        };
+                        var content = new FormUrlEncodedContent(pairs);
+                        var response = await handler.Options.Backchannel.PostAsync(GoogleDefaults.TokenEndpoint, content, context.RequestAborted);
```
HaoK: Are you only using the handler to get the options out?  If so, you can just directly ask for those via `IOptionsMonitor<GoogleOptions>.Get(GoogleDefaults.AuthenticationScheme))`

Tratcher: Yes, but I want to be able to look up all the providers without injecting them all.

Tratcher: I'm also trying to find a way to move this logic into the handlers, and may need direct access to them to invoke it.

PinpointTownes: > I'm also trying to find a way to move this logic into the handlers, and may need direct access to them to invoke it.

Adding such support in the handlers themselves is a good idea but directly accessing them sounds weird. A better flow would be to add a `RenewAsync()` method in `IAuthenticationService` just like what we have for authentication/sign-in/sign-out/challenge/forbid.

Tratcher: Agreed, this is more of an intermediate while I try to work out the generic version. It's not clear what the input or output of Renew would be, if it would operate on the SignInScheme, etc.. There would need to be a version where you're storing the tokens somewhere like Identity and pass in something like AuthProperties directly and get it back, then you manage the persistence. And you want the flow I already have where it's all managed for you in the default cookie.

PinpointTownes: > It's not clear what the input or output of Renew would be.

How about something like:

```csharp
AuthenticateResult RenewAsync(AuthenticationProperties properties);
```

> if it would operate on the SignInScheme

IMO, it should be consistent with what we already do in the callback part, so yeah, it should probably call `SignInAsync(SignInScheme)` to return a new cookie with the updated identity/claims/tokens.

Tratcher: There's at least two versions:
`AuthenticateResult RenewAsync();` Gets the current identity from AuthenticateAsync, stores the result via SignIn and returns it.
`AuthenticateResult RenewAsync(AuthenticationProperties properties);` Operates on the information give, returns the result, does not call SignIn. This is for the temp cookie scenario where you've moved the data elsewhere.
Also, returning a full AuthResult is odd because that is Ticket based, and we're trying to renew tokens, not identities.

PinpointTownes:
> AuthenticateResult RenewAsync(AuthenticationProperties properties); Operates on the information give, returns the result, does not call SignIn. This is for the temp cookie scenario where you've moved the data elsewhere.

For this very specific case, it would probably be easier for folks to directly send the refresh token request in their own code.

> Also, returning a full AuthResult is odd because that is Ticket based, and we're trying to renew tokens, not identities.

"renewing tokens" would be specific to OAuth/OIDC handlers and not a very good fit for a generic method in the handlers infrastructure.

Working at the identity level sounds better and would allow supporting identity refreshing with OIDC (an identity token can be returned in refresh token responses). Handlers that don't support that (like the OAuth2 handlers) could simply use the identity returned by `AuthenticateAsync()` and just update the authentication properties with the new tokens.

PinpointTownes:
> Handlers that don't support that (like the OAuth2 handlers) could simply use the identity returned by AuthenticateAsync() and just update the authentication properties with the new tokens.

Well, actually, we could even support such a scenario with the OAuth2 handlers, by sending a "userinfo request" using the new access token.

Tratcher:
One of the reasons to put code in the handlers is that's where all of the config already is. Even if it wasn't generic, just a method we expose from the GoogleHandler, that'd be an improvement on what we (don't) have now.

We haven't had asks for a full identity renewal, and the whole sign-in process is usually more involved than just fetching user info.

PinpointTownes:
> We haven't had asks for a full identity renewal, and the whole sign-in process is usually more involved than just fetching user info.

Why would you need a full sign-in? At the renewal stage, you have a refresh token that you redeem for an access token (you can use to send userinfo requests) and optionally, an identity token (that you must validate, BTW).

PinpointTownes:
It would be also interesting to see if the existing renewal logic in the cookies handler could be moved to this new thing so it's easier for folks to manually renew cookies.

Tratcher:
Rebased. I've now got samples for OIDC, Facebook, Google & MSA (Generic OAuth2). Twitter and Github don't support refresh, so that's all of the sample providers for now. I'm not going to try to develop a generic refresh API yet, let's review and commit this sample as is and then revisit the API approach later. @HaoK

samples/OpenIdConnectSample/Startup.cs
```diff
@@ -73,30 +77,34 @@ public void ConfigureServices(IServiceCollection services)
...
+                var response = context.Response;
+
                 if (context.Request.Path.Equals("/signedout"))
                 {
-                    await WriteHtmlAsync(context.Response, async res =>
+                    await WriteHtmlAsync(response, async res =>
                     {
                         await res.WriteAsync($"<h1>You have been signed out.</h1>");
-                        await res.WriteAsync("<a class=\"btn btn-link\" href=\"/\">Sign In</a>");
+                        await res.WriteAsync("<a class=\"btn btn-default\" href=\"/\">Sign In</a>");
```
HaoK: Are these buttons doing anything different?  Should the href be different for SignIn?

Tratcher: No, going to the home page auto-logs in. I'll remove them.

samples/OpenIdConnectSample/Startup.cs
```diff
@@ -151,15 +162,76 @@ await context.SignOutAsync(OpenIdConnectDefaults.AuthenticationScheme, new Authe
...
+                        await WriteHtmlAsync(response, async res =>
+                        {
+                            await res.WriteAsync($"No refresh_token is available.<br>");
+                            await res.WriteAsync("<a class=\"btn btn-link\" href=\"/signout\">Sign Out</a>");
+                        });
+
+                        return;
+                    }
+
+                    var handler = (AuthenticationHandler<OpenIdConnectOptions>)await authSchemes.GetHandlerAsync(context, OpenIdConnectDefaults.AuthenticationScheme);
```
HaoK: If you aren't using the handler yet, I wouldn't resolve the handler, I'd just get the options from IOptionsManager instead

samples/SocialSample/Startup.cs
```diff
@@ -281,24 +427,46 @@ public void Configure(IApplicationBuilder app)
...
             });
         }
+
+        private async Task<OAuthOptions> GetOAuthOptionsAsync(IAuthenticationHandlerProvider authSchemes, HttpContext context, string currentAuthType)
+        {
+            var handler = await authSchemes.GetHandlerAsync(context, currentAuthType);
+
+            if (string.Equals(GoogleDefaults.AuthenticationScheme, currentAuthType))
+            {
+                return ((AuthenticationHandler<GoogleOptions>)handler).Options;
```
HaoK: Since you already have a switch you should just do `IOptionsMonitor<GoogleOptions>.Get(GoogleDefaults.AuthenticationScheme)` and skip resolving the handler, same for the other cases

samples/SocialSample/Startup.cs
```diff
@@ -215,16 +217,158 @@ public void Configure(IApplicationBuilder app)
...
+                            authProperties.UpdateTokenValue("expires_at", expiresAt.ToString("o", CultureInfo.InvariantCulture));
+                        }
+                        await context.SignInAsync(user, authProperties);
+
+                        response.ContentType = "text/html";
+                        await response.WriteAsync("<html><body>");
+                        await response.WriteAsync("Refreshed.<br>");
+                        await response.WriteAsync(HtmlEncoder.Default.Encode(payload.ToString()).Replace(",", ",<br>") + "<br>");
+
+                        await response.WriteAsync("<br>Tokens:<br>");
```
HaoK: Move this token printing into a helper?  seems like there's 2-3 instances of the same wall of text?

HaoK:
Wherever this PR lands, you should add equivalent tests in each auth mocking the flows that the sample is exercising at the same time.

Tratcher:
There aren't tests to add, this functionality is all external to the handlers.

HaoK:
So this is just demonstrating the Token sugar methods we added a long time ago to AuthenticationProperties?

Tratcher:
Pretty much, and the refresh specifics of each provider.

Issue #1319, opened 2017-07-13
DisplayName in new AuthN Plumbing?


leastprivilege:
How do I set it? It's gone from the options class.

Tratcher:
https://github.com/aspnet/Security/blob/ff9f145a8e89c9756ea12ff10c6d47f2f7eb345f/src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs#L37
https://github.com/aspnet/Security/blob/df325deaf355f99239c1319932bc28eb3a837372/src/Microsoft.AspNetCore.Authentication/AuthenticationServiceCollectionExtensions.cs#L50
@haok is this exposed anywhere else? E.g. I don't see a way to change it for Google, OIDC, etc..

leastprivilege:
but not here

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectExtensions.cs#L20

Tratcher:
It's set here to match the scheme name: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectExtensions.cs#L23
Needs an overload to set the display name.

HaoK:
Yeah we don't have those overloads in the security layer right now, so you'd have to drop down and use the lower overload to tweak the display name

PinpointTownes:
If you decide to fix that in 2.0 (and I personally hope you will), please also fix https://github.com/aspnet/IISIntegration/issues/391 (I'm fine submitting a PR if you prefer).

HaoK:
Doubt we are doing that now, since its a non breaking change in 2.1

You can probably workaround this via directly tweaking the DisplayName via AuthenticationOptions.SchemeMap:

`services.Configure<AuthenticationOptions>(o => o.SchemeMap["OIDC"].DisplayName = whatever`

leastprivilege:
That's ridiculous.

PinpointTownes:
> Doubt we are doing that now, since its a non breaking change in 2.1

Well, it automatically means waiting maybe 5 or 6 months more before moving to 2.x.
And while you can indeed work around this limitation with the "classical" handlers, it's not possible with the host handlers.

HaoK:
What are you trying to do with display name that is so critical?  Check if its null?

PinpointTownes:
[...]

PR #1320, opened 2017-07-13
Add DisplayName overloads

Tratcher:
#1319 @leastprivilege @PinpointTownes 
No change to defaults.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookExtensions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookExtensions.cs
@@ -16,6 +16,9 @@ namespace Microsoft.Extensions.DependencyInjection
             => builder.AddFacebook(FacebookDefaults.AuthenticationScheme, configureOptions);
 
         public static AuthenticationBuilder AddFacebook(this AuthenticationBuilder builder, string authenticationScheme, Action<FacebookOptions> configureOptions)
-            => builder.AddOAuth<FacebookOptions, FacebookHandler>(authenticationScheme, configureOptions);
+            => builder.AddFacebook(authenticationScheme, FacebookDefaults.DisplayName, configureOptions);
+
+        public static AuthenticationBuilder AddFacebook(this AuthenticationBuilder builder, string authenticationScheme, string displayName, Action<FacebookOptions> configureOptions)
+            => builder.AddOAuth<FacebookOptions, FacebookHandler>(authenticationScheme, displayName, configureOptions);
     }
 }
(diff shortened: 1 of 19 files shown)


Tratcher:
@DamianEdwards

src/Microsoft.AspNetCore.Authentication.Facebook/FacebookExtensions.cs
```diff
@@ -16,6 +16,9 @@ public static AuthenticationBuilder AddFacebook(this AuthenticationBuilder build
             => builder.AddFacebook(FacebookDefaults.AuthenticationScheme, configureOptions);
 
         public static AuthenticationBuilder AddFacebook(this AuthenticationBuilder builder, string authenticationScheme, Action<FacebookOptions> configureOptions)
-            => builder.AddOAuth<FacebookOptions, FacebookHandler>(authenticationScheme, configureOptions);
+            => builder.AddFacebook(authenticationScheme, authenticationScheme, configureOptions);
```
PinpointTownes: That would be great to avoid using the scheme as the display name when we have better options (with the social providers, we can default to the provider name, like we used to do in 1.0)

Tratcher: For the `UseFacebook()` overload you'll get scheme name Facebook and display name Facebook. As soon as you add multiple schemes you'd want unique display names, no? or are you talking about having just one but with a non-default scheme name?

PinpointTownes: > As soon as you add multiple schemes you'd want unique display names, no?

In 1.0, `Facebook` is always used as the default display name, even if you use a different authentication scheme (so no, it's not unique). I think it would be preferable to have the same behavior in 2.0.

Tratcher: That's doable. You didn't actually answer the question though, would you really want duplicates in your login list?

PinpointTownes: > You didn't actually answer the question though, would you really want duplicates in your login list?

I'd prefer that over using my custom scheme as the display name :smile: 
To avoid that, I could simply use the longest overload.

My point is that you can use the `builder.AddFacebook("custom-scheme", options => { })` overload without necessarily having multiple Facebook handlers. In this case, it wouldn't make much sense to use `custom-scheme` as the display name.

Tratcher:
Updated default names.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectExtensions.cs
```diff
@@ -18,9 +18,12 @@ public static AuthenticationBuilder AddOpenIdConnect(this AuthenticationBuilder
             => builder.AddOpenIdConnect(OpenIdConnectDefaults.AuthenticationScheme, configureOptions);
 
         public static AuthenticationBuilder AddOpenIdConnect(this AuthenticationBuilder builder, string authenticationScheme, Action<OpenIdConnectOptions> configureOptions)
+            => builder.AddOpenIdConnect(authenticationScheme, OpenIdConnectDefaults.Caption, configureOptions);
```
PinpointTownes: Haha, wasn't this constant dead code? :sweat_smile: 

It would be nice to rename it to `DisplayName` and add similar constants in the other providers, for consistency.

brockallen:
Thank you for getting this feature back in.

leastprivilege:
thanks!

PR #1328, opened 2017-07-20
AuthZ PolicyEvalutor should take resource

HaoK:
@davidfowl 

Will file PRs for SignalR/MVC next


HaoK:
Replaced with PR against rel/2.0.0 https://github.com/aspnet/Security/pull/1330

PR #1363, opened 2017-08-12
Fix obsolete warning message

pksorensen:
Fixes the messaged shown to developers when they are updating their applications to 2.0. Bad idea to set the domain with some random naming text.


natemcmaster:
Thanks @pksorensen

Issue #1233, opened 2017-05-25
Add MaxAge to OpenIdConnectOptions


openidauthority:
In a single sign-on environment, sometimes a specific application needs to have a shorter session than the identity provider. For example, the identity provider might have a 10 hour session, but a high-security application might need to re-authenticate users after 15 minutes of inactivity. Setting a short expiration cookie on the client application is ineffective, because when the local session expires, the user is automatically logged back in due to the existing session at the identity provider.

The [OIDC spec](http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) has a max_age parameter that can be sent with an authorization request. If present, and the identity provider has not actively authenticated the user within the length of time specified, the user will be prompted to re-authenticate.

I propose that a MaxAge property be added to the OpenIdConnectOptions, and if the time span is one second or longer, that the max_age parameter would be set for the authenticate request.

I have tested this locally already and can submit a pull request.

openidauthority:
Should this be implemented as a TimeSpan, int, or nullable int?

Eilon:
This seems reasonable, we are working to finish the 2.0 release, but we will take a look at it for the 2.1 release (which isn't too much later).

As to the type, a nullable Timespan would be ideal.

Eilon:
@openidauthority - any chance you want to submit a PR for this? We've started work on the 2.1 release and this would be a good feature to have in there.

openidauthority:
I'll give it a shot a soon as I have a bit of time.


On Thu, Aug 10, 2017, at 03:42 PM, Eilon Lipton wrote:
> @openidauthority[1] - any chance you want to submit a PR for this?
> We've started work on the 2.1 release and this would be a good feature
> to have in there.> — You are receiving this because you were mentioned. Reply to this
> email directly, view it on GitHub[2], or mute the thread[3].> 


Links:

  1. https://github.com/openidauthority
  2. https://github.com/aspnet/Security/issues/1233#issuecomment-321693041

openidauthority:
@Eilon - you can assign it to me.

Eilon:
Hmm that must be a new feature - I don't recall GitHub allowing assigning an issue to someone not in the org! Or maybe I never noticed it before.

@openidauthority thanks for taking a crack at this!

PR #1374, opened 2017-08-17
Add MaxAge to OpenIdConnectOptions

openidauthority:
- max_age parameter added to the authentication request if MaxAge is 1 second or more
 - Fractions of seconds are ignored
 - See http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest for expected behavior

Addresses #1233

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
@@ -320,6 +320,14 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
                 Scope = string.Join(" ", Options.Scope)
             };
 
+            // Add the 'max_age' parameter to the authentication request if MaxAge is not null.
+            // See http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
+            if (Options.MaxAge != null)
+            {
+                message.MaxAge = Convert.ToInt64(Math.Floor(((TimeSpan)Options.MaxAge).TotalSeconds))
+                    .ToString(CultureInfo.InvariantCulture);
+            }
+
             // Omitting the response_mode parameter when it already corresponds to the default
             // response_mode used for the specified response_type is recommended by the specifications.
             // See http://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#ResponseModes
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
@@ -84,6 +84,11 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
         {
             base.Validate();
 
+            if (MaxAge != null && MaxAge.Value < TimeSpan.Zero)
+            {
+                throw new InvalidOperationException("MaxAge must not be a negative TimeSpan.");
+            }
+
             if (string.IsNullOrEmpty(ClientId))
             {
                 throw new ArgumentException("Options.ClientId must be provided", nameof(ClientId));
@@ -158,6 +163,13 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
             get => (OpenIdConnectEvents)base.Events;
             set => base.Events = value;
         }
+
+        /// <summary>
+        /// Gets or sets the 'max_age'. If set the 'max_age' parameter will be sent with the authentication request. If the identity
+        /// provider has not actively authenticated the user within the length of time specified, the user will be prompted to
+        /// re-authenticate. By default no max_age is specified.
+        /// </summary>
+        public TimeSpan? MaxAge { get; set; } = null;
 
         /// <summary>
         /// Gets or sets the <see cref="OpenIdConnectProtocolValidator"/> that is used to ensure that the 'id_token' received
--- a/test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/OpenIdConnectChallengeTests.cs
+++ b/test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/OpenIdConnectChallengeTests.cs
@@ -409,5 +409,46 @@ namespace Microsoft.AspNetCore.Authentication.Test.OpenIdConnect
             var exception = await Assert.ThrowsAsync<InvalidOperationException>(() => server.SendAsync(ChallengeEndpoint));
             Assert.Equal("Cannot redirect to the authorization endpoint, the configuration may be missing or invalid.", exception.Message);
         }
+
+        [Fact]
+        public async Task Challenge_WithDefaultMaxAge_HasExpectedMaxAgeParam()
+        {
+            var settings = new TestSettings(
+                opt => 
+                { 
+                    opt.ClientId = "Test Id";
+                    opt.Authority = TestServerBuilder.DefaultAuthority;
+                });
+
+            var server = settings.CreateTestServer();
+            var transaction = await server.SendAsync(ChallengeEndpoint);
+
+            var res = transaction.Response;
+
+            settings.ValidateChallengeRedirect(
+                res.Headers.Location,
+                OpenIdConnectParameterNames.MaxAge);
+        }
+
+        [Fact]
+        public async Task Challenge_WithSpecificMaxAge_HasExpectedMaxAgeParam()
+        {
+            var settings = new TestSettings(
+                opt => 
+                { 
+                    opt.ClientId = "Test Id";
+                    opt.Authority = TestServerBuilder.DefaultAuthority;
+                    opt.MaxAge = TimeSpan.FromMinutes(20);
+                });
+
+            var server = settings.CreateTestServer();
+            var transaction = await server.SendAsync(ChallengeEndpoint);
+
+            var res = transaction.Response;
+
+            settings.ValidateChallengeRedirect(
+                res.Headers.Location,
+                OpenIdConnectParameterNames.MaxAge);
+        }
     }
 }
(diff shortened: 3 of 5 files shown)


src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -320,6 +320,14 @@ protected virtual Task<bool> HandleSignOutCallbackAsync()
                 Scope = string.Join(" ", Options.Scope)
             };
 
+            string maxAge = OpenIdConnectHelper.GetMaxAgeParameterFromTimeSpan(Options.MaxAge);
+            // Add the 'max_age' parameter to the authentication request if MaxAge is at least one second.
```
PinpointTownes: `max_age=0` is not disallowed by the specification and can be seen as an equivalent to `prompt=login` (in OpenID 2.0 PAPE, `max_auth_age=0` was explicitly allowed and served this exact purpose).

That would be great to avoid making the ASP.NET Core implementation stricter than the specification, specially since this won't be trivial to debug for beginners (invalid values are silently ignored).

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
```diff
@@ -160,6 +160,13 @@ public new OpenIdConnectEvents Events
         }
 
         /// <summary>
+        /// Gets or sets the 'max_age'. If the <see cref="TimeSpan"/> is at least one second, the 'max_age' parameter will be sent with the
+        /// authentication request. If the identity provider has not actively authenticated the user within the length of time specified, the
+        /// user will be prompted to re-authenticate. By default no max_age is specified.
+        /// </summary>
+        public TimeSpan MaxAge { get; set; } = new TimeSpan(0, 0, 0);
```
PinpointTownes: It would be preferable to use a nullable property so the `max_age=0` value can be validly represented.

test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/OpenIdConnectTests.cs
```diff
@@ -294,6 +294,31 @@ public class OpenIdConnectTests
...
+        public void MaxAgeConversion()
+        {
+            // max_age not set for default value
+            var options = new OpenIdConnectOptions();
+            Assert.Null(OpenIdConnectHelper.GetMaxAgeParameterFromTimeSpan(options.MaxAge));
+
+            // max_age not set for values less than 1 second
+            Assert.Null(OpenIdConnectHelper.GetMaxAgeParameterFromTimeSpan(TimeSpan.FromSeconds(0.999)));
+            Assert.Null(OpenIdConnectHelper.GetMaxAgeParameterFromTimeSpan(new TimeSpan(0, 0, 0)));
+            Assert.Null(OpenIdConnectHelper.GetMaxAgeParameterFromTimeSpan(TimeSpan.FromMinutes(-10)));
```
PinpointTownes: IMHO it would be better to throw an exception when validating the OIDC configuration. You should consider adding a `if (options.MaxAge != null && options.MaxAge.Value < TimeSpan.Zero)` check in `OpenIdConnectPostConfigureOptions.PostConfigure`.

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectPostConfigureOptions.cs#L31

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHelper.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using System;
+using System.Globalization;
+
+namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
+{
+    /// <summary>
+    /// Contains static methods for converting OpenIdConnectOptions to OIDC parameter values.
+    /// </summary>
+    public static class OpenIdConnectHelper
```
PinpointTownes: Curious: is it worth having this type in the public contract?

Tratcher: There's no reason for this to be public.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
Tratcher: Yes, make it nullable

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHelper.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="maxAge"></param>
+        /// <returns></returns>
+        public static string GetMaxAgeParameterFromTimeSpan(TimeSpan maxAge)
+        {
+            var maxAgeSeconds = Convert.ToInt64(Math.Floor(maxAge.TotalSeconds));
+            if (maxAgeSeconds >= 1)
```
Tratcher: max_age=0 looks like a valid use case.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -320,6 +320,14 @@ protected virtual Task<bool> HandleSignOutCallbackAsync()
...
+            // See http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
+            if (maxAge != null)
+            {
+                message.MaxAge = maxAge;
```
Tratcher: OpenIdConnectMessage allows assigning null, you don't need to null check here. You can inline all of this right after Scope.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectPostConfigureOptions.cs
```diff
@@ -30,6 +30,11 @@ public OpenIdConnectPostConfigureOptions(IDataProtectionProvider dataProtection)
         /// <param name="options">The options instance to configure.</param>
         public void PostConfigure(string name, OpenIdConnectOptions options)
         {
+            if (options.MaxAge != null && options.MaxAge.Value < TimeSpan.Zero) 
+            {
+                throw new InvalidOperationException("MaxAge must not be a negative TimeSpan.");
```
Tratcher: This actually belongs over here:
https://github.com/aspnet/Security/blob/442df16e265ab1955d7b8468cf1d8b38b8d915c7/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs#L87

test/Microsoft.AspNetCore.Authentication.Test/Microsoft.AspNetCore.Authentication.Test.csproj
```diff
@@ -23,8 +23,9 @@
     <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.AspNetCore.Testing" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(TestSdkVersion)" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="$(XunitVersion)" />
-    <PackageReference Include="xunit" Version="$(XunitVersion)" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.3.0-beta4-build3742" />
```
Tratcher: revert

test/Microsoft.AspNetCore.Authentication.Test/MicrosoftAccountTests.cs
```diff
@@ -1,242 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved. See License.txt in the project root for license information.
-
-using System;
```
Tratcher: ??

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectPostConfigureOptions.cs
PinpointTownes: Well, both places are fine 'cause we do validation in the initializer too. I suggested removing the `Validate` method from the options, but @HaoK wasn't okay with that... :trollface:

Tratcher: OpenIdConnectPostConfigureOptions is for delayed initialization, not basic args validation.

HaoK: Yep, Validate on the Options class itself should be the source of truth for validation

PinpointTownes: > OpenIdConnectPostConfigureOptions is for delayed initialization, not basic args validation.
 
I still think the two places should be consolidated. There's no reason to repeat the same validation checks for each request again and again.

test/Microsoft.AspNetCore.Authentication.Test/MicrosoftAccountTests.cs
openidauthority: I was just making a half-baked commit to my repo. I didn't realize that it would become part of the pull request :) I can't get it to build properly on command line, VS Code, or VS...different error each place, so this is not ready.

Eilon:
@openidauthority - we're wondering if you had an update on this PR so that we can proceed?

openidauthority:
@Eilon - I'm having issues with xunit in this branch, but I finished the work in a new branch (https://github.com/openidauthority/Security/tree/openidauthority/max_age). Should I open a new pull request from that branch?

Tratcher:
You can force push to replace this branch.

Tratcher:
Rebased, updated, and merged. Thanks.

PR #1387, opened 2017-08-24
Using WebEncoders' Base64Url encode/decode functionality

kichalla:


src/Microsoft.AspNetCore.Authentication/Data/TextEncoder.cs
```diff
@@ -13,7 +13,7 @@ public static class Base64UrlTextEncoder
         /// <returns>Base64 encoded string modified with non-URL encodable characters</returns>
         public static string Encode(byte[] data)
         {
-            return WebUtilities.Base64UrlTextEncoder.Encode(data);
+            return WebUtilities.WebEncoders.Base64UrlEncode(data);
```
Tratcher: You can remove the namespace right?

kichalla:
bf685de16be9949d67e93cc058ef4393f005756b

PR #1390, opened 2017-08-27
Removed useless checks

DamirAinullin:
Removed useless checks


src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
```diff
@@ -142,7 +142,7 @@ protected string BuildRedirectUri(string targetPath)
                     Logger.AuthenticationSchemeNotAuthenticated(Scheme.Name);
                 }
             }
-            else if (result?.Failure != null)
+            else
```
davidfowl: result can never be null?

DamirAinullin: if it is null then first "if" will be true (line 133).

davidfowl: Ah!

Issue #1398, opened 2017-08-30
Support for OpenIdConnect prompt parameter


gb056:
This may be a duplicate of [Issue 99](https://github.com/aspnet/Security/issues/99).

It would be good if this could support the inclusion of the [prompt](http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) parameter as part of the `OpenIdConnectOptions`. Similar to how we currently pass through the ResponseType, Resource etc.

The fix in this case may be as simple as adding `Prompt = Options.Prompt` [here](https://github.com/aspnet/Security/blob/f00efc46348c5e89d4c1665e6892427524069880/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L314) and adding a `Prompt` property to `OpenIdConnectOptions`.

Happy to do a PR on this if required.

Eilon:
We would be happy to work with you on a PR for this!

We think adding a new `ICollection<string> Prompt { get; }` property would be great. It should be similar to the `Scope` property:
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs#L212-L215

brockallen:
Can we keep it as space delimited? The only reason scope was a collection was to normalize those in the ASP.NET programming model because of the inconsistencies in the OAuth2 implementations (mainly facebook as the oddball, IIRC). This is a whitespace delimited set of values and your MW will just pass it along verbatim. Is there really a need for a collection in the options?

Eilon:
I guess it's a matter of which is easier to use. I don't have a strong feeling about it. Though it also might depend a bit on what any downstream APIs use in IdentityModel.

Tratcher:
@brockallen Assigning more than one needs to be clean. How would you do that? Perhaps with constants that already contain the permutations?

Tratcher:
OpenIdConnectResponseType does exactly that: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/f10a045459e7cb07548806561a47ecdb3471f5f6/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectResponseType.cs
We could ask IdentityModel for similar constants for Prompt.

brockallen:
[...]

Issue #990, opened 2016-09-23
[OAuth] Allow developer to easily customize query parameters on authorization endpoint


troydai:
Today what parameter to be put on the query is hardcoded: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs#L210. There are two existing ways to customize the list of parameters to be added.
1. Override the `BuildChallengeUrl` method. That requires a deeper overwrite to the handler.
2. Override the event `RedirectToAuthorizationEndpoint`. It is simpler than the first approach but still impose unnecessary difficulties. 

We may be add a list of allowed parameter names to the OAuth options and let the OAuth handler to dynamically build the query string.

Tratcher:
The Google middleware has a feature where it allows you to specify parameters via the AuthProperties in the challenge.
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.Google/GoogleHandler.cs#L98
Unfortunately these are hard-coded and get out of date. Consider changing this to a list of KVP of kyes to take from the auth properties and add to the query.

PR #1401, opened 2017-09-01
Added ability to set prompt parameter in OpenIdConnectOptions

gb056:
- Set prompt parameter to value set in Options in OpenIdConnectHandler.HandleChallengeAsync()
 - See http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest for expected behavior

Addresses #1398

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
@@ -212,7 +212,7 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
         /// <summary>
         /// Gets or sets the 'prompt'.
         /// </summary>
-        public string Prompt { get; set; } = "none";
+        public string Prompt { get; set; }
 
         /// <summary>
         /// Gets the list of permissions to request.
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
@@ -210,6 +210,11 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
         public string ResponseType { get; set; } = OpenIdConnectResponseType.IdToken;
 
         /// <summary>
+        /// Gets or sets the 'prompt'.
+        /// </summary>
+        public string Prompt { get; set; } = "none";
+
+        /// <summary>
         /// Gets the list of permissions to request.
         /// </summary>
         public ICollection<string> Scope { get; } = new HashSet<string>();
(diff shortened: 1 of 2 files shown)


gb056:
Default value of `OpenIdConnectOptions.Prompt` is hard coded to "none" for now, it would be great if we could use constants for this as with `ResponseType`

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs
```diff
@@ -210,6 +210,11 @@ public new OpenIdConnectEvents Events
         public string ResponseType { get; set; } = OpenIdConnectResponseType.IdToken;
 
         /// <summary>
+        /// Gets or sets the 'prompt'.
+        /// </summary>
+        public string Prompt { get; set; } = "none";
```
Tratcher: Why a default rather than empty?

PinpointTownes:
> Default value of OpenIdConnectOptions.Prompt is hard coded to "none" for now.

I agree with @Tratcher, you definitely can't do that. `prompt=none` asks the IdP to "silently" process the authorization request without displaying any consent page nor redirecting the user to a login page. If it can't do that, it must return an error to the client application indicating why the request failed.

It would be a terrible regression: folks who are not already logged in with the IdP or haven't granted your ASP.NET Core app an access to their resources wouldn't be able to successfully log in.

gb056:
Of course, sorry clearly didn't read the spec properly there. Assumed `none` was the default without really thinking. Apologies, will fix now.

poke:
I just noticed that this functionality is coming in 2.1. Is there any way we could extend this to allow overwriting the prompt value when challenging the authentication?

I image that we could simply look for an item in the `AuthenticationProperties` and use that if there’s one. Something like this:

```
Prompt = properties.Items.TryGetValue("prompt", out var prompt) ? prompt : Options.Prompt,
```

Then, one could explicitly require a login prompt like this:

```
var props = new AuthenticationProperties();
props.Items.Add("prompt", "login");
await HttpContext.ChallengeAsync(OpenIdConnectDefaults.AuthenticationScheme, props);
```

Tratcher:
@poke see https://github.com/aspnet/Security/issues/990 for a related discussion

poke:
@Tratcher Thanks, that’s interesting. But I’m not sure if a solution that’s being created for that will have an impact on OpenIdConnect. The OIDC handler is not built on top of the OAuth handler, so we would need a separate implementation anway. And since OIDC is limited by the capabilities of [`OpenIdConnectMessage`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/master/src/Microsoft.IdentityModel.Protocols.OpenIdConnect/OpenIdConnectMessage.cs), a dynamic way to build the query arguments on this side will probably not work (unless we also adjust the upstream message object).

I would happily provide a pull request for something like I’ve shown above, if there’s a chance for that being accepted. Ideally to get this done still in time for 2.1.0 (if that’s even possible).

Tratcher:
I referenced it as conceptually similar, the Google Auth handler has been doing dynamic parameter overrides for a long time and we've considered expanding that to the other handlers.

Try a PR and we'll see how the pattern works out.

PR #1403, opened 2017-09-01
Minor improvements

agoretsky:


src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -246,7 +245,7 @@ private CookieOptions BuildCookieOptions()
             _signInCalled = true;
 
             // Process the request cookie to initialize members like _sessionKey.
-            var result = await EnsureCookieTicket();
+            await EnsureCookieTicket();
```
Tratcher: does this cause compiler warnings about ignoring return values?

agoretsky: I can't see these warnings in console after `dotnet build` and also there are no warnings in VS Error List

Eilon:
This looks good overall. @Tratcher can you complete your review & merge when it's ready?

PR #1410, opened 2017-09-06
[Prototype] AuthNPolicy[Provider] / AuthorizationRequest[Evaluator]

HaoK:
Rough prototype of some of what we were discussing for Auth 2.1 changes to further discussion:

- `AuthenticationPolicy`: gives us a named set of schemes, and the ability to resolve these on a per request basis (PolicyProvider)
- `AuthorizationRequest`: is the new unit of work for AuthZ, it contains the authentication policy name, the authorization requirements, and the resource being authorized.  This is the complete set of data needed to process an authorization
- `IAuthorizationRequestEvaluator` builds on top of both the new 2.0 PolicyEvaluator and the AuthenticationPolicyProvider, it has a single method:
`PolicyResult AuthorizeAsync(AuthorizationRequest, HttpContext)` and returns Success/Forbidden/Challenged 

The new authorization metadata would just need to produce the AuthorizationRequest

I focused mostly on the AuthZ side of things first, exposing the authNPolicy concept for authN will come later

@davidfowl @blowdart @Tratcher @rynowak @Eilon @ajcvickers


src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationRequestEvaluator.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            var result = await _authorization.AuthorizeAsync(principal, request.Resource, policy.Build());
+            if (result.Succeeded)
+            {
+                return PolicyAuthorizationResult.Success();
+            }
+
+            // If authentication was successful, return forbidden, otherwise challenge
+            return (authenticateSuccess) 
+                ? PolicyAuthorizationResult.Forbid()
```
HaoK: Should we just directly forbid/challenge using the authentication scheme instead of returning a result for the callee to take care of? (i.e. Mvc Challenge/ForbidResults)

src/Microsoft.AspNetCore.Authorization.Policy/AuthenticationPolicy.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        /// <summary>
+        /// The authentication schemes that should be used to construct the ClaimsPrincipal during authentication.
+        /// </summary>
+        public IEnumerable<string> AuthenticateSchemes { get; }
+
+        /// <summary>
+        /// The authentication schemes that should be challenged. 
+        /// </summary>
+        public IEnumerable<string> ChallengeSchemes { get; }
```
davidfowl: Should be one, no? Many challenge schemes is odd.

src/Microsoft.AspNetCore.Authorization.Policy/AuthenticationPolicy.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        /// <summary>
+        /// The authentication schemes that should be forbidden.
+        /// </summary>
+        public IEnumerable<string> ForbidSchemes { get; }
```
davidfowl: Same comment as challenge schemes

src/Microsoft.AspNetCore.Authorization.Policy/AuthenticationPolicyEvaluator.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// </summary>
+        /// <param name="context">The <see cref="HttpContext"/>.</param>
+        /// <param name="policyName">The name of the authentication policy.</param>
+        /// <returns>The result.</returns>
+        public virtual async Task<AuthenticateResult> AuthenticateAsync(HttpContext context, string policyName)
+        {
+            var policy = await Policies.GetAsync(context, policyName);
+            if (policy != null)
+            {
+                var count = policy.AuthenticateSchemes.Count();
```
davidfowl: Any

src/Microsoft.AspNetCore.Authorization.Policy/AuthenticationPolicyEvaluator.cs
```diff
@@ -0,0 +1,124 @@
...
+                        var auth = await context.AuthenticateAsync(s);
+                        if (auth.Succeeded)
+                        {
+                            if (principal == null)
+                            {
+                                principal = auth.Principal;
+                            }
+                            else
+                            {
+                                // REVIEW: what about the auth properties, we just dropped them...
```
davidfowl: How did this work before? Did it ever?

src/Microsoft.AspNetCore.Authorization.Policy/AuthenticationPolicyEvaluator.cs
```diff
@@ -0,0 +1,124 @@
...
+            var policy = await Policies.GetAsync(context, policyName);
+            if (policy != null)
+            {
+                var count = policy.AuthenticateSchemes.Count();
+                if (count > 0)
+                {
+                    ClaimsPrincipal principal = null;
+                    foreach (var s in policy.AuthenticateSchemes)
+                    {
+                        var auth = await context.AuthenticateAsync(s);
```
davidfowl: Isn't it cleaner to pass the IAuthenticationService to the ctor of this service then call `AuthenticateAsync` on that? Those extension methods are just for convenience...

davidfowl:
@HaoK can you outline which components use which APIs? Would the Authentication middleware be updated to use the `AuthenticationPolicyEvaluator`?

src/Microsoft.AspNetCore.Authorization.Policy/AuthenticationPolicyEvaluator.cs
HaoK: Nope, never worked, whenever we merged principals we would lose whatever was in the auth properties, but to be fair, you couldn't access the properties in 1.x unless you called the appropriate Authenticate overload anyways

src/Microsoft.AspNetCore.Authorization.Policy/AuthenticationPolicy.cs
HaoK: Yeah its odd, I believe one of the original reasons why we ever supported this was an upgrade flow, where a user would be signed in, but they might require more permissions which would result in additional challenge, @Tratcher @blowdart do you guys remember the details of why we ever wanted multiple authentication schemes for authorization (other than bearer + cookies)

src/Microsoft.AspNetCore.Authorization.Policy/AuthenticationPolicyEvaluator.cs
HaoK: Yeah I wasn't sure which was better, I went with the context extensions since all the sugar overloads only exist on httpContext...

HaoK:
Yeah we should consider switching all of the public authentication entrypoints to expose the policy concept.

So
- AuthenticationOptions would get a DefaultPolicy property which could be configured
- The middleware from UseAuthentication() would switch to using `IAuthenticationPolicyEvaluator.AuthenticateAsync(httpContext, policyName: null)` which would fall back to the current `context.AuthenticateAsync` behavior if no policies are configured.
- The authorization stack would switch to a new layer built on top of the old authZ layer (`IPolicyEvaluator`) + named authN policies (`IAuthenticationPolicyEvaluator`) = `IAuthorizationRequestEvaluator`)

I'm not sure the ability to name specific other non default authentication policies would be used outside of Authorization, but regardless it would allow dynamic per request resolution of the authentication policies via `IAuthenticationPolicyProvider`

HaoK:
And some kind of new attribute syntax like:

[Authorize(AuthenticationPolicy = "Cookies", AuthorizationPolicy = "Organization", Resource = Operation.Something)]

src/Microsoft.AspNetCore.Authorization.Policy/AuthenticationPolicy.cs
blowdart: Yea, the upgrade was definitely a scenario

blowdart:
Don't forget the imperative calls as well, not just the attribute.

src/Microsoft.AspNetCore.Authorization.Policy/AuthenticationPolicy.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        // REVIEW: Do we really need SignIn/SignOut schemes?
```
Tratcher: We need some way for remote providers to resolve their correct sign-in scheme. Or we make them specify it in their options, but it seems odd to leave out.

src/Microsoft.AspNetCore.Authorization.Policy/AuthenticationPolicy.cs
Tratcher: Challenging Basic, Bearer, Windows, on the same response works, they stack. That's what IIS does if you enable multiple auth types.

src/Microsoft.AspNetCore.Authorization.Policy/AuthenticationPolicy.cs
Tratcher: Yeah, this one is pretty strange. In theory you would forbid each scheme the user was signed in with, but I don't know of any implementations that stack here. It's somewhat unusual to sign in with multiple anyways.

HaoK: So we allow multiple authenticate/challenge and force them to pick a single forbid scheme?

src/Microsoft.AspNetCore.Authorization.Policy/AuthenticationPolicy.cs
HaoK: Remote providers don't need to change from what they are doing today, they require a sign in scheme, it will use the default if not specified, we could have that use the default policy if specified as a small change, but I don't think adding a SignInPolicy really helps since its only using the SignInScheme

HaoK:
@Tratcher consumption is via IAuthenticationPolicyEvaluator, you call Authenticate/Challenge/Forbid with the policy name.  That's what AuthZ will use as well.  I don't really think SignIn/SignOut make sense unless they supports multiple schemes as it would just be an exact call to the underlying IAuthenticationService, we can keep it as just passthrough if we really want, but doesn't seem that useful

HaoK:
Specifically picking between cookies/bearer isn't all that different, you can do it on [Authorize] explicitly via AuthenticationPolicy, or you can do it dynamically by using a single "BlowdartHatesThis" policy where `IAuthenticationPolicyProvider` resolves it to cookies/bearer based on something that would never be request path...

HaoK:
To summarize the main benefits of this change: you get the ability to associate a set of named challenge/forbid/authenticate schemes (including a default).  And the ability to dynamically resolve these named schemes on a per request basis.  This will allow us to tidy things up more and finally define some of the undefined behaviors with multiple authentication schemes in Authorize

HaoK:
@blowdart imperative calls will basically just be getting the appropriate service and calling it, with the lowest level something like:

`AuthorizeAsync(authenticationPolicy, authorizationRequirements, resource)` which returns Success/Forbidden(failure when authentication succeeded)/Challenge(failure when authentication failed)

Tratcher:
Specifically, can you show an example of defining two policies and a selector in Startup for the JWT vs Cookies scenario.

src/Microsoft.AspNetCore.Authorization.Policy/AuthenticationPolicy.cs
davidfowl: I know what it does but it feels like a pit of failure. How often is this usable?

HaoK:
Sure, it would look something like:

```C#
services.AddAuthentication(o => {
    o.AddPolicy("Cookies", p => p.UseDefault("Cookies"))); // sets all the schemes to cookies
    o.AddPolicy("Bearer", p => p.UseDefault("Bearer"))); // sets all the schemes to bearer
});
services.AddScoped<IAuthenticationPolicyProvider, MyPolicyProvider>();

class MyPolicyProvider : IAuthenticationPolicyProvider {
    AuthenticationOptions _options;
    
     MyPolicyProvider(IOptions<AuthenticationOptions> options) => _options = options.Value;

    public Task<AuthenticationPolicy> GetAsync(string name, HttpContext context) {
        // If no policy is specified, pick one
        if (name == null) {
            if (ShouldUseBearer(context)) return _options.Policies["Bearer"];
            else return _options.Policies["Cookie"];
        }
        return _options.Policies[name];
    }
}
```

src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationRequestEvaluator.cs
davidfowl: Maybe this should do it. The reason to allow both though is because you can inspect action results in the mvc pipeline.

HaoK: So this will take care of calling Forbid/Challenge but we'd still continue to return the results but we'd change MVC to no-op with the Challenge/Forbid results?

HaoK: That feels a bit weird to no-op, and probably a breaking change too, so would we have to introduce new results that no-op'd but signal'd that a challenge/forbid was done?

Tratcher:
The layering looks a little odd. I was expecting a policy name selector more like: `public Task<string> GetAsync(HttpContext context)`
And then a services or auth builder extension that reduces that to a lambda:
`services.AddAuthentication(...).AddPolicySelector((context => ShouldUseBearer(context) ? "Bearer" : "Cookies")

HaoK:
Sure, we can always add something to make selecting only the default easier

`AddDefaultPolicySelector(context =>  ShouldUseBearer(context) ? "Bearer" : "Cookies")`

and there'd be a matching IDefaultAuthenticationPolicySelector

HaoK:
Example usage:

```
               services.Configure<AuthenticationPolicyOptions>(o => 
                {   
                    o.DefaultPolicySelector = ctx => (string)ctx.Items["policy"];
                    o.AddPolicy("auth", p => p.DefaultScheme = "default");
                });
 ```

File: src/Microsoft.AspNetCore.Authorization.Policy/AuthenticationPolicy.cs (hunk 2 above)
Comment (HaoK) [reply src/Microsoft.AspNetCore.Authorization.Policy/AuthenticationPolicy.cs]: What's the verdict here @blowdart @Tratcher @davidfowl 

Can we limit Challenge/Forbid to a single scheme for an AuthenticationPolicy?  Multiples would still be supported on the APIs I guess.

**Comment by HaoK** [comment]
Are we happy with the current names (and planned homes) for this stuff?

New to `Microsoft.AspNet.Core.Authentication[.Abstractions]`:
`AuthenticationPolicy` - Authenticate/Challenge/Forbid schemes only (no sign in/sign out)
`AuthenticationPolicyBuilder` - Sugar to build policies
`[I]AuthenticationPolicyEvaluator` - used Authenticate/Challenge/Forbid via policy name (used instead of directly using `IAuthenticationService`)
`[I]AuthenticationPolicyProvider` - Per request resolution of name => AuthenticationPolicy
`[I]DefaultAuthenticationPolicySelector` - Per request resolution of default authentication policy

New to `Microsoft.AspNet.Core.Authorization.Policy`: (is this the right home?  we added it in 2.0 to link Authentication to Authorization):
`AuthorizationRequest` - Encapsulates the AuthZ request: authN policy name, resource, requirements
`[I]AuthorizationRequestEvaluator` - Evaluates the new `AuthorizationRequest` (replaces IAuthorizationPolicyEvaluator which itself replaces `IAuthorizationService`)

**Comment by HaoK** [comment]
So I've started work in https://github.com/aspnet/Identity/pull/1442 trying to switch Identity's SignInManager over to using authentication policies.  

Initial thoughts are:
There will be a "Identity" AuthenticationPolicy that is setup by default to:
`Authenticate/Forbid/ChallengeScheme`: "Identity.Application"

// There is no support for SignIn/SignOut right now, as they are kinda confusing
`SignInScheme`: We set this to "Identity.External" today but that is confusing since you'd expect SignIn to set the Application cookie, as we set this mostly for the external auth policies.
`SignOutScheme`: Identity needs to clear Application, External, and 2 two factor schemes

So I guess what we could do now is setup a new default "External" policy, which is what we use to set the SignInScheme to "Identity.External" in the templates for social logins, and I guess we'd just have to update the documentation since I don't think we can just make it all work with this, the external providers need to specify their authenticationPolicy somewhere, since there will be always be at least two authentication policies (Identity + External grrr)

**Comment by HaoK** [comment]
Or maybe we should just add additional new dedicated setting for just the remote authentication SignIn default, which is separate from the DefaultSignInScheme, i.e.

```
services.AddAuthentication(o => {
    o.DefaultPolicy = "Identity.Application";
    o.DefaultRemoteSignInScheme = "Identity.External";

// Google Handler resolves the default SignInScheme by trying to use DefaultRemoteSignInScheme first, before falling back to the existing logic
}).AddGoogle(); 

```

**Comment by Tratcher** [comment]
No for DefaultRemoteSignInScheme, DefaultSignInScheme is primarily used for remotes anyways.

**Comment by HaoK** [comment]
So if we don't add a new dedicated setting,I guess everything still hangs together for the template if we set the default AuthenticationPolicy to "Identity", then the remote auth handlers will resolve to the External cookie, and SignInManager never uses the default sign in scheme..  Its only users who call SignInAsync not specifying a scheme who will be surprised which hopefully won't be a huge number of people)

**Comment by HaoK** [comment]
So tentative startup will look something like:

```C#
    services.AddIdentityCore<ApplicationUser>() // + UserManager 
                .AddRoles<IdentityRole>() // + RoleManager
                .AddEntityFrameworkStores(); // + User/RoleStores
                .AddSignIn() // + SignInManager
                .AddDefaultCookies() // + Identity cookies, configures "Identity" auth policy
                .AddDefaultTokenProviders() // TFA stuff

    services.AddAuthentication(o.DefaultPolicy = IdentityConstants.DefaultPolicy)
            .AddGoogle()
            .AddJwtBearer() ...
```

src/Microsoft.AspNetCore.Authorization.Policy/AuthenticationPolicy.cs
HaoK: Ok so if we are going to switch the Remote providers to use the new `IAuthentication,PolicyEvaluator` to SignIn, then we'll need to add a SignInPolicy setting which I guess would be used if SignInScheme isn't specified?  Gets confusing fast :)

davidfowl:
Auth's relationship with funcs needs to go away. We need to use types like sane people 😄 . They are more DI friendly.

HaoK:
Went with virtual scheme instead: https://github.com/aspnet/Security/pull/1486

Issue #1378, opened 2017-08-20
Should RemoteAuthenticationOptions validate SignInScheme != null?


HaoK:
This should validation was lost at some point, SignInScheme is required for remote auth.

HaoK:
Or maybe we removed this validation to enable calling the default sign in scheme...

davidfowl:
Can we also validate and SignInScheme != Scheme, are the schemes stored on the instances of the handlers themselves?

HaoK:
We will likely have to validate that outside of the normal validation (not AuthenticationSchemeOptions.Validate, but inside of the base RemoteAuthenticationHandler itself to have access to the scheme name, and we'd likely have to also explicitly check for when its the default that fallback to itself as well.

PinpointTownes:
> Can we also validate and SignInScheme != Scheme, are the schemes stored on the instances of the handlers themselves?

Exactly what I had recommended here: https://github.com/aspnet/Security/issues/1264#issuecomment-309100351

> We will likely have to validate that outside of the normal validation

You can easily do that in `EnsureSignInScheme` as you have access to both names: https://github.com/aspnet/Security/blob/488eb44467eb677eab62bdc49aa6255cc1be3119/src/Microsoft.AspNetCore.Authentication/AuthenticationServiceCollectionExtensions.cs#L98-L101

Tratcher:
Related issue to address:
https://github.com/aspnet/Security/blob/488eb44467eb677eab62bdc49aa6255cc1be3119/src/Microsoft.AspNetCore.Authentication/AuthenticationServiceCollectionExtensions.cs#L98-L101
There should be an additional `?? _authOptions.DefaultScheme;` for when DefaultSignInScheme isn't set. That's what was starting some of the stack overlfows by passing nulls.

HaoK:
b9d9418f6d4fecd8e3b0c60538bb985c15e31fcc

PR #1422, opened 2017-09-13
Block SignInScheme = self

HaoK:
Fixes https://github.com/aspnet/Security/issues/1378

@Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
+++ b/src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
@@ -97,6 +97,10 @@ namespace Microsoft.AspNetCore.Authentication
             public void PostConfigure(string name, TOptions options)
             {
                 options.SignInScheme = options.SignInScheme ?? _authOptions.DefaultSignInScheme;
+                if (options.SignInScheme == name)
+                {
+                    throw new InvalidOperationException(Resources.Exception_RemoteSignInSchemeCannotBeSelf);
+                }
             }
         }
     }
--- a/src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
+++ b/src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
@@ -84,7 +84,7 @@ namespace Microsoft.AspNetCore.Authentication
             return AddScheme<TOptions, THandler>(authenticationScheme, displayName, configureOptions: configureOptions);
         }
 
-        // Used to ensure that there's always a default data protection provider
+        // Used to ensure that there's always a default sign in scheme that's not itself
         private class EnsureSignInScheme<TOptions> : IPostConfigureOptions<TOptions> where TOptions : RemoteAuthenticationOptions
         {
             private readonly AuthenticationOptions _authOptions;
--- a/src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
+++ b/src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
@@ -97,7 +97,7 @@ namespace Microsoft.AspNetCore.Authentication
             public void PostConfigure(string name, TOptions options)
             {
                 options.SignInScheme = options.SignInScheme ?? _authOptions.DefaultSignInScheme ?? _authOptions.DefaultScheme;
-                if (options.SignInScheme == name)
+                if (string.Equals(options.SignInScheme, name, StringComparison.Ordinal))
                 {
                     throw new InvalidOperationException(Resources.Exception_RemoteSignInSchemeCannotBeSelf);
                 }
--- a/src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
+++ b/src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
@@ -96,7 +96,7 @@ namespace Microsoft.AspNetCore.Authentication
 
             public void PostConfigure(string name, TOptions options)
             {
-                options.SignInScheme = options.SignInScheme ?? _authOptions.DefaultSignInScheme;
+                options.SignInScheme = options.SignInScheme ?? _authOptions.DefaultSignInScheme ?? _authOptions.DefaultScheme;
                 if (options.SignInScheme == name)
                 {
                     throw new InvalidOperationException(Resources.Exception_RemoteSignInSchemeCannotBeSelf);
--- a/src/Microsoft.AspNetCore.Authentication/Resources.resx
+++ b/src/Microsoft.AspNetCore.Authentication/Resources.resx
@@ -130,6 +130,6 @@
     <value>The '{0}' option must be provided.</value>
   </data>
   <data name="Exception_RemoteSignInSchemeCannotBeSelf" xml:space="preserve">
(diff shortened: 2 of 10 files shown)


src/Microsoft.AspNetCore.Authentication.Facebook/Properties/Resources.Designer.cs
```diff
@@ -15,16 +15,14 @@ internal static class Resources
         /// </summary>
         internal static string Exception_OptionMustBeProvided
         {
-            get { return GetString("Exception_OptionMustBeProvided"); }
+            get => GetString("Exception_OptionMustBeProvided");
```
Tratcher: revert? It's kindof weird to include these with your change. Especially when we want to backport it.

src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
```diff
@@ -97,6 +97,10 @@ public EnsureSignInScheme(IOptions<AuthenticationOptions> authOptions)
             public void PostConfigure(string name, TOptions options)
             {
                 options.SignInScheme = options.SignInScheme ?? _authOptions.DefaultSignInScheme;
+                if (options.SignInScheme == name)
+                {
+                    throw new InvalidOperationException(Resources.Exception_RemoteSignInSchemeCannotBeSelf);
```
HaoK: Note: this is technically in the wrong place, it SHOULD be in RemoteAuthenticationOptions.Validate....but we don't pass in the authenticationScheme name we are validating :(  whoops...   So this is the next best spot, its almost as good (but will block anyone from setting sign in scheme via a later PostConfigure)

src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
```diff
@@ -97,6 +97,10 @@ public EnsureSignInScheme(IOptions<AuthenticationOptions> authOptions)
             public void PostConfigure(string name, TOptions options)
             {
                 options.SignInScheme = options.SignInScheme ?? _authOptions.DefaultSignInScheme;
```
Tratcher: There should be an additional  `?? _authOptions.DefaultScheme;` for when DefaultSignInScheme isn't set. That's what was starting some of the stack overlfows.

src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
```diff
@@ -97,6 +97,10 @@ public EnsureSignInScheme(IOptions<AuthenticationOptions> authOptions)
...
+                if (options.SignInScheme == name)
```
Tratcher: string.equals with a specific comparer (ordinal?)

test/Microsoft.AspNetCore.Authentication.Test/TwitterTests.cs
```diff
@@ -18,6 +18,19 @@ namespace Microsoft.AspNetCore.Authentication.Twitter
     public class TwitterTests
     {
         [Fact]
+        public async Task VerifySignInSchemeCannotBeSetToSelf()
+        {
+            var server = CreateServer(o =>
+            {
+                o.ConsumerKey = "Test Consumer Key";
+                o.ConsumerSecret = "Test Consumer Secret";
+                o.SignInScheme = TwitterDefaults.AuthenticationScheme;
```
Tratcher: Any tests for users specifying DefaultSignInScheme wrong rather than the direct option?

src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
HaoK: Oops this is actually another bug, this should actually be using IAuthenticationSchemeProvider.DefaultSignInScheme I'll fix this now too

src/Microsoft.AspNetCore.Authentication.Facebook/Properties/Resources.Designer.cs
HaoK: I guess, we should fix this repo to be clean when you run ./build.cmd /t:resx which is what I did to generate the new resource

We aren't planning on porting this one back are we?  The other fix is important, this one just results in a better error message

src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
HaoK: Hrm so that's actually an interesting question, we don't define the mixed casing behavior anywhere today.  Options just flows the name as is, we use constants mostly so this behavior actually isn't clearly defined today.  I guess we probably should be case insensitive for the purposes of auth schemes?

Tratcher: I think it has always been explicitly case sensitive.

src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
Tratcher: The DefaultScheme bit needs to get backported, that's an otherwise valid configuration that currently fails.

HaoK: Fair enough so the real important fix out of all of this is for the PostConfigure to switch to using the SchemeProvider

HaoK: Ugh options being sync only bites again, so we can't cleanly use the provider since that's async, I guess we can just do the ghetto fallback using only the options

test/Microsoft.AspNetCore.Authentication.Test/TwitterTests.cs
HaoK: I added 2 variants to the FB tests to verify the default sign in / default scheme flow, facebook's CreateServer was the ugliest so it exposed the proper overloads needed :)

src/Microsoft.AspNetCore.Authentication.Facebook/Properties/Resources.Designer.cs
HaoK: I pulled the other resx changes into: https://github.com/aspnet/Security/pull/1424

HaoK: I'll get that in first, and then rebase so these resx diffs will go away

src/Microsoft.AspNetCore.Authentication/Resources.resx
```diff
@@ -129,4 +129,7 @@
   <data name="Exception_OptionMustBeProvided" xml:space="preserve">
     <value>The '{0}' option must be provided.</value>
   </data>
+  <data name="Exception_RemoteSignInSchemeCannotBeSelf" xml:space="preserve">
+    <value>The SignInScheme for a remote authentication handler cannot be set to itself.  If it was not explicitly set, the AuthenticationOptions.DefaultSignInScheme is used.</value>
```
Tratcher: Or DefaultScheme

Tratcher:
Also open a backport to 2.0.x bug for the DefaultScheme fallback, if not the whole PR.

Tratcher:
Filed https://github.com/aspnet/Security/issues/1435 for backport.

HaoK:
b9d9418f6d4fecd8e3b0c60538bb985c15e31fcc

Issue #1147, opened 2017-03-13
Consider removing of "profile" scope as a default in OpenIdConnectOptions


BorisWilhelms:
The OIDC Core Specc states the "profile" scope as an optional scope (http://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims). Therefor it should not be added by default to the scope hashset of the OpenIdconnectOptions. 
The specc only states only the "openid" scope as a required scope.

leastprivilege:
Yes. Absolutely agreed.

Eilon:
The Great Auth Redesign of 2017 should make it easier to remove/control the scopes, but we are not planning on changing the default list of scopes because we've seen from customer feedback that it's very commonly used.

PR #1429, opened 2017-09-14
Remove the profile scope

leastprivilege:
Remove the profile - because it is optional. If someone wants profile, it should be added explicitly. For most scenarios, profile is not appropriate.


PinpointTownes:
Correct me if I'm wrong, but I'm pretty sure @Eilon said a few months ago that this was not a change they were willing to make. Has something changed?

(it's worth noting that this is a functional breaking change... it couldn't be merged before 3.x)

PinpointTownes:
> The Great Auth Redesign of 2017 should make it easier to remove/control the scopes, but we are not planning on changing the default list of scopes because we've seen from customer feedback that it's very commonly used.

https://github.com/aspnet/Security/issues/1147

Eilon:
Indeed, we still think that for most users this is the ideal behavior. We try to make the defaults cover the common cases.

For that reason, I'm closing this PR.

leastprivilege:
@PinpointTownes just waned to make sure you are still here. You were so quiet recently.

Issue #1218, opened 2017-05-12
Consider making Auth handlers public


HaoK:
They are internal today (prior to 2.0 they weren't really easy to extend/reuse).  There's no reason they can't be public today.  Before events used to be the sole extensibility points, but handlers should now be able to be extended/replaced.

Eilon:
Let's wait until 2.1 so that we can come up with the right API surface for these.

brockallen:
@HaoK I don't recall if overriding the handler is how you suggested IdentityServer can inject the claims it needs to or not... but if it is, then delaying this to 2.1 causes an issue for us with IdentityServer.

HaoK:
Can you point me to the relevant thread for that if we had one?

brockallen:
Yea... might have been email... @leastprivilege might have an easier time of finding it. We'll look and get back. Thx.

HaoK:
Sounds like you should be good, decorating the default AuthenticationService implementation to validate/add your own claims for every call to SignIn.  That's not affected this issue of not exposing our OAuth handler implementations right now.

brockallen:
okies, thx

Eilon:
Closing because we have no plans to do this. Many extensibility points are available via existing events.

PR #1430, opened 2017-09-14
Make all handlers public

HaoK:
Easy half of https://github.com/aspnet/Security/issues/1218

Most are public already, so should be harmless being consistent with these.

Extensibility, aka 

`MyGoogle<MyGoogleOptions> : GoogleHandler` is something we can look at later

@Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Facebook/FacebookHandler.cs
@@ -16,7 +16,7 @@ using Newtonsoft.Json.Linq;
 
 namespace Microsoft.AspNetCore.Authentication.Facebook
 {
-    internal class FacebookHandler : OAuthHandler<FacebookOptions>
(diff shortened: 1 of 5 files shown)


Tratcher:
What's the other half?

We should go over the methods and see which ones may need to be overridden as well. I see several private methods in cookie auth that would be useful. Not to mention adding a sample where we derive from some handlers and see if we can make meaningful changes. No point in making them derivable if you can't actually do anything with them.

Tratcher:
e.g.
Facebook GenerateAppSecretProof
Google AddQueryString
JwtBearer CreateErrorDescription
OIDC _configuration, SaveTokens, Read/WriteNonceCookie, GetPropertiesFromState, ValidateToken
Twitter ObtainRequestTokenAsync, OptainAccessTokenAsync, RetrieveUserDetailsAsync

etc...

HaoK:
Yes the other half is most likely making all of the private methods protected virtual, and also probably adding `<TOptions> as a base class to all of these so people can extend adding their own options properties similar to the existing base handlers

Issue #43, opened 2014-08-21
Port the WsFederation middleware from Katana


Tratcher:
Depends on System.IdentityModel.Tokens.Jwt and Microsoft.IdentityModel.Protocol.Extensions.

xytyx:
Any news on WsFederation in ASP.NET 5 ?

blowdart:
The dependency still stands and that work has not yet been completed by the team that owns the IdentityModel pieces.

brockallen:
Any plans to add signoutcleanup support? That was missing in Katata v3's implementation. It's nice to have single sign-on, but single sign-out is also nice :)

brentschmaltz:
signoutcleanup is on the map.

WsFed in on the map also, no time frame yet.

guigobavaresco:
Hi, 
how about it, now?

blowdart:
We're still awaiting support from the WAAD team, they're the ones driving it.

brentschmaltz:
Yep, it is on our plates and radar. No promises till we drop OIDC.

lomithrani:
Is there any alternative to authentify using wsfederation or salm2 meanwhile ?

aredfox:
Is there any news on the feature / roadmap for this feature, as this is now holding back starting new applications in ASPNET5 for our organisation.

marc-mueller:
Is there any update so far? Not being able to support SAML within our ASP.NET 5 application is currently a showstopper for our project since we have some dependencies with existing systems.

Tratcher:
This is not happening for v1.0. We'll follow up afterwards.

marc-mueller:
So this means that the development has not started yet? Is there any estimation about the release?

Tratcher:
Correct. No, we haven't scheduled any of the post-v1 work yet.

rschiefer:
We use ADFS extensively for 20 or so internal web applications.  This would be a major blocker for us to migrate to ASP.NET Core.

Can someone post a link to the related dependencies so we can go show support for that work as well?

helmsb:
We've used ADFS as the core of our authentication for our internal application framework which is used throughout our organization. This is a huge blocker for us going to .NET Core.

Tratcher:
@blowdart @brentschmaltz is it this one? https://github.com/dotnet/corefx/issues/4278

MaximRouiller:
Just bouncing this again. I'm still seeing people trying the Katana bits with .NET Core. 

[...]

PR #1441, opened 2017-09-19
WsFederation review

Tratcher:
This PR is for review purposes only, we still need to work out the final branching structure.

This is a port of Microsoft.Owin.Security.WsFederation. It targets Core 2.0 dependencies so that we can release it sooner than 2.1. As such some of the project files are unusual. We're also using the IdentityModel 5.2.0-preview1 dependencies from nuget.org. The version for this package will start as 2.0.0-alpha1.

The behavior is approximately the same as the Katana version, with minor updates to mirror those we made to OIDC when it was ported over.  Changes:
- User state in the wctx is nested rather than appended to.
- CallbackPath is required and defaults to "/signin-wsfed". However it can be set to "/" along with `options.SkipUnrecognizedRequests = true;` to get some of the stranger legacy behavior apps have asked for in the past. OIDC uses the same mechanic.
- Better resolving of relative to absolute redirect urls.


Tratcher:
@brentschmaltz @leastprivilege @brockallen

Tratcher:
#43

Tratcher:
We have a build plan, @JunTaoLuo will be in charge of that. We'll start as soon as we get code sign-off from @HaoK.

samples/WsFedSample/Startup.cs
```diff
@@ -0,0 +1,169 @@
...
+
+        public IConfiguration Configuration { get; }
+
+        // This method gets called by the runtime. Use this method to add services to the container.
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddAuthentication(sharedOptions =>
+            {
+                sharedOptions.DefaultScheme = CookieAuthenticationDefaults.AuthenticationScheme;
+                sharedOptions.DefaultSignInScheme = CookieAuthenticationDefaults.AuthenticationScheme;
```
HaoK: Nit: doesn't DefaultSignInScheme fallback to DefaultScheme?

samples/WsFedSample/Startup.cs
```diff
@@ -0,0 +1,169 @@
...
+                    {
+                        await res.WriteAsync($"<h1>You have been signed out.</h1>");
+                        await res.WriteAsync("<a class=\"btn btn-link\" href=\"/\">Sign In</a>");
+                    });
+                    return;
+                }
+
+                if (context.Request.Path.Equals("/signout"))
+                {
+                    await context.SignOutAsync(CookieAuthenticationDefaults.AuthenticationScheme);
```
HaoK: Should this just use the empty overload since you are setting the defaults?

samples/WsFedSample/Startup.cs
Tratcher: Not correctly in the 2.0.0 branch we're targeting. That was just fixed in 2.1 and 2.0.x.

samples/WsFedSample/Startup.cs
Tratcher: We could. This just mirrors the template code.

samples/WsFedSample/Startup.cs
```diff
@@ -0,0 +1,169 @@
...
+                    await context.SignOutAsync(WsFederationDefaults.AuthenticationScheme, new AuthenticationProperties()
+                    {
+                        RedirectUri = "/signedout"
+                    });
+                    return;
+                }
+
+                if (context.Request.Path.Equals("/Account/AccessDenied"))
+                {
+                    await context.SignOutAsync(CookieAuthenticationDefaults.AuthenticationScheme);
```
HaoK: Why does AccessDenied sign out?

src/Microsoft.AspNetCore.Authentication.WsFederation/Events/AuthenticationFailedContext.cs
```diff
@@ -0,0 +1,35 @@
...
+using System;
+using Microsoft.AspNetCore.Http;
+using Microsoft.IdentityModel.Protocols.WsFederation;
+
+namespace Microsoft.AspNetCore.Authentication.WsFederation
+{
+    /// <summary>
+    /// The context object used in for <see cref="WsFederationEvents.AuthenticationFailed"/>.
+    /// </summary>
+    public class AuthenticationFailedContext : RemoteAuthenticationContext<WsFederationOptions>
```
HaoK: Nit: should this have a more specific name, i.e. WsFedFailedContext?

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
```diff
@@ -0,0 +1,347 @@
...
+                    return HandleRequestResult.SkipHandler();
+                }
+
+                return HandleRequestResult.Fail("No message.");
+            }
+            
+            try
+            {
+                // Retrieve our cached redirect uri
+                string state = wsFederationMessage.Wctx;
```
HaoK: Nit: var?

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
```diff
@@ -0,0 +1,347 @@
...
+                var properties = Options.StateDataFormat.Unprotect(state);
+
+                if (properties == null)
+                {
+                    properties = new AuthenticationProperties();
+                }
+                else
+                {
+                    // Extract the user state from properties and reset.
+                    properties.Items.TryGetValue(WsFederationDefaults.UserstatePropertiesKey, out var userState);
```
HaoK: Should this do something if TryGetvalue fails?  Or is that ok to just ignore?

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
```diff
@@ -0,0 +1,347 @@
...
+                await Options.Events.MessageReceived(messageReceivedContext);
+                if (messageReceivedContext.Result != null)
+                {
+                    return messageReceivedContext.Result;
+                }
+
+                if (wsFederationMessage.Wresult == null)
+                {
+                    Logger.LogWarning("Received a sign-in message without a WResult.");
+                    return (HandleRequestResult)HandleRequestResult.NoResult();
```
HaoK: Do we need the cast?

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
```diff
@@ -0,0 +1,347 @@
...
+                }
+
+                if (_configuration == null)
+                {
+                    _configuration = await Options.ConfigurationManager.GetConfigurationAsync(Context.RequestAborted);
+                }
+
+                // Copy and augment to avoid cross request race conditions for updated configurations.
+                var tvp = Options.TokenValidationParameters.Clone();
+                IEnumerable<string> issuers = new[] { _configuration.Issuer };
```
HaoK: nit: var?

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
```diff
@@ -0,0 +1,347 @@
...
+                SecurityToken parsedToken = null;
+                foreach (var validator in Options.SecurityTokenHandlers)
+                {
+                    if (validator.CanReadToken(token))
+                    {
+                        principal = validator.ValidateToken(token, tvp, out parsedToken);
+                        break;
+                    }
+                }
+
```
HaoK: Is the validator contract that if principal != null, parsedToken will be non null?  Should this check parsedToken as well?

src/Microsoft.AspNetCore.Authentication.WsFederation/Events/AuthenticationFailedContext.cs
Tratcher: *shrug* I don't see why it needs one. This type will rarely be referenced by name, and if you really need to and have a conflict you can do WsFederation.AuthenticationFailedContext.

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
Tratcher: No, it's optional.

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
Tratcher: Yes, NoResult returns the base type AuthenticateResult.

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationOptions.cs
```diff
@@ -0,0 +1,170 @@
...
+        public new WsFederationEvents Events
+        {
+            get => (WsFederationEvents)base.Events;
+            set => base.Events = value;
+        }
+
+        /// <summary>
+        /// Gets or sets the collection of <see cref="ISecurityTokenValidator"/> used to read and validate the <see cref="SecurityToken"/>s.
+        /// </summary>
+        [SuppressMessage("Microsoft.Usage", "CA2208:InstantiateArgumentExceptionsCorrectly", Justification = "By design")]
```
HaoK: Can we remove this supression and just throw the argument exception with parameter name correctly?

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
Tratcher: Yes, ValidateToken always returns both a ClaimsPrincipal and a parsedToken.

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationOptions.cs
```diff
@@ -0,0 +1,170 @@
...
+            set
+            {
+                _securityTokenHandlers = value ?? throw new ArgumentNullException("SecurityTokenHandlers");
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the AuthenticationType used when creating the <see cref="System.Security.Claims.ClaimsIdentity"/>.
+        /// </summary>
+        public string SignInAsAuthenticationType
```
HaoK: Nit: SignInAuthenticationType?  Does the As add anything?  This is what we use along with SignInScheme right?

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationOptions.cs
```diff
@@ -0,0 +1,170 @@
...
+        /// <summary>
+        /// Gets or sets the type used to secure data handled by the middleware.
+        /// </summary>
+        public ISecureDataFormat<AuthenticationProperties> StateDataFormat { get; set; }
+
+        /// <summary>
+        /// Gets or sets the <see cref="TokenValidationParameters"/>
+        /// </summary>
+        /// <exception cref="ArgumentNullException"> if 'TokenValidationParameters' is null.</exception>
+        [SuppressMessage("Microsoft.Usage", "CA2208:InstantiateArgumentExceptionsCorrectly", Justification = "Name of the property.")]
```
HaoK: Remove? use nameof for exceptions everywhere as well instead of constants.

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationOptions.cs
Tratcher: Obsolete, removed.

test/Microsoft.AspNetCore.Authentication.WsFederation.Test/CustomStateDataFormat.cs
```diff
@@ -0,0 +1,56 @@
...
+{
+    public class CustomStateDataFormat : ISecureDataFormat<AuthenticationProperties>
+    {
+        private string lastSavedAuthenticationProperties;
+        private DataContractSerializer serializer = new DataContractSerializer(typeof(AuthenticationProperties));
+
+        public string Protect(AuthenticationProperties data)
+        {
+            lastSavedAuthenticationProperties = Serialize(data);
+            return "ValidStateData";
```
HaoK: Nit: use a const?

test/Microsoft.AspNetCore.Authentication.WsFederation.Test/WsFederationTest.cs
```diff
@@ -0,0 +1,293 @@
...
+        [Fact]
+        public async Task InvalidTokenIsRejected()
+        {
+            var builder = new WebHostBuilder()
+                .ConfigureServices(ConfigureAppServices)
+                .Configure(ConfigureApp);
+            var server = new TestServer(builder);
+
+            var httpClient = server.CreateClient();
+
```
HaoK: Can we move all the builder/server/client creation into a `CreateClient()`, the tests all seem to only use httpClient right?

test/Microsoft.AspNetCore.Authentication.WsFederation.Test/WsFederationTest.cs
```diff
@@ -0,0 +1,293 @@
...
+            // Did the request end in the actual resource requested for
+            Assert.Equal("AuthenticationFailed", await response.Content.ReadAsStringAsync());
+        }
+
+        private void ConfigureAppServices(IServiceCollection services)
+        {
+            services.AddAuthentication(sharedOptions =>
+            {
+                sharedOptions.DefaultScheme = CookieAuthenticationDefaults.AuthenticationScheme;
+                sharedOptions.DefaultSignInScheme = CookieAuthenticationDefaults.AuthenticationScheme;
```
HaoK: Don't need this if its the same as the DefaultScheme?

Tratcher: We do today due to the 2.0.x patch bug.

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
HaoK: Still might be safer to just check principal != null && parsedToken != null

samples/WsFedSample/Startup.cs
HaoK: Ugh, yeah that's too bad

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
Tratcher: We only reference the token for the expiration values, so I'll check there.

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
PinpointTownes: Why using `NoResult` here? A WS-Fed sign-in response without a `wresult` parameter is illegal so this should return an errored result.

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
```diff
@@ -0,0 +1,347 @@
...
+                {
+                    Logger.LogWarning("Received a sign-in message without a WResult.");
+                    return (HandleRequestResult)HandleRequestResult.NoResult();
+                }
+
+                var token = wsFederationMessage.GetToken();
+                if (string.IsNullOrEmpty(token))
+                {
+                    Logger.LogWarning("Received a sign-in message without a token.");
+                    return (HandleRequestResult)HandleRequestResult.NoResult();
```
PinpointTownes: Same remark. Plus it's inconsistent with the OIDC handler.

PinpointTownes:
@Tratcher it's 2017 and CSRF/session fixation attacks are no longer cool so please consider adding built-in antiforgery support like what we have in the OIDC handler. To support unsolicited assertions, an opt-in option `AcceptUnsolicitedAssertions` could be be added.

Tratcher:
The anti-xsrf with an opt-out is an interesting suggestion. I'll open a new issue for that.

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
Tratcher: I don't see a direct equivalent in OIDC, it detects correct messages by looking for State and CorrelationId, both of which are optional for WsFed. That said, we should probably be using the SkipUnrecognizedRequests pattern here.

PinpointTownes: > That said, we should probably be using the SkipUnrecognizedRequests pattern here.

It's something you're already doing at line 129 by making an `IsSignInMessage` check, which ensures that a `wa` parameter is present and corresponds to a sign-in operation. Once this check is made, you can consider the request as a valid WS-Fed message. Every missing parameter that is marked as required by the spec should result in an error.

Tratcher: Fair.

samples/WsFedSample/WsFedSample.csproj
```diff
@@ -0,0 +1,27 @@
...
+  <PropertyGroup>
+    <TargetFrameworks>net461;netcoreapp2.0</TargetFrameworks>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Microsoft.AspNetCore.Authentication.WsFederation\Microsoft.AspNetCore.Authentication.WsFederation.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.Cookies" Version="2.0.0" />
```
natemcmaster: This is going to cause issues. You should always use a ProjectReference to anything that is built from the same repo.

cc @JunTaoLuo

JunTaoLuo: Ah I see why this fails universe. We will build this for this release without using Universe. I'll build locally instead. The reason we wanted to take this approach is because we wanted to use the release 2.0.0 packages as dependencies.

Issue #942, opened 2016-08-10
Add RedirectToPostLogoutRedirectUri to OpenIdConnect auth


troydai:
Proposed in PR: https://github.com/aspnet/Security/pull/925

Requires change to the IOpenIdConnectEvent: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/IOpenIdConnectEvents.cs.

Tratcher:
Note this is a breaking change to add interface members and cannot be done until 2.0. Consider removing the interface pattern entirely to avoid this limitation in the future.

PR #1446, opened 2017-09-21
Add the SignedOutCallbackRedirect event for OIDC

Tratcher:
#942 Back-filling an event we couldn't add before due to interface breaking change issues. The event interface was removed in 2.0 so we can add the events we were missing.
@PinpointTownes

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
@@ -122,6 +122,10 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
                     Logger.RemoteSignOutSkipped();
                     return false;
                 }
+                if (remoteSignOutContext.Result.Failure != null)
+                {
+                    throw new InvalidOperationException("An error was returned from the RemoteSignOut event.", remoteSignOutContext.Result.Failure);
+                }
             }
 
             if (message == null)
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -273,18 +273,31 @@ public override Task<bool> HandleRequestAsync()
...
         /// </summary>
         /// <returns>A task executing the callback procedure</returns>
-        protected virtual Task<bool> HandleSignOutCallbackAsync()
+        protected async virtual Task<bool> HandleSignOutCallbackAsync()
         {
-            if (Request.Query.TryGetValue(OpenIdConnectParameterNames.State, out StringValues protectedState))
+            Request.Query.TryGetValue(OpenIdConnectParameterNames.State, out StringValues protectedState);
+            var properties = Options.StateDataFormat.Unprotect(protectedState);
+
+            var redirectContext = new SignedOutRedirectContext(Context, Scheme, Options, properties);
```
PinpointTownes: To make custom parameters easier to retrieve from the OIDC post logout notification request, maybe it would be worth instantiating and flowing an `OpenIdConnectMessage` built from the query string parameters?

test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/OpenIdConnectEventTests.cs
```diff
@@ -62,6 +64,7 @@ public class OpenIdConnectEventTests
                 OnRedirectToIdentityProvider = RedirectNotImpl,
                 OnRedirectToIdentityProviderForSignOut = RedirectNotImpl,
                 OnRemoteSignOut = RemoteSignOutNotImpl,
+                OnSignedOutCallbackRedirect = RedirectToSignedOutNotImpl,
```
PinpointTownes: The fact you have to do that every time you add a new event makes me think this syntax is not the best one.

You could muchly simplify the OIDC and JWT tests if you use a delegate-based syntax:

```csharp
var server = CreateServer(options =>
{
    options.Events.OnMessageReceived = context => { ... };

    // Don't set the other events as a default "throw a NotImplementedException"
    // event handler is already registered by the CreateServer() test helper.
});
```

src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs
```diff
@@ -93,6 +93,10 @@ public virtual Task<bool> ShouldHandleRequestAsync()
                     {
                         return false;
                     }
+                    else if (errorContext.Result.Failure != null)
+                    {
+                        throw new InvalidOperationException("An error was returned from the RemoteFailure event.", errorContext.Result.Failure);
```
HaoK: Nit: Do we not use resources in Security?

Tratcher: Not many, it's really inconsistent

HaoK: If loc ever becomes important, that will be fun :)

Issue #1443, opened 2017-09-20
Anti-xsrf plus opt-out for WsFed


Tratcher:
WsFed has historically not used anti-xsrf cookies because it supports unsolicited logins. However anti-xsrf cookies could be used if there was an opt-out option for applications that required support for unsolicited logins.
https://github.com/aspnet/Security/pull/1441#issuecomment-330873428

@brentschmaltz how common is it for apps to rely on unsolicited logins?

PinpointTownes:
> @brentschmaltz how common is it for apps to rely on unsolicited logins?

Common or not, it should definitely be opt-out. XSRF/session fixation attacks are too dangerous to ignore them by default.

PinpointTownes:
BTW, it might be interesting to backport such a security feature to Katana 4 :sweat_smile:

Tratcher:
Frequency matters because if all apps need to allow unsolicited logins then this feature is counter productive.

PinpointTownes:
Luckily, not all apps need IdP or third-party initiated login (which is not the most common case from what I can say based on my own experience).

JSYK, the OASIS committee decided in 2012 to add an errata to the SAML2 specification to encourage implementations to offer a way to reject unsolicited assertions when it became clear that it represented a real security threat:

> Note that the use of unsolicited responses can lead to Cross-Site Request Forgery (CSRF) vulnerabilities due to the inability to ensure that a request from the client originated the SAML profile transaction. Service providers SHOULD have a means of disabling the acceptance of unsolicited responses if circumstances warrant. The use of solicited responses may also be vulnerable to such attacks, the use of cookies to correlate the issuance of SAML requests and responses with the same client being one possible solution. However, if unsolicited respones cannot be prevented, no improvement to the solicited case will be of use.

http://docs.oasis-open.org/security/saml/v2.0/errata05/os/saml-v2.0-errata05-os.html

WS-Fed's unsolicited responses feature works exactly the same way so the same attack vector unfortunately applies.

I'm tempted to think it would be better to make it safe by default.

blowdart:
damnit I agree with @PinpointTownes

PinpointTownes:
> damnit I agree with @PinpointTownes

We really need to do something to stop that... :trollface:

Tratcher:
Offline notes: @brentschmaltz also swore and agreed with @PinpointTownes.

brentschmaltz:
@Tratcher I did :-)

PR #1457, opened 2017-09-27
Block unsolicited wsfed logins by default.

Tratcher:
#1443 This adds the standard correlation cookie requirement to WsFed. If wctx or cookie are missing then the request will be rejected. AllowUnsolicitedLogins can be used to opt out of this check.
@brentschmaltz @PinpointTownes

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
@@ -20,6 +20,7 @@ namespace Microsoft.AspNetCore.Authentication.WsFederation
     /// </summary>
     public class WsFederationHandler : RemoteAuthenticationHandler<WsFederationOptions>, IAuthenticationSignOutHandler
     {
+        private const string CorrelationProperty = ".xsrf";
         private WsFederationConfiguration _configuration;
 
         /// <summary>
@@ -77,6 +78,8 @@ namespace Microsoft.AspNetCore.Authentication.WsFederation
             {
                 wsFederationMessage.Wreply = BuildRedirectUri(Options.CallbackPath);
             }
+
+            GenerateCorrelationId(properties);
 
             var redirectContext = new RedirectContext(Context, Scheme, Options, properties)
             {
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
```diff
@@ -164,6 +169,13 @@ public WsFederationHandler(IOptionsMonitor<WsFederationOptions> options, ILogger
                 {
                     return messageReceivedContext.Result;
                 }
+                wsFederationMessage = messageReceivedContext.ProtocolMessage;
+                properties = messageReceivedContext.Properties; // Provides a new instance if not set.
+
+                if (!ValidateCorrelationId(properties) && !Options.AllowUnsolicitedLogins)
```
PinpointTownes: I'd be tempted to think that when you have a non-null `properties` instance (for RP-initiated logins), the correlation identifier should be checked even if `Options.AllowUnsolicitedLogins` was set to `true`.

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationOptions.cs
```diff
@@ -153,5 +153,11 @@ public TokenValidationParameters TokenValidationParameters
         /// The default is true. This should be disabled only in development environments.
         /// </summary>
         public bool RequireHttpsMetadata { get; set; } = true;
+
+        /// <summary>
+        /// The Ws-Federation protocol allows the user to initiate logins without contacting the application for a Challenge first.
+        /// However, that flow is susceptible to XSRF and other attacks so it is disabled here by default.
+        /// </summary>
+        public bool AllowUnsolicitedLogins { get; set; } = false;
```
PinpointTownes: I can already hear @Eilon say that the `= false` part is unnecessary 'cause `false` is the default value :sweat_smile:

Eilon: Yes, yes it is. This is bad 😄

Tratcher:
Updated.

test/Microsoft.AspNetCore.Authentication.WsFederation.Test/WsFederationTest.cs
```diff
@@ -51,36 +52,44 @@ public class WsFederationTest
...
         }
 
         [Fact]
-        public async Task ValidUnsolicitedTokenIsAccepted()
+        public async Task ValidUnsolicitedTokenIsRefused()
         {
             var httpClient = CreateClient();
 
-            // Send an invalid token and verify that the token is not honored
+            var kvps = new List<KeyValuePair<string, string>>();
```
HaoK: Nit: maybe its worth moving a lot of the common kvps init into a BuildFormContent() helper

Eilon:
cc @blowdart because security.

Issue #1469, opened 2017-10-06
Core 2.0 multiple auth schemes problem: default one overrides all, and no default ignores all


vgribok:
The problem I am reporting seems to be similar to https://github.com/aspnet/Security/issues/1381.

# Summary
My TL;DR summary is that in ASP.NET Core 2,0 it seems impossible to have multiple authentication schemes handling requests with corresponding values of the `Authorization` header, while we had no trouble implementing this in 1.1 and earlier with OWIN.

# Details
Using ASP.NET Core 2.0, we are adding two custom authentication schemes, let's call them "Abc" and "Xyz", using `services.AddSchema()` method. When I send a request with `Authorization` header having either `Abc SomeAuthInfo` or `Xyz SomeOtherAuthInfo` value, I expect a corresponding handler to be called by the framework, as it happened in OWIN and later in Core 1.1. That does not happen. The only non-invasive way to make it work was specifying default schema via `services.AddAuthentication("Abc")`, which resulted in "Abc" scheme working and "Xyz" not working. 

Eventually, I have to do this:
```
di.AddScoped<IAuthenticationHandlerProvider, AuthSchemeMappingAuthenticationHandlerProvider>();
di.AddSingleton<IAuthenticationSchemeProvider, BugWorkaroundAuthenticationSchemeProvider>();
```
with those two hand-rolled classes hacking the default ASP.NET Core 2.0 logic letting us look at what Authorization header value starts with - the scheme - and map it the appropriate handler.

# The Question
My question is, is this a bug, or have I simply failed to find a legitimate way to map auth schemes to their handlers?

sschleicher208:
I have this exact same issue.  I have Jwt and Facebook authentication enabled (Jwt is default scheme).  When I pass Jwt token in the authorization header, works perfectly.  When I pass the access token in the authorization header that I received from facebook, it fails.  Did you ever find anything out about this?

vgribok:
So far I have not heard anything from ASP.NET dev team. I have implemented a workaround by replacing two classes as shown in my post, but I hope to replace it with a better solution. I'm hoping someone can suggest a better way.

PinpointTownes:
> My question is, is this a bug, or have I simply failed to find a legitimate way to map auth schemes to their handlers?

Good or bad, this is a deliberate change that was announced here: https://github.com/aspnet/Announcements/issues/262.
[...]

Issue #1199, opened 2017-05-04
Consider dropping Active in AuthorizeAttribute.ActiveAuthenticationSchemes 


HaoK:
The active was removed everywhere else, but left in the attribute property.  Would be nice to cleanup if possible.

PinpointTownes:
Ideally, you should consider removing the authentication part from `AuthorizeAttribute` and move that to a separate attribute, which would allow doing things like that:

```
[Authenticate("Bearer")]
public class ApiController : Controller
{
    public IActionResult ActionWithOptionalAuthentication() { ... }

    [Authorize]
    public IActionResult ActionWithRequiredAuthentication() { ... }

    [Authorize(Policy = "HasSuperPrivilege")]
    public IActionResult ActionWithRequiredAuthenticationAndSpecifiedPolicy() { ... }
}

[Authenticate("Cookie")]
public class WebController : Controller
{
    public IActionResult ActionWithOptionalAuthentication() { ... }

    [Authorize]
    public IActionResult ActionWithRequiredAuthentication() { ... }

    [Authorize(Policy = "HasSuperPrivilege")]
    public IActionResult ActionWithRequiredAuthenticationAndSpecifiedPolicy() { ... }
}
```

HaoK:
That doesn't sound unreasonable to me, @rynowak @blowdart @Tratcher thoughts?

HaoK:
Would we keep the existing semantic today of [Authorize] => [Authenticate] implied?

blowdart:
And what would happen if you put Authorize against a method without Authenticate anywhere near?

PinpointTownes:
> And what would happen if you put Authorize against a method without Authenticate anywhere near?

It would use the default authentication scheme handler, as specified in the authentication options. If no is specified, an exception would be thrown, exactly like what already happens in the 1.0 bits.

blowdart:
That's not what it looks like.

Consider the average dev.

I want authentication, so I'm going to put authenticate on my controller. What happens now when there's no authorize attribute? Do we assume one on every method? This looks confusing.

PinpointTownes:
> Would we keep the existing semantic today of [Authorize] => [Authenticate] implied?

Yeah, the default authorization policy should still use `RequireAuthenticatedUser()`.

PinpointTownes:
> I want authentication, so I'm going to put authenticate on my controller.

Authentication = determining who's the caller.
[...]

PR #1475, opened 2017-10-09
Prototype policy handler

HaoK:
Some rough prototype of what we were discussing in https://github.com/aspnet/Security/issues/1469

This would be a different direction than https://github.com/aspnet/Security/pull/1410

Thoughts @davidfowl @Tratcher 

Example:

```
                auth.AddPolicyScheme("policy1", "policy1", p =>
                    {
                        p.DefaultScheme = "auth1";
                    })
                    .AddPolicyScheme("policy2", "policy2", p =>
                    {
                        p.AuthenticateScheme = "auth2";
                    })
                    .AddPolicyScheme("dynamic", "dynamic", p =>
                    {
                        p.DefaultSchemeSelector = c => c.Request.QueryString.Value.Substring(1);
                    })
                    .AddScheme<TestOptions, TestHandler>("auth1", o => { })
                    .AddScheme<TestOptions, TestHandler>("auth2", o => { })
                    .AddScheme<TestOptions, TestHandler>("auth3", o => { });

```


Tratcher:
First glance: What's a policy scheme? You mean a scheme policy?

Tratcher:
I'd rather put this project on hold until we can build out those samples we're discussing, I think they'll help us approach this more holistically.

src/Microsoft.AspNetCore.Authentication/AuthenticationPolicyHandler.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// Forwards calls to another authentication scheme based on <see cref="AuthenticationPolicyOptions"/>.
+    /// </summary>
+    /// <typeparam name="TOptions"></typeparam>
+    public class AuthenticationPolicyHandler<TOptions> : AuthenticationHandler<TOptions>, IAuthenticationSignInHandler
+        where TOptions : AuthenticationPolicyOptions, new()
+    {
+        public AuthenticationPolicyHandler(IOptionsMonitor<TOptions> options, ILoggerFactory logger, UrlEncoder encoder, ISystemClock clock)
+            : base(options, logger, encoder, clock) { }
+
+        private string ResolveDefaultScheme(string scheme)
```
davidfowl: We need to make sure this never maps to the policy scheme itself (stackoverflow).

src/Microsoft.AspNetCore.Authentication/AuthenticationPolicyHandler.cs
```diff
@@ -0,0 +1,58 @@
...
+        public string AuthenticateScheme { get; set; }
+        public string ChallengeScheme { get; set; }
+        public string ForbidScheme { get; set; }
+        public string SignInScheme { get; set; }
+        public string SignOutScheme { get; set; }
+
+        /// <summary>
+        /// Used to 
+        /// </summary>
+        public Func<HttpContext, string> DefaultSchemeSelector { get; set; }
```
davidfowl: Isn't this a policy selector?

davidfowl:
@HaoK I'd wire up some more realistic policies (facebook, oidc, api, cookies). I'd like to see what they look like in sample code. 

Besides the samples, I think this is on the right track and doesn't introduce any new concepts!

src/Microsoft.AspNetCore.Authentication/AuthenticationPolicyHandler.cs
HaoK: Well this one is actually selecting what scheme to resolve to for this policy instance, as opposed to a policy selector.  

But naming is getting confusing here, since "policies" are just special authentication options that contain a full set of schemes to redirect the normal `IAuthenticationHandler` calls to...

HaoK:
// Identity/default template
```
AddAuthentication(o => o.DefaultScheme = "Identity")
   // AddIdentityCookies() or the equivalent
   .AddPolicy("Identity", policy => {
       policy.DefaultScheme = "Identity.Application";
       policy.SignInScheme = "Identity.External"
   }))
```

// OIDC template:
```
AddAuthentication(o => o.DefaultScheme = "OIDCPolicy")
   .AddOIDC()
   .AddCookies()
   .AddPolicy("OIDCPolicy", policy => {
       policy.DefaultScheme = "Cookies";
       policy.ChallengeScheme = "OIDC"
   }))
```

// Social template (no identity)
```
AddAuthentication(o => o.DefaultScheme = "Social")
   .AddFacebook()
   .AddCookies()
   .AddPolicy("Social", policy => {
       policy.DefaultScheme = "Cookies";
       policy.ChallengeScheme = "Facebook"
   }))
```
// API doesn't really need a policy since its a single bearer already, but for completeness 
```
AddAuthentication(o => o.DefaultScheme = "API")
   .AddJwtBearer()
   .AddPolicy("API", policy => {
       policy.DefaultScheme = "Bearer"
   }))
```

davidfowl:
@HaoK I tweaked the samples slightly. 

/cc @PinpointTownes @brockallen @leastprivilege

davidfowl:
@HaoK we also need to see API and cookies in the same application. That's the scenario that requires switching based on some per request logic.

HaoK:
Request based scheme selection:
```
AddAuthentication("Dynamic")
   .AddJwtBearer()
   .AddCookies()
   .AddPolicy("Dynamic", schemes => {
       schemes.DefaultSchemeSelector = httpContext => IsRequestApi(httpContext) ? "Bearer" : "Cookies"
   }))
```

src/Microsoft.AspNetCore.Authentication/AuthenticationPolicyHandler.cs
```diff
@@ -0,0 +1,58 @@
...
+            => scheme ?? Options.DefaultSchemeSelector?.Invoke(Context) ?? Options.DefaultScheme;
+
+        protected override Task<AuthenticateResult> HandleAuthenticateAsync()
+            => Context.AuthenticateAsync(ResolveDefaultScheme(Options.AuthenticateScheme));
+
+        protected override Task HandleChallengeAsync(AuthenticationProperties properties)
+            => Context.ChallengeAsync(ResolveDefaultScheme(Options.ChallengeScheme));
+
+        protected override Task HandleForbiddenAsync(AuthenticationProperties properties)
+            => Context.ForbidAsync(ResolveDefaultScheme(Options.ForbidScheme));
```
davidfowl: Do we also need to do the crazy fallback here? Forbid ?? Challenge ?? DefaultScheme

src/Microsoft.AspNetCore.Authentication/AuthenticationPolicyHandler.cs
Jingkun-Hu: I think it's better to give a us a general description on the concepts that are necessary and complete and come naturally. Like given a restricted resource, be it code, data, or anything else, it maps to "policy" name(string). Then in AuthorizationOptions, policy name maps to an AuthorizationPolicy that mainly list of requirements...

src/Microsoft.AspNetCore.Authentication/AuthenticationPolicyHandler.cs
HaoK: So the crazy fallback is all still there if no default is set, and the scheme is null, as ResolveDefaultScheme will be null, and then it will fall back to the underlying defaults.  

I think we should get rid of the crazy fallbacks for the policies if we can...  A single default scheme fallback seems sufficient from all of the examples so far...

HaoK: But maybe you are right and this would make things inconsistent :(

HaoK:
So is it safe to say this is a preferred option to where we were going in https://github.com/aspnet/Security/pull/1410 ?  If so I'll close that out as that seemed way more complex than this PR

src/Microsoft.AspNetCore.Authentication/AuthenticationPolicyHandler.cs
HaoK: Yeah so that's exactly why I don't think we should call this new thing Policy, as its totally not the same thing as Authorization policy today.  We need a new word to use to disambiguate...

HaoK: I'd suggest a weirder name like calling it `AddSchemeSet/Map()` instead of `AddPolicy`

PinpointTownes:
@davidfowl thanks for pinging me.

I took a brief look and I'm not sure I like it (yet). Why doing that at the handler level? Why can't we integrate it with the APIs we already have? (e.g [`IAuthenticationSchemeProvider`](https://stackoverflow.com/a/46475072/542757)).

FWIW, I like the concept of authentication policies (because well, I'm the one who suggested that, initially :trollface:), but if you adopt it, then I'd expect seeing similar APIs to the ones we have for the authorization stack.

HaoK:
@PinpointTownes you mean you want it more of a first class new thing?  More like what we tried in https://github.com/aspnet/Security/pull/1410 then?

HaoK:
So I had a thought... why don't we just roll this concept into the base AuthenticationSchemeOptions/Handler.  We already sorta have this concept with RemoteAuthenticationOptions/SocialAuth where it forwards SignIn/Forbid to a Cookie.

Instead of adding a new thing, we could just bake this directly into `AuthenticationSchemeOptions`, and have the `AuthenticationHandler` base classes do the scheme forwarding automatically if set.

And then we wouldn't need a a new 'policy' for authentication, as scheme is already the equivalent of a named thing that is a Policy in Authorization.

So then you updating the examples a bit, using a new `AddFowardScheme` method that just adds the base handler implementation that can only forward

// Identity/default template
```C#
AddAuthentication("Identity")
   // AddIdentityCookies() or the equivalent
    .AddForwardScheme("Identity", options => {
       options.DefaultForwardScheme = "Identity.Application";
       options.ForwardSignInScheme = "Identity.External"
   }))
```

// OIDC template:
```C#
AddAuthentication("Cookies")
   .AddOIDC()
   .AddCookies(options => options.ForwardChallengeScheme = "OIDC")
```

// Social template (no identity)

```C#
AddAuthentication("Cookies")
   .AddFacebook()
   .AddCookies(options => options.ForwardChallengeScheme = "Facebook")
```
// API doesn't really need a policy since its a single bearer already, but for completeness 

```C#
AddAuthentication("Bearer")
   .AddJwtBearer()
```

Request based scheme selection:
```C#
AddAuthentication("Dynamic")
   .AddJwtBearer()
   .AddCookies()
   .AddForwardScheme("Dynamic", schemes => {
       schemes.DefaultForwardSchemeSelector = httpContext => IsRequestApi(httpContext) ? "Bearer" : "Cookies"
   }))
```

PinpointTownes:
> @PinpointTownes you mean you want it more of a first class new thing? More like what we tried in #1410 then?

Yep. The other PR looks better, IMHO (and more consistent with existing concepts in the authorization stack).

HaoK:
Yeah but its also pretty complicated, personally I prefer the direction this PR is going, but its possible there are still some things we added in the other PR that are worth keeping

HaoK:
So summarizing my current thinking after playing with this for a while:

I don't think we need a new concept of AuthenticationPolicy.  An authentication scheme is exactly what we want already.  It has exactly the same operations a policy has: Authenticate/Challenge/SignIn[Out]/Forbid.  The problem today is its not easy at all to mix and match the operations from multiple authentication schemes for a single authorization.  But we can solve that by simply moving towards registering a "Virtual/Dynamic" authentication scheme, that is able to combine the operations of other schemes arbitrarily (and dynamically).

That's what kept bugging me about the other PR trying to introduce authentication scheme, it felt like a slightly different flavor of what we already had.

The benefit of this new pattern of using "virtual/dynamic" authentication schemes, is that we can move away from needing to support multiple authentication schemes entirely.  For any situation where executing multiple authentication handlers is necessary in a challenge/authenticate, you just plug in a custom authentication scheme to aggregate it explicitly.

Thoughts @davidfowl @Tratcher @PinpointTownes ?

Tratcher:
Overriding classes is a much higher usage barrier. We have some very common scenarios today like JWT + Cookies that shouldn't require that level of customization.

HaoK:
> We have some very common scenarios today like JWT + Cookies 

That doesn't require a custom authentication handler, its only if you wanted to challenge more than one scheme, or authenticate more than one scheme and aggregate.  So like if you wanted to challenge Windows + Basic or multiple Bearer schemes at once for example.  

The JWT +Cookie example would look like above, I'm not proposing doing nothing, just not adding anything resembling a 'policy':

```
AddAuthentication("Dynamic")
   .AddJwtBearer()
   .AddCookies()
   .AddVirtualScheme("Dynamic", schemes => {
       schemes.DefaultForwardSchemeSelector = httpContext => IsRequestApi(httpContext) ? "Bearer" : "Cookies"
   }))
```

Tratcher:
That's an interesting example. So AddVirtualScheme is something we'd provide out of the box?  (or we call it AddDynamicScheme and you don't have to specify the name)

HaoK:
Yes I'm proposing we go with the idea that we make it easy to virtualize/forward + resolve schemes dynamically (the last set of examples I showed above).  But no, you NEED to give it a real authentication scheme name.  It just doesn't have to do anything, we'd either add a new base handler that did this forwarding logic, or we roll it into the existing base handler class and make it non abstract (effectively the same thing)

HaoK:
Basically I'm arguing schemes are already policies...not that we don't want to do all the stuff in this PR :)

HaoK:
I'm guessing we need to create a VirtualAuthenticationHandler since we don't implement all the interfaces by default anymore...

davidfowl:
@HaoK I agree and I'm on the same page.

PinpointTownes:
> That doesn't require a custom authentication handler, its only if you wanted to challenge more than one scheme, or authenticate more than one scheme and aggregate. So like if you wanted to challenge Windows + Basic or multiple Bearer schemes at once for example.

This scenario was supported in 1.0 and it's now painful to do in 2.0 (the remarks on the announcement tickets are eloquent). I believe it should be supported natively and shouldn't require adding custom code.

The problem with the approach you suggest is that it's all or nothing (e.g all the challenge/forbid/signin/signout/authentication schemes points to "Cookie" or "Bearer"). Having a policy-based system would allow much more flexibility.

For me, something like that would be the ideal API:

```
{
    options.AddIdentityPolicy("Cookies");

    options.AddPolicy("JWT tokens only", policy =>
    {
        policy.UseAuthenticateShemes(JwtBearerDefaults.AuthenticationScheme);
        policy.UseChallengeSchemes(JwtBearerDefaults.AuthenticationScheme);
    });

    options.AddPolicy("JWT tokens + Data Protection tokens", policy =>
    {
        // Authenticate requests using both the JWT and the aspnet-contrib handlers.
        policy.UseAuthenticateShemes(
            JwtBearerDefaults.AuthenticationScheme, 
            OAuthValidationDefaults.AuthenticationScheme);

        // Both the JWT and aspnet-contrib are "Bearer" handlers. Apply a single
        // challenge to ensure no duplicate "Bearer" WWW-Authenticate is returned.
        policy.UseChallengeSchemes(JwtBearerDefaults.AuthenticationScheme);
    });

    options.DefaultPolicySelector = context =>
    {
        if (context.Request.Path.StartsWithSegments("/legacy-api"))
            return "JWT tokens only";

        if (context.Request.Path.StartsWithSegments("/api"))
            return "JWT tokens + Data Protection tokens";

        return "Cookies";
    };
})

.AddJwtBearer(...)

.AddOAuthValidation();
```

PinpointTownes:
You could even add the `[Authenticate]` attribute I suggested [here](https://github.com/aspnet/Security/issues/1199) to make the policy story super nice:

```
[Authenticate("Bearer")]
public class ApiController : Controller
{
    public IActionResult ActionWithOptionalAuthentication() { ... }

    [Authorize]
    public IActionResult ActionWithRequiredAuthentication() { ... }

    [Authorize(Policy = "HasSuperPrivilege")]
    public IActionResult ActionWithRequiredAuthenticationAndSpecifiedPolicy() { ... }
}

[Authenticate("Cookie")]
public class WebController : Controller
{
    public IActionResult ActionWithOptionalAuthentication() { ... }

    [Authorize]
    public IActionResult ActionWithRequiredAuthentication() { ... }

    [Authorize(Policy = "HasSuperPrivilege")]
    public IActionResult ActionWithRequiredAuthenticationAndSpecifiedPolicy() { ... }
}
```

davidfowl:
@PinpointTownes I think we're splitting hairs TBH. 

> Having a policy-based system would allow much more flexibility.

How doesn't this design achieve the same thing? It looks identical, what am I missing?

PinpointTownes:
> How doesn't this design achieve the same thing? It looks identical, what am I missing?

My proposal allows assigning multiple schemes per request and different authenticate/signin/signout/challenge/forbid schemes per request. @HaoK's approach doesn't.

HaoK:
Yes my hope is to drop any explicit support for automatic routing of authentication/challenge to multiple schemes entirely, leave it as a pure extensibility (code + custom authentication handler scenario), so there's no expectation it just works.  If we get repeated asks to make it easier to do complex things that involve things like merging multiple schemes for authentication or complex challenge behavior, we can always add it later.

PinpointTownes:
> Yes my hope is to drop any explicit support for automatic routing of authentication/challenge to multiple schemes entirely, leave it as a pure extensibility (code + custom authentication handler scenario), so there's no expectation it just works.

#fail.

HaoK:
Challenging multiple authentication schemes today is basically unsupported in many scenarios anyways, it is almost impossible to get it to work without dealing with quirks.  This puts things in a much better state going forward.

PinpointTownes:
> Multiple authentication scheme support today is basically unsupported anyways, it is almost impossible to get it to work without dealing with quirks anyways.

It was perfectly supported in 1.0, where you could mix Basic + Bearer, Basic + Negociate, Basic + Negociate + API Keys, etc...

HaoK:
You can still accomplish a virtual scheme that represents multiple schemes, you just need to plug in an auth handler that does it...  It's not going to be a lot of code.

HaoK:
Nothing was 'perfectly supported' in 1.0.  Everything barely worked...and everything broke in minor regressions with any changes...  Its 'perfectly supported' in 2.0 as well, when you write an auth handler which represents your "policy"

HaoK:
We don't ship basic auth today do we?

PinpointTownes:
> Nothing was 'perfectly supported' in 1.0.

Ask @Tratcher, mixing NTLM + Bearer (or any other non-interactive scheme) used to work just fine (as designed).

HaoK:
You mean how you can have multiple automatic authenticate = true in 1.0?

HaoK:
That was totally not how it was designed...we just had no way to enforce single automatic authenticate easily

PinpointTownes:
> We don't ship basic auth today do we?

Sure you do. Both IIS and HttpSysServer support it.

HaoK:
Yes and the hosts are already effectively using schemes as described in this PR

HaoK:
Define can't mix?

HaoK:
If you want to mix "Windows" with "Bearer", you would have to add your own "WindowsBearer" that authenticated/challenged both yes.  This is no different than how things stand in 2.0 anyways.  So we can split that into its own issue/PR if that's the case.

PinpointTownes:
> That was totally not how it was designed...we just had no way to enforce single automatic authenticate easily

You don't get it, obviously. In 1.0, multiple non-interactive authentication middleware could deliberately apply their challenges to a same response. That's exactly the reason why `JwtBearerHandler.HandleUnauthorizedAsync()` returns `false` in 1.x (again, ask @Tratcher if you don't trust me...).

HaoK:
I get it, you like the old chaining behavior we explicitly killed in 2.0.  You can chain, but you do yourself in IAuthenticationHandler.Challenge

PinpointTownes:
> This is no different than how things stand in 2.0 anyways.

Stop talking about 2.0, we all know the situation is not ideal and that many scenarios are no longer supported. Mixing NTLM and Bearer was supported in 1.0 and is not in 2.0. Call that whatever you want, but it's a regression.

Reading all the complaints on the discussions tickets should be enough to realize it :sweat_smile:

PinpointTownes:
> I get it, you like the old chaining behavior we explicitly killed in 2.0.

With a **huge** difference. You have to explicitly define a policy asking for multiple schemes to be used. In 1.0, most authentication middleware were configured to use automatic challenge.

```csharp
options.AddPolicy("JWT tokens + Data Protection tokens", policy =>
{
    // Authenticate requests using both the JWT and the aspnet-contrib handlers.
    policy.UseAuthenticateShemes(
        JwtBearerDefaults.AuthenticationScheme, 
        OAuthValidationDefaults.AuthenticationScheme);
    
    // Both the JWT and aspnet-contrib handlers are "Bearer" handlers. Apply a single
    // challenge to ensure no duplicate "Bearer" WWW-Authenticate is returned.
    policy.UseChallengeSchemes(JwtBearerDefaults.AuthenticationScheme);
});
```

HaoK:
Its really not that much more code, and merging principles really should be handled by the app itself:

```
services.AddAuthentication("JwtData")
            .AddJwtBearer()
            .AddOAuthValidation()
            .AddVirtualHandler<JwtDataAuth>("JwtData")


public class JwtDataAuth : VirtualAuthenticationHandler {
   public override async Task<string> HandleAuthenticate() 
      => MergeYourPrincipals(await Context.AuthenticateAsync("Bearer"), await Context.AuthenticateAsync("OAuthValidation");

     public override Task HandleChallengeAsync() => Context.ChallengeAsync("Bearer");     
   }
}
```

PinpointTownes:
> and merging principles really should be handled by the app itself

There's already code that handles that in ASP.NET Core, there's really no reason to do that yourself, is it?

I really hope you'll end up with a better design for such simple scenarios. Just saying folks "hey, you know what? Write custom code!" is not really satisfying.

HaoK:
That merging principal code was something we totally made up for this (we reorder things, and it also cannot be used with WindowsPrincipals at all)  So if you try to use it with Windows auth it will likely explode.

PinpointTownes:
> So if you try to use it with Windows auth it will likely explode.

The "merging code" works with Windows identities. It's just that the resulting principal is a simple `ClaimsPrincipal` so things like `IsInRole` don't work properly in this case.

HaoK:
Replacing with new iteration PR

Issue #1259, opened 2017-06-12
Stop logging any PII


mattmazzola:
Currently the JwtBearerMiddleware logs success or failure of authorization and token validation as seen in the below locations:

"Authorization was successful for user: {UserName}."
https://github.com/aspnet/Security/blob/22d2fe99c6fd9806b36025399a217a3a8b4e50f4/src/Microsoft.AspNetCore.Authorization/LoggingExtensions.cs#L18
Call to: _logger.UserAuthorizationSucceeded(GetUserNameForLogging(user));
https://github.com/aspnet/Security/blob/e940cdb36bc7617af3d762ff9cc853a8fef55b0a/src/Microsoft.AspNetCore.Authorization/DefaultAuthorizationService.cs#L101
 
"Failed to validate the token {Token}."
https://github.com/aspnet/Security/blob/22d2fe99c6fd9806b36025399a217a3a8b4e50f4/src/Microsoft.AspNetCore.Authentication.JwtBearer/LoggingExtensions.cs#L19
Call to: Logger.TokenValidationFailed(token, ex);
https://github.com/aspnet/Security/blob/e940cdb36bc7617af3d762ff9cc853a8fef55b0a/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs#L122

For our particular usage of this middleware we don't want to be liable for this PII data in our logs.

In other libraries such as ADAL (Active Directory Authentication Library) which also deal with tokens they hash the token and username so that they are still unique within the logs to allow metrics of particular user without identifying them or exposing token.

See example of hashing:
`string accessTokenHash = PlatformPlugin.CryptographyHelper.CreateSha256Hash(result.AccessToken);`
https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/8f6d560fbede2247ec0e217a21f6929d4375dcaa/src/ADAL.PCL/Flows/AcquireTokenHandlerBase.cs#L351

In the future versions perhaps this PII is hashed by default to match behavior of other microsoft libraries, but for backwards compatibility perhaps add new flag to the `JwtBearerOptions` class that's used to configure the middleware's logging such as:

```
var options = new JwtBearerOptions
{
  ...
  HashPersonallyIdentifiableInformation = true
};
```
And the logging would then be:

"Authorization was successful for user: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855."
Instead of:
[...]

PR #1478, opened 2017-10-10
Remove username from AuthZ logging

HaoK:
Fixes https://github.com/aspnet/Security/issues/1259

@Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
@@ -110,7 +110,7 @@ namespace Microsoft.AspNetCore.Authentication.JwtBearer
                         }
                         catch (Exception ex)
                         {
-                            Logger.TokenValidationFailed(token, ex);
+                            Logger.TokenValidationFailed(ex);
 
                             // Refresh the configuration for exceptions that may be caused by key rollovers. The user can also request a refresh in the event.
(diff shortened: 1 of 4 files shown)


Tratcher:
It's a start, but @blowdart had a laundry list of things we should remove beyond this item.

HaoK:
Also removed token logging, @blowdart what else did you want us to stop logging?

PinpointTownes:
If you had to debug real world apps having authentication issues every day, you'd never dare remove that... :trollface:

blowdart:
I know, but GPDR says no personal identifying information without protecting it, and we cannot know how logs are protected.

However I will email the lawyers just to check.

blowdart:
I can see the other side "Microsoft software helped my company get fined 10% of our global income". I'm going to leave this up to legal :D

HaoK:
@blowdart what's the call here?

blowdart:
Still waiting. In theory, "if we do not retain the data it is our customers responsibility to service DSR requests."

Now we'd have to then add something into logging to allow us to search in the templates. Which is a challenge.

leastprivilege:
Move it to trace logging - that can by definition (aka according to your docs) contain "sensitive" data.

PinpointTownes:
@leastprivilege you're a little late to the party... this was discussed in the main thread :sweat_smile: 

https://github.com/aspnet/Security/issues/1259#issuecomment-335627995

Tadimsky:
Any updates on this PR?

HaoK:
ba1eb281d135400436c52c17edc71307bc038ec0

Issue #1353, opened 2017-08-07
Cookies and OIDC emit invalid Expires header values


Tratcher:
CookieHandler and OpenIdConnectHandler set `Response.Headers[HeaderNames.Expires] = "-1";` for responses that should not be cached. However, `-1` is an invalid value for the expires header which should be a full date string `Expires: Sun, 06 Nov 1994 08:49:37 GMT`.
https://tools.ietf.org/html/rfc7234#section-5.3
It's abusing one line in the spec that says `A cache recipient MUST interpret invalid date formats, especially the value "0", as representing a time in the past (i.e., "already expired").`

Some tools hide/drop invalid headers, making it harder to inspect them. (E.g. HttpClient).

We should use the epoc date used by Response.Cookies.Delete instead. ` new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc),`
https://github.com/aspnet/HttpAbstractions/blob/e8123db21ece289e30abf174a37a908b9e5e1e98/src/Microsoft.AspNetCore.Http/Internal/ResponseCookies.cs#L131

Tratcher:
@JunTaoLuo ResponseCaching doesn't handle this special case of an invalid expires headers does it?

JunTaoLuo:
ResponseCaching relies on `HeaderUtilities.TryParseDate` which returns `DateTimeOffset.MinValue` when given an invalid value. For the purpose of ResponseCaching logic, this ensures that invalid date formats are treated as time in the past.

PR #1481, opened 2017-10-11
Use a compliant Expires header value for expiration.

Tratcher:
#1353 -1 is an invalid value for the Expires header, it should be a date in the past.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -20,7 +20,7 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
         IAuthenticationSignOutHandler
     {
         private const string HeaderValueNoCache = "no-cache";
-        private const string HeaderValueMinusOne = "-1";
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -20,7 +20,7 @@ public class CookieAuthenticationHandler :
         IAuthenticationSignOutHandler
     {
         private const string HeaderValueNoCache = "no-cache";
-        private const string HeaderValueMinusOne = "-1";
+        private const string HeaderValueEpocDate = "Thu, 01 Jan 1970 00:00:00 GMT";
```
HaoK: Nit: why not just call this HeaderValueExpired so if you change this again, you don't have to rename it :)

HaoK: Also if this is only used in one place...these don't really even need to be consts do they?

Tratcher: I'd rather have the name describe the value than where it gets used. It makes the usage code more obvious.
The constants also let you describe the value, and keep mics strings out of the code.

Issue #1349, opened 2017-08-04
Make default value of ChunkingCookieManager.ThrowForPartialCookies be 'false'


Tratcher:
When an auth cookie gets chunked it's possible that the client may drop some of those chunks due to storage capacity limits. When ChunkingCookieManager then tries to reassemble the chunks and fails it defaults to throwing a format exception. This exception is not caught by the cookie auth handler and can result in a 500 response.
https://github.com/aspnet/Security/blob/5d802a798847f61c8c32368f5690a48caefbca97/shared/Microsoft.AspNetCore.ChunkingCookieManager.Sources/ChunkingCookieManager.cs#L116

Alternatively the cookie auth handler could catch and log the exception.
https://github.com/aspnet/Security/blob/5d802a798847f61c8c32368f5690a48caefbca97/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs#L107

In practice customers end up disabling this option (if they find it).
```
var cookieManager = new ChunkingCookieManager
{
    ThrowForPartialCookies = false
};
```

Eilon:
I think this goes well with our guidance of making it hard for a client to inadvertently "poison" themselves by having some bad cookies.

PR #1482, opened 2017-10-11
Don't throw for partial cookies

Tratcher:
#1349 In practice customers end up disabling this option (if they find it).

### DIFF
--- a/shared/Microsoft.AspNetCore.ChunkingCookieManager.Sources/ChunkingCookieManager.cs
+++ b/shared/Microsoft.AspNetCore.ChunkingCookieManager.Sources/ChunkingCookieManager.cs
@@ -44,7 +44,6 @@ namespace Microsoft.AspNetCore.Internal
             // See http://browsercookielimits.x64.me/.
             // Leave at least 40 in case CookiePolicy tries to add 'secure', 'samesite=strict' and/or 'httponly'.
             ChunkSize = DefaultChunkSize;
-            ThrowForPartialCookies = true;
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


shared/Microsoft.AspNetCore.ChunkingCookieManager.Sources/ChunkingCookieManager.cs
```diff
@@ -44,7 +44,7 @@ public ChunkingCookieManager()
             // See http://browsercookielimits.x64.me/.
             // Leave at least 40 in case CookiePolicy tries to add 'secure', 'samesite=strict' and/or 'httponly'.
             ChunkSize = DefaultChunkSize;
-            ThrowForPartialCookies = true;
+            ThrowForPartialCookies = false;
```
PinpointTownes: You could remove that since you're assigning the default value :sweat_smile:

src/Microsoft.Owin.Security.Interop/ChunkingCookieManager.cs
```diff
@@ -25,7 +25,7 @@ public ChunkingCookieManager()
             // See http://browsercookielimits.x64.me/.
             // Leave at least 20 in case CookiePolicy tries to add 'secure' and/or 'httponly'.
             ChunkSize = 4070;
-            ThrowForPartialCookies = true;
+            ThrowForPartialCookies = false;
```
PinpointTownes: Same.

PinpointTownes:
Now... I need to find a cool "party hard" GIF. 'cause this change is great.

Tratcher:
There was already a negative test that was overriding this setting.

PR #1486, opened 2017-10-12
Iteration 2: options.SchemeForwarding + AddVirtualScheme

HaoK:
Previous iteration: https://github.com/aspnet/Security/pull/1475

- Added SchemeForwardOptions to AuthenticationSchemeOptions, enables all handlers the ability to redirect schemes.
- Added AddVirtualScheme which is a pure redirect handler (also automatically enables the SchemeForwarding option)

Updated Examples: 

```C#
// Identity/default template
AddAuthentication("Identity")
   // AddIdentityCookies() or the equivalent
    .AddVirtualScheme("Identity", forwardOptions => {
       forwardOptions.DefaultTarget = "Identity.Application";
       forwardOptions.SignInTarget = "Identity.External"
   }))
```

```C#
// OIDC/Social/external auth templates:

AddAuthentication("Auth")
   .AddVirtualScheme("Auth", options => {
        options.Default = "Cookies";
        options.Challenge = "OIDC/Facebook/Google";
   }
   .AddOIDC/Facebook/Google()
   .AddCookies();

```

```C#
// API doesn't really need a policy since its a single bearer already, but for completeness 
AddAuthentication("Bearer")
   .AddJwtBearer()
```

```C#
// Request based scheme selection:
AddAuthentication(o => DefaultScheme = "Dynamic", DefaultSignInScheme = "Cookies")
   .AddJwtBearer()
   .AddCookies()
   .AddGoogle()
   .AddVirtualScheme("Dynamic", options => {
       options.DefaultSelector = httpContext => IsRequestApi(httpContext) ? "Bearer" : "Google"
   }))
```

@Tratcher @davidfowl


Tratcher:
This is an interesting concept, but it should be isolated to the virtual handler. The part about putting it in the base class and allowing any handler to forward to any other does not makes sense. Virtual handlers should be able to manage the indirection on behalf of other handlers. Your option 1 above is redundant with option 2.

HaoK:
Yeah I considered doing that, but I felt it was better to start this way to leave all the options on the table...  The difference between option 1 and option 2, is there's no additional virtual scheme needed.  So it's just cookies + social, where you configure cookie to forward the challenge to social.  As opposed to having 3 schemes, "Cookies", "<the social one>", and the "<virtual>" one you actually use.

That's the main difference and why I went with this approach for now to see if its worth it to give it for 'free' to all handlers.

Tratcher:
I'd rather keep them separate for the first pass. We can vet the virtual handler as an independent compoent before we decide if it makes sense to integrate the functionality into existing handlers.

HaoK:
Fair enough, updated with tests and made it standalone, should be more or less ready for real review now, we actually could move this to HttpAbstractions/Core authentication as it doesn't depend on the security stack at all now

src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
```diff
@@ -84,6 +90,17 @@ public AuthenticationBuilder(IServiceCollection services)
...
 
+        /// <summary>
+        /// Adds a <see cref="VirtualAuthenticationHandler"/> based authentication handler which can be used to 
+        /// redirect to other authentication schemes.
+        /// </summary>
+        /// <param name="authenticationScheme">The name of this scheme.</param>
+        /// <param name="displayName">The display name of this scheme.</param>
+        /// <param name="configureOptions">Used to configure the scheme options.</param>
+        /// <returns>The builder.</returns>
+        public virtual AuthenticationBuilder AddVirtualScheme(string authenticationScheme, string displayName, Action<VirtualSchemeOptions> configureOptions)
```
Tratcher: Extension?

src/Microsoft.AspNetCore.Authentication/VirtualSchemeOptions.cs
```diff
@@ -0,0 +1,27 @@
...
+        public string AuthenticateTarget { get; set; }
+        public string ChallengeTarget { get; set; }
+        public string ForbidTarget { get; set; }
+        public string SignInTarget { get; set; }
+        public string SignOutTarget { get; set; }
+
+        /// <summary>
+        /// Used to select a default target scheme based on the request.
+        /// </summary>
+        public Func<HttpContext, string> DefaultTargetSelector { get; set; }
```
Tratcher: does this need an additional parameter saying what action you're resolving so you can return different results for Challenge, Forbid, etc.? An enum maybe?

test/Microsoft.AspNetCore.Authentication.Test/GoogleTests.cs
```diff
@@ -1061,19 +1061,19 @@ private static TestServer CreateServer(Action<GoogleOptions> configureOptions, F
...
-                    services.AddAuthentication()
+                    services.AddAuthentication("Auth")
+                        .AddVirtualScheme("Auth", "Auth", o =>
+                        {
+                            o.DefaultTarget = TestExtensions.CookieAuthenticationScheme;
+                            o.ChallengeTarget = GoogleDefaults.AuthenticationScheme;
+                        })
                         .AddCookie(TestExtensions.CookieAuthenticationScheme)
                         .AddGoogle(configureOptions)
                         .AddFacebook(o =>
```
Tratcher: Why do the google tests include facebook?

src/Microsoft.AspNetCore.Authentication/VirtualSchemeOptions.cs
```diff
@@ -0,0 +1,27 @@
...
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    /// <summary>
+    /// Used to redirect authentication methods to another scheme
+    /// </summary>
+    public class VirtualSchemeOptions
+    {
+        public string DefaultTarget { get; set; }
```
davidfowl: These names bug me

test/Microsoft.AspNetCore.Authentication.Test/GoogleTests.cs
HaoK: There's a test that ensures that when you ask for google but get a facebook identity it throws I think,

src/Microsoft.AspNetCore.Authentication/VirtualSchemeOptions.cs
HaoK: Yeah not in love with the current Target naming either, but DefaultScheme didn't seem to great...    Redirect/Forward were other names I considered...

src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
HaoK: The other methods were public virtual so I just did the same for consistency...  This also seems slightly easier to customize if someone wanted to return their own derived custom AuthenticationBuilder from an AddXyz() that changed the behavior of `AddVirtualScheme`

src/Microsoft.AspNetCore.Authentication/VirtualSchemeOptions.cs
HaoK: You can always setup multiple virtual schemes and target that for each verb, i.e. "ChallengeScheme", "ForbidScheme" and use the selector like that, I want to avoid too much craziness if we can.

HaoK: Or I guess if we want to move away from funcs, we could make this a Type we resolve from DI and have a real interface and have it take an enum

brockallen:
I am jumping in here half-way thru changes that are being made and I have no context... but I wonder who the target is for these features. Is there a spec? What's driving these changes/enhancements? Are these band-aids for the recent authN redesign?

All just honest questions. But given that I think I know how 2.0 works, and then looking at these proposals I get nervous.

HaoK:
@brockallen this is basically sugar targetting the multiple mixed schemes scenario (i.e. cookies + bearer, and also making it easier to dynamically select a scheme using the request).  Its purely additive to 2.0.  This was driven by most of the upgrade/migration issues would have been solved by using some variation of a "virtual scheme".  This feature will allow us to give them a consistent story/answer with a combination of using virtual schemes as is + writing their own custom virtual scheme handler implementation for the most advanced scenarios.

src/Microsoft.AspNetCore.Authentication/VirtualSchemeOptions.cs
HaoK: So names...  could just go back to Schemes I guess with VirtualSchemeOptions having Authenticate/Challenge/Forbid/SignIn/SignOutScheme properties.

HaoK:
Ping @davidfowl @Tratcher for final review for initial commit so we can start using these in the new samples repo, I think the only real open issue left is if naming the scheme properties in `VirtualSchemeOptions` do we keep the *Target suffix or go back to just calling them `Authenticate/Challenge/SignIn/OutScheme`

src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
Tratcher: I still don't see why virtual is getting special treatment here, from the outside it's not mechanically different from any other auth handler. Let's start with an extension method like the other providers.

src/Microsoft.AspNetCore.Authentication/VirtualAuthenticationHandler.cs
```diff
@@ -0,0 +1,69 @@
...
+            }
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            Scheme = scheme;
+            Context = context;
+
+            Options = OptionsMonitor.Get(Scheme.Name) ?? new VirtualSchemeOptions();
```
Tratcher: You dropped the Validate call that other handlers use?

Tratcher: The fallback to a default doesn't make sense in this case, the default is unusable. Throw? Validate would also cover that. It seems like you must set at least one target and/or DefaultTargetSelector, or else what's the point of adding this handler?

src/Microsoft.AspNetCore.Authentication/VirtualSchemeOptions.cs
Tratcher: This still needs work. The context of the action you're trying to resolve matters.

src/Microsoft.AspNetCore.Authentication/VirtualAuthenticationHandler.cs
```diff
@@ -0,0 +1,69 @@
...
+            => Context.SignInAsync(ResolveTarget(Options.SignInTarget), user, properties);
+
+        public virtual Task SignOutAsync(AuthenticationProperties properties)
+            => Context.SignOutAsync(ResolveTarget(Options.SignOutTarget), properties);
+
+        public Task ChallengeAsync(AuthenticationProperties properties)
+            => Context.ChallengeAsync(ResolveTarget(Options.ChallengeTarget), properties);
+
+        public Task ForbidAsync(AuthenticationProperties properties)
+            => Context.ForbidAsync(ResolveTarget(Options.ForbidTarget), properties);
```
Tratcher: This should follow the same fallback rules as AuthSchemeProvider.
E.g. Forbid -> Challenge -> Default.
https://github.com/aspnet/HttpAbstractions/blob/c0f937239a0a099b73c67c96ab9e1c875952f67f/src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs#L71-L74

src/Microsoft.AspNetCore.Authentication/VirtualSchemeOptions.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public string AuthenticateTarget { get; set; }
```
Tratcher: Taking a second look at this I wonder if this shouldn't be two different handlers. One that has a list of targets/schemes, and one that only has the selector. Is there a functional advantage to combining them?

src/Microsoft.AspNetCore.Authentication/VirtualAuthenticationHandler.cs
Tratcher: It's interesting to think that this creates two levels of fallback. VirtualForbid -> VirtualChallenge -> VirtualDefault -> DefaultForbid -> DefaultChallenge -> DefaultDefault.

src/Microsoft.AspNetCore.Authentication/AuthenticationBuilder.cs
HaoK: Its not special, the other handler in the core package (RemoteAuthenticationHandler) is already also a virtual methods on the builder as opposed to an extension, this is just continuing that pattern.

HaoK: We could have written AddScheme itself as an extension method, at this point I think any handlers we add in core Authentication should go on the AuthenticationBuilder as a virtual, while anything added in another package would be an extension method (unless they are returning a custom AuthenticationBuilder in their `AddXyz()`).  I think the VirtualScheme is more of a core concept like RemoteAuth, unlike something narrow and specific like a cookie/oauth handler

HaoK: I would actually almost argue that we should consider moving VirtualScheme into the HttpAbstractions real core auth layer, but we aren't setup to do that easily right now...

src/Microsoft.AspNetCore.Authentication/VirtualSchemeOptions.cs
HaoK: We can iterate on this on the next go round once we start working on the samples repo.  I think it'll be easier for us to play with having a basic default selector to start, and we can tweak it in the next PR, if we remove it entirely, you won't be able to do request based selection at all.

src/Microsoft.AspNetCore.Authentication/VirtualAuthenticationHandler.cs
HaoK: The default is usable, it just means use the existing defaults for this virtual scheme.  So a use case could be, you want a virtual scheme that only kicks in for some very specific requests, otherwise it falls back to the existing schemes/defaults.  i.e. you have a selector that only targets "/specialmagichacks/" paths

But yeah I guess the Validate on the options is still a useful pattern to have for people who extend the handler/options.  I'll add that back as a no-op

src/Microsoft.AspNetCore.Authentication/VirtualAuthenticationHandler.cs
HaoK: It doesn't need to do the explicit fallbacks since it defaults to null which already captures the other fallbacks.

HaoK: That was the whole point of default: null, external callers to Context's auth apis will have the default logic taken care of already.

src/Microsoft.AspNetCore.Authentication/VirtualAuthenticationHandler.cs
Tratcher: The default VirtualSchemeOptions doesn't have a sector either. It's useless without at least one option set.

src/Microsoft.AspNetCore.Authentication/VirtualAuthenticationHandler.cs
Tratcher: So: VirtualForbid -> VirtualDefault -> DefaultForbid -> DefaultChallenge -> DefaultDefault.
It seems strange to leave VirtualChallenge out of that chain when VirtualDefault is included.

HaoK: I don't think we want this to mimic any of the crazy fallback chains, this is a purely virtual scheme. There should be a main default which you can easily set so you don't have to set every property.  There should be one which takes precedence for each action.  

But if we do find a need for the extra fallbacks in the virtual schemes in the upcoming sample/scenarios, we can then bring them back as needed

src/Microsoft.AspNetCore.Authentication/VirtualSchemeOptions.cs
HaoK: Just two concept counts instead of one, the meta concept count is authentication policy...so sorta...but there's no reason we couldn't split the behavior in different ways...

Lets keep them together for this iteration, if we don't like how the samples look, we can always split it apart into `AddVirtualScheme/AddSelectorScheme` or something like that in the next iteration

HaoK:
Re: DefaultTargetSelector

> The context of the action you're trying to resolve matters.

For this iteration of the selector, the only scenario is you have one virtual scheme per context, if you wanted to go crazy have have a different target selector per verb, you'd have to add one virtual scheme per selector, and glue it together via a top level "Uber" scheme that pointed each verb at the corresponding virtual scheme.  

Rather than trying to invent that complexity up front, lets leave that for trying in samples before trying to invent some 'sugar(or poison in this case)' for that

Tratcher:
You keep saying you want to check this in so you can start building samples, but there's nothing that prevents you from building basic samples for this in the repo and in this PR.

Tratcher:
Let's take this offline. The concept is certainly interesting but we're still disagreeing on some fundamental design points.

HaoK:
Sure, but the basic samples are already in the comments of the PR at the top.  The ones you are asking about are not the main scenarios.  But sure we can take this offline

HaoK:
Discussed offline with @Tratcher and for the initial commit, we settled on dropping the Target for the names, so

`VirtualSchemeOptions` with properties that target other schemes:
`Authenticate`
`SignIn`
`SignOut`
`Forbid`
`Challenge`
`Default`
`DefaultSelector`

There won't be any support for making it easy to use the selector to target different schemes per verb.  If we get requests for that, we can add more support later.  

And the fallback logic is
1. Specific Property is used if non null (i.e. Authenticate/Challenge)
2. DefaultSelector is run with the HttpContextand used if non null
3. Default is used (which if not set, and null, means we fall back to existing default behavior when no scheme is specified)

Specifically this means there's no similar forbid falling back to challenge for virtual scheme options.

HaoK:
Updated with the results of the previous discussions

test/Microsoft.AspNetCore.Authorization.Test/AuthorizationPolicyFacts.cs
```diff
@@ -20,6 +20,24 @@ public void RequireRoleThrowsIfEmpty()
         }
 
         [Fact]
+        public async Task TestHao()
```
Tratcher: ?

test/Microsoft.AspNetCore.Authorization.Test/PolicyEvaluatorTests.cs
```diff
@@ -104,6 +105,30 @@ public class PolicyEvaluatorTests
         }
 
         [Fact]
+        public async Task Challenge()
```
Tratcher: Aren't these for a different issue?

HaoK: Oops yeah was just testing stuff, didn't mean to check these in, good catch, will remove

HaoK:
72e1cb1385c982e6dba90c839d1a1455e2813e6c

PR #1501, opened 2017-10-20
Added support for multiple values (arrays) in default claim action

tstojecki:
Claims like role might come in a form of an array with multiple values. When using the default Oidc MW role option isn't mapped by default, so that mapping has to be added.
```o.ClaimActions.Add(new JsonKeyClaimAction("role", "role", "role"));```

The default mappers only process the single string value claims. This can be useful for any multivalue claim.

In the case of roles, as each claim is mapped to the same key, User.IsInRole works as expected. If you expand the list of claims for ClaimsIdentity it would looks something like this
```
{name: user}
{role: Foo}
{role: Bar}
```


Tratcher:
This needs test for the added behavior.

johnkors:
@Tratcher What Nuget and what version is this fix in?

Tratcher:
It looks like this will go into 2.1.0-preview one which we are preparing to release in the next few weeks.

johnkors:
Ok, thanks! 👍

Issue #1027, opened 2016-11-07
OIDC TokenValidated event fires in a different order depending on the flow


Tratcher:
The OIDC TokenValidated event fires in a different order based on the flow you're using.
1) for id_token it fires  first / alone
2) for code flow it fires after TokenResponseReceived
3) for code+id_token it fires before the code is redeemed, but then there's no event fired after the second id_token from token endpoint is validated

Proposal (Breaking):
Rename TokenValidated to something authorization endpoint specific like AuthorizationEndpointIdTokenValidated. Then add a new event TokenEndpointIdTokenValidated. Now the order would not vary, only the events based on your flow.

Eilon:
For case (3) we should fire the event twice, and on the second time add the additional info.

Tratcher:
I'm still concerned that firing the same event for two different scenarios will confuse and break people. The first time it fires it's primarily a chance for people to do additional validation and claims transformation. The new case we're considering is very different, it's about reconciling a secondary token/principal.

New proposal:
Keep TokenValidated as is for the first token processed, the app's implementation isn't likely to vary based on the endpoint the token was received from. Existing apps will be un-effected.
Add a new event SecondaryTokenValidated that only fires if a second token is received (code+id_token). This event presents both sets of tokens and principals and allows the user to reconcile them (see #1028).

muratg:
We'll potentially punt this if we can't get this in by early next week.

Tratcher:
@PinpointTownes any thoughts on adding an additional event here:
https://github.com/aspnet/Security/blob/824539f47ac7f57a824092b04795e25198e1438f/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L666-L673
As described in https://github.com/aspnet/Security/issues/1027#issuecomment-263073109

See https://github.com/aspnet/Security/issues/1028 about reconciling the two tokens.

PinpointTownes:
Sounds like a good idea, fairly cheap to implement and with no chance of breaking existing apps, no?

Tratcher:
Yes, it should be easy enough. The main question is how useful is that second token.

PinpointTownes:
> The main question is how useful is that second token.

[...]

PR #1515, opened 2017-10-26
Add OIDC SecondaryTokenValidated event. Refactor tests.

Tratcher:
#1027 This adds a new event for a previous gap. If you received an id_token and code from the authorization endpoint then you would not get an event where you could examine the validated id_token output from the token endpoint. No functional changes have been made, we're only exposing additional information that the user can take advantage of.

I've also refactored the event tests to reduce redundancy.

@PinpointTownes

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/OpenIdConnectEvents.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/OpenIdConnectEvents.cs
@@ -59,6 +59,12 @@ namespace Microsoft.AspNetCore.Authentication.OpenIdConnect
         public Func<TokenValidatedContext, Task> OnTokenValidated { get; set; } = context => Task.CompletedTask;
 
         /// <summary>
+        /// Invoked when an IdToken from the token endpoint (in addition to the one from the authorization endpoint)
+        /// has been validated and produced an AuthenticationTicket.
+        /// </summary>
+        public Func<SecondaryTokenValidatedContext, Task> OnSecondaryTokenValidated { get; set; } = context => Task.CompletedTask;
+
+        /// <summary>
         /// Invoked when user information is retrieved from the UserInfoEndpoint.
         /// </summary>
         public Func<UserInformationReceivedContext, Task> OnUserInformationReceived { get; set; } = context => Task.CompletedTask;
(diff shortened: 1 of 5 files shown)


PinpointTownes:
Looks fine.

Out of curiosity: instead of this super-specialized event, have you considered introducing a new event invoked just before a successful result is returned that would expose both the authorization/token/userinfo responses/principals? (e.g `AuthenticationCompleted`).

Tratcher:
I'd expect a catch-all event to be harder to use as what properties were set would depend on the flow. You'd end up with a lot of nulls in the simpler flows.

PinpointTownes:
As long as you have docs to warn about the fact certain properties may be null, it shouldn't be a problem. It would certainly be way more flexible as you could decide to pick specific claims from the 1st/2st identity token or from the userinfo response.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/Events/OpenIdConnectEvents.cs
```diff
@@ -59,6 +59,12 @@ public class OpenIdConnectEvents : RemoteAuthenticationEvents
         public Func<TokenValidatedContext, Task> OnTokenValidated { get; set; } = context => Task.CompletedTask;
 
         /// <summary>
+        /// Invoked when an IdToken from the token endpoint (in addition to the one from the authorization endpoint)
+        /// has been validated and produced an AuthenticationTicket.
+        /// </summary>
+        public Func<SecondaryTokenValidatedContext, Task> OnSecondaryTokenValidated { get; set; } = context => Task.CompletedTask;
```
HaoK: This is a pretty meh name, can we do better?

HaoK: Is there a more precise name for this other than "SecondaryTokenValidated"?

HaoK:
New event seems ok, "SecondaryToken" seems subpar as a name though...

HaoK:
I'd prefer a more generic name like @PinpointTownes ala OnEverythingDone

Tratcher:
I agree the naming is tricky on this one. It's not just for id_tokens from the token endpoint, it's for id_tokens from the token endpoint when you already had one from the authorization endpoint.

@PinpointTownes isn't recommending a different name, he's proposing a whole different event that would run at the end of Authenticate and expose all of the possible protocol objects generated. I see two problems with that approach. 1) Very different protocol objects are generated for the various flows. You would end up with a lot of null properties and people would have to know a lot about the flow they're using to know which properties they can work with. 2) the later you run, the less control you have over the process. By the time you reach the end of Authenticate you have almost no control over the process, at best you can re-arrange the produced ClaimsPrincipal.

brockallen:
BackChannelIdToken vs FrontChannelIdToken

Tratcher:
@brockallen That's the worst part of this, it's not just the BackChannelIdToken. The original complaint was that TokenValidated may fire for either the frontchannel or backchannel token depending on your flow. We can't actually change that right now without breaking people. Heince adding SecondaryTokenValidated.

I'm starting to think this is more trouble than it's worth, we haven't had pressing requests for this scenario.

brentschmaltz:
@Tratcher the validation of the id_token from the AuthorizeEndpoint is about signature, issuer, etc. the id_token from the TokenEndpoint is more about ProtocolValidation. Do we have something like: OnProtocolValidation.

HaoK:
If no-one has asked for this event, seems better to wait until there's demand no?

PR #1516, opened 2017-10-27
[WIP DO NOT MERGE] sketching some events for AuthenticationMiddleware

anurse:
This PR is **incomplete** ~~because of a quirk of EventListener. I'm going to iterate a bit on trying to figure that out. I've described the problem in a comment, it's a bit nasty but only affects "testing" the event source~~. It's also mostly a sketch right now, and I plan to take things out of it and into shared-source components in Common before merging anything. Those are noted by comments in the code.

This shows adding a single trio of Events (AuthenticationMiddlewareStart, AuthenticationMiddlewareEnd, AuthenticationMiddlewareFailure) to specifically track the impact of the Authentication Middleware on the request pipeline. There is also an EventCounter: "AuthenticationMiddlewareDuration" to track the time.

Also included in a version of `ValueStopwatch` (which will be moved to shared source package before we actually make any changes in dev) and `EventSpan` which can be used to track events that have a start and stop time without additional allocation (unless of course, the code in the middle goes async, at which point the `struct`s will box)

/cc @Tratcher @JunTaoLuo @davidfowl @rynowak


test/Microsoft.AspNetCore.Authentication.Test/CollectingEventListener.cs
```diff
@@ -0,0 +1,74 @@
...
+    // REVIEW: Also to go to some testing or shared-source assembly
+    // This can only collect from EventSources created AFTER this listener is created
+    public class CollectingEventListener : EventListener
+    {
+        private ConcurrentQueue<EventWrittenEventArgs> _events = new ConcurrentQueue<EventWrittenEventArgs>();
+        private HashSet<string> _eventSources;
+
+        public IReadOnlyList<EventWrittenEventArgs> EventsWritten => _events.ToArray();
+
+        public CollectingEventListener(params string[] eventSourceNames)
```
anurse: So. Here's the thing with `EventListener`. Because `EventSource`s can be created at any time, `OnEventSourceCreated` is used to make `EventListener`s "work". At construction time, it will be called once for each existing `EventSource`, then it will be called as each new `EventSource` is initialized. However, this is all done by the base class constructor, which means that when `OnEventSourceCreated` is called for `EventSources` that **already exist** at the time of construction, the derived constructor body **has not yet been run**, meaning `_eventSources` is unavailable.

This "works" right now in that I only tested running the single test I added, which means that the `AuthenticationEventSource` was only created **after** the `EventListener` was initialized. However, if it ran as part of a batch of tests where any previous test initialized the `AuthenticationEventSource`, it would all fall apart.

I'm not a fan of the idea of creating a listener that hooks every `EventSource` as there is a noticible perf impact to enabling any single `EventSource` and there are a few in the framework (plus as we add them to our components...).

One option is to just make `CollectingEventListener` abstract and require creating subclasses that hard-code the sources they care about.

Another option would be to make the `AuthenticationEventSource` (and friends) "pubternal" (or even public with all `internal` methods). `OnEventSourceCreated` only needs to be used when your listener doesn't have access to the actual `EventSource` type, you only have the name. If they do have that type, they can just call `.EnableEvents` and pass it in. I'm kinda partial to this as it would also make building arbitrary consumers a little easier (when the consumer is willing to reference our code directly) as they have a concrete type to use instead of a stringly-typed name.

anurse: We could also try to build something over `IObservable` like the SLAB of yore (https://msdn.microsoft.com/en-us/library/dn440729(v=pandp.60).aspx).

anurse: @pakrym and I may have a solution using devious locks :)

anurse: Solution found!

anurse:
FYI, let's focus here on the patterns for writing events rather than the particular event. I do not plan on merging this specific PR or this specific Event. We can discuss the value of the event itself in a separate thread, when it's time to actually add events :). I would like to get reviews/approvals for the pattern though.

src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs
```diff
@@ -31,34 +32,47 @@ public AuthenticationMiddleware(RequestDelegate next, IAuthenticationSchemeProvi
 
         public async Task Invoke(HttpContext context)
         {
-            context.Features.Set<IAuthenticationFeature>(new AuthenticationFeature
+            var span = AuthenticationEventSource.Log.AuthenticationMiddlewareStart(context.TraceIdentifier, context.Request.Path);
```
HaoK: Any reason not to just pass the context and let the event extract the trace identifier/path as needed?

src/Microsoft.AspNetCore.Authentication/Internal/AuthenticationEventSource.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        [NonEvent]
+        internal EventSpan<AuthenticationEventSource> AuthenticationMiddlewareStart(string traceIdentifier, PathString path)
+        {
+            if (IsEnabled(EventLevel.Informational, EventKeywords.None))
+            {
+                AuthenticationMiddlewareStart(traceIdentifier, path.Value);
+            }
+
+            return EventSpan.Create(this, (self, duration) => self.AuthenticationMiddlewareEnd(traceIdentifier, path, duration));
```
Tratcher: Allocating a closer per event isn't promising.

src/Microsoft.AspNetCore.Authentication/Internal/AuthenticationEventSource.cs
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        [NonEvent]
+        internal void AuthenticationMiddlewareEnd(string traceIdentifier, PathString path, TimeSpan duration)
+        {
+            if (IsEnabled())
```
pakrym: Why `if (IsEnabled())` here but not in other methods?

src/Microsoft.AspNetCore.Authentication/Internal/EventSpan.cs
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        public static ValueStopwatch StartNew() => new ValueStopwatch(Stopwatch.GetTimestamp());
+
+        private TimeSpan ComputeElapsedTime()
+        {
+            // Start timestamp can't be zero. Like... it would have to be literally the first thing executed when the machine boots to be 0.
+            if (_startTimestamp == 0)
+            {
+                throw new InvalidOperationException("An uninitialized, or 'default', ValueStopwatch cannot be used");
```
pakrym: Extract method

src/Microsoft.AspNetCore.Authentication/Internal/EventSpan.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Diagnostics;
+
+namespace Microsoft.AspNetCore.Authentication.Internal
+{
+    // REVIEW: This file will be in a common shared-source package.
+    internal struct ValueStopwatch
+    {
+        private static readonly double TimestampToTicks = TimeSpan.TicksPerSecond / (double)Stopwatch.Frequency;
+
+        private bool _enabled;
```
pakrym: Do you really need `_enabled`?

src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs
anurse: That's reasonable too, it depends on the specific event. When I started this, I didn't have the pattern of `internal` with `[NonEvent]` and `private` with `[Event]` so I had to have only primitive types in the signature (required by the `[Event]` attribute)

src/Microsoft.AspNetCore.Authentication/Internal/AuthenticationEventSource.cs
anurse: Yeah. This needs revision for sure.

src/Microsoft.AspNetCore.Authentication/Internal/AuthenticationEventSource.cs
anurse: Because the others call `IsEnabled(EventLevel, EventKeywords)` which also does the `IsEnabled()` check.

anurse: Ah, actually in this case it's because the counter is conditional on the provider being active but not at any particular level. Open to discussion on that though.

src/Microsoft.AspNetCore.Authentication/Internal/EventSpan.cs
anurse: Probably not, now that I just decided to use `_startTimestamp == 0` to indicate uninitialized. I could just do the same with `_endTimestamp == 0` implying Enabled.

anurse: Unlike `Stopwatch`, `ValueStopwatch` is single-use and once it's stopped it can't be resumed or reset (both of which `Stopwatch` does allow)

src/Microsoft.AspNetCore.Authentication/Internal/EventSpan.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            _endTimestamp = Stopwatch.GetTimestamp();
+            _enabled = false;
+        }
+
+        public static ValueStopwatch StartNew() => new ValueStopwatch(Stopwatch.GetTimestamp());
+
+        private TimeSpan ComputeElapsedTime()
+        {
+            // Start timestamp can't be zero. Like... it would have to be literally the first thing executed when the machine boots to be 0.
```
rynowak: `default(ValueStopwatch)` begs to differ

rynowak: OH I see you mean valid cases.

src/Microsoft.AspNetCore.Authentication/Internal/EventSpan.cs
```diff
@@ -0,0 +1,83 @@
...
+            if (_startTimestamp == 0)
+            {
+                throw new InvalidOperationException("An uninitialized, or 'default', ValueStopwatch cannot be used");
+            }
+
+            var end = _endTimestamp == 0 ? Stopwatch.GetTimestamp() : _endTimestamp;
+            var timestampDelta = end - _startTimestamp;
+            var ticks = (long)(TimestampToTicks * timestampDelta);
+            return new TimeSpan(ticks);
+        }
```
rynowak: Why do we need a field for endtimestamp at all? Just don't call it twice. This type could just be a wrapper around long and still provide the same functionality

src/Microsoft.AspNetCore.Authentication/Internal/EventSpan.cs
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    internal static class EventSpan
+    {
+        public static EventSpan<T> Create<T>(T state, Action<T, TimeSpan> endAction) => new EventSpan<T>(endAction, state);
+    }
+
+    internal struct EventSpan<T>
+    {
+        private readonly Action<T, TimeSpan> _endAction;
```
rynowak: I know it's mentioned elsewhere, but this pattern can't really be optimized to remove the allocation without writing a lot of code. You need to either write a struct per 'span', or cache a delegate per span, or just don't bother with this.

Personally I, lean towards just leaving this type out and putting the `ValueStopWatch` in code.

test/Microsoft.AspNetCore.Authentication.Test/CollectingEventListener.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics.Tracing;
+using System.Linq;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Authentication.Test
+{
+    // REVIEW: Also to go to some testing or shared-source assembly
+    // This can only collect from EventSources created AFTER this listener is created
```
rynowak: Is this a **lie** then?

>Called for all existing event sources when the event listener is created and when a new event source is attached to the listener.

https://msdn.microsoft.com/en-us/library/system.diagnostics.tracing.eventlistener.oneventsourcecreated(v=vs.110).aspx

test/Microsoft.AspNetCore.Authentication.Test/CollectingEventListener.cs
```diff
@@ -0,0 +1,98 @@
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics.Tracing;
+using System.Linq;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Authentication.Test
```
rynowak: How important do we think testing of these events really is?

src/Microsoft.AspNetCore.Authentication/Internal/EventSpan.cs
anurse: Sure. Just trying to see where we can match `Stopwatch` functionality, but given the other comments on `EventSpan` that sounds reasonable.

test/Microsoft.AspNetCore.Authentication.Test/CollectingEventListener.cs
anurse: The comment is out of date. The problem is that the event is called during the base class constructor which means that you can't access any constructor arguments (making it non-trivial to dynamically attach to event sources created before the listener was created by name). We were able to refactor around some locks to allow it to collect events from sources created before the listener was created.

test/Microsoft.AspNetCore.Authentication.Test/CollectingEventListener.cs
anurse: That's part of the discussions here. I don't know if we really care about testing the events but I wanted to sketch out what it would look like. It's not much different from testing logging and I don't think we do that very much either,.

test/Microsoft.AspNetCore.Authentication.Test/CollectingEventListener.cs
rynowak: ah, the good old fragile base class problem

test/Microsoft.AspNetCore.Authentication.Test/CollectingEventListener.cs
rynowak: yeah I'm not afraid to take a stand against testing these. I think the pattern that you showed here is reasonable, but I think overall it's kinda low value. Doing a good job testing these usually involve injecting error into places and since we have a really simple error handling story we usually don't do that today so we have to invent it.

src/Microsoft.AspNetCore.Authentication/Internal/EventSpan.cs
anurse: Yep, that's become fairly obvious from this. The amount of work to create a "clean" consumption API per event site is the same or worse so it doesn't really yield anything useful.

test/Microsoft.AspNetCore.Authentication.Test/CollectingEventListener.cs
pakrym: How do we know that we didn't regress things?

PinpointTownes:
Out of curiosity, why doing that at the authentication middleware level? Why not directly in the authentication service, so that passive authentication operations are also logged?

anurse:
> Out of curiosity, why doing that at the authentication middleware level? Why not directly in the authentication service, so that passive authentication operations are also logged?

See my above comment about the focus of this PR. That's a good topic for discussion when we start iterating on the specific events we want to add. We do plan to log things in the service as well, this is literally just a minimum effort to explore the **pattern** of event tracing in our code rather than the specific event.

anurse:
It's looking like people are coming to consensus on the tracing patterns, any further thoughts there? If not I'll close and start an issue/PR to track adding some specific events to this repo to serve as a guide for other repos.

PinpointTownes:
@anurse good to know, thanks.

anurse:
Related PR for ValueStopwatch now open: https://github.com/aspnet/Common/pull/279

anurse:
I'm going to close this now. I think one modification I'm going to make to the pattern is to have the EventSource return a `ValueStopwatch`, allowing it to return `default(ValueStopwatch)` if it is disabled and avoid the call to `Stopwatch.GetTimestamp()`. "End" events would then also take in the `ValueStopwatch` and could no-op if it was never actually initialized.

PR #1519, opened 2017-10-31
Refactor OIDC event tests

Tratcher:
This is useful OIDC events test refactoring from an otherwise aborted PR (https://github.com/aspnet/Security/pull/1515). It makes it easier to add more events in the future.

### DIFF
--- a/test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/OpenIdConnectEventTests.cs
+++ b/test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/OpenIdConnectEventTests.cs
@@ -28,67 +28,44 @@ namespace Microsoft.AspNetCore.Authentication.Test.OpenIdConnect
 {
     public class OpenIdConnectEventTests
     {
-        private readonly Func<MessageReceivedContext, Task> MessageNotImpl = context => { throw new NotImplementedException("Message"); };
-        private readonly Func<TokenValidatedContext, Task> TokenNotImpl = context => { throw new NotImplementedException("Token"); };
-        private readonly Func<AuthorizationCodeReceivedContext, Task> CodeNotImpl = context => { throw new NotImplementedException("Code"); };
-        private readonly Func<TokenResponseReceivedContext, Task> TokenResponseNotImpl = context => { throw new NotImplementedException("TokenResponse"); };
-        private readonly Func<UserInformationReceivedContext, Task> UserNotImpl = context => { throw new NotImplementedException("User"); };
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/OpenIdConnectEventTests.cs
```diff
@@ -28,1579 +28,938 @@ namespace Microsoft.AspNetCore.Authentication.Test.OpenIdConnect
...
 
         [Fact]
         public async Task OnAuthenticationFailed_HandledWithoutTicket_NoMoreEventsRun()
         {
-            var messageReceived = false;
-            var tokenValidated = false;
-            var codeReceived = false;
-            var tokenResponseReceived = false;
-            var userInfoReceived = false;
-            var a
```
HaoK: Nit: I can't tell if there's a common set of these which you are setting to true in all the tests, but maybe its worth adding a helper to slim these down to something more like:

var events = ExpectedOidcEvents.CreateDefaultExpectations();
// only set the one or two test specific ones. (which hopefully is none for many tests?)

test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/OpenIdConnectEventTests.cs
```diff
@@ -1669,41 +1024,41 @@ public class OpenIdConnectEventTests
...
-                    context.SkipHandler();
-                    return Task.CompletedTask;
-                };
-            }),
-            context =>
+                ExpectRemoteSignOut = true,
+            };
+            events.OnRemoteSignOut = context =>
+            {
+                events.InvokedRemoteSignOut = true;
```
HaoK: Is there any way to clean up this pattern to avoid needing to remember to set the invoked?  Maybe just add a new set of virtuals in the custom event class to override in the test...

Forgetting to set the bool will cause the validation to fail right?  Meaning that line is required always right?

HaoK: Basically something like `events.TestXyzEvent(Action<XyzContext> test)` method for the test to use instead which takes care of the InvokeXyz = true.

test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/OpenIdConnectEventTests.cs
Tratcher: It's progressive. You start with only one and work your way though until you've hit almost all of them.  There's no subset for all tests.

Issue #1517, opened 2017-10-28
Change log level of Cookie handler authentication message


leastprivilege:
Right now the cookie handler is using information level for this:

[10:31:24 Information] Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationHandler
AuthenticationScheme: [scheme] was successfully authenticated.

This happens way too often (at least once per request) and fills up the log file with unnecessary information. Can we move that to debug level?

Eilon:
Agreed. Leaving it up-for-grabs for anyone to fix, or we'll get to it when we can.

leastprivilege:
Done https://github.com/aspnet/Security/pull/1522

PinpointTownes:
It's worth noting that this change doesn't only affect the cookies handler, but all the handlers inheriting from `AuthenticationHandler` (third-party providers included).

Tratcher:
True, but only handlers that can directly produce an identity like cookies, bearer, basic, etc... None of the remote providers are affected.

PinpointTownes:
> None of the remote providers are affected.

That's not exact actually, as you can call `AuthenticateAsync("Facebook")`. Currently, it logs an `Information` message in both the sign-in scheme handler AND the remote handler.

I have no particular opinion about that, 'tho I think it will make debugging harder for beginners. I just wanted to point out that it may be considered as a functional breaking change (probably an acceptable one, though).

Eilon:
Fixed in https://github.com/aspnet/Security/pull/1522.

niemyjski:
@Eilon Is there any reason why I'm still seeing this even if my min level was info but I had an override for System and Microsoft for min level of warning? Looking at the extension methods I don't see where it's getting the log source from.

PR #1522, opened 2017-11-03
Change LogLevel from Information to Debug (see #1517)

leastprivilege:


src/Microsoft.AspNetCore.Authentication/LoggingExtensions.cs
```diff
@@ -39,7 +39,7 @@ static LoggingExtensions()
                 formatString: "{AuthenticationScheme} was not authenticated. Failure message: {FailureMessage}");
             _authSchemeAuthenticated = LoggerMessage.Define<string>(
                 eventId: 8,
-                logLevel: LogLevel.Information,
```
Eilon: >Information [](start = 35, length = 11)

@Tratcher check this out.

PR #1544, opened 2017-11-17
Add consent policies to CookiePolicy

Tratcher:
Here is the baseline feature for discussion. CookiePurpose needs attention, we need to decide what the categories are, if there's a default policy allowing some categories, etc..

I updated CookieAuth to work, but not any of the remote auth providers.

Depends on https://github.com/aspnet/HttpAbstractions/commit/0bd9bd8f76665f64c3e412de94f10fe3703aca50

### DIFF
--- a/samples/CookiePolicySample/Startup.cs
+++ b/samples/CookiePolicySample/Startup.cs
@@ -1,0 +1,118 @@
+using System;
+using System.Linq;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Authentication.Cookies;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Net.Http.Headers;
+
+namespace CookiePolicySample
+{
+    public class Startup
+    {
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddAuthentication(CookieAuthenticationDefaults.AuthenticationScheme)
+                .AddCookie();
+            services.Configure<CookiePolicyOptions>(options =>
+            {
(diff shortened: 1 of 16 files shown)


glennc:
for http://github.com/aspnet/Security/issues/1561

samples/CookiePolicySample/Startup.cs
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        private async Task HomePage(HttpContext context)
+        {
+            var response = context.Response;
+            var cookies = context.Request.Cookies;
+            response.ContentType = "text/html";
+            await response.WriteAsync("<html><body>\r\n");
+
+            await response.WriteAsync($"<a href=\"{context.Request.PathBase}/\">Home</a><br>\r\n");
```
anurse: Just use StaticFiles?

Tratcher:
Updated to drop cookie purposes and use IsEssential instead.

@blowdart said offline this is adequate for preview1, but is still mulling over the naming of a few properties.

The non-empty templates will add
```
            services.Configure<CookiePolicyOptions>(options =>
             {
                 options.CheckConsentPolicyNeeded = context => true;
             });
```
and
`app.UseCookiePolicy();`

samples/CookiePolicySample/Startup.cs
Tratcher: It's using dynamic paths.

src/Microsoft.AspNetCore.CookiePolicy/CookiePolicyOptions.cs
```diff
@@ -27,6 +27,18 @@ public class CookiePolicyOptions
         /// </summary>
         public CookieSecurePolicy Secure { get; set; } = CookieSecurePolicy.None;
 
+        public CookieBuilder ConsentCookie { get; set; } = new CookieBuilder()
+        {
+            Name = ".AspNet.Consent",
```
Tratcher: @blowdart  this cookie could use a better name, it doesn't say what is being consented to.

Tratcher:
Tested and ready for review.

PR #1549, opened 2017-11-22
Updated logging extension to remove the period on ValidationFailed

muqeet-khan:
_tokenValidationFailed format string includes the JWT token followed by a period, which if a dev troubleshooting copies incorrectly to the EOL will make the JWT invalid. (https://github.com/aspnet/Security/blob/a9482d1d03fc22f62e42cf31b8adb32c5550835e/src/Microsoft.AspNetCore.Authentication.JwtBearer/LoggingExtensions.cs#L19)

Current: Failed to validate the token eyJhbGc.......HCwFmw.
Proposed: Failed to validate the token eyJhbGc.......HCwFmw


Eilon:
Thanks for the PR!

Issue #1359, opened 2017-08-10
2.1 AuthZ improvements


HaoK:
This issue will track all the various Authorization improvements we are looking at for 2.1.

## Some initial thoughts:

From @davidfowl 

We've had a bunch of feedback about our authz system with respect to flowing context from the authorize attribute to the authorization handler:
 
- https://github.com/aspnet/Security/issues/670
- https://github.com/aspnet/Mvc/issues/5607
- https://github.com/aspnet/Security/issues/917
- https://github.com/aspnet/Security/issues/1352

Today the Authorize attribute supports has enough metadata to describe the policy but it doesn't let you specify the resource (via IAuthorizeData). I think if we solve that, it might solve a bulk of the issues since people will be able to write custom attributes that flow the relevant context to the authorization handler. 
 
Today that's only possible when doing imperative authz. I was thinking something like:
 
```
public interface IAuthorizeResource
{
    object Resource { get; }
}
```
 
If the attribute implemented this, we would flow that as the resource to the handler. This way you implement permissions of whatever you please via resources.
 
From @rynowak 

Look into providing overriding semantics in MVC, maybe a marker interface: `IAuthorizeMetadata`. Any attributes that implement that interface on the endpoint could be flowed. Then it’s up to developers to build whatever they want.  (Also look into flowing single objects vs many objects + requirements VS resources)

We will also look into making it possible to specify requirements via Attributes similar to imperative AuthZ so you don't have to preconstruct policies for attributes.

rynowak:
My additional thoughts on this. We're looking at making the concept of an **endpoint** available at a low level from the dispatcher. Endpoints will have arbitrary metadata in an ordered list. This is the equivalent to what MVC does today by passing the action descriptor.

-------------------------------------

Additionally in the case of MVC, the current `[Authorize]` has some surprising behaviors compared to other filters. All authorize filters will execute rather than supporting overriding.  Additionally, `[AllowAnonymous]` cannot be overridden and turns off **all** authorize attributes. These might be consistent with the past, but they don't really fit with the *zen* of MVC filters.

HaoK:
Interesting, well if we have a new endpoint concept with a clean slate for metadata, we should probably just expose all of the basic building blocks 

[...]

Issue #1338, opened 2017-07-24
Discussion for Auth 2.0


HaoK:
Related discussion thread of issues/questions about https://github.com/aspnet/Announcements/issues/262

joeaudette:
Hi,

In the announcement for the breaking changes you said:

"In 1.0, it was possible to configure different Authentication middleware with branching, this is no longer possible with a single middleware and shared services across all branches. A workaround could be to use different schemes/options for each branch."

I am facing this problem and would appreciate some elaboration on specifically how I can use different schemes per branch. My solution can provision new tenants at runtime, so iterating through tenants in startup to add schemes is not a solution for me. 

Wondering if you could provide some guidance on what I could override to dynamically resolve schemes and options per request. [PostConfigureCookieAuthenticationOptions](https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.Cookies/PostConfigureCookieAuthenticationOptions.cs), looks like a potential candidate that might allow me to do that, but not sure if that allows me to adjust the options per request. IAuthenticationSchemeProvider sounds like another potential thing to implement. 

Could you offer a suggestion of the right thing to implement/override to achieve a way to resolve auth scheme/options on a per request basis?

Thanks in advance for any advice!

HaoK:
Alternatively you could also try replacing the 'IAuthenticationHandlerProvider` to return the appropriate instance for your branch and use the same name in each branch.  The default implementation of this is just a scoped dictionary cache: https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Authentication.Core/AuthenticationHandlerProvider.cs

If you had a implementation that was branch aware, that should be a step in the right direction.  You'll still have to figure out how to resolve the appropriate options, but you could do that via plugging in your own IOptionsMonitor that is branch aware.

joeaudette:
Thank you! Really appreciate the prompt response! I will give IAuthenticationHandlerProvider a try.

Can you tell me what is the purpose of IOptionsMonitor? what is it used for? I'm familiar with IOptions of T but IOptionsMonitor is new to me.

HaoK:
Its basically the same thing, it just adds the ability to name options `IOptionsMonitor.Get(authenticationScheme)` is what you will need to implement

joeaudette:
Thanks! 
[...]

PR #1550, opened 2017-11-23
Add virtual scheme functionality into base

HaoK:
- Removed VirtualSchemeOptions
- Added to AuthenticationSchemeOptions as
Forward[Authenticate/Challenge/Forbid/Selector]
- Added abstract SignIn/OutAuthenticationHandler base classes
- VirtualAuthenticationHandler uses these, will update existing handlers
to use the new base classes once things are stable

This makes social auth no longer even need a 3rd 'virtual' scheme.  Instead, the default cookie scheme can set ForwardChallenge = "Google".

@Tratcher @davidfowl


src/Microsoft.AspNetCore.Authentication/AuthenticationSchemeOptions.cs
```diff
@@ -36,5 +37,21 @@ public virtual void Validate(string scheme)
         /// If set, will be used as the service type to get the Events instance instead of the property.
         /// </summary>
         public Type EventsType { get; set; }
+
+        // Scheme forwarding properties
+
+        public string ForwardDefault { get; set; }
```
Tratcher: These are going to need thorough explanations, they're very meta.

Does ForwardDefault make sense for anything but the virtual handler? If you set it on something like Google then none of Google's functions would work, right?

src/Microsoft.AspNetCore.Authentication/AuthenticationSchemeOptions.cs
```diff
@@ -36,5 +37,21 @@ public virtual void Validate(string scheme)
...
+        public string ForwardAuthenticate { get; set; }
+        public string ForwardChallenge { get; set; }
+        public string ForwardForbid { get; set; }
+        public string ForwardSignIn { get; set; }
+        public string ForwardSignOut { get; set; }
+
+        /// <summary>
+        /// Used to select a default scheme to target based on the request.
+        /// </summary>
+        public Func<HttpContext, string> ForwardDefaultSelector { get; set; }
```
Tratcher: The selector also only makes sense for the virtual handler. Using it on any other handler prevents that handler from ever working.

HaoK: Every handler is a virtual handler now, so a better way to think about this is to think of the forwarding as a selective opt out of that handlers functionality, so lets say you have a google handler that you want to do something else for /custom/ requests, that's what you would do.

That's basically what this feature is, if you set the default selector and it returns something non null, that's basically saying to the handler to get out of the way and let another scheme handle everything.

src/Microsoft.AspNetCore.Authentication/VirtualAuthenticationHandler.cs
```diff
@@ -1,71 +1,29 @@
...
-        protected HttpContext Context { get; private set; }
+        public VirtualAuthenticationHandler(IOptionsMonitor<AuthenticationSchemeOptions> options, ILoggerFactory logger, UrlEncoder encoder, ISystemClock clock) : base(options, logger, encoder, clock)
+        { }
 
-        public VirtualAuthenticationHandler(IOptionsMonitor<VirtualSchemeOptions> options)
-        {
-            OptionsMonitor = options;
-        }
+        protected override Task<AuthenticateResult> HandleAuthenticateAsync()
+            => Context.AuthenticateAsync();
```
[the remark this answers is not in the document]
HaoK: Yeah I'll have to add some checks somewhere to blowup with a better exception for the circular refs

src/Microsoft.AspNetCore.Authentication/AuthenticationSchemeOptions.cs
HaoK: A more concrete example would be for apps with a single social auth, lets say google.

That could be setup now without a 3rd virtual scheme as:
```
   services.AddAuthentication("Cookies").AddCookie(o => o.ForwardChallenge = "Google").AddGoogle()
```

HaoK: Note: that's basically what the new `services.AddAuthentication().AddGoogleSignIn()` sugar will setup by default (add a cookie if it doesn't exist, and then point it at "Google" for challenges, the equivalent for AddIdentity())

Tratcher: So an app that wanted to do API vs Cookies would make Cookies the default and set the ForwardDefaultSelector on CookieAuthOptions?

How about ForwardDefault? Is that ever useful outside of the virtual handler?

leastprivilege:
Can someone describe the real world scenario that this change solves?

HaoK:
This is mostly targetting making it easier for these two scenarios:
- Mix different auth schemes (i.e. cookies and bearer)
- Today's AuthenticationSchemes property on AuthorizationPolicies is not powerful enough to express what we were trying to enable, what's actually needed there was the ability to configure a coherent set of challenge/authenticate/forbid behaviors for a specific authorization policy (which is what we are calling a virtual scheme, but its basically an authentication policy to match an authorization policy)

src/Microsoft.AspNetCore.Authentication/AuthenticationSchemeOptions.cs
HaoK: I think its still useful as the idea is all handlers are now able to be used as virtual schemes, but I see the problem, now that this is merged, we need a way to opt out of the ForwardDefault.

So I think the tweak is simply to set ForwardAuthenticate to yourself

```
   services.AddAuthentication("Google").AddGoogle(o => {
      o.ForwardDefault = "Cookies"
      o.ForwardChallenge = "Google"
   }
```

This would block falling back to ForwardDefault, and I can add a scheme name check to call into HandleXyzAsync() if forward is targeting self.

leastprivilege:
> Mix different auth schemes (i.e. cookies and bearer)

What's the exact problem with that today? Besides that they shouldn't be mixed?

davidfowl:
> What's the exact problem with that today? Besides that they shouldn't be mixed?

1. It makes it possible to simplify authentication schemes that require a different combination of values for Challenge/SignIn/SignOut/Authenticate etc. It means we can introduce higher level schemes that basically combine lower level ones. A concrete example of this would be a Google+Cookies authentication scheme that has Challenge = Google, EverythingElse = Cookies.

2. It makes it possible to dynamically select a scheme on a per request basis. There's been consistent feedback that what we have today is extremely limiting (specifying authentication schemes on the authorize attribute). See https://github.com/aspnet/Security/issues/1338. This was another concrete scenario driving this discussion https://github.com/aspnet/Security/issues/1469

leastprivilege:
> It means we can introduce higher level schemes that basically combine lower level ones

Oh well - sure even more shielding of devs from the way it works. Besides that Google+Cookies is never the answer. But that's a different discussion

> It makes it possible to dynamically select a scheme on a per request basis.

OK - if that makes it easier to select the right validator for a bearer token, then this is an improvement. Not sure if you had that use case in mind...

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
```diff
@@ -118,8 +118,17 @@ protected AuthenticationHandler(IOptionsMonitor<TOptions> options, ILoggerFactor
         protected string BuildRedirectUri(string targetPath)
             => Request.Scheme + "://" + Request.Host + OriginalPathBase + targetPath;
 
+        protected virtual string ResolveTarget(string scheme)
+            => scheme ?? Options.ForwardDefaultSelector?.Invoke(Context) ?? Options.ForwardDefault;
```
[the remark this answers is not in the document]
HaoK: Updated added tests and checks to prevent reentering the methods, we don't support concurrency in our handlers today right?  Each request should have its own handler by default...

HaoK:
Updated PR adding self forwarding checks and a test that verifies we get the proper error for all of the existing handlers.  Its a shame we didn't add the base SignIn/SignOut handlers in 2.0, can't use the new base handler pattern since the existing handlers expose the interface methods as public virtual :(

HaoK:
One more update, allowed explicit target to Scheme.Name to skip forwarding, which allows setting of ForwardDefault and selectively pointing at yourself to opt out of some forwarding

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -240,119 +243,160 @@ private CookieOptions BuildCookieOptions()
                 throw new ArgumentNullException(nameof(user));
             }
 
-            properties = properties ?? new AuthenticationProperties();
+            if (_inSignIn)
```
Tratcher: ???? 😕 this smells terrible.
Did you repro this error? How?

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
```diff
@@ -118,27 +121,49 @@ protected AuthenticationHandler(IOptionsMonitor<TOptions> options, ILoggerFactor
...
+                throw new InvalidOperationException("Authenticate for scheme:[" + Scheme.Name + "] resulted in a recursive call back to itself.");
+            }
+
+            _inAuthenticate = true;
+            try
             {
-                var ticket = result?.Ticket;
-                if (ticket?.Principal != null)
+                var target = ResolveTarget(Options.ForwardAuthenticate);
+                if (target != null && !string.Equals(target, Scheme.Name, StringComparison.Ordinal))
```
Tratcher: helper method?
Why didn't you do this self-reference check in CookieAuth?

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
HaoK: Yeah without this guard its trivial to hit, all it takes is a virtual scheme loop of any sort we have indirection now...  so 

Scheme A => Scheme B => Scheme A...  I added the alias to every test which is the minimal cycle.

HaoK: And yes I totally agree its kinda gross

src/Microsoft.AspNetCore.Authentication/AuthenticationHandler.cs
HaoK: yeah I guess I can push the name comparison into Resolve and have it return null when its self referencing

leastprivilege:
> This makes social auth no longer even need a 3rd 'virtual' scheme. Instead, the default cookie scheme can set ForwardChallenge = "Google".

You are aware that this is not how it works in the real world? For 90% of the time you want to show some sort of page before challenging the external provider? And this will then probably also use the external cookie for signin...

I have the feeling you are building something here where you don't understand the real world requirements.

leastprivilege:
@davidfowl @HaoK 

Will this feature help me to invoke different authentication handlers based on the "format" or "shape" of an incoming bearer token?

This is something that is really needed.

We had to jump through a couple of hoops to make it work

https://github.com/IdentityServer/IdentityServer4.AccessTokenValidation/blob/c5631a3fa91aeeb97e40d58a8ed534c321acd5f6/src/IdentityServer4.AccessTokenValidation/IdentityServerAuthenticationHandler.cs#L51-L61

HaoK:
> You are aware that this is not how it works in the real world? For 90% of the time you want to show some sort of page before challenging the external provider? And this will then probably also use the external cookie for signin...

Complex flows like you describe here (like what identity needs) will always have to build their own custom authentication "policy", that's not really any different than today, except it can be expressed using a named virtual scheme instead of shoehorning things into the single set of defaults.  This would also let authentication frameworks target their own named scheme, i.e. Identity can target the "Identity" scheme instead of hardcoding a bunch of constant schemes.

HaoK:
It gets you closer, you are given the ability to dispatch based on the context, but you would still have to extract the token and return the appropriate scheme name, so...

```
   // Configure the selector in your IdentityServerAuthenticationOptions
  o.ForwardDefaultSelector = ctx => {
     var token =  Options.TokenRetriever(Context.Request);
     if (token != null) {
        if (token.Contains('.') && Options.SupportsJwt) 
            return Scheme.Name + IdentityServerAuthenticationDefaults.JwtAuthenticationScheme;
        else if (Options.SupportsIntrospection)
            return Scheme.Name + IdentityServerAuthenticationDefaults.IntrospectionAuthenticationScheme;
        return null;
     }
  }

   // And then you can update HandlerAuthenticateAsync to return NoResult always.
```

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
muratg: Will stack look informative when they get this exception?

HaoK:
Updated cleaning up the ResolveTarget helper to also take care of the self targeting check.  Unfortunately the circular reference issue is unavoidable with scheme forwarding in general, I could check for it using a hashset with the scheme name instead of the bool per action handler, that might look a little better

HaoK:
Actually never mind, tracking visited schemes via hashset is a lot messier than the flag per handler

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
HaoK: Its not too bad:
```
Authenticate for scheme:[Google] resulted in a recursive call back to itself.
   at Microsoft.AspNetCore.Authentication.AuthenticationHandler`1.<AuthenticateAsync>d__51.MoveNext() in C:\Github\Security\src\Microsoft.AspNetCore.Authentication\AuthenticationHandler.cs:line 141
   at Microsoft.AspNetCore.Authentication.AuthenticationService.<AuthenticateAsync>d__10.MoveNext()
   at Microsoft.AspNetCore.Authentication.AuthenticationHandler`1.<AuthenticateAsync>d__51.MoveNext() in C:\Github\Security\src\Microsoft.AspNetCore.Authentication\AuthenticationHandler.cs:line 147
   at Microsoft.AspNetCore.Authentication.AuthenticationService.<AuthenticateAsync>d__10.MoveNext()
   at Microsoft.AspNetCore.Authentication.AuthenticationHandler`1.<AuthenticateAsync>d__51.MoveNext() in C:\Github\Security\src\Microsoft.AspNetCore.Authentication\AuthenticationHandler.cs:line 147
   at Microsoft.AspNetCore.Authentication.AuthenticationService.<AuthenticateAsync>d__10.MoveNext()
   at Xunit.Assert.<RecordExceptionAsync>d__79.MoveNext() in C:\Dev\xunit\xunit\src\xunit.assert\Asserts\Record.cs:line 82
```

HaoK: I'll update the error messages to talk about the appropriate option, so something more explicit like:

"Authenticate for scheme:[Google] resulted in a recursive call back to itself. Check for cycles in either ForwardAuthenticate, ForwardDefault, or ForwardDefaultSelector."

Tratcher:
These cycle problems a deal breaker. We need to step back and redesign this. I think we can achieve most of our goals using the current forwarding handler without introducing these cycle complexities. I'll be in the office tomorrow and we can go over it.

HaoK:
Just the concept of a forward handler creates cycle detection issues, that logic needs to exist in the forwarding code regardless, since you can always forward A to B which forwards back to A.

So I'm not sure how we can design away the issue

Tratcher:
The issue becomes worse the deeper you integrate the forwarding. For the forwarding handler in dev the risks are much lower because you primarily forward from one virtual handler to as series of concrete handlers, and those handlers have no mechanic for forwarding back to you. In the worst case you point two forwarding handlers at eachother. Even if we needed this kind of cycle detection for those cases it would be contained to the virtual handler, not spread through every call path.

I think we can go back to using the distinct virtual handler and still achieve our design goals without introducing these complexities to the rest of the stack.

HaoK:
So is your main concern the gross _inSignIn checks in every method being part of the base methods instead of only the virtual handler class?  

To me, the existence of that code is painful regardless of where it lives, but if it is going to exist anyways, deep integration of virtual schemes/handlers at least gives the benefits of making it possible to glue two schemes together without introducing a 3rd pure virtual scheme (i.e. mix cookies and an oidc via forwarding, instead of cookies + oidc + virtual).  

When forwarding isn't being used, the checks actually still have some value, since now that I think about it, didn't we have cases in the past where event code tries to call Authenticate on itself would blow up?  I'll see if I can reproduce that issue tomorrow.

HaoK:
Filed https://github.com/aspnet/Security/issues/1578 to track hardening against reentrancy/circular references in general, as this isn't something unique to virtual handlers, events and sign in scheme already have the same issue.  I'll remove the checks and tests from this PR

HaoK:
Updated removing all the safeguard checks

HaoK:
Updated the doc comments:

ForwardDefault:
```
        /// <summary>
        /// If set, this specifies a default scheme that authentication handlers should forward all authentication operations to
        /// by default. The default forwarding logic will check the most specific ForwardAuthenticate/Challenge/Forbid/SignIn/SignOut 
        /// setting first, followed by checking the ForwardDefaultSelector, followed by ForwardDefault. The first non null result
        /// will be used as the target scheme to forward to.
        /// </summary>
```

ForwardXyz:
```
        /// <summary>
        /// If set, this specifies the target scheme that this scheme should forward AuthenticateAsync calls to.
        /// For example Context.AuthenticateAsync("ThisScheme") => Context.AuthenticateAsync("ForwardAuthenticateValue");
        /// Set the target to the current scheme to disable forwarding and allow normal processing.
        /// </summary>
```

ForwardDefaultSelector:
```
        /// <summary>
        /// Used to select a default scheme for the current request that authentication handlers should forward all authentication operations to
        /// by default. The default forwarding logic will check the most specific ForwardAuthenticate/Challenge/Forbid/SignIn/SignOut 
        /// setting first, followed by checking the ForwardDefaultSelector, followed by ForwardDefault. The first non null result
        /// will be used as the target scheme to forward to.
        /// </summary>
```

src/Microsoft.AspNetCore.Authentication/VirtualAuthenticationHandler.cs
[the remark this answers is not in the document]
Tratcher: Shouldn't this throw an InvalidOperationException? In what case would it make sense to call the virtual handler's HandleAuthenticateAsync method? Or any of its Handle methods? All forwarding should be handled in the base class.

test/Microsoft.AspNetCore.Authentication.Test/MicrosoftAccountTests.cs
```diff
@@ -53,6 +53,58 @@ public class MicrosoftAccountTests
             Assert.Equal(MicrosoftAccountDefaults.AuthenticationScheme, scheme.DisplayName);
         }
 
+        // Stack overflow checking is not currently implemented
+        //[Fact]
+        //public async Task TargetsSelfDoesntStackOverflow()
```
Tratcher: remove?

test/Microsoft.AspNetCore.Authentication.Test/OAuthTests.cs
```diff
@@ -48,6 +49,61 @@ public class OAuthTests
             Assert.Equal(OAuthDefaults.DisplayName, scheme.DisplayName);
         }
 
+        // Stack overflow checking is not currently implemented
+        //[Fact]
+        //public async Task TargetsSelfDoesntStackOverflow()
```
Tratcher: remove?

test/Microsoft.AspNetCore.Authentication.Test/OpenIdConnect/OpenIdConnectConfigurationTests.cs
```diff
@@ -3,10 +3,12 @@
 
 using System;
 using System.Net;
+using System.Security.Claims;
```
Tratcher: ?

test/Microsoft.AspNetCore.Authentication.Test/VirtualHandlerTests.cs
```diff
@@ -277,21 +280,65 @@ public class VirtualHandlerTests
...
+            var services = new ServiceCollection().AddOptions().AddLogging();
+
+            services.AddAuthentication(o =>
+            {
+                o.AddScheme<TestHandler>("auth1", "auth1");
+                o.DefaultScheme = "auth1";
+            })
+            .AddScheme("forward", "forward", p =>
+            {
+                p.ForwardDefault = "forward";
```
Tratcher: This looks invalid and should throw as discussed in the virtual handler.

test/Microsoft.AspNetCore.Authentication.Test/VirtualHandlerTests.cs
```diff
@@ -335,6 +382,66 @@ public class VirtualHandlerTests
...
+            var services = new ServiceCollection().AddOptions().AddLogging();
+
+            services.AddAuthentication(o =>
+            {
+                o.AddScheme<TestHandler>("auth1", "auth1");
+            })
+            .AddScheme<AuthenticationSchemeOptions, TestAuthHandler>("forward", "forward", p =>
+            {
+                p.ForwardDefault = "auth1";
+                p.ForwardAuthenticate = "forward";
```
Tratcher: Scenario? this seems invalid.

HaoK: This enables setting the remote auth as the default which forwards everything to the cookie, except challenge which it continues to handle.:

```
services.AddAuthentication("Google").AddGoogle(o => {
  o.ForwardDefault = "Cookies";
  o.ForwardChallenge = "Google";
}).AddCookies();
```

But more generally, this is the only way to turn off forwarding for a specific action when ForwardDefault is set, you have to set it to yourself to opt out of forwarding.

test/Microsoft.AspNetCore.Authentication.Test/VirtualHandlerTests.cs
HaoK: Yeah this might not be needed anymore now that the virtual logic has been merged into all handlers

HaoK: But if we go that route, we probably need to always require a ForwardDefault being set for a pure virtual handler scheme

Tratcher: Or a selector.

test/Microsoft.AspNetCore.Authentication.Test/VirtualHandlerTests.cs
Tratcher: 👃

Tratcher: Your reverse example was cleaner:
```
services.AddAuthentication("Cookies").AddCookies(o => {
  o.ForwardChallenge = "Google";
}).AddGoogle();
```

HaoK: yes you would set it up that way typically, but the inverse should work as well.

test/Microsoft.AspNetCore.Authentication.Test/VirtualHandlerTests.cs
HaoK: Nah the selector doesn't guarantee a fallback, since the selector can return null, I think we need to force a default target which we can do via the AddScheme overload, since this type exists and has different behavior, we should probably just go back to calling it AddVirtualScheme("virtualScheme", "displayName", "defaultTarget")

test/Microsoft.AspNetCore.Authentication.Test/VirtualHandlerTests.cs
Tratcher: I don't know that it needs to. I'd be easier to explain if self-references were consistently treated as error cases.

test/Microsoft.AspNetCore.Authentication.Test/VirtualHandlerTests.cs
Tratcher: Or throw if anything ever actually reaches the virtual handler's HandleX methods.

test/Microsoft.AspNetCore.Authentication.Test/VirtualHandlerTests.cs
HaoK: That's not really the issue, this was to prevent adding a enableFowardXyz to turn off the forwarding logic per action, setting it to yourself is that switch.

Tratcher: This is a very confusing mechanic.

This is OK:
```
.AddGoogle(o => {
  o.ForwardDefault = "Cookies";
  o.ForwardChallenge = "Google";
})
```
But this causes stack overflows:
```
.AddGoogle(o => {
  o.ForwardChallenge = "Google";
})
```

Trying to support this use case is a landmine. We'd be better off blocking it and requiring you do do one of the other two ways to do this, 1) the inverted example of challenging Google from Cookies shown above, or 2) don't use ForwardDefault in this scenario, set all of the other Forwards to Cookies. It wouldn't disable functionality, it would require a more coherent config.

Tratcher:
Wasn't the big thing Fowler wanted with this PR new combine extensions for .AddGoogle/Facebook/Etc. that included Cookies for you? I think it would be more effective to start with those and work backwards.

test/Microsoft.AspNetCore.Authentication.Test/VirtualHandlerTests.cs
HaoK: No, you don't get stack overflows with self targetting ever, think of self targeting only as a mechanism for turning off the forward.  So your example is a no-op

```
.AddGoogle(o => {
  o.ForwardChallenge = "Google";
})
```

The effective forwards are computed in this order until a non null is found:
1. ForwardXyz
2. ForwardSelector
3. ForwardDefault

Without the special case for self forwarding meaning don't forward, there would be no way to selectively turn forwarding off.

HaoK:
I have those extensions in another PR, the main issue there is where they should live, currently the auth packages don't depend on cookies, we need some package that depends on auth + cookies, unless just depend on cookies for everything

There's nothing too interesting about those methods, they basically just all look like:

```C#
   AddGoogleSignIn(this AuthenticationBuilder builder, Action<GoogleOptions> configure) { 
      builder.AddAuthentication("Cookies")
            // Don't add the default cookie if it already exists, but always configure it
           .TryAddCookie(o = o.ForwardDefault = "Google")
           .AddGoogle(configure)
   }
```

Typically you'd only call one of each, the templates wouldn't use any of these except for maybe `AddOpenIdConnectSignIn`

Tratcher:
Does this do about the same using the checked in code?
```
        public static IServiceCollection AddOpenIdConnectSignIn(this IServiceCollection services, Action<OpenIdConnectOptions> configure)
        {
            {
                options.DefaultScheme = "Cookies";
                options.DefaultChallengeScheme = "OpenIdConnect";
            })
            // .TryAddCookies()
            .AddOpenIdConnect(configure);

            return services;
        }
or
        public static IServiceCollection AddOpenIdConnectSignIn(this IServiceCollection services, Action<OpenIdConnectOptions> configure)
        {
            {
                options.DefaultScheme = "OIDCSignIn";
                options.DefaultSignInScheme = "Cookies";
            })
            // .TryAddCookies()
            .AddOpenIdConnect("OpenIdConnect", "", configure)
            .AddVirtualScheme("OIDCSignIn", "OpenIdConnect", options =>
            {
                options.Default = "Cookies";
                options.Challenge = "OpenIdConnect";
            });

            return services;
        }
```

HaoK:
Yup pretty much a (much) cleaner way to express that

Tratcher:
At the cost of adding unnecessary complexity to the core classes. I'm putting this on hold until we can review it in person with Eilon.

HaoK:
That's fine, I can schedule an in person review with @davidfowl and @Eilon when everyone is back

HaoK:
But I'd say its better to look at the whole virtual scheme/forwarding feature as more of a new core concept rather than some tangential functionality limited to a special virtual handlers.

Tratcher:
We're having a hard time keeping track of what their for though. The two scenarios I'm aware of are API + Cookies which is handled by the existing DefaultSelector, and AddOpenIdConnectSignIn which doesn't require virtual schemes at all.

HaoK:
So remember what's driving this change, its not meant to be a targeted thing, but a brand new concept.  This is coming from the main AuthZ improvements issue: https://github.com/aspnet/Security/issues/1359, and https://github.com/aspnet/Security/issues/1469 for more context

The new core concept we've been iterating on is AuthenticationPolicies and the ability to have logical schemes, aka VirtualSchemes, the two scenarios you listed are specific examples of why you would use policies.   Basically keeping VirtualSchemes a separate thing means authentication policies are only "VirtualSchemes", while integrating them means any auth handler using our base classes (including our built in handlers, and all existing custom handlers) will be able to take advantage of the new policy concept without having to add any new schemes. 

The main thing that is fundamentally broken today with `[Authorize]` without AuthenticationPolicies is the whole multiple AuthenticationSchemes being unable to effectively express what we want.  Going forward now that we have policies, we can safely state that `[Authorize]` and AuthorizationPolicies in general should only ever point to a single AuthenticationScheme, if apps need to compose different schemes or do anything more advanced, they can use the new AuthenticationPolicy/VirtualScheme/Forwarding stuff we are adding. And in 3.0, we can probably look at deprecating the comma splitting/ability to specify multiple schemes entirely from `[Authorize]`/`AuthorizationPolicy`

HaoK:
Note, the default schemes we have in AuthenticationOptions are similarly broken to [Authorize], we could consider deprecating those as well in 3.0, and only support setting a single DefaultScheme, and require them to use the virtual scheme stuff instead.

test/Microsoft.AspNetCore.Authentication.Test/VirtualHandlerTests.cs
HaoK: We can discuss this more in the design review, but as soon as we force a default forward for any virtual scheme, there actually is no reason to even support pure virtual schemes.  They can just use whatever the default forward and configure that as desired, the only slight difference would be if they wanted to use a non cookie default and attempt to forward something like sign-in which isn't always supported.

HaoK: AuthSamples.PathBasedSelection would simply become:

```
            services.AddAuthentication("DynamicCookie")
                .AddScheme<AuthenticationSchemeOptions, ApiAuthHandler>("Api", o => { })
                .AddCookie("DynamicCookie", options =>
                {
                    o.DefaultSelector = ctx => ctx.Request.Path.StartsWithSegments("/api") ? "Api" : "DynamicCookie";
                    options.AccessDeniedPath = "/account/denied";
                    options.LoginPath = "/account/login";
                })
```

HaoK:
We reviewed this feature with @Eilon @blowdart @davidfowl @Tratcher yesterday and we're going with "schemes as policies" as a feature implemented by the base AuthenticationHandlers for preview2.

I don't think we have any more issues in this PR regarding the logic in AuthenticationHandler itself.  @Tratcher can you double check to make sure all of the concerns you had there have been addressed?

I believe all of the remaining issues to resolve are with the VirtualHandler.  But I'm proposing we table those for now, I'd like to get this PR in first, then update AuthSamples to stop using VirtualSchemes and switch to the new pattern.  Then I'll do a follow up PR deleting the VirtualAuthenticationHandler and AddScheme overload and adding any tweaks needed to make sure we don't miss anything (I believe the only thing we might need to keep from the stand along virtual scheme stuff is potentially the new SignIn/SignOutAuthenticationHandler base classes, I'm not sure this has much value since non of our existing handlers would be able to use these classes, it would only help future handlers or custom ones...

Tratcher:
My concerns have been overruled. You'll need @davidfowl or @Eilon to sign off on this.

HaoK:
I believe in the final state, the only additions we have from 2.0 will be just the additional of the 6 new forward options in `AuthenticationSchemeOptions`, a new protected helper in AuthenticationHandler that resolves the forwarding option, and using that helper method in the implementations of Authenticate/SignIn/SignOut/Challenge/Forbid.  (AuthenticationHandler + the sign in /sign out handlers like cookies/oidc)

HaoK:
Replaced by https://github.com/aspnet/Security/pull/1625 and subsequent PRs

PR #1587, opened 2018-01-03
[Design] Example UseGoogleSignIn

HaoK:
Starting to play with the shape of the proposed Use/AddXyzSignIn...

```C#
services.AddAuthentication()
    // Pick one, last one will win (set to DefaultScheme) if more than one called (hence Use over Add)
   .UseGoogleSignIn(o => { })
   .UseFacebookSignIn(o => { })
   .UseOAuthSignIn(o => { })
   .UseOpenIdConnectSignIn(o => { })
   .UseTwitterSignIn(o => { })
   .UseIdentitySignIn()

// If you want to custom configure the cookie, you need to know which cookie
services.Configure<CookieAuthenticationOptions>(CookieAuthenticationDefaults.AuthenticationScheme, o => o.ExpireTimeSpan = 100);
// or with identity
services.ConfigureApplicationCookie(o => o.ExpireTimeSpan = 100);
```

Each Use is responsible for configuring the defaults (and adding any cookies as needed).  All of the social sign ins would use the default cookie.  Identity would probably continue to use its existing Identity.Application cookie as the default

We could also consider standardizing on a new constant "SignInCookie" as the scheme name for the cookie so these are all consistent, this would require some work for Identity to make it work, since SignInManager hardcodes IdentityConstants.ApplicationScheme today...

@Tratcher @davidfowl


samples/SocialSample/SignInExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+                // Override instead of default since this method is opinionated on the cookie scheme name.
+                o.SignInScheme = CookieAuthenticationDefaults.AuthenticationScheme;
+            });
+
+            builder.ConfigureCookie(GoogleDefaults.AuthenticationScheme);
+            return builder;
+        }
+
+        // Add the defalut cookie if needed, set it to default and configure the cookie to forward challenge
+        private static void ConfigureCookie(this AuthenticationBuilder builder, string forwardChallenge)
```
Tratcher: Aren't these also supposed to set cookies as the overall default scheme? Or how else does it work with `[Authroize]`

HaoK: It does on line 31:

`o.DefaultScheme = CookieAuthenticationDefaults.AuthenticationScheme;`

Tratcher: Ah, I mistook that for the new virtual DefaultScheme.

samples/SocialSample/SignInExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+                o.DefaultScheme = CookieAuthenticationDefaults.AuthenticationScheme;
+                if (!o.SchemeMap.ContainsKey(CookieAuthenticationDefaults.AuthenticationScheme))
+                {
+                    // TODO: make it easier to reuse AddCookie
+                    builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IPostConfigureOptions<CookieAuthenticationOptions>, PostConfigureCookieAuthenticationOptions>());
+                    builder.AddScheme<CookieAuthenticationOptions, CookieAuthenticationHandler>(CookieAuthenticationDefaults.AuthenticationScheme, CookieAuthenticationDefaults.AuthenticationScheme, _ => { });
+                }
+            });
+            // Always configure the cookie's challenge
+            builder.Services.Configure<CookieAuthenticationOptions>(CookieAuthenticationDefaults.AuthenticationScheme, o => o.ForwardChallenge = forwardChallenge);
```
Tratcher: How is this better than setting DefaultChallengeScheme adjacent to DefaultScheme above?

HaoK: That only affects empty Challenge()...

This means all challenges that end up going to Cookies (including [Authorize]'s default), as well as Challenge("Cookies") go to google. which is what you want.

HaoK: For all intensive purposes this makes it impossible to challenge the cookie

HaoK:
Replaced by https://github.com/aspnet/Security/pull/1644

PR #1591, opened 2018-01-04
Make consent cookie value public

### LINKED ISSUE #1590: Add API to the ITrackingConsentFeature to generate the cookie for javascript use
Damian wants to set the consent cookie from a javascript action, but wants to honor all of the middleware settings for it. Add an API to the feature interface that generates the full cookie for you.

user1336:
Fixes #1590

### DIFF
--- a/src/Microsoft.AspNetCore.CookiePolicy/ResponseCookiesWrapper.cs
+++ b/src/Microsoft.AspNetCore.CookiePolicy/ResponseCookiesWrapper.cs
@@ -10,8 +10,6 @@ namespace Microsoft.AspNetCore.CookiePolicy
 {
     internal class ResponseCookiesWrapper : IResponseCookies, ITrackingConsentFeature
     {
-        private const string ConsentValue = "yes";
-
         private bool? _isConsentNeeded;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.CookiePolicy/ResponseCookiesWrapper.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.AspNetCore.CookiePolicy
 {
     internal class ResponseCookiesWrapper : IResponseCookies, ITrackingConsentFeature
```
HaoK: The class is internal so this won't help much

user1336: How did I miss that, should I add a Defaults class?

Tratcher: Wait, Damian gave me some additional feedback yesterday that may make this irrelevant.

Tratcher:
Closing this for now as the requirements have significantly changed.

PR #1595, opened 2018-01-05
Implement ITrackingConsentFeature.CreateConsentCookie()

Tratcher:
#1590 @DamianEdwards 
This reduces the amount of work that needs to be done in the template when they want to set the consent cookie on the client via javascript.

See https://github.com/aspnet/templating/pull/213/files#diff-947841bf97b345569d23143ef4af84af

This requires https://github.com/aspnet/HttpAbstractions/commit/ce6934126513f1da9e46b8d695fedb2ddf261ae1


Tratcher:
None that I know of.

Issue #1448, opened 2017-09-22
OIDC should use AuthenticationProtocolMessage.BuildFormPost()


Tratcher:
When OIDC does redirects using the FormPost mode it generates an HTML form redirect.
https://github.com/aspnet/Security/blob/31dd4ebd9c0c26fc9592aa08d7fb3804bf34011b/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L237-L252

However, `AuthenticationProtocolMessage.BuildFormPost()` already has this code so we can remove our redundant copy.

PinpointTownes:
This doesn't sound like a good idea.

When I introduced POST authorization requests support in the OIDC middleware 2 years ago, I deliberately avoided using `BuildFormPost()` for a simple reason: [it was - and is still - broken as it URLencodes the form action attribute](https://github.com/aspnet/Security/pull/392#r36568911).

Tratcher:
Fair, but we can fix it first.
https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/755
@brentschmaltz

brentschmaltz:
@Tratcher @PinpointTownes ,  @616b2f fixed OIDC 755

Tratcher:
This depends on getting updated versions. See https://github.com/aspnet/Security/issues/1468

PR #1597, opened 2018-01-08
OIDC: Use IdentityModel redirect form generator

Tratcher:
#1448 @PinpointTownes 
We don't need this code, IdentityModel can do it for us.


PinpointTownes:
Should @blowdart sign off this change? Unlike `OpenIdConnectHandler`, the fixed `AuthenticationProtocolMessage` uses `WebUtility.HtmlEncode` instead of the new web encoders, which may have security implications.

Tratcher:
They're both doing HTML encoding, they only differ on some of their defaults for how aggressively they encode.

PinpointTownes:
> They're both doing HTML encoding, they only differ on some of their defaults for how aggressively they encode.

That was my point. That said, after briefly looking at the source code, it looks like the only difference is that the newer encoders encode `+` to prevent subtile attacks that require forcing the page encoding to UTF-7. Since it shouldn't possible in modern browsers, I guess it's not a big deal.

PR #1611, opened 2018-01-17
Don't overwrite pre-authenticated identities.

berdon:
Seems 2.0 introduced a bug where the Authentication middleware will overwrite any current identities on the user principal.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs
@@ -56,10 +56,7 @@ namespace Microsoft.AspNetCore.Authentication
                 {
                     var identities = context.User.Identities;
                     context.User = result.Principal;
-                    if (identities != null)
-                    {
-                        context.User.AddIdentities(identities);
-                    }
+                    context.User.AddIdentities(identities);
                 }
             }
 
--- a/src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs
@@ -54,7 +54,7 @@ namespace Microsoft.AspNetCore.Authentication
                 var result = await context.AuthenticateAsync(defaultAuthenticate.Name);
                 if (result?.Principal != null)
                 {
-                    var identities = context.User?.Identities;
+                    var identities = context.User.Identities;
                     context.User = result.Principal;
                     if (identities != null)
                     {
--- a/src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs
@@ -54,7 +54,12 @@ namespace Microsoft.AspNetCore.Authentication
                 var result = await context.AuthenticateAsync(defaultAuthenticate.Name);
                 if (result?.Principal != null)
                 {
+                    var identities = context.User?.Identities;
                     context.User = result.Principal;
+                    if (identities != null)
+                    {
+                        context.User.AddIdentities(identities);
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs
```diff
@@ -54,11 +54,16 @@ public AuthenticationMiddleware(RequestDelegate next, IAuthenticationSchemeProvi
                 var result = await context.AuthenticateAsync(defaultAuthenticate.Name);
                 if (result?.Principal != null)
                 {
+                    var identities = context.User?.Identities;
```
Tratcher: HttpContext.User never returns null, it at least returns an anonymous identity.

berdon: Good to know!

HaoK:
What behavior are you expecting in this situation you expect the identities to be merged?

HaoK:
In this situation I would just not use the defaultAuthenticate and control what goes into httpContext.User explicitly yourself

berdon:
@HaoK I'm expecting the same behavior as before.

PreAuthMiddleware adds IdentityX onto the HttpContext.User
(Cookie|Jwt|Etc)Middleware gets called and "authenticates"

HttpContext.User.Identities contains IdentityX.

This was true pre 2.0 - the 2.0 change broke this. Is this a faulty assumption that those identities should exist post Authentication middleware?

HaoK:
Yes 1.0 did merging, but that was also "AutomaticAuthenticate" and more than one could be true.  It doesn't make as much sense to merge a "defaultAuthenticate".  Merging is really something the app should be doing anyways as order of identities matters in some cases.

HaoK:
@berdon so in 2.0, its probably easier to do it as a PostAuth middleware that adds whatever identity you want to the HttpContext.User

HaoK:
Alternatively you might be able to just do this as a IClaimsTransformation too, as that would get applied automatically after all Authenticate calls...  https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Authentication.Abstractions/IClaimsTransformation.cs

berdon:
"Merging is really something the app should be doing anyways as order of identities matters in some cases."

Yeah, that. What if the order here does matter and now authenticate is clearing the identities?

HaoK:
Your app set defaultAuthenticate to something, that results in the user having that identity after UseAuthentication.  Clear the option if you don't want that behavior

berdon:
Will an authentication handler, ie. Jwt, fire still without that set?

HaoK:
No, you will need to call authenticate yourself, i.e. when you merge the identities.

HaoK:
Alternatively you can trigger it via [Authorize(AuthenticationScheme = "Bearer")] as well

berdon:
That doesn't seem like an actual solution then.

I guess if it's understood that the authentication middleware is now going to overwrite the HttpContext.User then that's all well and fine. But it's a deviation from 1.0's behavior.

HaoK:
Yes this is an intentional change from 2.0, no more merging in general

HaoK:
We only merge in one case now, when multiple authentication schemes are requested in [Authorize] but even that is going to be deprecated soon

HaoK:
What are you actually trying to do in your app?  Have a principal with an identity from Jwt and another custom one from your own middleware?

berdon:
Yeah, the solution was to move the middleware post authentication middleware. It just seemed like that change shouldn't have to be made if authentication behaves the same way. Thus if this was intended then it's fine - if it was not intended and a bug then I wanted it fixed.

Eilon:
This was indeed a deliberate change that was made for the 2.0 release, so closing this PR.

Issue #1589, opened 2018-01-03
Raise the CookiePolicy consent cookie lifetime


Tratcher:
Current lifetime: 90 days
Suggested: 'forever' for some value of forever?

Eilon:
Suggested time is 1 year.

PR #1621, opened 2018-01-25
Raise the CookiePolicy consent cookie lifetime to one year

Tratcher:
#1589


test/Microsoft.AspNetCore.CookiePolicy.Test/CookieConsentTests.cs
```diff
@@ -221,6 +221,8 @@ public class CookieConsentTests
             var consentCookie = cookies[0];
             Assert.Equal(".AspNet.Consent", consentCookie.Name);
             Assert.Equal("yes", consentCookie.Value);
+            Assert.True(consentCookie.Expires.HasValue);
+            Assert.True(consentCookie.Expires.Value > DateTimeOffset.Now + TimeSpan.FromDays(360));
```
HaoK: Nit: why is this checking only 360 days from now?

Tratcher: 365 would be cutting the comparison too close. 364?

Issue #1613, opened 2018-01-18
Regression: AuthMiddleware can't terminate the request based on the results of AuthenticateAsync for the default scheme


Tratcher:
1.1:
https://github.com/aspnet/Security/blob/9ab189856dbe61bf792707b278255823d2658f51/src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs#L71-L74
2.0:
https://github.com/aspnet/Security/blob/ba1eb281d135400436c52c17edc71307bc038ec0/src/Microsoft.AspNetCore.Authentication/AuthenticationMiddleware.cs#L41-L61

The JwtBearer handle has an OnAuthenticationFailed event which used to be able to generate a response and call context.HandleResponse() to prevent the pipeline from continuing. This doesn't work anymore in 2.0. OIDC still works because it runs under HandleRequestAsync which still has flow control. The AuthMiddleware ignores all non-success results from the default handler, even Fail. See https://github.com/aspnet/Security/issues/1338#issuecomment-358134979

Part of the trouble is that a handler can't tell if it's the default and if it's being invoked from a call path that makes sense to generate error responses from. The OAuth & OIDC handlers always can because they run on dedicated endpoints. JwtBearer, Cookie, Basic, etc. don't know if they're running as the default handler, if they were invoked imperatively, or if they're being invoked by Authorization. The middleware should at least honor Fail though.

PinpointTownes:
That's the expected behavior: ˋFail` means "authentication failed", not "stop processing the request" (which is not always possible since flow control is not always guaranteed for ˋAuthenticateAsync`).

It's definitely something you won't be able to change in 2.1 as it will be a major breaking change. 

Of course, there would be no ambiguity if you had not overruled the name I had proposed (i.e ˋRejectAuthentication()`)...

HaoK:
Yup Fail is return a failed auth result, its not control flow

HaoK:
The new behavior looks reasonable to me, the handler is given a chance to handle the request, if it doesn't.... the middleware looks to do default authenticate, at this point its too late for the handler to stop control flow.

HaoK:
Basically what I'm saying is this regression is by design

Tratcher:
Our JwtBearer sample does exactly this. We should figure out the correct pattern to use and update it.
https://github.com/aspnet/Security/blob/ba1eb281d135400436c52c17edc71307bc038ec0/samples/JwtBearerSample/Startup.cs#L58-L69

[...]

PR #1623, opened 2018-01-26
Update JwtBearer sample error handling

Tratcher:
#1613 The sample was not updated after the 2.0 refactors. The handler cannot directly return an error response, it must rely on a secondary component for that.

### DIFF
--- a/samples/JwtBearerSample/Startup.cs
+++ b/samples/JwtBearerSample/Startup.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Runtime.ExceptionServices;
 using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Authentication.JwtBearer;
 using Microsoft.AspNetCore.Builder;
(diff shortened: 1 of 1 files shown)


samples/JwtBearerSample/Startup.cs
```diff
@@ -86,13 +67,16 @@ public void Configure(IApplicationBuilder app)
...
-                if (user?.Identity?.IsAuthenticated ?? false)
+                var authResult = await context.AuthenticateAsync(JwtBearerDefaults.AuthenticationScheme);
+                if (authResult.Succeeded && authResult.Principal.Identity.IsAuthenticated)
                 {
                     await next();
                 }
+                else if (authResult.Failure != null)
+                {
+                    // Rethrow, let the exception page handle it.
+                    ExceptionDispatchInfo.Capture(authResult.Failure).Throw();
```
PinpointTownes: Do you really want to encourage developers to throw an exception for non-exceptional things like expired tokens? :sweat_smile:

Tratcher: For debugging 😁

PinpointTownes: Without the usual "don't do that in a real world app", I'm convinced folks will blindly copy this snippet without realizing it's not really a good idea :smile:

PR #1625, opened 2018-01-29
Add scheme forwarding (authN policies)

HaoK:
Replaces https://github.com/aspnet/Security/pull/1550

This PR will only focus on adding the new forwarding behavior.

The fate of pure VirtualAuthenticationHandlers/behavior will be addressed in a separate PR.

### DIFF
--- a/test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs
+++ b/test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Diagnostics;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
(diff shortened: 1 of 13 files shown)


test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs
```diff
@@ -26,6 +26,416 @@ public class CookieTests
...
 
+        [Fact]
+        public async Task CanForwardDefault()
+        {
+            var services = new ServiceCollection().AddLogging();
+
+            services.AddAuthentication(o =>
+            {
+                o.DefaultScheme = CookieAuthenticationDefaults.AuthenticationScheme;
+                o.AddScheme<TestHandler>("auth1", "auth1");
```
davidfowl: Why is this call nested inside of AddAuthenticaton instead of being:

```C#
services.AddAuthentication(CookieAuthenticationDefaults.AuthenticationScheme)
        .AddScheme<TestHandler>("auth1", "auth1")
        .AddCookie(o => o.ForwardDefault = "auth1");
```

HaoK: It could go in either place for the purposes of these tests, but the original intent was to exercising that you can forward to non-security based IAuthenticationHandlers

HaoK: The difference between the two places is that adding schemes directly to the option is coming from Core Authentication (httpabstractions), while the AuthenticationBuilder is something that is part of Security and adds a bunch of constraints like deriving from `AuthenticationSchemeOptions/AuthenticationHandler`

PR #1626, opened 2018-01-30
Fix Logger.UnexpectedCorrelationCookieValue

keqli:
The log format string/logic is inconsistent with the parameters.

I am now working on a project referencing AspNetCore.Security 2.0.x so I hope this can be merged into next 2.0.x version.

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication/LoggingExtensions.cs
+++ b/src/Microsoft.AspNetCore.Authentication/LoggingExtensions.cs
@@ -64,7 +64,7 @@ namespace Microsoft.Extensions.Logging
             _unexpectedCorrelationCookieValue = LoggerMessage.Define<string, string>(
                eventId: 16,
                logLevel: LogLevel.Warning,
-               formatString: "The correlation cookie value '{CorrelationCookieName}' did not match the expected value '{CorrelationCookieValue}'.");
+               formatString: "The correlation cookie value '{CookieValue}' did not match the expected value '{ExpectedCookieValue}'.");
         }
 
         public static void AuthenticationSchemeAuthenticated(this ILogger logger, string authenticationScheme)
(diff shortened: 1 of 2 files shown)


Tratcher:
Actually the log was correct, even if the code was confusing. The cookies use unique names to avoid collisions and the important data is all embedded in the name. The cookie values are only placeholders. See https://github.com/keqli/Security/blob/271c15e93a3285131175f09087b94bd8e9c2be9c/src/Microsoft.AspNetCore.Authentication/RemoteAuthenticationHandler.cs#L203-L205

Tratcher:
What output did you see that lead you to changing this?

keqli:
@Tratcher I was writing a WeChat OAuth Authentication Middleware in my project that needs to override `GenerateCorrelationId()` and `ValidateCorrelationId()`, then I found this.

Tratcher:
What was the output?

keqli:
@Tratcher I don't have a real-world case that hits this log. I found it by reading the code.

Tratcher:
Closing this for now, the correct content is bing output already.

keqli:
@Tratcher No, the correct content is not being output. 

Here is an example: Let's say we are using the default MicrosoftAccount Authentication Middleware.The expected cookie is like `.AspNetCore.Correlation.Microsoft.M6mB5dpBscKWSlXaMaTcSBDS82BSkL_wMiU0mHXRlbk=N`. If someone else manually modify the cookie like `.AspNetCore.Correlation.Microsoft.M6mB5dpBscKWSlXaMaTcSBDS82BSkL_wMiU0mHXRlbk=Foo`, now the log will be 

> The correlation cookie value '.AspNetCore.Correlation.Microsoft.M6mB5dpBscKWSlXaMaTcSBDS82BSkL_wMiU0mHXRlbk' did not match the expected value 'Foo'.

The log looks weird. The 2 parameters {CorrelationCookieName} and {CorrelationCookieValue} don't match the semantics of the log sentence or the code context:
`if (!string.Equals(correlationCookie, CorrelationMarker, StringComparison.Ordinal))`
The correct log should be

> The correlation cookie value 'Foo' did not match the expected value 'N'.

Tratcher:
I see. Rather than removing the cookie name, how about rephrasing the message?
`The correlation cookie '.AspNetCore.Correlation.Microsoft.M6mB5dpBscKWSlXaMaTcSBDS82BSkL_wMiU0mHXRlbk' did not have the expected value 'N'.`

Again, the cookie name contains the important details, not the value. The value only exists because client's delete cookies without values.

PR #1628, opened 2018-01-30
WsFed OOB Update versions, dependencies.

Tratcher:
Rebased on 2.0.1.
Updating the IdentityModel dependency to 5.2.0 RTM. This was our largest ship blocker.
Change the package version to 2.0.0 RTM.


Tratcher:
That's the idea. After this we'll merge it into the patch & dev branches and update all of the versions accordingly.

Issue #1629, opened 2018-01-30
[Flaky Test] Microsoft.AspNetCore.CookiePolicy.Test.CookieConsentTests.CreateConsentCookieMatchesGrantConsentCookie


Tratcher:
```
[10:50:14][Step 6/7]   Failed   Microsoft.AspNetCore.CookiePolicy.Test.CookieConsentTests.CreateConsentCookieMatchesGrantConsentCookie
[10:50:14][Step 6/7]   Error Message:
[10:50:14][Step 6/7]    Assert.Equal() Failure
[10:50:14][Step 6/7]   Expected: StringValues [".AspNet.Consent=yes; expires=Wed, 30 Jan 2019 18:5"...]
[10:50:14][Step 6/7]   Actual:   StringValues [".AspNet.Consent=yes; expires=Wed, 30 Jan 2019 18:5"...]
[10:50:14][Step 6/7]   Stack Trace:
[10:50:14][Step 6/7]      at Microsoft.AspNetCore.CookiePolicy.Test.CookieConsentTests.<CreateConsentCookieMatchesGrantConsentCookie>d__18.MoveNext() in C:\b\w\33bdfc1cae7b2a38\modules\Security\test\Microsoft.AspNetCore.CookiePolicy.Test\CookieConsentTests.cs:line 582
[10:50:14][Step 6/7]   --- End of stack trace from previous location where exception was thrown ---
[10:50:14][Step 6/7]      at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[10:50:14][Step 6/7]      at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[10:50:14][Step 6/7]   --- End of stack trace from previous location where exception was thrown ---
[10:50:14][Step 6/7]      at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[10:50:14][Step 6/7]      at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[10:50:14][Step 6/7]   --- End of stack trace from previous location where exception was thrown ---
[10:50:14][Step 6/7]      at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[10:50:14][Step 6/7]      at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```

[...]

PR #1630, opened 2018-01-30
Fix flaky cookie policy tests 

Tratcher:
#1629 The granted cookie and the manually generated cookie should match, but they're created milliseconds apart so the date may drift. I changed the test to a logical equality check instead of a string equality.

### DIFF
--- a/test/Microsoft.AspNetCore.CookiePolicy.Test/CookieConsentTests.cs
+++ b/test/Microsoft.AspNetCore.CookiePolicy.Test/CookieConsentTests.cs
@@ -579,7 +579,13 @@ namespace Microsoft.AspNetCore.CookiePolicy.Test
             Assert.Equal(Net.Http.Headers.SameSiteMode.Lax, consentCookie.SameSite);
             Assert.NotNull(consentCookie.Expires);
 
-            Assert.Equal(httpContext.Response.Headers[HeaderNames.SetCookie], httpContext.Response.Headers["ManualCookie"]);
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNetCore.CookiePolicy.Test/CookieConsentTests.cs
```diff
@@ -626,7 +632,13 @@ public class CookieConsentTests
...
             Assert.NotNull(consentCookie.Expires);
 
-            Assert.Equal(httpContext.Response.Headers[HeaderNames.SetCookie], httpContext.Response.Headers["ManualCookie"]);
+            cookies = SetCookieHeaderValue.ParseList(httpContext.Response.Headers["ManualCookie"]);
+            Assert.Equal(1, cookies.Count);
+            var manualCookie = cookies[0];
+            Assert.Equal(consentCookie.Name, manualCookie.Name);
+            Assert.Equal(consentCookie.Value, manualCookie.Value);
+            Assert.Equal(consentCookie.SameSite, manualCookie.SameSite);
+            Assert.NotNull(manualCookie.Expires); // Expires may not exactly match to the second.
```
HaoK: Nit: do you care about doing a rough match at least?  Parse and check for same day or something?

Tratcher: The entire field is unstable as an added second could roll any or all fields if you run it on the minute, hour, day, month, or year boundary. Admittedly each stage gets less likely.

Issue #1609, opened 2018-01-12
OIDC add support for mapping all user info


Tratcher:
OIDC in 1.0 mapped all user info into claims by default. 
https://github.com/aspnet/Security/blob/79a60504a824cf8d9d0af33282bd141deb74fa6c/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L735-L763
This could cause a lot of cookie bloat so in 2.0 it was moved to use the new ClaimActions infrastructure to map only a specific set. The same approach is used for all of the OAuth providers.
https://github.com/aspnet/Security/blob/dde7671c06da64e4a7a290c37ed86e9a9bdd0dd7/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectOptions.cs#L63-L69

There's been some community feedback that this is hard to use for some scenarios:
https://twitter.com/BrockLAllen/status/951222624792834049
https://leastprivilege.com/2017/11/15/missing-claims-in-the-asp-net-core-2-openid-connect-handler/

We can't re-design this until 3.0 and it's not clear what that would look like. In the meantime we can provide some helpers that cover this scenario.

Proposal:
Add `ClaimActions.MapAllClaims()` that emulates the 1.0 behavior. Also `ClaimActions.MapAllClaimsExcept(params string[] exclusions)` that does the same but excludes certain claim types like Roles (Azure can add way too many).
Make sure there's a rational way to handle duplicate claims.

@BrockAllen @leastprivilege

leastprivilege:
sounds good to me.

Not sure it is worth throwing exceptions for duplicate claims - it's an application concern.

We found that e.g. Azure with "Live ID" login emits two name claims which breaks the handler in federation gateway scenarios depending which external provider is used.

brockallen:
My use case for `MapAllClaimsExcept` is not to reduce the claims in the cookie (since there are other approaches for managing cookie size), but to simply filter out the protocol claims coming from the `id_token` When claims are returned from the userinfo endpoint, typically a client app wants all of those claims (since they asked for them in the first place via the `scope`s requested).

Tratcher:
That's a bit trickier. We don't map the id token claims at our layer, that happens in Identity model during the validation. Hence those DeleteClaim actions that retroactively unmap.

leastprivilege:
Assume we always disable the stupid mapping in the JWT handler...and that's what you should do too when using the handler..

[...]

PR #1636, opened 2018-02-02
Add ClaimActions for bulk add and remove

Tratcher:
#1609 @leastprivilege @brockallen how does this look? It covers your asks in three parts:
- Clear the JWT claims type map.
- Map all claims
- bulk delete protocol claims

### DIFF
--- a/samples/OpenIdConnectSample/Startup.cs
+++ b/samples/OpenIdConnectSample/Startup.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.IdentityModel.Tokens.Jwt;
 using System.Linq;
 using System.Net.Http;
 using System.Text.Encodings.Web;
@@ -44,6 +45,8 @@ namespace OpenIdConnectSample
 
         public void ConfigureServices(IServiceCollection services)
         {
+            JwtSecurityTokenHandler.DefaultInboundClaimTypeMap.Clear();
+
             services.AddAuthentication(sharedOptions =>
             {
                 sharedOptions.DefaultAuthenticateScheme = CookieAuthenticationDefaults.AuthenticationScheme;
@@ -56,9 +59,13 @@ namespace OpenIdConnectSample
                 o.ClientId = Configuration["oidc:clientid"];
                 o.ClientSecret = Configuration["oidc:clientsecret"]; // for code flow
                 o.Authority = Configuration["oidc:authority"];
+
                 o.ResponseType = OpenIdConnectResponseType.CodeIdToken;
                 o.SaveTokens = true;
                 o.GetClaimsFromUserInfoEndpoint = true;
+
+                o.ClaimActions.MapAllExcept("aud", "iss", "iat", "nbf", "exp", "aio", "c_hash", "uti", "nonce");
+
                 o.Events = new OpenIdConnectEvents()
                 {
                     OnAuthenticationFailed = c =>
--- a/src/Microsoft.AspNetCore.Authentication.OAuth/Claims/ClaimActionCollectionMapExtensions.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OAuth/Claims/ClaimActionCollectionMapExtensions.cs
@@ -88,6 +88,27 @@ namespace Microsoft.AspNetCore.Authentication
         }
 
         /// <summary>
+        /// Clears any current ClaimsActions and maps all values from the json user data as claims, excluding duplicates.
+        /// </summary>
+        /// <param name="collection"></param>
+        public static void MapAll(this ClaimActionCollection collection)
+        {
+            collection.Clear();
+            collection.Add(new MapAllClaimsAction());
+        }
+
+        /// <summary>
+        /// Clears any current ClaimsActions and maps all values from the json user data as claims, excluding the specified types.
+        /// </summary>
+        /// <param name="collection"></param>
+        /// <param name="exclusions"></param>
+        public static void MapAllExcept(this ClaimActionCollection collection, params string[] exclusions)
+        {
+            collection.MapAll();
+            collection.DeleteClaims(exclusions);
+        }
+
+        /// <summary>
         /// Delete all claims from the given ClaimsIdentity with the given ClaimType.
         /// </summary>
         /// <param name="collection"></param>
@@ -95,6 +116,24 @@ namespace Microsoft.AspNetCore.Authentication
         public static void DeleteClaim(this ClaimActionCollection collection, string claimType)
         {
             collection.Add(new DeleteClaimAction(claimType));
+        }
+
+        /// <summary>
+        /// Delete all claims from the ClaimsIdentity with the given claimTypes.
+        /// </summary>
+        /// <param name="collection"></param>
+        /// <param name="claimTypes"></param>
+        public static void DeleteClaims(this ClaimActionCollection collection, params string[] claimTypes)
+        {
+            if (claimTypes == null)
+            {
+                throw new ArgumentNullException(nameof(claimTypes));
+            }
+
+            foreach (var claimType in claimTypes)
+            {
+                collection.Add(new DeleteClaimAction(claimType));
+            }
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Authentication.OAuth/Claims/MapAllClaimsAction.cs
+++ b/src/Microsoft.AspNetCore.Authentication.OAuth/Claims/MapAllClaimsAction.cs
@@ -1,0 +1,42 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Security.Claims;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.AspNetCore.Authentication.OAuth.Claims
+{
+    /// <summary>
+    /// A ClaimAction that selects all top level values from the json user data and adds them as Claims.
+    /// This excludes duplicate sets of names and values.
+    /// </summary>
+    public class MapAllClaimsAction : ClaimAction
+    {
+        public MapAllClaimsAction() : base("All", ClaimValueTypes.String)
+        {
+        }
+
+        public override void Run(JObject userData, ClaimsIdentity identity, string issuer)
+        {
+            if (userData == null)
+            {
+                return;
+            }
+            foreach (var pair in userData)
+            {
+                var claimValue = userData.TryGetValue(pair.Key, out var value) ? value.ToString() : null;
+
+                // Avoid adding a claim if there's a duplicate name and value. This often happens in OIDC when claims are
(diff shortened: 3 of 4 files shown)


src/Microsoft.AspNetCore.Authentication.OAuth/Claims/ClaimActionCollectionMapExtensions.cs
```diff
@@ -87,6 +87,15 @@ public static void MapCustomJson(this ClaimActionCollection collection, string c
             collection.Add(new CustomJsonClaimAction(claimType, valueType, resolver));
         }
 
+        /// <summary>
+        /// Maps all values from the json user data as claims, excluding duplicates.
+        /// </summary>
+        /// <param name="collection"></param>
+        public static void MapAllClaims(this ClaimActionCollection collection)
```
HaoK: Shorten to MapAll to be consistent with all the other methods that don't have Claims suffix?

src/Microsoft.AspNetCore.Authentication.OAuth/Claims/ClaimActionCollectionMapExtensions.cs
```diff
@@ -96,5 +105,23 @@ public static void DeleteClaim(this ClaimActionCollection collection, string cla
         {
             collection.Add(new DeleteClaimAction(claimType));
         }
+
+        /// <summary>
+        /// Delete all claims from the ClaimsIdentity with the given claimTypes.
+        /// </summary>
+        /// <param name="collection"></param>
+        /// <param name="claimTypes"></param>
+        public static void DeleteClaims(this ClaimActionCollection collection, params string[] claimTypes)
```
HaoK: Delete?

src/Microsoft.AspNetCore.Authentication.OAuth/Claims/MapAllClaimsAction.cs
```diff
@@ -0,0 +1,45 @@
...
+        public override void Run(JObject userData, ClaimsIdentity identity, string issuer)
+        {
+            foreach (var pair in userData)
+            {
+                var claimValue = userData.TryGetValue(pair.Key, out var value) ? value.ToString() : null;
+
+                // Avoid adding a claim if there's a duplicate name and value. This often happens in OIDC when claims are
+                // retrieved both from the id_token and from the user-info endpoint.
+                var duplicate = false;
+                foreach (var existingClaim in identity.FindAll(pair.Key))
```
HaoK: Nit: can you something like this?

```
var duplicate = identity.FindFirst(c => c.Key == pair.Key && string.Equals(c.Value, claimValue, StringComparison.Ordinal) != null
```

src/Microsoft.AspNetCore.Authentication.OAuth/Claims/ClaimActionCollectionMapExtensions.cs
```diff
@@ -96,5 +105,23 @@ public static void DeleteClaim(this ClaimActionCollection collection, string cla
...
+        {
+            if (claimTypes == null)
+            {
+                throw new ArgumentNullException(nameof(claimTypes));
+            }
+
+            foreach (var claimType in claimTypes)
+            {
+                collection.Add(new DeleteClaimAction(claimType));
```
HaoK: Consider enhancing DeleteClaimAction type to take a list of claims so you don't have to add one action per type?

Tratcher: Performance wise it's the same, you still need to loop through all of the names.

Adding them individually preserves the functionality of methods like ClaimActionsCollection.Remove(claimType) that removes the action by name.

src/Microsoft.AspNetCore.Authentication.OAuth/Claims/ClaimActionCollectionMapExtensions.cs
Tratcher: People have been very confused by Remove vs Delete. We need to avoid aggravating that.

src/Microsoft.AspNetCore.Authentication.OAuth/Claims/ClaimActionCollectionMapExtensions.cs
HaoK: Pretty sure that's exactly the behavior that's confusing today.

options.ClaimActions.Remove("something") you would think is the same as options.ClaimActions.DeleteClaim("something").

I think this api should be mostly focusing on making it easy to do the common things.

MapJsonKey
MapAll
Delete

HaoK: The list of delegate/actions is fine as an implementation detail, but I really don't think generally developers should have to understand/deal with that... when all they want to do is include/ignore some set of claims.  Only someone who is trying to plug in a complex mapping should have to understand the whole claim action stuff

src/Microsoft.AspNetCore.Authentication.OAuth/Claims/ClaimActionCollectionMapExtensions.cs
HaoK: you could also just change DeleteClaim to take a params string[] right?

Tratcher: In theory, but it would be binary breaking.

samples/OpenIdConnectSample/Startup.cs
```diff
@@ -59,6 +60,16 @@ public void ConfigureServices(IServiceCollection services)
...
                 o.GetClaimsFromUserInfoEndpoint = true;
+
+                // TODO: Revert
+                var handler = new JwtSecurityTokenHandler();
+                handler.InboundClaimTypeMap.Clear();
+                o.SecurityTokenValidator = handler;
+                o.ClaimActions.Clear();
+                o.ClaimActions.MapAllClaims();
+                // Note some of these protocol claims also show up in the user info
+                o.ClaimActions.DeleteClaims("aud", "iss", "iat", "nbf", "exp", "aio", "c_hash", "uti", "nonce");
```
Tratcher: RemoveProtocolClaims
MapAllUserInfo

leastprivilege:
just to be really clear

this
```
var handler = new JwtSecurityTokenHandler();
handler.InboundClaimTypeMap.Clear();
o.SecurityTokenValidator = handler;
```

is not really an improvement over just turning of the inbound claims mapping globally via their static `DefaultInboundClaimTypeMap`.

I would like to see no claims mapping at all.

Tratcher:
@leastprivilege @brockallen Updated. MapAllExcept combines both operations like you asked.

leastprivilege:
..and what about the claim type mappings - will this be turned off at the handler level by default?

Tratcher:
That's a breaking change, I can't do that right now.

brockallen:
I have a VM with preview-1. I assume I can just pull the source for this and test that.

leastprivilege:
> That's a breaking change, I can't do that right now.

Two things you can do:

* Do the mapping of standard to proprietary claims in the default claim actions
* Allow setting/clearing the inbound claim type map directly on the OIDC handler options

Tratcher:
@brockallen the code should even work against 2.0 RTM.

Tratcher:
@leastprivilege 
> Do the mapping of standard to proprietary claims in the default claim actions

I don't follow. Can you give an example?

> Allow setting/clearing the inbound claim type map directly on the OIDC handler options

So the code you want to simplify is:
```
                var handler = new JwtSecurityTokenHandler();
                handler.InboundClaimTypeMap.Clear();
                o.SecurityTokenValidator = handler;
```
Down to something like `o.InboundClaimTypeMap.Clear();`? That might be doable but it would rely on downcasting `ISecurityTokenValidator SecurityTokenValidator` to it's default type `JwtSecurityTokenHandler`.

brockallen:
> @brockallen the code should even work against 2.0 RTM.

Even easier. I'll do it first thing tomorrow.

leastprivilege:
> I don't follow. Can you give an example?

Well - you could have a `MakeYourClaimsWorse` (naming is up to you ;)) option. And when it is set - whenever you run a claim action, you use the dictionary provided by the JwtSecurityTokenHandler to do a mapping if present.

This way the mapping behavior would be an OIDC handler thing...

brockallen:
I was unable to get this branch working on either 2.0 or 2.1-preview-1. Mainly build issues on the Security repo.

Tratcher:
@brockallen no need to build this repo, you should be able to copy the MapAllClaimsAction class directly into your app. You could also copy the new extension methods if you like.

src/Microsoft.AspNetCore.Authentication.OAuth/Claims/MapAllClaimsAction.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// </summary>
+    public class MapAllClaimsAction : ClaimAction
+    {
+        public MapAllClaimsAction() : base("All", ClaimValueTypes.String)
+        {
+        }
+
+        public override void Run(JObject userData, ClaimsIdentity identity, string issuer)
+        {
+            foreach (var pair in userData)
```
brockallen: When I use the new `MapAll` I get a null reference exception on userData.

brockallen:
And just to repeat since I don't know if you saw it:

When I use the new `MapAll` I get a null reference exception on `userData`. This fails when I have an implicit flow only client, meaning `ResponseType="id_token"`.

Tratcher:
Updated

brockallen:
Cool, thanks. I think with those changes, this now becomes a nice one-liner to allow controlling which claims are kept/removed. There's of course the claim type map issue, but that's another topic/thread I assume?

Tratcher:
Agreed, the mapping is orthogonal.

Tratcher:
@HaoK please sanity check

src/Microsoft.AspNetCore.Authentication.OAuth/Claims/ClaimActionCollectionMapExtensions.cs
```diff
@@ -87,6 +87,28 @@ public static void MapCustomJson(this ClaimActionCollection collection, string c
...
+
+        /// <summary>
+        /// Clears any current ClaimsActions and maps all values from the json user data as claims, excluding the specified types.
+        /// </summary>
+        /// <param name="collection"></param>
+        /// <param name="exclusions"></param>
+        public static void MapAllExcept(this ClaimActionCollection collection, params string[] exclusions)
+        {
+            collection.Clear();
+            collection.Add(new MapAllClaimsAction());
```
HaoK: This should call MapAll

Tratcher:
@leastprivilege second guessing the mappings is problematic, you're better off clearing JWT's mappings. I'm going to merge this PR as your suggestions along that line are independent. I'll file a new issue to simplify clearing the JWT type map.

Issue #1634, opened 2018-02-01
[2.0.x] Merge WsFed into the patch branch


Tratcher:
WsFed is getting an out of band 2.0 release targeting 2.0.1 dependencies. Once that's done we should merge it into the 2.0 patch branch for long term servicing.

Considerations:
- Do not add it to any store, metapackage, etc.. This is a ship, exclude package that will only be available from nuget.org until 2.1.
- This depends on a higher but compatible version of IdentityModel (5.2.0) than OpenIdConnect (5.1.4). Do we keep the dependency jagged or raise the other packages? Jagged might be OK as these packages won't be used side-by-side much.
- Update the in-repo dependencies from packages to projects.

@muratg @Eilon

Eilon:
We should keep the dependencies as-is because we don't want to raise the minor version dependency on the 2.0.x stuff that already shipped.

PR #1637, opened 2018-02-02
Merge WsFed into the 2.0 patch branch

Tratcher:
#1634 This will enable long term servicing. If 2.0.6 is re-spun then this will be included.
@Eilon 

We'll also need to work out what changes are required in Universe for this. They should be minimal as this isn't going to be added to any metapackages. We will need to declare the new external dependencies.


build/dependencies.props
```diff
@@ -29,13 +29,15 @@
     <MicrosoftExtensionsWebEncodersPackageVersion>2.0.0</MicrosoftExtensionsWebEncodersPackageVersion>
     <MicrosoftIdentityModelClientsActiveDirectoryPackageVersion>3.14.1</MicrosoftIdentityModelClientsActiveDirectoryPackageVersion>
     <MicrosoftIdentityModelProtocolsOpenIdConnectPackageVersion>2.1.4</MicrosoftIdentityModelProtocolsOpenIdConnectPackageVersion>
+    <MicrosoftIdentityModelProtocolsWsFederationPackageVersion>5.2.0</MicrosoftIdentityModelProtocolsWsFederationPackageVersion>
```
JunTaoLuo: @natemcmaster Does this need to be added to https://github.com/aspnet/Universe/blob/release/2.0/build/dependencies.props?

natemcmaster: Yup. Chris opened a PR to universe with this.

JunTaoLuo: Ah I see there is already a PR for this: https://github.com/aspnet/Universe/pull/844

Issue #1635, opened 2018-02-01
Merge WsFed package from patch branch into dev


Eilon:
Depends on https://github.com/aspnet/Security/issues/1634

Starting in 2.1.0-preview2, the WsFed support will be just another part of the release.

We need to check whether this also needs to go into the `.App` package. Presumably yes, but check with the Powers That Be.

Issue #1581, opened 2017-12-29
WsFed remote signout doesn't work with ADFS4


Tratcher:
@brockallen reported that singout cleanup messages don't work with ADFS4 because you can't configure both CallbackPath and RemoteSignOutPath to the same value. We may have to redesign how these are routed in HandleRequestAsync.

"ADFS has an odd requirement that the redirect URL after SLO must be beneath the URL used for signin response (for example "/wsfed" for wreply and for redirect after signout would need to be "/wsfed/" -- notice extra slash)."

We'll need to set up a repro.

Issue #1520, opened 2017-11-02
WsFederationOptions.EventsType pattern unusable because WsFederationEvents class is missing virtual members.


lzandman:
I've placed my custom cookie authentication events inside a custom class that inherits from CookieAuthenticationEvents and specify that class using the EventsType property of CookieAuthenticationOptions as specified in [this Microsoft document](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/cookie?tabs=aspnetcore2x#reacting-to-back-end-changes).

I tried to apply the same pattern to the WsFederationEvents class, but noticed it doesn't contain any virtual members that I can override in my own custom class. It only inherits one overrideable virtual member, TicketReceived, from its base class RemoteAuthenticationEvents. So it seems this pattern isn't applicable and the WsFederationOptions.EventsType property mechanism cannot be used.

I'm using the latest nightly build of the WsFederation 2.0.0-preview1 package.

PinpointTownes:
Actually, the missing virtual methods are not the only issue: `WsFederationHandler` directly uses `Options.Events`, which means the `EventsType` option is never used.

/cc @Tratcher

PR #1639, opened 2018-02-05
Add Microsoft.AspNetCore.Authentication.WsFederation, samples, and tests.

Tratcher:
#1635 Merge the new WsFed package from patch branch into dev
The only diffs from the patch branch involve reacting to the new scheme forwarders in SignOut.

Squashed commits from the patch branch:
#43 Add Microsoft.AspNetCore.Authentication.WsFederation, samples, and tests.
#1443 Block unsolicited wsfed logins by default.
#1520 Update WsFed to use the 2.0 event structure
#1425 Implement WsFed remote signout cleanup
Rework WsFed RemoteSignOutPath logic to work with ADFS #1581
Update versions, dependencies.


src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationOptions.cs
```diff
@@ -0,0 +1,179 @@
...
+        /// </summary>
+        public bool RefreshOnIssuerKeyNotFound { get; set; } = true;
+
+        /// <summary>
+        /// Indicates if requests to the CallbackPath may also be for other components. If enabled the handler will pass
+        /// requests through that do not contain OpenIdConnect authentication responses. Disabling this and setting the
+        /// CallbackPath to a dedicated endpoint may provide better error handling.
+        /// This is disabled by default.
+        /// </summary>
+        public bool SkipUnrecognizedRequests { get; set; } = false;
```
HaoK: nit: you don't set some of the other defaults to false

test/Microsoft.AspNetCore.Authentication.WsFederation.Test/Microsoft.AspNetCore.Authentication.WsFederation.Test.csproj
```diff
@@ -0,0 +1,33 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
HaoK: Today we only have a single test project, do you want to merge this into a new WsFed folder similar to OIDC?

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
```diff
@@ -0,0 +1,435 @@
...
+        /// </summary>
+        /// <returns></returns>
+        protected override async Task HandleChallengeAsync(AuthenticationProperties properties)
+        {
+            if (_configuration == null)
+            {
+                _configuration = await Options.ConfigurationManager.GetConfigurationAsync(Context.RequestAborted);
+            }
+
+            string baseUri =
```
HaoK: nit: var?

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
```diff
@@ -0,0 +1,435 @@
...
+                    Request.Scheme +
+                    Uri.SchemeDelimiter +
+                    Request.Host +
+                    Request.PathBase;
+
+            string currentUri =
```
HaoK: nit: var?

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
```diff
@@ -0,0 +1,435 @@
...
+                Request.Path +
+                Request.QueryString;
+
+            // Save the original challenge URI so we can redirect back to it when we're done.
+            if (string.IsNullOrEmpty(properties.RedirectUri))
+            {
+                properties.RedirectUri = currentUri;
+            }
+
+            WsFederationMessage wsFederationMessage = new WsFederationMessage()
```
HaoK: nit: var?

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
```diff
@@ -0,0 +1,435 @@
...
+            wsFederationMessage = redirectContext.ProtocolMessage;
+
+            if (!string.IsNullOrEmpty(wsFederationMessage.Wctx))
+            {
+                properties.Items[WsFederationDefaults.UserstatePropertiesKey] = wsFederationMessage.Wctx;
+            }
+
+            wsFederationMessage.Wctx = Uri.EscapeDataString(Options.StateDataFormat.Protect(properties));
+
+            string redirectUri = wsFederationMessage.CreateSignInUrl();
```
HaoK: nit: var?

test/Microsoft.AspNetCore.Authentication.WsFederation.Test/Microsoft.AspNetCore.Authentication.WsFederation.Test.csproj
Tratcher: We've been trying to separate out OIDC into it's own project for a while but the work keeps getting backlogged. I'm not sure what we'd gain by merging this one.

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationOptions.cs
```diff
@@ -0,0 +1,179 @@
...
+        /// <summary>
+        /// Requests received on this path will cause the handler to invoke SignOut using the SignOutScheme.
+        /// </summary>
+        public PathString RemoteSignOutPath { get; set; }
+
+        /// <summary>
+        /// The Authentication Scheme to use with SignOutAsync from RemoteSignOutPath. SignInScheme will be used if this
+        /// is not set.
+        /// </summary>
+        public string SignOutScheme { get; set; }
```
HaoK: Can consider using ForwardSignOut instead..  But I guess this is fine for now, we can clean up the whole signin/signout scheme stuff for the remote handlers in 2.2/3.0

test/Microsoft.AspNetCore.Authentication.WsFederation.Test/Microsoft.AspNetCore.Authentication.WsFederation.Test.csproj
HaoK: I didn't think that was a goal for OIDC, just to refactor/clean up tests.  I don't really care one way or another really long term (between one test project per package or one uber test project), but today since there's only one uber test project for all of auth, it seems weird to have this be stand alone.

Tratcher: https://github.com/aspnet/Security/issues/349

HaoK: Personally I think having all the tests in one project are useful (when doing refactoring especially), running them all together doesn't even take long right now.  If these ever become slow then splitting them apart would be more useful, but if we are putting the functional tests in other repos like AuthSamples/Identity/templates, I don't think there's much value in splitting them up

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
Tratcher: Note: Use base.CurrentUri

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationOptions.cs
Tratcher: Let's wait until we figure that out system-wide.

Tratcher:
Updated

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
```diff
@@ -0,0 +1,425 @@
...
+
+                return HandleRequestResult.Fail(exception, properties);
+            }
+        }
+
+        /// <summary>
+        /// Handles Signout
+        /// </summary>
+        /// <returns></returns>
+        public async virtual Task SignOutAsync(AuthenticationProperties properties)
```
HaoK: Do you think we should add a SignOutHandler and have this derive from that so you'd only have to implement HandleSignOut like the other helpers?   I've been debating adding those base classes for SignIn/SignOut.

test/Microsoft.AspNetCore.Authentication.Test/WsFederation/WsFederationTest.cs
```diff
@@ -0,0 +1,357 @@
...
+
+            var request = new HttpRequestMessage(HttpMethod.Post, queryItems["wreply"]);
+            CopyCookies(response, request);
+            request.Content = CreateSignInContent("WsFederation/InvalidToken.xml", queryItems["wctx"]);
+            response = await httpClient.SendAsync(request);
+
+            // Did the request end in the actual resource requested for
+            Assert.Equal("AuthenticationFailed", await response.Content.ReadAsStringAsync());
+        }
+
```
HaoK: Consider adding a test to verify that map'd urls are picked up in redirects, see https://github.com/aspnet/Security/blob/dev/test/Microsoft.AspNetCore.Authentication.Test/FacebookTests.cs#L597

We actually have a large set of common tests now, we should probably consider adding a base handler specification test rather than cloning all of the tests that really should be common across all the handlers.  In the mean time you might want to consider copying a lot of the tests from one of the other common ones from facebook like ThrowsIfAppIdMissing variants, VerifySchemeDefaults, VerifySignInSchemeCannotBeSetToSelfUsingDefaultSignInScheme etc.

HaoK: I created https://github.com/aspnet/Security/issues/1640 to track moving as many as possible into base classes, so I'd just do the ones specific to wsfed, like VerifySchemeDefaults and the appropriate validations that are required on the options, and the redirect + map test

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationHandler.cs
Tratcher: Oh, did those get dropped from the final version? Yeah, it's awkward that all implementors need to include this for the forwarders feature to work. File a followup issue?

HaoK: Not dropped, I just merged the minimal changes in for now, nothing was using the new base classes yet so I didn't put them in...  we should probably add them if only for wsfed to use...  Its a breaking change to change the other ones so I couldn't fix the existing handers to use them...

Tratcher: It may also be a breaking change for WsFed then, this code was merged into 2.0 first.

HaoK: We shipped WsFed already?  I just meant its a breaking change on the handlers since we would be making SignOut non virtual and replacing it with protected virtual HandleSignOut etc.

HaoK: So WsFed can't depend on 2.1?

Tratcher: We've nearly shipped WsFed 2.0.0 and an out of band release. We're only waiting for verification sign-off on the final package. It's also been merged into the next patch release.

src/Microsoft.AspNetCore.Authentication.WsFederation/WsFederationOptions.cs
```diff
@@ -0,0 +1,179 @@
...
+
+        /// <summary>
+        /// Gets or sets if a metadata refresh should be attempted after a SecurityTokenSignatureKeyNotFoundException. This allows for automatic
+        /// recovery in the event of a signature key rollover. This is enabled by default.
+        /// </summary>
+        public bool RefreshOnIssuerKeyNotFound { get; set; } = true;
+
+        /// <summary>
+        /// Indicates if requests to the CallbackPath may also be for other components. If enabled the handler will pass
+        /// requests through that do not contain OpenIdConnect authentication responses. Disabling this and setting the
```
Tratcher: Justin offline: OpenIdConnect -> WsFederation

test/Microsoft.AspNetCore.Authentication.Test/WsFederation/WsFederationTest.cs
Tratcher: Rebased and added a few tests.

PR #1644, opened 2018-02-10
[WIP] PolicySchemes / UseSignIn

HaoK:
Starting to play with the shape of the proposed Use/AddXyzSignIn...

```C#
services.AddAuthentication()
    // Pick one, last one will win (set to DefaultScheme) if more than one called (hence Use over Add)
   .UseGoogleSignIn(o => { })
   .UseFacebookSignIn(o => { })
   .UseOAuthSignIn(o => { })
   .UseOpenIdConnectSignIn(o => { })
   .UseTwitterSignIn(o => { })
   .UseIdentitySignIn()

// If you want to custom configure the cookie, you need to know which cookie
services.Configure<CookieAuthenticationOptions>(CookieAuthenticationDefaults.AuthenticationScheme, o => o.ExpireTimeSpan = 100);
// or with identity
services.ConfigureApplicationCookie(o => o.ExpireTimeSpan = 100);
```

Each Use is responsible for configuring the defaults (and adding any cookies as needed).  All of the social sign ins would use the default cookie.  Identity would probably continue to use its existing Identity.Application cookie as the default

-  VirtualSchemeHandler => PolicySchemes
- new TryAddCookie extension method that attempts to add a cookie
- UseXyzSignIn will AddXyz, TryAddCookie, set that to the default, set forward challenge to Xyz.
- related UseIdentitySignIn/refactor PR: https://github.com/aspnet/Identity/pull/1442

----------------
Open questions:
- Should policy schemes fall back to default scheme behavior? aka `AddPolicyScheme("This<=>Default", o => { })`
- Should we pick a new "SignIn" constant scheme value to use for the name of the sign in cookie (and switch identity to use this too?
- Do we want to enable sign in/ sign out for existing handlers so they would forward or throw.  Today they don't implement the sign in/sign out interfaces so the service will block them from being called.


HaoK:
Split this PR, moving the virtual scheme rename/cleanup into https://github.com/aspnet/Security/pull/1665/files will only address the UseSignIn/additional new apis in this PR

HaoK:
Closing since we aren't going to do anything in 2.1, will revisit with more time in 2.2

Issue #589, opened 2015-11-27
More efficient Authentication ticket serialization


jods4:
Serialized auth tickets can grow fairly large very quickly, even in basic scenarios. If you have moderately many roles, they all get serialized with the long claim type name and that's how your tickets becomes several Ko in size before you even know it.

In the common scenario where a ticket has a few of roles, today each one is adding more than 60 bytes just for its name (unencoded, actually more than that after encryption and encoding in base64). Having roles counted in tens results in cookies of size in thousands.

Previously the payload was compressed, but because of security concerns this was removed.

I noticed that in the beta-8 or rc-1 timeframe cookies chunking was added. That's a very good thing because it prevents failures caused by too large cookies. But it's not enough, because even if they don't fail, 10Ko cookies are not great for perf.

I know your guideline for very large tickets is to keep them on the server in session state. It's nice to have a solution to support scenarios where there is no way you could decently round-trip your state. But when you have a stateless server, you would like to keep it that way as much as possible as it is a lot simpler and safer.

With a more efficient serialization format, common auth. tickets could be a lot smaller and not require session state. I hope you'll pursue this option in addition to what has already been done.

Here's a simple suggestion: serialize claims grouped by claim type, so that you only have to serialize each claim type name only once. Something like this (pseudo-format):

```
http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name
1
jods
http://schemas.microsoft.com/ws/2008/06/identity/claims/role
2
Admin
PowerUser
```

(Nitpicker corner: I know some claim types are multi-values, so it's more something like this: `Roles 2 1 Admin 1 PowerUser`, etc. But you get the idea.)

brockallen:
> I know your guideline for very large tickets is to keep them on the server in session state.

Where did you see this guidance?

As to your issue -- I think the general consensus is that if you're putting that many claims into the cookie, then you might be designing your system incorrectly. Claims in the cookie should be identity data, not permissions data. You then use the identity data at runtime to locate additional data you at need for authorization. Additionally, you can always add claims transformation to your pipeline after the cookie middleware runs and make your user look as if it has all the roles in the cookie. And if your lookups are a concern, then do caching in the app layer.

Tratcher:
@brockallen I assume by session he's referring to this: https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNet.Authentication.Cookies/CookieAuthenticationOptions.cs#L156-L160

jods4:
[...]

PR #1654, opened 2018-02-15
Add TicketSerializer property to CookieAuthenticaitonOptions

Zoxive:
Provide the ability to set the TicketSerializer without having to build up TicketDataFormat yourself.

This then allows users to more easily compress cookies. https://github.com/aspnet/Security/issues/589

Would be nice to even have Compression built in, and allow users to opt-in via a boolean option.


Tratcher:
What kind of compression? The data is intentionally not compressed as there's a common attack against compressed encrypted data. See https://en.wikipedia.org/wiki/CRIME.

Eilon:
Hi, we are closing this because we explicitly don't want to make it easier to compress the contents of these cookies because it makes them more susceptible to certain types of attacks, which @Tratcher linked to

PR #1658, opened 2018-02-20
Include scope when saving tokens in OpenIdConnect

duckblaster:


Eilon:
@duckblaster thanks for the PR, we'll review it. Is there an issue logged that describes why this behavior change is desirable? If so, which one? And if not, can you log an issue with details so that we can consider the implications of this change?

Eilon:
Closing because it is unclear that this would add sufficient value to be worth the risk. This change can potentially increase cookie sizes beyond their limits, or cause other unexpected behavior in applications. At this time we recommend that applications copy the tokens/scopes in their own logic.

PR #1665, opened 2018-02-22
Rename VirtualScheme => PolicyScheme

HaoK:
Update VirtualScheme to use the new base forward options, also introduce SignIn/Out base abstract handlers which handle forwarding.  Cookie is able to use these, but not OpenIdConnect since its using RemoteAuthenticationHandler

@Tratcher @davidfowl


src/Microsoft.AspNetCore.Authentication.Cookies/breakingchanges.netcore.json
```diff
@@ -0,0 +1,6 @@
+  [
+    {
+      "TypeId": "public class Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationHandler : Microsoft.AspNetCore.Authentication.AuthenticationHandler<Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationOptions>, Microsoft.AspNetCore.Authentication.IAuthenticationSignInHandler",
```
Eilon: Breaking change?

HaoK: New base class, shouldn't be a real breaking change.

Old signature derived from AuthenticationHandler
Now it derives from SignInAuthenticationHandler who continues to have a virtual SignOut method.

Eilon: Oh so not a real breaking change? Bug in the tool?

HaoK: The tool has always worked this way as far as I can tell, it just baselines the signature of all the public types/classes, so if you modify the inheritance or move stuff around it will generate diffs that require exceptions

Eilon: It probably is a bit smarter than that, but it does have limitations, so this one here is fine as long as it isn't a real breaking change.

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -15,8 +15,7 @@
 namespace Microsoft.AspNetCore.Authentication.Cookies
 {
     public class CookieAuthenticationHandler :
-        AuthenticationHandler<CookieAuthenticationOptions>,
-        IAuthenticationSignInHandler,
+        SignInAuthenticationHandler<CookieAuthenticationOptions>,
         IAuthenticationSignOutHandler
```
Tratcher: Do you still need IAuthenticationSignOutHandler? Isn't it transitive?

src/Microsoft.AspNetCore.Authentication/PolicySchemeHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    /// <summary>
+    /// PolicySchemes are used to redirect authentication methods to another scheme.
+    /// </summary>
+    public class PolicySchemeHandler : SignInAuthenticationHandler<PolicySchemeOptions>
+    {
+        public PolicySchemeHandler(IOptionsMonitor<PolicySchemeOptions> options, ILoggerFactory logger, UrlEncoder encoder, ISystemClock clock) : base(options, logger, encoder, clock)
+        { }
+
+        protected override Task HandleChallengeAsync(AuthenticationProperties properties)
```
Tratcher: This works because the base class handles all of the forwarding so this only gets invoked if there is no forwarding? At which point you fall back to the system default?  Which may be this handler? 

It seems like it would be a lot safer to require forwarding for this handler and throw if it ever fell through to the local implementation.

HaoK: Yeah so previous iteration of this required an 'anchor' basically a default forward.  The reason I'm not sure that's what we want anymore, is this should also be the base class of choice for plugging in custom 'policies' that are just gluing other schemes together, like if you want to have a "AllBearer" scheme that implements authentication by merging Bearer1 + Bearer2 together.  Requiring something like that to forward seems wrong.

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
HaoK: Yeah nope, good catch don't need it anymore

src/Microsoft.AspNetCore.Authentication/PolicySchemeHandler.cs
Tratcher: Why would they extend this class vs SignInAuthenticationHandler

HaoK: Because they don't might not want to implement SignIn?  I mean that works too, but this one isn't abstract, so they don't have to implement anything

Tratcher: Make it two classes? What for the build in forwarders that requires a default, and one for customization that doesn't?

Tratcher: Yes, throwing NotImplemented is my main suggestion.

HaoK: Ok that's fine, derived policies can always add fallback to the default schemes if they want anyways... updated

Issue #1306, opened 2017-07-10
Update Facebook API to 2.12


Tratcher:
https://developers.facebook.com/docs/apps/changelog
https://developers.facebook.com/docs/facebook-login/manually-build-a-login-flow
The package currently defaults to 2.6 which is available until July 13, 2018.

From a quick test there doesn't appear to be any breaking issues between the current 2.6 and 2.9. 

https://github.com/aspnet/Security/issues/92

PR #1673, opened 2018-02-28
Update facebook API version to 2.12

Tratcher:
#1306 No behavioral changes.
Our current version (2.6) will expire July 13, 2018.

### DIFF
--- a/test/Microsoft.AspNetCore.Authentication.Test/FacebookTests.cs
+++ b/test/Microsoft.AspNetCore.Authentication.Test/FacebookTests.cs
@@ -585,7 +585,7 @@ namespace Microsoft.AspNetCore.Authentication.Facebook
             var transaction = await server.SendAsync("http://example.com/base/login");
             Assert.Equal(HttpStatusCode.Redirect, transaction.Response.StatusCode);
             var location = transaction.Response.Headers.Location.AbsoluteUri;
-            Assert.Contains("https://www.facebook.com/v2.6/dialog/oauth", location);
+            Assert.Contains("https://www.facebook.com/v2.12/dialog/oauth", location);
             Assert.Contains("response_type=code", location);
             Assert.Contains("client_id=", location);
             Assert.Contains("redirect_uri=" + UrlEncoder.Default.Encode("http://example.com/base/signin-facebook"), location);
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Authentication.Test/FacebookTests.cs
```diff
@@ -585,7 +585,7 @@ private void ConfigureDefaults(FacebookOptions o)
             var transaction = await server.SendAsync("http://example.com/base/login");
             Assert.Equal(HttpStatusCode.Redirect, transaction.Response.StatusCode);
             var location = transaction.Response.Headers.Location.AbsoluteUri;
-            Assert.Contains("https://www.facebook.com/v2.6/dialog/oauth", location);
+            Assert.Contains("https://www.facebook.com/v2.12/dialog/oauth", location);
```
HaoK: Nit: Assert using the constants?

Eilon: I could go either way on that. There are pros/cons to using the fields vs. repeating the data.

HaoK: Yeah but in this case isn't it basically a always bug if these ever are inconsistent since we are exposing this in our Defaults class?

Eilon: Pros to using const's:
* Less work to maintain because you make the change in one place

Cons to using const's:
* Less clear when you make a src change to see what effect it has on the product


So, yes, it'd be a bug if it's inconsistent, and this test would tell you that, and you have to fix it (of course). Using a const would avoid the need to have to change the test.

HaoK: But mainly the upside is it makes future updates trivial, as its just updating the constant

Eilon: And that's why I'm fine either way 😄

Tratcher: I actually prefer updating it separately in the tests. It's like having to enter your e-mail address twice to confirm you did it right.

Issue #1588, opened 2018-01-03
Add CookiePolicy logging


Tratcher:
Any changes the policy makes to a given cookie should be logged as Debug, especially if an cookie is ever suppressed by the new consent feature.

Eilon:
Should be using the `Debug` log level.

Tratcher:
Blocked by https://github.com/aspnet/Home/issues/2871.

PR #1678, opened 2018-03-02
Add CookiePolicy logging 

Tratcher:
#1588
I had to work around https://github.com/aspnet/Home/issues/2871 by putting the constructors in a specific order.

@blowdart I did find an is issue where a CookieSecurePolicy.SameAsRequest policy would cause a Secure cookie to be downgraded if it was being sent over HTTP rather than HTTPS. However, users shouldn't be sending secure cookies over HTTP anyways. I changed it to make sure it never downgrades.

### DIFF
--- a/src/Microsoft.AspNetCore.CookiePolicy/LoggingExtensions.cs
+++ b/src/Microsoft.AspNetCore.CookiePolicy/LoggingExtensions.cs
@@ -1,0 +1,105 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Logging
+{
+    internal static class LoggingExtensions
+    {
+        private static Action<ILogger, bool, Exception> _needsConsent;
+        private static Action<ILogger, bool, Exception> _hasConsent;
+        private static Action<ILogger, Exception> _consentGranted;
+        private static Action<ILogger, Exception> _consentWithdrawn;
+        private static Action<ILogger, string, Exception> _cookieSuppressed;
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.CookiePolicy/LoggingExtensions.cs
```diff
@@ -0,0 +1,105 @@
...
+                logLevel: LogLevel.Debug,
+                formatString: "Cookie '{key}' same site mode upgraded to '{mode}'.");
+            _upgradedToHttpOnly = LoggerMessage.Define<string>(
+                eventId: 9,
+                logLevel: LogLevel.Debug,
+                formatString: "Cookie '{key}' upgraded to 'httponly'.");
+        }
+
+        public static void NeedsConsent(this ILogger logger, bool needsConsent)
+        {
```
HaoK: nit: why no => love?

test/Microsoft.AspNetCore.CookiePolicy.Test/CookiePolicyTests.cs
```diff
@@ -102,7 +102,7 @@ public class CookiePolicyTests
                         Assert.Equal("A=A; path=/; samesite=lax", transaction.SetCookie[0]);
                         Assert.Equal("B=B; path=/; samesite=lax", transaction.SetCookie[1]);
                         Assert.Equal("C=C; path=/; samesite=lax", transaction.SetCookie[2]);
-                        Assert.Equal("D=D; path=/; samesite=lax", transaction.SetCookie[3]);
+                        Assert.Equal("D=D; path=/; secure; samesite=lax", transaction.SetCookie[3]);
                     }),
```
HaoK: Nit: are we supposed to test logging at all?

Issue #1607, opened 2018-01-11
Way to add identities to context.User without it getting serialized into cookie


NinoFloris:
Basically there is some very insidious behavior right now that only triggers on refresh or login.

We have extra identities we want to add to the httpcontext ClaimsPrincipal, we do this by pulling them out of a cache on request start and adding them to the ClaimsPrincipal. 

However this goes wrong in the case of a refresh (in our case, as we do work in `ValidatePrincipal`) as the `CookieAuthHandler` references the same instance in the auth ticket as is eventually set on `httpcontext.User` and so will serialize our later additions as well.

Is there any way to stop the `CookieAuthHandler` from doing this? 
(Besides cloning the cookie ClaimsPrincipal, setting that clone as new `httpcontext.User` and then working on just that instance? It feels dirty and the problem feels like an oversight.)

blowdart:
@HaoK @Tratcher 

So is the problem here that changing the identity inside ValidatePrincipal doesn't attach itself to the context until the next request?

NinoFloris:
It doesn't clone the principal it has to write out to the response when ShouldRenew is on, this results in the `httpcontext.User` instance and the `CookieValidationContext` principal instance being the same instance. This means any changes on context.User you'd think are 'request only' are actually sometimes (when ShouldRenew is true) written to cookie. 

I see it like this, `OnValidatePrincipal` should be about creating a principal you want to write to the cookie. Like [SecurityStampValidator](https://github.com/aspnet/Identity/blob/85f8a49aef68bf9763cd9854ce1dd4a26a7c5d3c/src/Identity/SecurityStampValidator.cs) does with delegating that work to `UserClaimsPrincipalFactory`.  

All changes after it ran should be request lifetime and absolutely have no effect on what gets written to cookie. However as it currently shares the instance any changes actually do end up in your cookie when the `Response.OnStarting` action writes it out at the end of the pipeline, way after either event.

Eilon:
@NinoFloris can you share more info about exactly where the request-only changes are made? We think that cloning the principal is the right approach here because that way the cookie serialization won't see the modifications. However, it isn't clear to us at what point the changes are being made.

Tratcher:
No, there's no way to change CookieHandler's behavior here. Your best bet would be to move this functionality from ValidatePrincipal to IClaimsTransformation or a followup middleware and clone the principal before using it.

NinoFloris:
[...]

PR #1679, opened 2018-03-02
Clone tickets for refresh

Tratcher:
#1607 When we are asked to refresh a cookie either via sliding refresh or manual, clone the ticket so that any app modifications are not persisted.
@NinoFloris

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -31,6 +31,7 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
         private DateTimeOffset? _refreshExpiresUtc;
         private string _sessionKey;
         private Task<AuthenticateResult> _readCookieTask;
+        private AuthenticationTicket _refreshTicket;
 
         public CookieAuthenticationHandler(IOptionsMonitor<CookieAuthenticationOptions> options, ILoggerFactory logger, UrlEncoder encoder, ISystemClock clock)
             : base(options, logger, encoder, clock)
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -99,9 +100,27 @@ private void RequestRefresh(AuthenticationTicket ticket)
                 _refreshIssuedUtc = currentUtc;
                 var timeSpan = expiresUtc.Value.Subtract(issuedUtc.Value);
                 _refreshExpiresUtc = currentUtc.Add(timeSpan);
+                _refreshTicket = CloneTicket(ticket);
             }
         }
 
+        private AuthenticationTicket CloneTicket(AuthenticationTicket ticket)
+        {
```
HaoK: Does this introduce any sideeffects now that we are cloning the principal/identities?

test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs
```diff
@@ -1114,6 +1116,51 @@ private Task SignOutAsWrong(HttpContext context)
...
+            var server = CreateServer(o =>
+            {
+                o.ExpireTimeSpan = TimeSpan.FromMinutes(10);
+                o.SlidingExpiration = true;
+                o.Events.OnValidatePrincipal = c =>
+                {
+                    // https://github.com/aspnet/Security/issues/1607
+                    // On sliding refresh the transformed principal should not be serialized into the cookie, only the original principal.
+                    Assert.Single(c.Principal.Identities);
+                    Assert.True(c.Principal.Identities.First().HasClaim("marker", "true"));
```
HaoK: Are they still able to manipulate the principal that gets serialized via some other mechanism?

NinoFloris: Through OnValidatePrincipal like we do atm. And if that's not enough it might need a new event like OnTicketResponse which gets called by the Response.OnStarting handler that writes out the principal.

src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
Tratcher: It's yet another failure point for people that want to use custom principal and identity types, but that's never worked anyways. Even custom identities would mostly work if they override Clone.

test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs
Tratcher: Right, only OnValidatePrincipal is the only touch point where they can modify a deserialized user in such a way that it will be persisted, and only if they set ShouldRenew.

HaoK:
Ok I think the PR is fine, but can we add a few tests to 'spec' the behavior around custom identities and the update validate principal behavior so we at least have some coverage of the existing behavior going forward?

Tratcher:
What spec for custom identities? We've always said it isn't supported. At best you'd have to replace the ticket serializer.

I can add a test for modifications inside validate principal.

Tratcher:
Updated

kanadaj:
@HaoK @Tratcher I believe this change has broken automatic claims refreshes for Identity, so role and claims changes aren't applied to users

kanadaj:
Nevermind that, though I'd recommend adding a separate callback for Identity so as to not modify the original identity pipeline when adding logic for the callback doesn't break the original pipeline

kanadaj:
I believe the issue is because ReplacePrincipal replaces the principal object inside the validation context, but the clone method uses the principal inside the ticket, and after OnValidate is called context.Principal is modified, not context.Ticket.Principal.

Tratcher:
@kanadaj please open a new issue with a repro for your scenario.

Issue #581, opened 2015-11-18
Question about ITicketStore and ApplicationCookie.SessionStore injection


rustd:
Moving this bug from https://github.com/aspnet/Identity/issues/624
@Schlurcher FYI

jusbuc2k:
I'm running into the same issue trying to use a database connection, entity framework, etc. in my ticket store implementation. It seems like the setupAction parameter on the AddIdentity() service collection needs to provide an instance of IServiceProvider in a similiar manner to AddScoped(), AddTransient(), etc. does when using a function to create the service instance.

jusbuc2k:
After looking at this more, it occurred to me that you can just register an IOptions<IdentityOptions>> *before* the call to AddIdentity() in order to get services as part of configuring identity options. For example:

```csharp
services.AddScoped<IOptions<IdentityOptions>>((sp) =>
            {
                var opts = new IdentityOptions();

                opts.Cookies.ApplicationCookie.SessionStore = new MyCustomDatabaseTicketStore(sp.GetService<ApplicationDbContext>());

                return Options.Create(opts);
            });

services.AddIdentity<ApplicationUser, IdentityRole>()
                .AddEntityFrameworkStores<ApplicationDbContext>()                
                .AddDefaultTokenProviders();
```

This works because internally all the identity classes use the DI container/service provider to get IdentityOptions via dependencies on IOptions<IdentityOptions>.

KolyaNET:
In the version of `Microsoft.AspNetCore.Identity 1.1.2`, I get an error when using `Microsoft.AspNetCore.Identity.UserManager.GenerateUserTokenAsync`

> NotSupportedException: No IUserTokenProvider named 'Default' is registered.

As soon as I comment on `services.AddScoped<IOptions<IdentityOptions>((sp) =>...` the error disappears and for example, resetting the password or confirming the email work correctly.

izzyonstage:
If your `MyCustomDatabaseTicketStore` is registered for DI then you could use something similar to what is described here [https://andrewlock.net/access-services-inside-options-and-startup-using-configureoptions/]. This would allow options to be configured that use other DI entities rather than trying to wrap the options yourself.

Personally I have never liked adding `IOptions<T>` things to DI using the "Add..." methods. I have always added options using `services.Configure<T>(...)`. However, that does not have access to the ServiceProvider, so as shown in the link above there is the IConfigureOptions<T> mechanism that can be used to do this.

Eilon:
[...]

PR #1719, opened 2018-04-06
Enable using a Cookie Session Store from DI

couven92:
A `CookieAuthenticationHandler` will use a `ITicketStore` from DI if and only if the following conditions are met:
1. The `UseSessionStoreFromDI` property of the handler's `CookieAuthenticationOptions` is `true`.
2. The DI container has a registered service for the `ITicketStore` interface, and the returned instance is non-`null`.

Since `UseSessionStoreFromDI` is a new property and new behaviour only is applied when its value is `true`, these changes remain fully backwards compatible.

Code Changes:
* Adds a new boolean property `UseSessionStoreFromDI` to `Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationOptions`.
* `Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationHandler` requests an `ITicketStore` in its constructor with an optional argument.  
   Because the argument is optional, DI can provide the argument as `null` if no `ITicketStore` service is registered. This will ensure backwards-compatability.
* In every case where `CookieAuthenticationHandler` previously got the `ITicketStore` from its `Options.SessionStore`, it now first checks the new `Options.UseSessionStoreFromDI` property and uses the non-null DI-injected `ITicketStore` instance if `UseSessionStoreFromDI` is `true`.

fixes https://github.com/aspnet/Security/issues/581

### DIFF
--- a/samples/CookieSessionSample/MemoryCacheTicketStore.cs
+++ b/samples/CookieSessionSample/MemoryCacheTicketStore.cs
@@ -11,9 +11,9 @@ namespace CookieSessionSample
         private const string KeyPrefix = "AuthSessionStore-";
         private IMemoryCache _cache;
 
-        public MemoryCacheTicketStore()
+        public MemoryCacheTicketStore(IMemoryCache cache)
         {
-            _cache = new MemoryCache(new MemoryCacheOptions());
+            _cache = cache ?? new MemoryCache(new MemoryCacheOptions());
         }
 
         public async Task<string> StoreAsync(AuthenticationTicket ticket)
--- a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
@@ -29,10 +29,17 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
         private string _sessionKey;
         private Task<AuthenticateResult> _readCookieTask;
         private AuthenticationTicket _refreshTicket;
+        private readonly ITicketStore _diSessionStore;
 
-        public CookieAuthenticationHandler(IOptionsMonitor<CookieAuthenticationOptions> options, ILoggerFactory logger, UrlEncoder encoder, ISystemClock clock)
+        private ITicketStore GetSessionStore() => Options.UseSessionStoreFromDI
+            ? _diSessionStore ?? Options.SessionStore
+            : Options.SessionStore;
+
+        public CookieAuthenticationHandler(IOptionsMonitor<CookieAuthenticationOptions> options, ILoggerFactory logger, UrlEncoder encoder, ISystemClock clock, ITicketStore diSessionStore = default)
             : base(options, logger, encoder, clock)
-        { }
+        {
+            _diSessionStore = diSessionStore;
+        }
 
         /// <summary>
         /// The handler calls methods on the events which give the application control at certain points where processing is occurring.
@@ -132,7 +139,8 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
                 return AuthenticateResult.Fail("Unprotect ticket failed");
             }
 
-            if (Options.SessionStore != null)
+            var sessionStore = GetSessionStore();
+            if (sessionStore != null)
             {
                 var claim = ticket.Principal.Claims.FirstOrDefault(c => c.Type.Equals(SessionIdClaim));
                 if (claim == null)
@@ -140,7 +148,7 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
                     return AuthenticateResult.Fail("SessionId missing");
                 }
                 _sessionKey = claim.Value;
-                ticket = await Options.SessionStore.RetrieveAsync(_sessionKey);
+                ticket = await sessionStore.RetrieveAsync(_sessionKey);
                 if (ticket == null)
                 {
                     return AuthenticateResult.Fail("Identity missing in session store");
@@ -152,9 +160,9 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
 
             if (expiresUtc != null && expiresUtc.Value < currentUtc)
             {
-                if (Options.SessionStore != null)
+                if (sessionStore != null)
                 {
-                    await Options.SessionStore.RemoveAsync(_sessionKey);
+                    await sessionStore.RemoveAsync(_sessionKey);
                 }
                 return AuthenticateResult.Fail("Ticket expired");
             }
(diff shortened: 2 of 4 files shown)


src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs
```diff
@@ -29,10 +29,17 @@ public class CookieAuthenticationHandler : SignInAuthenticationHandler<CookieAut
...
         private Task<AuthenticateResult> _readCookieTask;
         private AuthenticationTicket _refreshTicket;
+        private readonly ITicketStore _diSessionStore;
 
-        public CookieAuthenticationHandler(IOptionsMonitor<CookieAuthenticationOptions> options, ILoggerFactory logger, UrlEncoder encoder, ISystemClock clock)
+        private ITicketStore GetSessionStore() => Options.UseSessionStoreFromDI
+            ? _diSessionStore ?? Options.SessionStore
+            : Options.SessionStore;
+
+        public CookieAuthenticationHandler(IOptionsMonitor<CookieAuthenticationOptions> options, ILoggerFactory logger, UrlEncoder encoder, ISystemClock clock, ITicketStore diSessionStore = default)
```
Tratcher: No API signature changes are allowed in 2.1. You'll have to add a new constructor.

Tratcher: @pakrym does DI support optional parameters like this? We don't usually use them. We tend to use `IEnumerable<T>` for optional services.

pakrym: It does, usually, we use multiple constructors to achieve optional args. Usage of IEnumerable optional services is generally discouraged.

samples/CookieSessionSample/MemoryCacheTicketStore.cs
```diff
@@ -11,9 +11,9 @@ public class MemoryCacheTicketStore : ITicketStore
         private const string KeyPrefix = "AuthSessionStore-";
         private IMemoryCache _cache;
 
-        public MemoryCacheTicketStore()
+        public MemoryCacheTicketStore(IMemoryCache cache)
         {
-            _cache = new MemoryCache(new MemoryCacheOptions());
+            _cache = cache ?? new MemoryCache(new MemoryCacheOptions());
```
pakrym: If this ctor gets injected into `cache` would never be null.

pakrym:
Did we ever put `ITicketStore` to DI? Can't we just enable the feature without the flag? If options.Store == null resolve from DI

Tratcher:
We never did put it in DI though users might have. There's also the issue of having multiple cookie middleware, they don't all necessarily want the same store. The option may still be redundant though, it may be sufficient to prefer the instance on Options over the instance from DI.

HaoK:
Do we really need this functionality at all?  You can already use a custom IConfigureNamedOptions that gets ITicketStore from DI and use that to set the instance...

```
   public class CookieConfig : IConfigureNamedOptions<CookieAuthenticationOptions> {
       public CookieConfig(ITicketStore store) => _store = store;
       public void Configure(string name, CookieAuthenticationOptions o) { o.TicketStore = store; }
   }
```

HaoK:
That seems good enough to me

Tratcher:
Defining your own class and registering that in DI is more work and less discoverable than just registering the ITokenStore in DI. Though if you also have to set an option then the discoverability suffers there too.

Tratcher:
File a doc bug to add that example to https://docs.microsoft.com/en-us/aspnet/core/security/authentication/cookie?tabs=aspnetcore2x ?

HaoK:
That's just the general pattern of how to get things from DI onto options, its not particularly specific to cookies, @scottaddie @Rick-Anderson I think @Tratcher has a good point that we should have an explicit section recommending this pattern of how to configure options instances with things from DI in this doc maybe?  https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options

```
public class UseSomethingFromDI : IConfigureNamedOptions<YourOptions> {
       public CookieConfig(ISomethingFromDI thing) => _thing = thing;
       public void Configure(string name, YourOptions o) { o.Thing = thing; }
   }
   services.AddTransient<IConfigureOptions<YourOptions>, UseSomethingFromDI>();
```

HaoK:
Also Its certainly more work in the long run if we adopted something like this everywhere (useXyzFromDI flags on all services?), instead of just always using the options instance, and relying on configuring the options instance appropriately.

Rick-Anderson:
https://github.com/aspnet/Docs/issues/5922  Add doc section - Enable using a Cookie Session Store from DI

PinpointTownes:
Instead of a `UseSessionStoreFromDI` option, why not `SessionStoreType`, modeled after `EventsType`? This way, you could register multiple stores in the DI and each handler would pick the one set in the options.

HaoK:
I still say we should do nothing special here, since this already can be done via options already

HaoK:
EventsType was added before named options was fully baked, I don't think we'd have added that either since you can accomplish that via a custom named IConfigureOptions as well

HaoK:
So this is actually even easier in 2.1: services.AddOptions<CookieAuthenticationOptions>(CookieAuthenticationDefaults.AuthenticationScheme).Configure<ITicketStore>(o,t => o.TicketStore = t);

PinpointTownes:
> EventsType was added before named options was fully baked, I don't think we'd have added that either since you can accomplish that via a custom named IConfigureOptions as well

No, it was added because there was a need for scoped "services as options", that you can't do with options, where everything is a singleton.

HaoK:
Ah right events are recreated per request, but that's not what ticket store needs here

PinpointTownes:
If the ticket stored uses EF Core, it needs to be scoped, since `DbContext` itself must be scoped.

HaoK:
You probably can make a particular options type scoped for an IOptionsMonitor by plugging in a no-op cache:

https://github.com/aspnet/Options/blob/dev/src/Microsoft.Extensions.Options/IOptionsMonitorCache.cs
```
services.AddSingleton<IOptionsMonitorCache<CookieAuthenticationOptions>, AlwaysCallFactoryCache>();
```

PinpointTownes:
> You probably can make a particular options type scoped for an IOptionsMonitor by plugging in a no-op cache:

Heh, that means the entire options and all the "services" attached are re-created per request, including the `IDataFormat` and the underlying data protector. Rather inefficient.

The real question is: is using scoped ticket stores common enough to introduce DI support?

Eilon:
Closing because we are not planning to add direct support for this feature.

To work around this, you can use the service located pattern to get the required services imperatively instead of through typical DI.

Issue #1695, opened 2018-03-13
Twitter Handler OAuth Signing code is incorrect


speige:
I'm not aware of a specific bug caused by this. My guess is it work 99% of the time with the current api calls, but if it were used for any other twitter api calls, they'd all fail.

If you read the documentation here:
https://developer.twitter.com/en/docs/basics/authentication/guides/creating-a-signature
you'll see that the code here doesn't match:
https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.Twitter/TwitterHandler.cs

One example is: All uses of UrlEncoder.Encode should be replaced with Uri.EscapeDataString
There are several others. I found this because I copied the code in order to call other twitter apis & they failed until I made corrections.

In TwitterHandler.cs all the methods have a slightly different copy/pasted OAuth code. I've extracted it to a generic method. I've tested & it is working on several different twitter apis, I haven't tested it on all of them, so it's possible there's still an issue somewhere (for example, I expect all data to be in the query params, but some apis may have a body instead).

The existing requests in TwitterHandler would need to be refactored to use this generic method. I haven't coded that since I'm using the built-in version for authentication & my version for extra api calls.

```
using System;
using System.Collections.Generic;
using System.Collections.Immutable;
using System.Globalization;
using System.Linq;
using System.Net.Http;
using System.Security.Claims;
using System.Security.Cryptography;
using System.Text;
using System.Text.Encodings.Web;
using System.Threading.Tasks;
using Microsoft.AspNetCore.Authentication.Twitter;
using Newtonsoft.Json.Linq;

namespace Microsoft.AspNetCore.Authentication.Twitter
{
	public class TwitterHandler
	{
		protected string ComputeSignature(string consumerSecret, string tokenSecret, string signatureData)
		{
			using (HMACSHA1 algorithm = new HMACSHA1())
			{
				algorithm.Key = Encoding.ASCII.GetBytes(string.Format(CultureInfo.InvariantCulture, "{0}&{1}", Uri.EscapeDataString(consumerSecret), string.IsNullOrEmpty(tokenSecret) ? string.Empty : Uri.EscapeDataString(tokenSecret)));
				return Convert.ToBase64String(algorithm.ComputeHash(Encoding.ASCII.GetBytes(signatureData)));
			}
		}

[...]

PR #1720, opened 2018-04-06
Correct Twitter OAuth signing code. Also, refactor copy/pasted code & make reusable TwitterHandler.ExecuteRequestAsync public

speige:
https://github.com/aspnet/Security/issues/1695

Tested in my app & working

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication.Twitter/TwitterHandler.cs
+++ b/src/Microsoft.AspNetCore.Authentication.Twitter/TwitterHandler.cs
@@ -148,53 +148,92 @@ namespace Microsoft.AspNetCore.Authentication.Twitter
             await Events.RedirectToAuthorizationEndpoint(redirectContext);
         }
 
-        private async Task<RequestToken> ObtainRequestTokenAsync(string callBackUri, AuthenticationProperties properties)
+        private async Task<HttpResponseMessage> ExecuteRequestAsync(string url, HttpMethod httpMethod, RequestToken accessToken = null, Dictionary<string, string> extraOAuthPairs = null, Dictionary<string, string> queryParameters = null, Dictionary<string, string> formData = null)
         {
-            Logger.ObtainRequestToken();
-
-            var nonce = Guid.NewGuid().ToString("N");
-
-            var authorizationParts = new SortedDictionary<string, string>
+            var authorizationParts = new SortedDictionary<string, string>(extraOAuthPairs ?? new Dictionary<string, string>())
             {
-                { "oauth_callback", callBackUri },
                 { "oauth_consumer_key", Options.ConsumerKey },
-                { "oauth_nonce", nonce },
+                { "oauth_nonce", Guid.NewGuid().ToString("N") },
                 { "oauth_signature_method", "HMAC-SHA1" },
                 { "oauth_timestamp", GenerateTimeStamp() },
                 { "oauth_version", "1.0" }
             };
 
-            var parameterBuilder = new StringBuilder();
-            foreach (var authorizationKey in authorizationParts)
+            if (accessToken != null)
             {
-                parameterBuilder.AppendFormat("{0}={1}&", UrlEncoder.Encode(authorizationKey.Key), UrlEncoder.Encode(authorizationKey.Value));
+                authorizationParts.Add("oauth_token", accessToken.Token);
+            }
+
+            var signatureParts = new SortedDictionary<string, string>(authorizationParts);
+            if (queryParameters != null)
+            {
+                foreach (var queryParameter in queryParameters)
+                {
+                    signatureParts.Add(queryParameter.Key, queryParameter.Value);
+                }
+            }
+            if (formData != null)
+            {
+                foreach (var formItem in formData)
+                {
+                    signatureParts.Add(formItem.Key, formItem.Value);
+                }
+            }
+
+            var parameterBuilder = new StringBuilder();
+            foreach (var signaturePart in signatureParts)
+            {
+                parameterBuilder.AppendFormat("{0}={1}&", Uri.EscapeDataString(signaturePart.Key), Uri.EscapeDataString(signaturePart.Value));
             }
             parameterBuilder.Length--;
             var parameterString = parameterBuilder.ToString();
 
             var canonicalizedRequestBuilder = new StringBuilder();
-            canonicalizedRequestBuilder.Append(HttpMethod.Post.Method);
+            canonicalizedRequestBuilder.Append(httpMethod.Method);
             canonicalizedRequestBuilder.Append("&");
-            canonicalizedRequestBuilder.Append(UrlEncoder.Encode(RequestTokenEndpoint));
+            canonicalizedRequestBuilder.Append(Uri.EscapeDataString(url));
             canonicalizedRequestBuilder.Append("&");
-            canonicalizedRequestBuilder.Append(UrlEncoder.Encode(parameterString));
+            canonicalizedRequestBuilder.Append(Uri.EscapeDataString(parameterString));
 
-            var signature = ComputeSignature(Options.ConsumerSecret, null, canonicalizedRequestBuilder.ToString());
+            var signature = ComputeSignature(Options.ConsumerSecret, accessToken?.TokenSecret, canonicalizedRequestBuilder.ToString());
             authorizationParts.Add("oauth_signature", signature);
+
+            var queryString = "";
+            if (queryParameters != null)
+            {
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Authentication.Twitter/TwitterHandler.cs
```diff
@@ -19,353 +20,257 @@
 
 namespace Microsoft.AspNetCore.Authentication.Twitter
 {
-    public class TwitterHandler : RemoteAuthenticationHandler<TwitterOptions>
+  public class TwitterHandler : RemoteAuthenticationHandler<TwitterOptions>
```
Tratcher: Spacing changed? Revert.

src/Microsoft.AspNetCore.Authentication.Twitter/TwitterHandler.cs
```diff
@@ -19,353 +20,257 @@
...
+
+      Response.Cookies.Append(Options.StateCookie.Name, Options.StateDataFormat.Protect(requestToken), cookieOptions);
+
+      var redirectContext = new RedirectContext<TwitterOptions>(Context, Scheme, Options, properties, twitterAuthenticationEndpoint);
+      await Events.RedirectToAuthorizationEndpoint(redirectContext);
+    }
+
+		public async Task<HttpResponseMessage> ExecuteRequestAsync(string url, HttpMethod httpMethod, RequestToken accessToken = null, Dictionary<string, string> extraOAuthPairs = null, Dictionary<string, string> queryParameters = null, Dictionary<string, string> formData = null)
+    {
+      StringBuilder canonicalizedRequestBuilder = new StringBuilder();
```
Eilon: Also use `var` everywhere in the new code.

speige:
Accidental whitespace changes removed

speige:
updated based on request to use "var" instead of specific types

speige:
@Tratcher 
The original intent of the PR wasn't to make the reusable method public (although that's an added bonus). It was because the code didn't match Twitter's API Spec (based on the issue I submitted). It seemed to work most of the time, but I was nervous there would be random hard to debug errors.

To get an instance of TwitterHandler via the constructor is painful. However, with DI it's not too bad.

I'm use .net core with SimpleInjector instead of the base DI. You have to use CrossWire to make classes registered with the built-in DI available to your custom classes. I assume it's trivial with the built-in DI, so I'm leaving out the DI code itself.

After DI of IHttpContextAccessor & AuthenticationHandlerProvider.
```
	        TwitterHandler twitterHandler = (TwitterHandler)await _authenticationHandlerProvider.GetHandlerAsync(_httpContextAccessor.HttpContext, "Twitter");
```

You could also DI TwitterHandler, but the Scheme won't be set, so you have to call InitializeAsync (which also requires DI of HttpContext).

speige:
I'm confused. Pretty much every test in TwitterTests.cs already calls ExecuteRequestAsync. If you add a breakpoint to the method & debug unit tests, you'll see it.

I'm kindof braindead when it comes to unit testing, so I'm not confident I can write a test that will be satisfactory. Would you mind coding that portion if it's important for the pull request to be accepted? Sorry :(

Tratcher:
Sorry, I meant that shows what it's like to call the API directly like an end user would.

PinpointTownes:
Having public APIs in the handlers classes reminds me of a cool story about Pandora and a Box... :trollface: 

IMHO, pointing folks to user-friendly Twitter clients like https://github.com/JoeMayo/LinqToTwitter or https://github.com/linvi/tweetinvi would be much better than exposing an obscure method in the handler itself (both will take care of generating the `Authorization` header with all the signing madness for you)

speige:
I'm not opposed to leaving the method private if it raises concerns. The initial intent of the pull request was because the current signing method did not match the spec. However, in that case, why have aspnet/Security implement the signing code itself instead of also referencing a 3rd party library?

Eilon:
Let's keep it private for now because it might otherwise be difficult code to maintain forever.

src/Microsoft.AspNetCore.Authentication.Twitter/TwitterHandler.cs
```diff
@@ -148,53 +149,52 @@ public TwitterHandler(IOptionsMonitor<TwitterOptions> options, ILoggerFactory lo
...
-            request.Headers.Add("Authorization", authorizationHeaderBuilder.ToString());
+            return await Backchannel.SendAsync(request, Context.RequestAborted);
+        }
+
+        private async Task<RequestToken> ObtainRequestTokenAsync(string callBackUri, AuthenticationProperties properties)
+        {
+            Logger.ObtainRequestToken();
 
-            var response = await Backchannel.SendAsync(request, Context.RequestAborted);
+            var response = await ExecuteRequestAsync(RequestTokenEndpoint, HttpMethod.Post, null, new Dictionary<string, string>() { { "oauth_callback", callBackUri } });
```
Tratcher: You can drop the null param, it has a default, and use a named parameter for the last param. The same goes for the other calls to ExecuteRequestAsync.

src/Microsoft.AspNetCore.Authentication.Twitter/TwitterHandler.cs
```diff
@@ -148,53 +149,52 @@ public TwitterHandler(IOptionsMonitor<TwitterOptions> options, ILoggerFactory lo
             await Events.RedirectToAuthorizationEndpoint(redirectContext);
         }
 
-        private async Task<RequestToken> ObtainRequestTokenAsync(string callBackUri, AuthenticationProperties properties)
+        public async Task<HttpResponseMessage> ExecuteRequestAsync(string url, HttpMethod httpMethod, RequestToken accessToken = null, Dictionary<string, string> extraOAuthPairs = null, Dictionary<string, string> queryParameters = null, Dictionary<string, string> formData = null)
```
Tratcher: Let's make this private for now.

Wrap long lines over ~160 chars

src/Microsoft.AspNetCore.Authentication.Twitter/TwitterHandler.cs
```diff
@@ -148,53 +149,52 @@ public TwitterHandler(IOptionsMonitor<TwitterOptions> options, ILoggerFactory lo
...
             }
-            parameterBuilder.Length--;
-            var parameterString = parameterBuilder.ToString();
 
-            var canonicalizedRequestBuilder = new StringBuilder();
-            canonicalizedRequestBuilder.Append(HttpMethod.Post.Method);
-            canonicalizedRequestBuilder.Append("&");
-            canonicalizedRequestBuilder.Append(UrlEncoder.Encode(RequestTokenEndpoint));
+            var signatureParts = new SortedDictionary<string, string>((queryParameters ?? new Dictionary<string, string>()).Union(authorizationParts).Union(formData ?? new Dictionary<string, string>()).ToDictionary(x => x.Key, x => x.Value));
+            string parameterString = signatureParts.Select(x => string.Format("{0}={1}", Uri.EscapeDataString(x.Key), Uri.EscapeDataString(x.Value))).Aggregate((x, y) => string.Format("{0}&{1}", x, y));
```
Tratcher: Please revert the complex linq expressions. It reduces readability and code efficiency compared to the StringBuilder pattern that was used before.

src/Microsoft.AspNetCore.Authentication.Twitter/TwitterHandler.cs
```diff
@@ -148,53 +149,52 @@ public TwitterHandler(IOptionsMonitor<TwitterOptions> options, ILoggerFactory lo
...
+                authorizationParts.Add("oauth_token", accessToken.Token);
             }
-            parameterBuilder.Length--;
-            var parameterString = parameterBuilder.ToString();
 
-            var canonicalizedRequestBuilder = new StringBuilder();
-            canonicalizedRequestBuilder.Append(HttpMethod.Post.Method);
-            canonicalizedRequestBuilder.Append("&");
-            canonicalizedRequestBuilder.Append(UrlEncoder.Encode(RequestTokenEndpoint));
+            var signatureParts = new SortedDictionary<string, string>((queryParameters ?? new Dictionary<string, string>()).Union(authorizationParts).Union(formData ?? new Dictionary<string, string>()).ToDictionary(x => x.Key, x => x.Value));
```
Tratcher: This is the one linq expression that it may make sense to keep, but it needs a more readable layout rather than stringing it out all on one line.

Tratcher:
Thanks for the help

PR #1742, opened 2018-04-30
Code improvements of JwtBearer module

AMalininHere:
Refactoring of `JwtBearerHandler` to simplify code


src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs
```diff
@@ -283,37 +282,32 @@ private static string CreateErrorDescription(Exception authFailure)
             {
                 // Order sensitive, some of these exceptions derive from others
```
Tratcher: Is this comment still relevant? Does the switch select the most specific match in a type hierarchy?

AMalininHere:
Not at all

Issue #1764, opened 2018-05-22
AuthorizationPolicyBuilder ArgumentNullException when using a PolicyProvider


nrandell:
I've got a situation where I'm trying to use an IAuthorizationPolicyProvider to provide a policy for the logged in user.

However I keep seeing ArgumentNullException in Authorizationpolicybuilder.Combine

```
System.ArgumentNullException: Value cannot be null.
Parameter name: policy
   at Microsoft.AspNetCore.Authorization.AuthorizationPolicyBuilder.Combine(AuthorizationPolicy policy)
   at Microsoft.AspNetCore.Authorization.AuthorizationPolicyBuilder..ctor(AuthorizationPolicy policy)
   at Microsoft.AspNetCore.Mvc.Authorization.AuthorizeFilter.GetEffectivePolicyAsync(AuthorizationFilterContext context)
   at Microsoft.AspNetCore.Mvc.Authorization.AuthorizeFilter.OnAuthorizationAsync(AuthorizationFilterContext context)
   at Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker.InvokeFilterPipelineAsync()
   at Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker.InvokeAsync()
   at Microsoft.AspNetCore.Builder.RouterMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.StaticFiles.StaticFileMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore.MigrationsEndPointMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore.DatabaseErrorPageMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore.DatabaseErrorPageMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware.Invoke(HttpContext context)
```

As far as I can tell, the issue is caused in `GetEffectivePolicyAsync`

As the filter was created using an `IAuthorizationPolicyProvider`, Policy is null.
[...]

PR #1766, opened 2018-05-22
AuthZ: Combine should ignore null policies

HaoK:
Fixes https://github.com/aspnet/Security/issues/1764 and https://github.com/aspnet/Mvc/issues/7809

cc @Tratcher @ajcvickers 

I'll file a separate PR that adds a test for this scenario in MVC that depends on this PR to pass


HaoK:
We've decided to leave this behavior alone, since we will fix the MVC issue that was resulting in trying to combine null policies

PR #1767, opened 2018-05-23
Include AuthenticationTicket.Properties in Authentication success

### LINKED ISSUE #1765: HttpContext.GetTokenAsync fails after Upgrade to 2.1.0
Hello,
I have some code that's calling `await HttpContext.GetTokenAsync("access_token")`. When using `Microsoft.AspNetCore.Authentication.JwtBearer` version 2.0.4 it works, but upgrading to 2.1.0-preview1-final or newer results in null being returned.

I believe this is because on line 73 in [Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs](https://github.com/aspnet/HttpAbstractions/blob/bcb7bcbb7a6461204f4cefd9848e2a782ebb0a6b/src/Microsoft.AspNetCore.Authentication.Core/AuthenticationService.cs#L73)
`return AuthenticateResult.Success(new AuthenticationTicket(transformed, result.Properties, result.Ticket.AuthenticationScheme));` result.Properties is null.

When using JwtBearer auth, result.Properties is never set - however, the properties that are expected to be there are present on result.Ticket. Changing line 73 to `return AuthenticateResult.Success(new AuthenticationTicket(transformed, result.Ticket.Properties, result.Ticket.AuthenticationScheme));` appears to fix the issue for me, is there any reason not to do so?
Thanks

TimHess:
Fixes #1765

### DIFF
--- a/src/Microsoft.AspNetCore.Authentication/HandleRequestResult.cs
+++ b/src/Microsoft.AspNetCore.Authentication/HandleRequestResult.cs
@@ -33,7 +33,7 @@ namespace Microsoft.AspNetCore.Authentication
             {
                 throw new ArgumentNullException(nameof(ticket));
             }
-            return new HandleRequestResult() { Ticket = ticket };
+            return new HandleRequestResult() { Ticket = ticket, Properties = ticket.Properties };
         }
 
         /// <summary>


HaoK:
@Tratcher we should have a test for this in security

Tratcher:
For SaveToken overall, yes.

Tratcher:
Filed https://github.com/aspnet/Security/issues/1809

PR #1775, opened 2018-06-07
Fixed keyword order.

zippy1981:
This is the preferred order according to ReSharper.


Tratcher:
Sorry, we don't take PRs for asthetic only changes. I will take a look at your docs PR though.

PR #1782, opened 2018-06-13
Set 2.1 baselines

ryanbrandenburg:


src/Microsoft.AspNetCore.Authentication.Cookies/baseline.netcore.json
```diff
@@ -239,10 +239,8 @@
       "Name": "Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationHandler",
       "Visibility": "Public",
       "Kind": "Class",
-      "BaseType": "Microsoft.AspNetCore.Authentication.AuthenticationHandler<Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationOptions>",
-      "ImplementedInterfaces": [
-        "Microsoft.AspNetCore.Authentication.IAuthenticationSignInHandler"
-      ],
+      "BaseType": "Microsoft.AspNetCore.Authentication.SignInAuthenticationHandler<Microsoft.AspNetCore.Authentication.Cookies.CookieAuthenticationOptions>",
```
ryanbrandenburg: @HaoK this change in base class and the accompanying change in method names are breaking changes in 2.1, I assume they were approved?

HaoK: Yeah its not as bad of a breaking change as the diffs appear.

SignInAsync is implemented as a virtual base class, and we added a new virtual protected HandleSignInAsync that is called by default. So the end result the old public surface area is the same, its just refactored into different base classes

HaoK: This was discussed in the PR with Eilon so we should be good, see https://github.com/aspnet/Security/pull/1665

Issue #1788, opened 2018-06-21
[Regression] Role/claim changes in identity aren't persisted into the auth cookie


kanadaj:
When ASP.NET Core Identity tries to update the current user's claims, it updates the Principal of the current request, however the update to the claims is not persisted to the cookie.

I've also made an issue for this in aspnet/Identity before I've managed to figure out the details: https://github.com/aspnet/Identity/issues/1845

I'm not sure which repo this issue really belongs to considering this seems like an interop issue.

Repro repo with steps: https://github.com/kanadaj/aspnetrefreshrolesrepro/tree/master

kanadaj:
@Tratcher here you go. I still think the problem is that Identity modifies `context.Principal` but Security persists `context.Ticket.Principal`; a quick verification of this would be welcome since I'm running a system in production and it's causing issues for the users.

kanadaj:
https://github.com/aspnet/Security/blob/1df139eb6d8ef1c1e74a14dd20c9dd13eb6e0e23/test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs#L958
This test doesn't cover the case where ctx.ReplacePrincipal() is used to replace the principal with a new one, since you're using the reference directly.

kanadaj:
As a quick and dirty workaround, this seems to work
```
options.OnRefreshingPrincipal = context =>
{
    // Any original code in OnRefreshingPrincipal
    // ...

    // Copy the new principal claims into the old value to keep the reference to context.ticket.Principal
    var oldIdentity = context.CurrentPrincipal.Identities.First();
    var newIdentity = context.NewPrincipal.Identities.First();

    Claim claim;
    while ((claim = oldIdentity.Claims.FirstOrDefault()) != null)
    {
        oldIdentity.RemoveClaim(claim);
    }

    foreach (var newClaim in newIdentity.Claims)
    {
        oldIdentity.AddClaim(newClaim);
    }

    return Task.FromResult(0);
};
```

Tratcher:
@HaoK can you take a look? If this regressed then it's a patch candidate.

HaoK:
Sure, I wouldn't be surprised if this was a regression from the cloning change

HaoK:
Well so this is interesting, looks like the security stamp might have been relying on the old behavior before to persist updates to the cookie:

See https://github.com/aspnet/Identity/blob/dev/src/Identity/SecurityStampValidator.cs#L80

[...]

PR #1795, opened 2018-06-28
Fix regression in Cookies

HaoK:
Fix for https://github.com/aspnet/Security/issues/1788

CookieCanBeUpdatedByValidatorDuringRefresh is the new regression test that fails for the identity scenario.

@Tratcher @ajcvickers @blowdart

### DIFF
--- a/test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs
+++ b/test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs
@@ -901,6 +901,76 @@ namespace Microsoft.AspNetCore.Authentication.Cookies
         }
 
         [Fact]
+        public async Task CookieCanBeReplacedByValidator()
+        {
+            var server = CreateServer(o =>
+            {
+                o.Events = new CookieAuthenticationEvents
+                {
+                    OnValidatePrincipal = ctx =>
+                    {
+                        ctx.ShouldRenew = true;
+                        ctx.ReplacePrincipal(new ClaimsPrincipal(new ClaimsIdentity(new GenericIdentity("Alice2", "Cookies2"))));
+                        return Task.FromResult(0);
+                    }
+                };
+            },
+            context =>
+                context.SignInAsync("Cookies",
+                    new ClaimsPrincipal(new ClaimsIdentity(new GenericIdentity("Alice", "Cookies")))));
+
+            var transaction1 = await SendAsync(server, "http://example.com/testpath");
+
+            var transaction2 = await SendAsync(server, "http://example.com/me/Cookies", transaction1.CookieNameValue);
+            Assert.NotNull(transaction2.SetCookie);
+            Assert.Equal("Alice2", FindClaimValue(transaction2, ClaimTypes.Name));
+        }
+
+        [Fact]
+        public async Task CookieCanBeUpdatedByValidatorDuringRefresh()
+        {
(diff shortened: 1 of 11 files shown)


test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs
```diff
@@ -900,6 +900,76 @@ private Task SignOutAsWrong(HttpContext context)
...
+        public async Task CookieCanBeReplacedByValidator()
+        {
+            var server = CreateServer(o =>
+            {
+                o.Events = new CookieAuthenticationEvents
+                {
+                    OnValidatePrincipal = ctx =>
+                    {
+                        ctx.ShouldRenew = true;
+                        ctx.ReplacePrincipal(new ClaimsPrincipal(new ClaimsIdentity(new GenericIdentity("Alice2", "Cookies2"))));
```
Tratcher: Is it not possible to update the AuthProperties at the same time?

HaoK: Hrm, we might have a similar bug with properties I guess due to the cloning change.  Renew probably is stomping on any changes that were applied in the context

blowdart:
Do it.

poke:
I’ve touched on this in https://github.com/aspnet/Mvc/issues/8011, and just want to check:

Do we really want to make a security stamp validation with a successful result refresh the identity, making it update all claims? Currently, we didn’t have this behavior before and an identity would always be the same until we sign in again. But with this, we are essentially updating the claims identity on every security stamp validation interval, which might be a nice but similarly also confusing side-effect.

HaoK:
@poke this is how this security stamp feature was originally designed to work, identity would take care of ensuring a fresh set of user claims are regenerated every so often (validationInterval), which should be determined based on a performance/staleness tradeoff since as you pointed out, its not cheap to hit the database/generate user claims.

Tratcher:
More specifically, it worked that way in 2.0 but regressed in 2.1.

test/Microsoft.AspNetCore.Authentication.Test/CookieTests.cs
HaoK: There's no issue with properties because the context ends up taking the ticket properties 

https://github.com/aspnet/Security/blob/master/src/Microsoft.AspNetCore.Authentication.Cookies/Events/CookieValidatePrincipalContext.cs#L23

so updates to the ticket's properties show up automatically when they get copied:

https://github.com/aspnet/Security/blob/master/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs#L113

since ticket.Properties == context.Properties

I'll update the test to set a property, but I don't think we need to change the fix

HaoK:
@ajcvickers I think the fix in this PR is good for patch approval now

HaoK:
@natemcmaster this was approved for patch, should I merge this to rel/2.1 and then port it back to master?

natemcmaster:
Yes. Change this PR to target release/2.1 instead of master. Then, merge release/2.1 into release/2.2 (the bot hasn't automated this one yet). After that, the bot will open a PR to merge your release/2.2 change into master.

HaoK:
Replaced by https://github.com/aspnet/Security/pull/1811 since the rebase didn't go well

PR #1817, opened 2018-07-13
Auth policy does not call authority server in order to obtain auth manifest

bielu000:
With the code below, JwtBearer does not automaticly call authority service in order to obtain authentication / authorization manifest eg. public key, and does not occur any errors like "unable to retrive document".

//ConfigureServices
`services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
.AddJwtBearer(options =>
{
    options.Authority = "http://localhost:5050";
    options.Audience = "{yourAudience}";
});`

//Configure
`app.UseAuthentication`


Tratcher:
I don't think  you intended this to be a pull request as it does not contain any new commits. Please open an issue at https://github.com/aspnet/Security/issues with more details about your scenario.

PR #1831, opened 2018-07-31
Change description of role in RequireRole method

woeterman94:
The description of the pram roles feels ambiguous. Make it clear that the user should belong to one of the roles and not all of them.
As documented: https://docs.microsoft.com/en-us/aspnet/core/security/authorization/roles?view=aspnetcore-2.1


Eilon:
@blowdart can you share your thoughts on this new phrasing?

I agree the old phrasing was misleading. I'm not sure I'm crazy about the new wording either, but I do think it's better.

blowdart:
It's better. How about "The collection/array of roles where membership of at least one of the roles fulfills the requirement"? Or is that too wordy?

woeterman94:
That's too wordy, and perhaps more confusing.

Eilon:
Discussed in triage - this change is better than what was previously there. Thanks for the PR!

PR #1863, opened 2018-09-25
Update Facebook to v3.1 endpoints

Tratcher:
#92 Facebook has changed its API version from v2.12 to v3.1. It has also removed the public_profile scope, this permission is available by default. See https://developers.facebook.com/docs/graph-api/changelog/version3.0#login

None of the other providers have made any relevant API changes. I did add some doc links for next time.

### DIFF
--- a/samples/SocialSample/Startup.cs
+++ b/samples/SocialSample/Startup.cs
@@ -59,6 +59,7 @@ namespace SocialSample
                 .AddCookie(o => o.LoginPath = new PathString("/login"))
                 // You must first create an app with Facebook and add its ID and Secret to your user-secrets.
                 // https://developers.facebook.com/apps/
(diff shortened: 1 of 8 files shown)


HaoK:
Looks fine, why is the VSTS CI failing tho?

natemcmaster:
Same cause: the tests rely on global machine state. This is always a bad thing for tests. The fix is to change the test project to persist dataprotection keys to a temporary directory. https://docs.microsoft.com/en-us/aspnet/core/security/data-protection/configuration/overview?view=aspnetcore-2.1&tabs=aspnetcore2x#persistkeystofilesystem

Tratcher:
I'm am dubious of that explanation. DataProtection was designed to be be shared between multiple instances of an application and should be handling concurrency issues like this. We've used it this way for  years without issue, why the sudden regression?

Issue #1730, opened 2018-04-17
Map causes wrong ReturnUrl parameter for Cookies


Tratcher:
```
            app.UseAuthentication();

            app.Map("/hello", map =>
            {
                map.Run(async context =>
                {
                    if (!context.User.Identities.Any(identity => identity.IsAuthenticated))
                    {
                        await context.ChallengeAsync();
                        return;
                    }

                    context.Response.ContentType = "text/plain";
                    await context.Response.WriteAsync("Hello old timer");
                });
            });
```

If UseAuthentication is outside of the Map, and ChallengeAsync is called inside the Map, then CookieAuth generates the wrong ReturnUrl parameter. 
https://github.com/aspnet/Security/blob/2e3a1071a884de19d33daa00bbdc2b7614d73646/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs#L436

In the example `http://localhost:12345/hello/world?query`, it uses the OriginalPathBase (empty) but the current request Path "/world" and redirects to "/world?query". It should use the OriginalPathBase "" and the OriginalPath "/hello/world" rather than the current Path.

muratg:
Workaround:
```c#
                    await context.ChallengeAsync(new AuthenticationProperties()
                    {
                        // https://github.com/aspnet/Security/issues/1730
                        // Return here after authenticating
                        RedirectUri = context.Request.PathBase + context.Request.Path + context.Request.QueryString
                    });
```

Eilon:
Won't Fix because there is not enough demand for this.

PR #1887, opened 2018-10-22
Add AccessDeniedPath support to the OIDC/OAuth2/Twitter providers

PinpointTownes:
Fixes https://github.com/aspnet/Security/issues/710 (see also https://github.com/aspnet/Security/pull/1582#issuecomment-355060722 for more information).

This PR introduces native - but opt-in, to avoid a behavioral breaking change - support for automatic redirection for `access_denied` (OIDC/OAuth2) and `denied` (Twitter) errors.

Open questions:
  - Should we flow the `RedirectUri` attached to the `AuthenticationProperties` (when present) up to the access denied endpoint?
  - Do we want a new event allowing to control how the redirection is applied?

Note: I didn't add a test for OIDC because there's no existing test for the callback part (which is crazy, but hem...). Looks like I'm also going to contribute to the technical debt :smile: 

/cc @Eilon @Tratcher


Eilon:
Looks legit to me, assuming that this is more than just a Twitter thing, and actually a proper OIDC thing. @PinpointTownes - can you confirm that?

@Tratcher - can you take a look and we can discuss further on Thursday if needed?

samples/OpenIdConnectSample/Startup.cs
```diff
@@ -126,6 +127,17 @@ await context.SignOutAsync(OpenIdConnectDefaults.AuthenticationScheme, new Authe
                     return;
                 }
 
+                if (context.Request.Path.Equals("/access-denied-from-remote"))
+                {
+                    await context.SignOutAsync(CookieAuthenticationDefaults.AuthenticationScheme);
```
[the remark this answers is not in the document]
PinpointTownes: Bad copy/paste, I'll remove it :sweat_smile:

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -63,6 +63,17 @@ protected override async Task<HandleRequestResult> HandleRemoteAuthenticateAsync
             var error = query["error"];
             if (!StringValues.IsNullOrEmpty(error))
             {
+                // Note: access_denied errors are special protocol errors indicating the user didn't
+                // approve the authorization demand requested by the remote authorization server.
+                // Since it's a frequent scenario (that is not caused by incorrect configuration),
+                // access_denied errors are handled differently if AccessDeniedPath was populated.
+                if (Options.AccessDeniedPath.HasValue && StringValues.Equals(error, "access_denied"))
+                {
+                    Response.Redirect(BuildRedirectUri(Options.AccessDeniedPath));
```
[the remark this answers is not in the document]
PinpointTownes: Yep, these were the 2 open questions:

> Should we flow the RedirectUri attached to the AuthenticationProperties (when present) up to the access denied endpoint?

> Do we want a new event allowing to control how the redirection is applied?

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -63,6 +63,17 @@ protected override async Task<HandleRequestResult> HandleRemoteAuthenticateAsync
             var error = query["error"];
             if (!StringValues.IsNullOrEmpty(error))
             {
+                // Note: access_denied errors are special protocol errors indicating the user didn't
```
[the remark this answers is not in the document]
PinpointTownes: Sure, I'll add it.

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -63,6 +63,17 @@ protected override async Task<HandleRequestResult> HandleRemoteAuthenticateAsync
...
+                // approve the authorization demand requested by the remote authorization server.
+                // Since it's a frequent scenario (that is not caused by incorrect configuration),
+                // access_denied errors are handled differently if AccessDeniedPath was populated.
+                if (Options.AccessDeniedPath.HasValue && StringValues.Equals(error, "access_denied"))
```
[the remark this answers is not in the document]
PinpointTownes: I'm fine changing that if you have something better in mind.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -520,6 +520,18 @@ protected override async Task<HandleRequestResult> HandleRemoteAuthenticateAsync
                 // if any of the error fields are set, throw error null
                 if (!string.IsNullOrEmpty(authorizationResponse.Error))
                 {
+                    // Note: access_denied errors are special protocol errors indicating the user didn't
```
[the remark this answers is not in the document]
PinpointTownes: Will do.

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthOptions.cs
```diff
@@ -81,6 +79,14 @@ public override void Validate()
         /// </summary>
         public string UserInformationEndpoint { get; set; }
 
+        /// <summary>
+        /// Gets or sets the optional path the user agent is redirected to if the user
+        /// doesn't approve the authorization demand requested by the remote server.
+        /// If this property is not populated, an exception is thrown if an access_denied
+        /// response is returned by the remote authorization server to the callback endpoint.
```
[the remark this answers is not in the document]
PinpointTownes: Will add.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -520,6 +520,18 @@ protected override async Task<HandleRequestResult> HandleRemoteAuthenticateAsync
...
+                    // approve the authorization demand requested by the remote authorization server.
+                    // Since it's a frequent scenario (that is not caused by incorrect configuration),
+                    // access_denied errors are handled differently if AccessDeniedPath was populated.
+                    if (Options.AccessDeniedPath.HasValue &&
+                        string.Equals(authorizationResponse.Error, "access_denied", StringComparison.Ordinal))
```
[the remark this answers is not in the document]
PinpointTownes: Because OIDC uses `string.Equals` while OAuth uses `StringValues.Equals`, that doesn't offer an overload accepting an explicit comparison type. I'm fine changing that if you don't like it.

PinpointTownes:
> It loses the AuthProperties context so the app can't flow data across the login attempt.

We could certainly flow the state, but in practice, the access denied path would likely point either to the home page or to the login page (so that the user can start a new authentication flow), not really an endpoint controlled by the OIDC/social handlers. So if we flowed the state, developers would have to instantiate the properties reader + the data protection stuff to deserialize it and get back the `AuthenticationProperties`. Not exactly a trivial thing to do.

The most interesting data is probably the return URL, that we could flow like what the cookies handler does.

> This is almost the same functionality as the ExceptionHandler middleware except it's tailored to one specific error and a lot less flexible in how it's handled.

You've once told me that exceptions were not for flow control: it seems gross to throw (and log) an exception every time a user doesn't consent to the authorization demand :smile: 

The fact the handlers throws an `Exception` and not something super-specific indicating it's an access denied "error" makes using the exception handler a PITA if you want simply want to redirect the user agent to the login page.

> Events would be useful. OIDC has MessageReceived where you could already pre-empt this, OAuth and Twitter don't. Implementing the whole thing as a distinct event would be an alternative to the new property and might help with the AuthProperties and flexibility issues.

The existing alternative is to use the `RemoteFailure` event to override the default error handling logic. I've shared a snippet doing that multiple times here and elsewhere: the fact you have to repeat that in every project for every provider is a real PITA (hence this PR).

If you only want something generic, we already have that in place. My question was more about something very specific (typically an `AccessDenied`/`OnAccessDenied` event).

> We should talk about what support for this looks like in the Identity and OIDC templates. You'd want to populate options and land on a page that said "We're sorry, we really need that information to continue, click here to try again?"

Yep, pretty much what I had in mind. Flowing the `returnUrl` in the query string would allow be a good start to make restarting the authentication process easier. We could also flow the provider scheme, but I'm not so sure it's a safe approach (except if we only use it as a hint, maybe).

Tratcher:
>> Events would be useful. OIDC has MessageReceived where you could already pre-empt this, OAuth and Twitter don't. Implementing the whole thing as a distinct event would be an alternative to the new property and might help with the AuthProperties and flexibility issues.

> The existing alternative is to use the RemoteFailure event to override the default error handling logic. I've shared a snippet doing that multiple times here and elsewhere: the fact you have to repeat that in every project for every provider is a real PITA (hence this PR).

> If you only want something generic, we already have that in place. My question was more about something very specific (typically an AccessDenied/OnAccessDenied event).

I was proposing a new dedicated event for the access denied scenario and the ability to intercept the response prior to an exception being thrown.

PinpointTownes:
I took a look at how the different Security components handle URIs and woooo, things are insanely inconsistent:

  - The cookie handler stores the current URL (when the challenge is triggered) in `AuthenticationProperties.RedirectUri` as a relative URL.
  - The OIDC/OAuth2/Twitter/WS-Fed store the current URL in `AuthenticationProperties.RedirectUri` as an absolute URL.
  - The cookie handler ensures the login/logout URLs eventually passed to `Response.Redirect()` are absolute using the `BuildRedirectUri` helper.
  - The cookie handler doesn't do that conversion when redirecting the user agent back to the `AuthenticationProperties.RedirectUri`.
  - The base remote authentication handler doesn't have a preferred format and `Response.Redirect()` may be called with either an absolute URL (when `AuthenticationProperties.RedirectUri` is itself an absolute URL) or with a relative one (e.g when `AuthenticationProperties.RedirectUri` is null, as it defaults to `/`).

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
```diff
@@ -194,7 +205,7 @@ protected override async Task HandleChallengeAsync(AuthenticationProperties prop
         {
             if (string.IsNullOrEmpty(properties.RedirectUri))
             {
-                properties.RedirectUri = CurrentUri;
+                properties.RedirectUri = OriginalPathBase + Request.Path + Request.QueryString;
```
PinpointTownes: @Tratcher stricto sensu, these changes are not required, but if we end up flowing the `ReturnUrl` up to the access denied endpoint, we'll likely want it to be captured as a relative URL so helpers like `IUrlHelper.IsLocalUrl` and `RedirectToLocal()` work correctly (they don't deal with absolute URLs, even if the domain matches the current one).

Tratcher:
Any guess what the WsFed variant of this would look like for https://github.com/aspnet/Security/issues/1891?

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
Tratcher: This should be OriginalPathBase + OriginalPath.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -520,6 +520,17 @@ protected override async Task<HandleRequestResult> HandleRemoteAuthenticateAsync
...
                 if (!string.IsNullOrEmpty(authorizationResponse.Error))
                 {
+                    // Note: access_denied errors are special protocol errors indicating the user didn't
+                    // approve the authorization demand requested by the remote authorization server.
+                    // Since it's a frequent scenario (that is not caused by incorrect configuration),
+                    // denied errors are handled differently using a special "access denied" exception.
+                    // Visit https://tools.ietf.org/html/rfc6749#section-4.1.2.1 for more information.
+                    if (string.Equals(authorizationResponse.Error, "access_denied", StringComparison.Ordinal))
+                    {
+                        return HandleRequestResult.Fail(new AccessDeniedException(
```
Tratcher: I'd rather invoke the AccessDenied event here than create a new AccessDeniedException that's never actually thrown.

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
PinpointTownes: I copied it from the cookie handler: https://github.com/aspnet/Security/blob/master/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationHandler.cs#L425

Should I also update the cookie handler?

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
PinpointTownes: That's possible, but it means we'll have the same (duplicated) logic in all the handlers. It will also be a little weird, as `RemoteAuthenticationEvents`/`RemoteAuthenticationOptions` will have events/options for something that is actually implemented in subclasses.

It's unfortunate, but errors are represented by `Exception`s in 2.0. I'd also have preferred a proper `AuthentiationError`, but I guess that's too late for that.

PinpointTownes:
> Any guess what the WsFed variant of this would look like for #1891?

I'll have to re-read the WS-Fed as I don't remember if there's anything specific for access denied callbacks.

PinpointTownes:
> I'll have to re-read the WS-Fed specification as I don't remember if there's anything specific for access denied callbacks.

I read it during the weekend (it's always a terrible experience!) and I didn't find any trace of such a thing (AFAICT, there's even no error callback mechanism).

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: Calling the event in a subclass is fine, OAuth CreatingTicket is a lot like that. You could share the code via a protected method if you like.

src/Microsoft.AspNetCore.Authentication.OAuth/OAuthHandler.cs
Tratcher: Sure. FYI: https://github.com/aspnet/Security/issues/1730

PinpointTownes:
@Tratcher based on your suggestion, I added a protected virtual `HandleAccessDeniedErrorAsync()` helper in `RemoteAuthenticationHandler`.

Once the final design is approved, I'll add some tests for the new event.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
```diff
@@ -520,6 +520,16 @@ protected override async Task<HandleRequestResult> HandleRemoteAuthenticateAsync
...
                 if (!string.IsNullOrEmpty(authorizationResponse.Error))
                 {
+                    // Note: access_denied errors are special protocol errors indicating the user didn't
+                    // approve the authorization demand requested by the remote authorization server.
+                    // Since it's a frequent scenario (that is not caused by incorrect configuration),
+                    // denied errors are handled differently using HandleAccessDeniedErrorAsync().
+                    // Visit https://tools.ietf.org/html/rfc6749#section-4.1.2.1 for more information.
+                    if (string.Equals(authorizationResponse.Error, "access_denied", StringComparison.Ordinal))
+                    {
+                        return await HandleAccessDeniedErrorAsync(properties);
```
PinpointTownes: @Tratcher pretty much like `RemoteFailure`, the fact `AccessDenied` is protocol-agnostic prevents us from flowing things like the OIDC authorization response, which may limit extensibility. Not sure it's really a problem, but let me know if you'd prefer having multiple (package-specific) `AccessDenied` events instead of a global one.

Tratcher: There's no other parameters in the message we'd be missing out on, right?

The most awkward bit seems to be having to downcast the Options if you wanted anything specific. Or could you flow the generic TOptions? I remember trying that and having lots of issues before.

Tratcher:
The design looks good. Go ahead and finish the tests.

PinpointTownes:
> This is the same as setting Properties.RedirectUri, right?

Yup. It just has a better name and you'll be able to set manually if even if `properties` is null.

src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs
Tratcher: I can see those being useful if properly populated. It seems like you'd want to flow a generic property bag to the event rather than hardcoded parameters. I guess you could pass them through AuthProperties or some HttpContext field if you really needed to.

PinpointTownes: > The most awkward bit seems to be having to downcast the Options if you wanted anything specific. Or could you flow the generic TOptions? I remember trying that and having lots of issues before.

Yep. `RemoteAuthenticationEvents` is not generic so we can't easily flow `TOptions`. We could have a generic `AccessDeniedContext<TOptions>` class derived from `AccessDeniedContext` but it wouldn't help much as it would still be exposed as `AccessDeniedContext` by the events class.

> It seems like you'd want to flow a generic property bag to the event rather than hardcoded parameters. I guess you could pass them through AuthProperties or some HttpContext field if you really needed to.

We can certainly do that, but I'm a bit reluctant as it would be fairly inconsistent (we don't do that anywhere else).

PinpointTownes:
@Tratcher assuming this PR won't be part of 2.2 (it's too late for that, right?), we should consider setting `RemoteAuthenticationOptions.AccessDeniedPath` to `/Account/Login` by default to match the default login page path. This way, users would be automatically redirected to the login page OOTB without having to configure anything.

Tratcher:
Yeah, this is coming in too late for 2.2.  If there were to be a default I'd rather it go to the error page, or better yet a new login error page. That's something we'll need to work out with @blowdart and the templates.

PinpointTownes:
> Yeah, this is coming in too late for 2.2. If there were to be a default I'd rather it go to the error page, or better yet a new login error page.

Not a huge fan of the default error page option as it gives absolutely no indication of what's wrong with the request, which is precisely what this PR tries to avoid. A separate "try again" page is probably a reasonable tradeoff.

> That's something we'll need to work out with @blowdart and the templates.

@blowdart care to chime in or you're too busy posting random stuff on Twitter? :trollface:

Eilon:
@Tratcher - can you go ahead and merge this?

Also please log a bug in the Docs repo to suggest adding/updating docs regarding this new feature.

@PinpointTownes - thank you!

PinpointTownes:
> @PinpointTownes - thank you!

My pleasure :smile:

PR #1894, opened 2018-10-31
Authorization middleware

JamesNK:
https://github.com/aspnet/Routing/issues/537

~~Will add tests in a different PR~~

### DIFF
--- a/Security.sln
+++ b/Security.sln
@@ -79,6 +79,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Authen
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WsFedSample", "samples\WsFedSample\WsFedSample.csproj", "{5EC2E398-E46A-430D-8E4B-E91C8FC3E800}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "benchmarks", "benchmarks", "{55052FE3-F8C2-4E6C-97B0-C02ED1DBEA62}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Security.Performance", "benchmarks\Microsoft.AspNetCore.Security.Performance\Microsoft.AspNetCore.Security.Performance.csproj", "{556C4FAA-F4B1-4EA9-8921-CB1DF7D94C2A}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -517,6 +521,22 @@ Global
 		{5EC2E398-E46A-430D-8E4B-E91C8FC3E800}.Release|x64.Build.0 = Release|Any CPU
 		{5EC2E398-E46A-430D-8E4B-E91C8FC3E800}.Release|x86.ActiveCfg = Release|Any CPU
 		{5EC2E398-E46A-430D-8E4B-E91C8FC3E800}.Release|x86.Build.0 = Release|Any CPU
+		{556C4FAA-F4B1-4EA9-8921-CB1DF7D94C2A}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{556C4FAA-F4B1-4EA9-8921-CB1DF7D94C2A}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{556C4FAA-F4B1-4EA9-8921-CB1DF7D94C2A}.Debug|Mixed Platforms.ActiveCfg = Debug|Any CPU
+		{556C4FAA-F4B1-4EA9-8921-CB1DF7D94C2A}.Debug|Mixed Platforms.Build.0 = Debug|Any CPU
+		{556C4FAA-F4B1-4EA9-8921-CB1DF7D94C2A}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{556C4FAA-F4B1-4EA9-8921-CB1DF7D94C2A}.Debug|x64.Build.0 = Debug|Any CPU
+		{556C4FAA-F4B1-4EA9-8921-CB1DF7D94C2A}.Debug|x86.ActiveCfg = Debug|Any CPU
+		{556C4FAA-F4B1-4EA9-8921-CB1DF7D94C2A}.Debug|x86.Build.0 = Debug|Any CPU
+		{556C4FAA-F4B1-4EA9-8921-CB1DF7D94C2A}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{556C4FAA-F4B1-4EA9-8921-CB1DF7D94C2A}.Release|Any CPU.Build.0 = Release|Any CPU
+		{556C4FAA-F4B1-4EA9-8921-CB1DF7D94C2A}.Release|Mixed Platforms.ActiveCfg = Release|Any CPU
+		{556C4FAA-F4B1-4EA9-8921-CB1DF7D94C2A}.Release|Mixed Platforms.Build.0 = Release|Any CPU
+		{556C4FAA-F4B1-4EA9-8921-CB1DF7D94C2A}.Release|x64.ActiveCfg = Release|Any CPU
+		{556C4FAA-F4B1-4EA9-8921-CB1DF7D94C2A}.Release|x64.Build.0 = Release|Any CPU
+		{556C4FAA-F4B1-4EA9-8921-CB1DF7D94C2A}.Release|x86.ActiveCfg = Release|Any CPU
+		{556C4FAA-F4B1-4EA9-8921-CB1DF7D94C2A}.Release|x86.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -549,6 +569,7 @@ Global
 		{24A28F5D-E5A9-4CA8-B0D2-924A1F8BE14E} = {F8C0AA27-F3FB-4286-8E4C-47EF86B539FF}
 		{B1FC6AAF-9BF2-4CDA-84A2-AA8BF7603F29} = {4D2B6A51-2F9F-44F5-8131-EA5CAC053652}
 		{5EC2E398-E46A-430D-8E4B-E91C8FC3E800} = {F8C0AA27-F3FB-4286-8E4C-47EF86B539FF}
+		{556C4FAA-F4B1-4EA9-8921-CB1DF7D94C2A} = {55052FE3-F8C2-4E6C-97B0-C02ED1DBEA62}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {ABF8089E-43D0-4010-84A7-7A9DCFE49357}
--- a/src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationAppBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationAppBuilderExtensions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using Microsoft.AspNetCore.Authorization;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.Builder
 {
@@ -16,6 +17,21 @@ namespace Microsoft.AspNetCore.Builder
             }
 
             return app.UseMiddleware<AuthorizationMiddleware>();
+        }
+
+        public static IApplicationBuilder UseAuthorization(this IApplicationBuilder app, AuthorizationOptions options)
+        {
+            if (app == null)
+            {
+                throw new ArgumentNullException(nameof(app));
+            }
+
+            if (options == null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            return app.UseMiddleware<AuthorizationMiddleware>(new DefaultAuthorizationPolicyProvider(Options.Create(options)));
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationAppBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationAppBuilderExtensions.cs
@@ -3,12 +3,19 @@
 
 using System;
 using Microsoft.AspNetCore.Authorization;
-using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.Builder
 {
+    /// <summary>
+    /// Extension methods to add authorization capabilities to an HTTP application pipeline.
+    /// </summary>
     public static class AuthorizationAppBuilderExtensions
     {
+        /// <summary>
+        /// Adds the <see cref="AuthorizationMiddleware"/> to the specified <see cref="IApplicationBuilder"/>, which enables authorization capabilities.
+        /// </summary>
+        /// <param name="app">The <see cref="IApplicationBuilder"/> to add the middleware to.</param>
+        /// <returns>A reference to this instance after the operation has completed.</returns>
         public static IApplicationBuilder UseAuthorization(this IApplicationBuilder app)
         {
             if (app == null)
@@ -17,21 +24,6 @@ namespace Microsoft.AspNetCore.Builder
             }
 
             return app.UseMiddleware<AuthorizationMiddleware>();
-        }
-
-        public static IApplicationBuilder UseAuthorization(this IApplicationBuilder app, AuthorizationOptions options)
-        {
-            if (app == null)
-            {
-                throw new ArgumentNullException(nameof(app));
-            }
-
-            if (options == null)
-            {
-                throw new ArgumentNullException(nameof(options));
-            }
-
-            return app.UseMiddleware<AuthorizationMiddleware>(new DefaultAuthorizationPolicyProvider(Options.Create(options)));
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationAppBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationAppBuilderExtensions.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Authorization;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    public static class AuthorizationAppBuilderExtensions
+    {
+        public static IApplicationBuilder UseAuthorization(this IApplicationBuilder app)
+        {
+            if (app == null)
+            {
+                throw new ArgumentNullException(nameof(app));
+            }
+
+            return app.UseMiddleware<AuthorizationMiddleware>();
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationEndpointConventionBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationEndpointConventionBuilderExtensions.cs
@@ -1,0 +1,50 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Routing;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    public static class AuthorizationEndpointConventionBuilderExtensions
+    {
+        public static IEndpointConventionBuilder RequireAuthorization(this IEndpointConventionBuilder builder, params IAuthorizeData[] authorizeData)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            if (authorizeData == null)
+            {
+                throw new ArgumentNullException(nameof(authorizeData));
+            }
+
+            builder.Apply(endpointBuilder =>
+            {
+                foreach (var data in authorizeData)
+                {
+                    endpointBuilder.Metadata.Add(data);
+                }
+            });
+            return builder;
+        }
+
+        public static IEndpointConventionBuilder RequireAuthorization(this IEndpointConventionBuilder builder, params string[] policyNames)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            if (policyNames == null)
+            {
+                throw new ArgumentNullException(nameof(policyNames));
+            }
+
+            return builder.RequireAuthorization(policyNames.Select(n => new AuthorizeAttribute(n)).ToArray());
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationMiddleware.cs
@@ -7,6 +7,7 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Authorization.Policy;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Endpoints;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.Extensions.DependencyInjection;
 
@@ -40,7 +41,7 @@ namespace Microsoft.AspNetCore.Authorization
                 throw new ArgumentNullException(nameof(context));
             }
 
-            var endpoint = context.Features.Get<IEndpointFeature>()?.Endpoint;
+            var endpoint = context.GetEndpoint();
             var authorizeData = endpoint?.Metadata.GetOrderedMetadata<IAuthorizeData>() ?? Array.Empty<IAuthorizeData>();
             var policy = await AuthorizationPolicy.CombineAsync(_policyProvider, authorizeData);
             if (policy == null)
--- a/src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationMiddleware.cs
@@ -40,8 +40,6 @@ namespace Microsoft.AspNetCore.Authorization
                 throw new ArgumentNullException(nameof(context));
             }
 
-            var policyEvaluator = context.RequestServices.GetRequiredService<IPolicyEvaluator>();
-
             var endpoint = context.Features.Get<IEndpointFeature>()?.Endpoint;
             var authorizeData = endpoint?.Metadata.GetOrderedMetadata<IAuthorizeData>() ?? Array.Empty<IAuthorizeData>();
             var policy = await AuthorizationPolicy.CombineAsync(_policyProvider, authorizeData);
@@ -50,6 +48,8 @@ namespace Microsoft.AspNetCore.Authorization
                 await _next(context);
                 return;
             }
+
+            var policyEvaluator = context.RequestServices.GetRequiredService<IPolicyEvaluator>();
 
             var authenticateResult = await policyEvaluator.AuthenticateAsync(policy, context);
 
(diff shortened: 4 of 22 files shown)


HaoK:
Code looks fine, do we want consider just having this live in the Authorization.Policy package that the evaluator lives in?  There isn't very much in that package today, and since this middleware depends on that, it doesn't seem like a terrible fit...

JamesNK:
I chose not to because Authorization.Policy has no dependency on Http, and the namespace for every file is Authorization.Policy. It feels like an odd fit to put middleware in it.

HaoK:
It does though since I created it basically to be able to use Authorization and Http together, the reason you don't see http in the namespaces, is due to authentication not having it even though its lives in httpabstractions:

https://github.com/aspnet/HttpAbstractions/tree/master/src/Microsoft.AspNetCore.Authentication.Abstractions

Also for example, the main PolicyEvaluator AuthenticateAsync method takes httpContext as a parameter:

https://github.com/aspnet/HttpAbstractions/tree/master/src/Microsoft.AspNetCore.Authentication.Abstractions

Security.sln
```diff
@@ -79,6 +79,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Authen
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "WsFedSample", "samples\WsFedSample\WsFedSample.csproj", "{5EC2E398-E46A-430D-8E4B-E91C8FC3E800}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Authorization.Middleware", "src\Microsoft.AspNetCore.Authorization.Middleware\Microsoft.AspNetCore.Authorization.Middleware.csproj", "{E2D9B067-CD74-4EA8-9CBA-D303DB63C1AC}"
```
rynowak: Do we need a project for this? Is the reason that we don't have a project that ties together HTTP and AuthZ?

Tratcher:
I'll read through this tomorrow, but I do want you to try to replace this [sample](https://github.com/aspnet/AuthSamples/tree/master/samples/StaticFilesAuth) as an acceptance test.

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationAppBuilderExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+    public static class AuthorizationAppBuilderExtensions
+    {
+        public static IApplicationBuilder UseAuthorization(this IApplicationBuilder app)
+        {
+            if (app == null)
+            {
+                throw new ArgumentNullException(nameof(app));
+            }
+
+            return app.UseMiddleware<AuthorizationMiddleware>();
```
rynowak: is there any sort of options object that should be passed in? or is it always one from DI?

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationEndpointConventionBuilderExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Routing;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    public static class AuthorizationEndpointConventionBuilderExtensions
+    {
+        // REVIEW: What other overloads should exist?
+        // e.g. params string[] policies, or params string[] roles
+        public static IEndpointConventionBuilder RequireAuthorization(this IEndpointConventionBuilder builder, params IAuthorizeData[] authorizeData)
```
rynowak: I'd suggest starting with `string` (policyname) and a single `IAuthorizeData` here. I don't think passing in multiple datum is that common, and I would expect that a policy name is the most common.

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationMiddleware.cs
```diff
@@ -0,0 +1,108 @@
...
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.Authorization
+{
+    public class AuthorizationMiddleware
+    {
+        private readonly RequestDelegate _next;
+        private readonly IAuthorizationPolicyProvider _policyProvider;
+
+        public AuthorizationMiddleware(RequestDelegate next, IAuthorizationPolicyProvider policyProvider)
```
rynowak: Does it need to be public?

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationMiddleware.cs
```diff
@@ -0,0 +1,108 @@
...
+            if (endpoint == null)
+            {
+                await _next(context);
+                return;
+            }
+
+            var policyEvaluator = context.RequestServices.GetRequiredService<IPolicyEvaluator>();
+
+            // REVIEW: Does order matter? Is ordered metadata in the correct order?
+            var authorizeData = endpoint.Metadata.GetOrderedMetadata<IAuthorizeData>();
```
rynowak: I'll let @HaoK respond as to whether order matters when combining policies. `GetOrderedMetadata` returns items from most specific to least specific.

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationMiddleware.cs
```diff
@@ -0,0 +1,108 @@
...
+            var policy = await AuthorizationPolicy.CombineAsync(_policyProvider, authorizeData);
+            if (policy == null)
```
rynowak: Does this really ever return null? What if the metadata is empty?

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationMiddleware.cs
```diff
@@ -0,0 +1,108 @@
...
+            {
+                await _next(context);
+                return;
+            }
+
+            var authenticateResult = await policyEvaluator.AuthenticateAsync(policy, context);
+
+            // Allow Anonymous skips all authorization
+            if (endpoint.Metadata.GetMetadata<IAllowAnonymous>() != null)
```
rynowak: Are y'all in love with how this behaves (@blowdart @HaoK)? This is what MVC does but it's never made sense to me.

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationMiddleware.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        public async Task Invoke(HttpContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            var endpoint = context.Features.Get<IEndpointFeature>()?.Endpoint;
+            if (endpoint == null)
```
rynowak: how about a default policy?

Security.sln
JamesNK: Authorization.Policy has both

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationMiddleware.cs
JamesNK: The type doesn't have to be public.

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationAppBuilderExtensions.cs
JamesNK: IAuthorizationPolicyProvider currently always comes from DI

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationMiddleware.cs
Tratcher: All of our middleware classes are public. The extension methods don't always suite people and it's good that they can register the middleware themselves.

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationMiddleware.cs
JamesNK: It can return null if there is no `IAuthorizeData` metadata.

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationMiddleware.cs
JamesNK: What do you mean by default? I get the impression from the filter code that the default comes into play when there is an attribute with no policy or roles, i.e. `[Authorize]`

If there is no endpoint then there is no `AuthorizeAttribute`, and so the default policy doesn't come into play. Correct: yay/nay?

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationMiddleware.cs
blowdart: What do you think is wrong with it?

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationAppBuilderExtensions.cs
Tratcher: Is there a need to be able to add multiple instances of this? E.g. `UseAuthorization("policyA")`? It might be useful for that StaticFile example.

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationMiddleware.cs
Tratcher: That's true sofar, but being middleware allows for a new scenario. For example lots of people write middleware today that denies anonymous requests, or applies policy like conditions. See the StaticFile example.

Above I proposed `UseAuthorization("default")` to opt into a specific policy regardless of the endpoint.

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationAppBuilderExtensions.cs
HaoK: Yeah, that's the general extensibility point for if you want dynamic policies for example.

Tratcher:
https://github.com/aspnet/Security/issues/1734

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationEndpointConventionBuilderExtensions.cs
HaoK: So IAuthorizeData already has policy name inside it, so maybe you just need 2 overloads to start one that takes policyName, and another that takes IAuthorizeData

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationMiddleware.cs
HaoK: So if there's no endpoint and no authorize attribute, the authZ middleware is a no-op?

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationMiddleware.cs
HaoK: Order generally doesn't matter more combining policies with the default implementations since all requirements in all policies must be satisfied.  It might affect the order in which authentication builds the claims principal as well, but that shouldn't be order sensitive for the resulting claim set, just you would see slight differences in the order of execution, so I think its fine.

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationMiddleware.cs
HaoK: Yeah so I was told that this is how it always works 🤷‍♂️

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationAppBuilderExtensions.cs
HaoK: Yeah I asked about what the intended usage of this middleware is, and rather than it being analogous to a single `[Authorize(policyName)]` invocation, its intended to be a single effective authorization check that uses the endpoint to compute the effective policy for the entire request.

HaoK: I'm assuming if multiple UseAuthorization() end up in the pipeline, they would become no-ops similar to [Authorize] attributes in MVC right?

Tratcher: I was referring to branched pipelines.

Tratcher: > Yeah I asked about what the intended usage of this middleware is, and rather than it being analogous to a single `[Authorize(policyName)]` invocation, its intended to be a single effective authorization check that uses the endpoint to compute the effective policy for the entire request.

Why not both? Allowing us to specify a policy to run would help extend AuthZ to non-endpoint aware middleware.

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationMiddleware.cs
JamesNK: Right now, yes

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationAppBuilderExtensions.cs
JamesNK: I can see `UseAuthorization("policyA")` having value. Will add.

src/Microsoft.AspNetCore.Authorization.Middleware/AuthorizationEndpointConventionBuilderExtensions.cs
JamesNK: I'll setup two overloads:
* `params IAuthorizeData[] authorizeData`
* `params string[] policyNames`

I think param arrays give a good experience for passing in one, and passing in multiple.

rynowak: but why tho? Like why do you want multiple? Why are you **so** in love with complexity?

JamesNK: You can have multiple `AuthorizeAttribute`s on a method. It makes sense that the auth helper method would also allow multiple.

JamesNK:
🆙 📅 

The remaining ask a policy that runs even if there is no endpoint matched, or there is no `IAuthorizeData` on the endpoint.

Authorization today has the concept of `DefaultPolicy`, but that refers to the policy used when there is `IAuthorizeData` with no policy name on an endpoint.

What about adding the idea of `GlobalPolicy`? It runs on all requests that pass through `AuthorizationMiddleware` (and should add it to the MVC authz filter for consistency). If the request has an endpoint with policies then they are evaluated together with the global policy. The default GlobalPolicy is no op.

Questions:
* Does GlobalPolicy live on AuthorizationOptions?
* If the policy lives on AuthorizationOptions, should it be passed as an argument to the middleware? Will allow different globals in different branches
* If AuthorizationOptions is passed to the middleware, should ALL the options on it be local to the route? That could allow "Policy1" to mean one thing in one instance of the middleware, and something else in another instance.

Alternative suggestions welcome!

HaoK:
Just to clarify the meaning of GlobalPolicy, this is basically a required policy that always must be satisfied, as compared to DefaultPolicy which is only used if no policy is specified...  So really the globalPolicy is more of a requiredPolicy then which is always applied no matter what?

HaoK:
Also if we want to introduce something like this, maybe it would make more sense to bake in at the lowest level so imperative calls to IAuthorizationService.Authorize would pick it up as well, instead of something that only middleware/endpoints could take advantage of (this would align with exposing it in AuthorizationOptions)

HaoK:
Basically the default authZ service implementation would just always include the requirements from the Global/Required policy in addition to any other requirements passed in for evaluation.

JamesNK:
> If that's the case I think I prefer a name closer to RequiredPolicy since its always applied no matter what?

Ok!

> Also if we want to introduce something like this, maybe it would make more sense to bake in at the lowest level so imperative calls to IAuthorizationService.Authorize would pick it up as well

Good idea! I think it would feel cleanest if it was exposed from IAuthorizationPolicyProvider: `IAuthorizationPolicyProvider.GetRequiredPolicyAsync()`. The middleware already takes a policy provider so customizing the required policy per middleware would be simple enough. On the other hand it would be a breaking change for 3.0. The alternative option is to read it directly from AuthorizationOptions.

HaoK:
That's what major versions changes are for :)  Yeah I agree it should just behave just like DefaultPolicy so it should get a method on the policy provider and we can just have the default implementation read it from the options like with the default policy.

JamesNK:
Updated with RequiredPolicy + tests.

JamesNK:
@natemcmaster @ryanbrandenburg Hi. Could one of y'all enable feature branch builds for https://dotnet.myget.org/feed/aspnetcore-dev/package/nuget/Microsoft.AspNetCore.Authorization.Policy ?

I'd like to use a package over here: https://github.com/aspnet/AuthSamples

src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationMiddleware.cs
```diff
@@ -0,0 +1,102 @@
...
+            // Note that the resource will be null if there is no matched endpoint
+            var authorizeResult = await policyEvaluator.AuthorizeAsync(policy, authenticateResult, context, resource: endpoint);
+
+            if (authorizeResult.Challenged)
+            {
+                if (policy.AuthenticationSchemes.Any())
+                {
+                    foreach (var scheme in policy.AuthenticationSchemes)
+                    {
+                        await context.ChallengeAsync(scheme);
```
JamesNK: Should `ChallengeAsync` and `ForbidAsync` be passed the authentication properties from `authenticateResult`?

src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationMiddleware.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        public async Task Invoke(HttpContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            var policyEvaluator = context.RequestServices.GetRequiredService<IPolicyEvaluator>();
```
davidfowl: Why isn't this being injected in the constructor?

src/Microsoft.AspNetCore.Authorization.Policy/Microsoft.AspNetCore.Authorization.Policy.csproj
```diff
@@ -14,6 +14,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Authentication.Abstractions" Version="$(MicrosoftAspNetCoreAuthenticationAbstractionsPackageVersion)" />
+    <PackageReference Include="Microsoft.AspNetCore.Routing" Version="$(MicrosoftAspNetCoreRoutingPackageVersion)" />
```
davidfowl: Hmm, what's this for?

src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationMiddleware.cs
HaoK: I am not sure that makes sense, you likely got nothing back in the challenge case, and forbid means that was probably not the right identity.

The properties are more metadata associated with that particular result/request.  Things typically set in the properties for forbid/challenge are usually more like parameters to control behavior like urls

natemcmaster:
> Could one of y'all enable feature branch builds for 
done

src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationMiddleware.cs
JamesNK: The reason I brought it up is there is a bug in cookies auth around the redirect URL + app.Map - https://github.com/aspnet/AuthSamples/pull/65#issuecomment-435260693

I noticed that `ChallengeAsync` for cookies auth can read the `RedirectUri` from the authentication properties, so a fix for the bad redirect URL could be to hack in a new implementation of `IPolicyEvaluator.AuthenticateAsync` that returns a fixed `RedirectUri` in the result properties, and `ChallengeAsync` would then pick it up.

The bug is being fixed anyway, and this is a pretty esoteric scenario, just wondering if it should be done. My knowledge of ASP.NET Core security is high-level.

src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationMiddleware.cs
JamesNK: PolicyEvaluator is transient. @HaoK could you give a brief explanation why? I'll add a comment.

src/Microsoft.AspNetCore.Authorization.Policy/Microsoft.AspNetCore.Authorization.Policy.csproj
JamesNK: The extension methods for adding authorization metadata to an endpoint - `RequireAuthorization` - use `IEndpointConventionBuilder` and `EndpointModel` which come from routing.

JamesNK: We could move these builder types to `Microsoft.AspNetCore.Http.Abstractions`. They are pretty similar to `IApplicationBuilder` which lives there.

src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationMiddleware.cs
HaoK: Ah yes, people wanted the ability to evaluate policies per request, which actually brings up a similar question about the IPolicyProvider, should that also be resolved from RequestServices in case they have their own custom one that is scoped?

src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationAppBuilderExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            if (app == null)
+            {
+                throw new ArgumentNullException(nameof(app));
+            }
+
+            return app.UseMiddleware<AuthorizationMiddleware>();
+        }
+
+        public static IApplicationBuilder UseAuthorization(this IApplicationBuilder app, AuthorizationOptions options)
```
HaoK: Why do we need this overload?  I think its just got the potential for confusion when this overload will ignore the authorization options configured normally, is this needed for the static files sample?

src/Microsoft.AspNetCore.Authorization.Policy/Microsoft.AspNetCore.Authorization.Policy.csproj
davidfowl: We should probably do that.

JamesNK:
Two open questions to resolve:

1. Should there be a `UseAuthorization` overload that takes `AuthorizationOptions`? And if so, should the options completely replace the options configured globally?
2. Do we still want `RequiredPolicy` as a new concept? I have updated the static files auth sample so that it is no longer nessessary and I think the new approach is better.

I don't have a strong opinion either way about adding or removing these.

src/Microsoft.AspNetCore.Authorization.Policy/Microsoft.AspNetCore.Authorization.Policy.csproj
JamesNK: https://github.com/aspnet/HttpAbstractions/pull/1058

HaoK:
If you don't need either of these anymore, maybe leaving them out to start is best.  `AuthenticationMiddleware` is analogous to the AuthZMiddleware with its options + scheme provider and it doesn't have an overload that lets you replace the options, also its pretty easy to implement required policy in a custom IAuthorizationService even if we don't bake it in.

src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationMiddleware.cs
Tratcher: No need, this challenge is instigating a new auth flow.

src/Microsoft.AspNetCore.Authorization/AuthorizationOptions.cs
```diff
@@ -27,6 +27,11 @@ public class AuthorizationOptions
         /// </remarks>
         public AuthorizationPolicy DefaultPolicy { get; set; } = new AuthorizationPolicyBuilder().RequireAuthenticatedUser().Build();
 
+        /// <summary>
+        /// Gets or sets the required authorization policy. Defaults to null.
+        /// </summary>
+        public AuthorizationPolicy RequiredPolicy { get; set; }
```
Tratcher: So when would I use RequiredPolicy vs DefaultPolicy?

test/Microsoft.AspNetCore.Authorization.Test/AuthorizationAppBuilderExtensionsTests.cs
```diff
@@ -0,0 +1,66 @@
...
+
+            var app = new ApplicationBuilder(services);
+
+            app.UseAuthorization();
+
+            var appFunc = app.Build();
+
+            var endpointFeature = Mock.Of<IEndpointFeature>();
+            endpointFeature.Endpoint = new Endpoint(
+                context => Task.CompletedTask,
```
Tratcher: null

test/Microsoft.AspNetCore.Authorization.Test/AuthorizationAppBuilderExtensionsTests.cs
```diff
@@ -0,0 +1,66 @@
...
+            var authenticationService = new TestAuthenticationService();
+            var services = CreateServices(authenticationService);
+
+            var app = new ApplicationBuilder(services);
+
+            app.UseAuthorization();
+
+            var appFunc = app.Build();
+
+            var endpointFeature = Mock.Of<IEndpointFeature>();
```
Tratcher: Mock isn't needed here, the new SetEndpoint discussed elsewhere would handle this. We've avoided Mock unless absolutely necessary.

src/Microsoft.AspNetCore.Authorization/AuthorizationOptions.cs
JamesNK: DefaultPolicy = run on endpoint with `[Authorize]` (no policy name)
RequiredPolicy = run on ALL endpoints (no authorize)

RequiredPolicy was... required... in the static files sample before I changed it to set an endpoint with IAuthorizeData metadata on it.

Do you have an opinion on whether RequiredPolicy is a useful thing to have? We could remove it.

src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationEndpointConventionBuilderExtensions.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Routing;
+
+namespace Microsoft.AspNetCore.Builder
```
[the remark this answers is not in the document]
JamesNK: https://github.com/aspnet/HttpAbstractions/pull/1058/files

It is used in Startup.cs. I think this is the right namespace.

src/Microsoft.AspNetCore.Authorization.Policy/AuthorizationMiddleware.cs
JamesNK: ok!

src/Microsoft.AspNetCore.Authorization/AuthorizationOptions.cs
Tratcher: I've seen enough people write their own Authorization middleware that says "Deny all anonymous requests beyond this point", so yes. E.g. It's a global Authorize attribute.

JamesNK:
🆙 📅

HaoK (approved):

src/Microsoft.AspNetCore.Authorization/AuthorizationOptions.cs
Tratcher: Add that explanation to the doc comments please.

Tratcher (approved):

PR #1896, opened 2018-10-31
[WIP] Add AuthorizationMiddleware tests

JamesNK:
Port of relevant authz filter tests. Will add more to cover scenarios specific to the new authz middleware.


test/Microsoft.AspNetCore.Authorization.Test/AuthorizationMiddlewareTests.cs
```diff
@@ -0,0 +1,412 @@
...
+            public int CalledCount { get; private set; }
+
+            public Task RequestDelegate(HttpContext context)
+            {
+                CalledCount++;
+                context.Response.StatusCode = _statusCode;
+                return Task.CompletedTask;
+            }
+
+            public TestRequestDelegate(int statusCode = 200)
```
Tratcher: Move the constructor above Invoke.

Issue #1883, opened 2018-10-19
RequireClaim method signature is a little confusing


benmccallum:
```csharp
public AuthorizationPolicyBuilder RequireClaim(string claimType, params string[] requiredValues);
public AuthorizationPolicyBuilder RequireClaim(string claimType, IEnumerable<string> requiredValues);
```

"requiredValues" is a confusing name as the method in reality only checks for the presence of one of these values, not all. The parameter description is perfect, but I'm not sure if that's enough. I certainly got confused until I confirmed it in the docs/saw comment. 

It's a petty thing I know, but maybe something more generic like "values" would suffice? Or go super explicit like "anyOfTheseValues"?

HaoK:
Sure, do you want to submit a PR for this change?

benmccallum:
Sure thing. Any preference on a new param name?

dgarciarubio:
How about "possibleValues"? just "values" seems vague, but "anyOfTheseValues" sounds too informal to me.

HaoK:
acceptedValues maybe?

benmccallum:
How about valuesAllowed? Or along the lines of auth, valuesAuthorized?

HaoK:
allowedValues is fine

benmccallum:
PR up with allowedValues. Cheers!

HaoK:
Merged into master for 3.0, thanks @benmccallum !

PR #1908, opened 2018-11-09
refactor: Rename param from requiredValues to allowedValues to better…

benmccallum:
… indicate usage. Addresses #1883 .


HaoK:
Thanks @benmccallum !

Issue #1421, opened 2017-09-13
Remove unused Authentication.AddScheme/RemoteScheme methods


HaoK:
These are no longer used, as we introduced the AuthenticationBuilder pattern, but a few of the old extension methods were left behind (duplicated dead code basically)

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication/AuthenticationServiceCollectionExtensions.cs#L50

I discovered this when I edited the wrong EnsureSignInScheme :(

If we can't delete in 2.1 (due to potential breaking change?), we should at least obsolete

Eilon:
We'll obsolete them in 2.1 in https://github.com/aspnet/Security/issues/1431

And then in 3.0.0 we'll delete them.

Please post an announcement when we delete them.

Issue #1631, opened 2018-02-01
Remove unused dependency IDataProtectionProvider from JwtBearerHandler class


maximpashuk:
Hi.

JwtBearerHandler class constructor has unused dependency from IDataProtectionProvider interface

https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerHandler.cs#L25

Please remove this dependency from constructor.

Tratcher:
Why do you ask?

Changing/Removing API is breaking and can only be done in major versions.

Eilon:
It would be a breaking change to do this, but we will consider it for 3.0.0.

Issue #1292, opened 2017-07-05
Remove obsolete cookie options API


natemcmaster:
Follow up to https://github.com/aspnet/Security/pull/1285

Remove these obsoleted API in the next major release
```
CookieAuthenticationOptions
  .CookieName
  .CookieDomain
  .CookiePath
  .CookieHttpOnly
  .CookieSecure

```
~~.ExpireTimeSpan~~ - https://github.com/aspnet/Security/pull/1296

Tratcher:
Not ExpireTimeSpan: https://github.com/aspnet/Security/issues/1293

muratg:
@HaoK forgot to close this one? :)

Issue #1221, opened 2017-05-18
Remove obsolete AuthorizeAttribute.ActiveAuthenticationSchemes property


Eilon:
See https://github.com/aspnet/Security/issues/1199 where we marked it as obsolete, and added a new, better-named property.

muratg:
@HaoK forgot to close this one? :)

PR #1915, opened 2018-11-16
Remove obsolete apis

HaoK:
Fixes https://github.com/aspnet/Security/issues/1421
https://github.com/aspnet/Security/issues/1631
https://github.com/aspnet/Security/issues/1292
https://github.com/aspnet/Security/issues/1221


Tratcher:
Post an announcement.

Not waiting for reviews?

HaoK:
Not if it has to go in today :)  I figure any review feedback I can address afterwards

HaoK:
This has potential to break things, so I'd rather see any issues early in the build before a weekend is the main reason

HaoK:
I'll post an announcement post mondo jump

